PR #6, opened 2016-08-04
Don't blow away static input data

fogleman:
@martinken @jourdain

This line of code is blowing away my static input data, where I don't have an input connection. My code works when I comment this line out. Ken mentioned that this line might not even be needed. Just wanted to confirm before checking in a change to remove it. Let me know.

There's another potential issue here:

https://github.com/Kitware/vtk-js/blob/static-input-data/Sources/macro.js#L313

But I'm not sure when `update` is actually called.

### DIFF
--- a/Sources/macro.js
+++ b/Sources/macro.js
@@ -257,7 +257,7 @@ export function algo(publicAPI, model, numberOfInputs, numberOfOutputs) {
   }
 
   function getInputData(port = 0) {
-    if (!model.inputData[port] && model.inputConnection[port]) {
+    if (model.inputConnection[port]) {
       model.inputData[port] = model.inputConnection[port]();
     }
     return model.inputData[port];
@@ -310,9 +310,6 @@ export function algo(publicAPI, model, numberOfInputs, numberOfOutputs) {
     if (numberOfInputs) {
       let count = 0;
       while (count < numberOfInputs) {
-        if (model.inputConnection[count]) {
-          model.inputData[count] = null;
-        }
         ins[count] = publicAPI.getInputData(count);
         count++;
       }
--- a/Sources/macro.js
+++ b/Sources/macro.js
@@ -257,7 +257,7 @@ export function algo(publicAPI, model, numberOfInputs, numberOfOutputs) {
   }
 
   function getInputData(port = 0) {
-    if (!model.inputData[port]) {
+    if (!model.inputData[port] && model.inputConnection[port]) {
       model.inputData[port] = model.inputConnection[port]();
     }
     return model.inputData[port];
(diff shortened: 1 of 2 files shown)


jourdain:
@martinken you seems to have wrote those pipeline function. 

I think it has to do with the implementation of 

```
function getInputData(port = 0) {
    if (!model.inputData[port]) {
      model.inputData[port] = model.inputConnection[port]();
    }
    return model.inputData[port];
  }
```

which should probably be (so we can remove the push of null into the input data)

```
function getInputData(port = 0) {
    if (model.inputConnection[port]) {
       model.inputData[port] = model.inputConnection[port]();
    }
    return model.inputData[port];
  }
```

fogleman:
OK, I can make that change.

fogleman:
Please review my changes. `getInputData` will invoke the connection if it exists and if we don't already have data. `update` blows away the data so we get an update, but only if a connection exists.

Sources/macro.js
```diff
@@ -257,7 +257,7 @@ export function algo(publicAPI, model, numberOfInputs, numberOfOutputs) {
   }
 
   function getInputData(port = 0) {
-    if (!model.inputData[port]) {
+    if (!model.inputData[port] && model.inputConnection[port]) {
```
jourdain: It should be
```
if (model.inputConnection[port]) {
```

Otherwise you will still need to push that ```null```

jourdain: You also need to update the ```update()``` in macro to not push the ```null```

jourdain:
I've added some inline comments

Sources/macro.js
fogleman: It seemed like `getInputData` was trying to use cached data. You're saying it should always invoke the connection if one exists?

fogleman: (In that case, what is the purpose of `update`?)

jourdain: Yes, because update should try to get the new output of its input. When you have static data, you already have it. If you have a connection, you should ask for it (always).

If the pipeline didn't change, then the ```model.inputConnection[port]()``` is almost a NoOp() as it directly return its output. But if the filter figure out that it has some changes, then it will recompute itself in order to give you the proper output.

jourdain:
LGTM

fogleman:
There are two different `update` functions here: one in macro.js and one in the ImageMapper. Hope this isn't confusing our discussion.

Anyway, I made the change as you described... but it still seems to me that the original code for `getInputData` intended to return a cached value if it had one, and you would call `update` when you wanted to force a refresh.

jourdain:
```update()``` is a ```NoOp()``` when nothing as changed.

PR #11, opened 2016-10-12
Some rendering fixes

scottwittenburg:


jourdain:
Looks good to me but @martinken should look closer too...

Sources/Common/Core/ScalarsToColors/index.js
```diff
@@ -125,25 +125,26 @@ function vtkScalarsToColors(publicAPI, model) {
...
     for (let i = component; i < size; i += tuple) {
-      newColors[(count * 4)] = convtFun(values[i]);
-      newColors[(count * 4) + 1] = convtFun(values[i + 1]);
-      newColors[(count * 4) + 2] = convtFun(values[i + 2]);
-      newColors[(count * 4) + 3] = a;
+      newValues[(count * 4)] = convtFun(values[i]);
+      newValues[(count * 4) + 1] = convtFun(values[i + 1]);
+      newValues[(count * 4) + 2] = convtFun(values[i + 2]);
+      newValues[(count * 4) + 3] = a;
       count++;
```
martinken: Why operate on newValues as opposed to newColors? The method above is still using newColors so I am unsure what is behind the change.

martinken:
+2

Sources/Common/Core/ScalarsToColors/index.js
```diff
@@ -125,25 +125,26 @@ function vtkScalarsToColors(publicAPI, model) {
       newColors[count] = l;
       newColors[count + 1] = l;
       newColors[count + 2] = l;
-      newColors[count + 3] = convtFun(values[i + 1]);
+      newColors[count + 3] = convtFun(values[i + 1]) * alpha;
```
jourdain: @martinken do you agree with the logic of Alpha multiplication here? It seems that it was done for all other use case except this one. So we were wondering if it was missing or not.

martinken: I think it is correct. The code in VTK is
      outPtr[3] = static_cast<unsigned char>(a*alpha + 0.5);
which has the multiplication

Sources/Common/Core/ScalarsToColors/index.js
scottwittenburg: Good catch, I will fix the rest of those methods and push again.

PR #19, opened 2016-10-18
Better handling for points

scottwittenburg:


scottwittenburg:
@martinken Does the use of the 'p' key to set representation to points seem reasonable?  Since I was making sure point size was properly exported, it seemed like a natural thing to want to do in the viewer.

Sources/Rendering/Core/InteractorStyle/index.js
```diff
@@ -150,6 +150,21 @@ function vtkInteractorStyle(publicAPI, model) {
...
+          const ac = model.currentRenderer.getActors();
+          ac.forEach((anActor) => {
+            anActor.getProperty().setRepresentationToPoints();
+          });
+        } else {
+          vtkWarningMacro('no current renderer on the interactor style.');
+        }
+        rwi.render();
+        break;
+
```
scottwittenburg: Gotcha.  I amended that commit to use 'v' instead.

PR #23, opened 2016-11-06
fix can't remove temporary files for convert python script on windows

MaxPal:
On Windows filesystem a file should be closed before removed else system throws an error.


jourdain:
Thanks @MaxPal

PR #31, opened 2016-11-18
fix(Rendering): fix alpha blending and point lighting

martinken:
The blend function was not being set so translucent
objects appeared opaque.

When points were rendered in surface mode lighting was
being applied even if the points had no normals. Fixed
so that lighting is only applied on points with normals.


jourdain:
Working great, thanks Ken!

PR #64, opened 2016-12-21
Fix class hierarchy

jourdain:


jourdain:
@martinken make sure you follow that pattern next time.

aronhelser:
How about a test to illustrate the problem? :)

jourdain:
Good luck with that.

jourdain:
You will have to test each class individually.

For the multi `macro.obj` that does not create any issue just useless creation of methods that get garbage collected.

PR #69, opened 2017-01-03
fix(Rendering): Handle LookupTable Mtime in Mapper

martinken:
The mapper was not considering its LookupTable when
computing it mtime causing changes in the LookupTable
to be ignored.


jourdain:
Awesome, thanks!

martinken:
It is a needed fix, I'm not sure it will fix all of our issue. If it doesn't let me know and I'll look for the next step.

Thanks!

jourdain:
I've tried again on the example online, but the issue seems to be still here.

PR #73, opened 2017-01-06
fix(Rendering): the wcvc matrix was getting corrupted

jourdain:
It seems that on some systems the matrices passed to
webgl may be used after they are passed resulting in
potentially bad matrices being used.


jourdain:
Seems to fix it for me. Could not reproduced the issue.

PR #78, opened 2017-01-07
Doc proof

TristanWright:
If you'd like to read over the general changes they are spread across four commits in this branch.


jourdain:
This is great! Thanks for taking the time.

PR #81, opened 2017-01-13
feat(ImageStreamline): Add vtkImageStreamline filter

jourdain:
This allow computation of streamline within an ImageData


jourdain:
@berkgeveci  this is the compact one. You can merge it if you want.

Otherwise the full history one is here: https://github.com/Kitware/vtk-js/pull/80

But that second one does not have the proper commit message format to take benefit of the changelog.

jourdain:
Merging it... We can always fix things later...

PR #88, opened 2017-01-17
feat(IO): add texture coordinates to elevation reader

martinken:
Also fix a couple issues that turned up in OpenGL
with texture maps added after the first render.


jourdain:
Thanks @martinken , this is great...

PR #97, opened 2017-01-23
Molecular rendering pdb reader 2017.01.17

MayeulChassagnard:
Render molecules from .pdb files

For now only the atoms (with their radius, color and position) based on [elements.json](https://github.com/Kitware/vtk-js/blob/master/Utilities/XMLConverter/chemistry/elements.json) are displayed.

Next step will be to display the sticks (links between atoms).


Sources/IO/Misc/PDBReader/index.js
```diff
@@ -0,0 +1,257 @@
+import * as macro from '../../../macro';
+import vtkPolyData from '../../../Common/DataModel/PolyData';
+import vtkDataArray from '../../../Common/Core/DataArray';
+import DynamicTypedArray from '../../../Common/Core/DynamicTypedArray';
+
+import DataAccessHelper from '../../Core/DataAccessHelper';
+
+import atomElem from '../../../../Utilities/XMLConverter/chemistry/elements.json';
```
jourdain: You are not using the compact representation

Sources/IO/Misc/PDBReader/index.js
```diff
@@ -0,0 +1,257 @@
...
+
+
+// ----
+// Globals
+// ----
+
+const ATOMS = {};
+atomElem.atoms.forEach((a) => { ATOMS[a.id] = a; });
```
jourdain: You should use the compact representation.

Sources/Filters/General/MoleculeToRepresentation/index.js
```diff
@@ -0,0 +1,116 @@
+import * as macro     from '../../../macro';
+import vtkPolyData    from '../../../Common/DataModel/PolyData';
+import vtkDataArray   from '../../../Common/Core/DataArray';
+
+
```
jourdain: You should register that class into Sources/Filters/General/index.js

Sources/IO/Misc/PDBReader/index.js
```diff
@@ -0,0 +1,257 @@
...
+import DynamicTypedArray from '../../../Common/Core/DynamicTypedArray';
+
+import DataAccessHelper from '../../Core/DataAccessHelper';
+
+import atomElem from '../../../../Utilities/XMLConverter/chemistry/elements.json';
+
+
+// ----
+// Globals
+// ----
```
jourdain: use same // ------------

Sources/IO/Misc/PDBReader/index.js
```diff
@@ -0,0 +1,257 @@
...
+    const polydata = vtkPolyData.newInstance();
+
+    if (model.molecule) {
+      const jSize = model.molecule.length;
+
+      const points = polydata.getPoints();
+      const pointdata = polydata.getPointData();
+
+      // atom position
+      const pointValues = new DynamicTypedArray({ chunkSize: 65500, arrayType: 'Float32Array' }); // points.getData();
```
jourdain: Do you know the size? You may want to use plain JavaScript array and convert them after that.

Sources/IO/Misc/PDBReader/index.js
```diff
@@ -0,0 +1,257 @@
+import * as macro from '../../../macro';
```
jourdain: don't forget to register your new class into the index.js of the module directory.

jourdain:
Couple of comments but otherwise looks good.

MayeulChassagnard:
@jourdain Thanks for the review!

Sources/IO/Misc/PDBReader/index.js
jourdain: I'm talking about that one. https://github.com/Kitware/vtk-js/blob/master/Utilities/XMLConverter/chemistry-mapper/elements.json

MayeulChassagnard: Yes because we need finally more than just the `elementColor` and the `radiusCovalent`.
But I agree that we definitely don' t need the `nameOrigin` and other descriptive things.

jourdain: You can edit https://github.com/Kitware/vtk-js/blob/master/Utilities/XMLConverter/chemistry-mapper/post-process.js#L3

jourdain: And generate the data you need.

Sources/IO/Misc/PDBReader/index.js
MayeulChassagnard: We can know the size notwithstanding to do a loop in addition (Check the number of 'ATOMS' command and count all atoms which are not Hydrogen if we want to stick to the [PDBReader vtk version](https://github.com/Kitware/VTK/blob/master/IO/Geometry/vtkPDBReader.cxx#L91-L92))

jourdain: I was thinking you could simply use `pointValues = []`.
Then once the data is filled up:

vtkDataArray({ values: pointValues, dataType: 'Float32Array', ... })

Sources/IO/Misc/PDBReader/index.js
```diff
@@ -0,0 +1,253 @@
...
+  // return Busy state
+  publicAPI.isBusy = () => !!model.requestCount;
+}
+
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {
+  NumberOfAtoms: 0,
```
jourdain: should start with a lower case name

Sources/IO/Misc/index.js
```diff
@@ -0,0 +1,7 @@
+import ElevationReader    from './ElevationReader';
+import PDBReader          from './PDBReader';
+
+export default {
+  ElevationReader,
```
jourdain: the names should be prefixed by vtk.

```
import vtkElevationReader    from './ElevationReader';
import vtkPDBReader          from './PDBReader';
```

Sources/IO/Misc/PDBReader/index.js
```diff
@@ -0,0 +1,253 @@
...
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  // Build VTK API
+  macro.obj(publicAPI, model);
+  macro.get(publicAPI, model, [
+    'url',
+    'baseURL',
+    'NumberOfAtoms',
```
jourdain: numberOfAtoms

Sources/IO/Misc/PDBReader/test/testMolecule.js
```diff
@@ -0,0 +1,86 @@
...
+  // -----------------------------------------------------------
+
+  // create something to view it, in this case webgl
+  const glwindow = vtkOpenGLRenderWindow.newInstance();
+  glwindow.setContainer(renderWindowContainer);
+  renderWindow.addView(glwindow);
+  glwindow.setSize(400, 400);
+
+  // fetch caffeine.pdb file from Girder
+  reader.setUrl('https://data.kitware.com/api/v1/item/588652298d777f4f3f30849e/download').then(() => {
```
jourdain: Do you have a file in the repo? If so, and if you add a method `reader.parse('...file content...')`.

Then the test, can import that test pdb file. (We will have to add a loader so that file get processed as a text file)

Or we can see if we can define a proxy in the test runner using the local server.

Once merged, I can investigate and improve it.

MayeulChassagnard: Yes the file does exist at `vtk-js/Data/pdb/caffeine.pdb`.
Why do we need a parse method? I thought [setUrl](https://github.com/Kitware/vtk-js/pull/97/files#diff-0898b2950a2093dc15347d7b971be6d9R30) and [loadData](https://github.com/Kitware/vtk-js/pull/97/files#diff-0898b2950a2093dc15347d7b971be6d9R48) was doing this job to give the path no?

jourdain: Because I was thinking doing the code below

```
import caffeine from '.../pdb/caffeine.pdb';
[...]
reader.parse(caffeine);
[...]
```

This way we don't rely on a network call for testing.

jourdain:
don't know

MayeulChassagnard:
the commit https://github.com/Kitware/vtk-js/pull/97/commits/073403c5effae8236f14ecfbee239b4900247097 appeared with a misrebase from my side... 

Do we need to redo the merge? I've pushed (properly this time) [here](https://github.com/MayeulChassagnard/vtk-js/commits/molecular-rendering-pdbReader-2017.01.17)

Sorry for my mistake.

jourdain:
do a diff between your local branch and master after a fetch to be sure there is nothing.

Then push the data, since you forget it.

MayeulChassagnard:
I did:
```
$ git fetch upstream
$ git diff upstream/master origin/molecular-rendering-pdbReader-2017.01.17
$ git diff upstream/master molecular-rendering-pdbReader-2017.01.17
```
Both show nothing, so does that mean I should create another PR based on [this](https://github.com/MayeulChassagnard/vtk-js/commits/molecular-rendering-pdbReader-2017.01.17) branch?

MayeulChassagnard:
Because on [vtk-js/master](https://github.com/Kitware/vtk-js/commits/master) we can see this commit twice...
log-problem

jourdain:
Don't worry about that. But create a new branch (from master) to push the missing data so the example could actually work.

PR #103, opened 2017-01-27
More clean up

jourdain:


jourdain:
@martinken this is probably what you need.

PR #105, opened 2017-01-27
feat(Rendering): add hardware selector and framebuffers

martinken:
This topic adds a hardwareslector class that can be used
to perform hardware picks and return the prop and propid.
In the future we can extend it to support the other passes
such as composite id and attribute id.

To support this we added a SelectionNode class that contains
information and properties about a selection.

Since webgl doesn't really have an option to turn of bufferswapping
we add a framebuffer class so that the hardware picking can take
place in the framebuffer instead of the visible buffer. The
framebuffer class will of course be useful for a lot of other
cases.

Fixed a bug where is a class initialized mtime to zero it
would not be honored and would instead use the global mtime.

### DIFF
--- a/Sources/macro.js
+++ b/Sources/macro.js
@@ -31,7 +31,7 @@ export function getStateArrayMapFunc(item) {
 
 export function obj(publicAPI = {}, model = {}) {
   const callbacks = [];
-  model.mtime = model.mtime || ++globalMTime;
+  model.mtime = (Number.isInteger(model.mtime) ? model.mtime : ++globalMTime);
   model.classHierarchy = ['vtkObject'];
 
   function off(index) {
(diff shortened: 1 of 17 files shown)


martinken:
@jourdain   I changed a line in macro.js please look at that in case I screwed it up. I was trying to initialize mtime to zero and it was not working.

Sources/macro.js
```diff
@@ -31,7 +31,7 @@ export function getStateArrayMapFunc(item) {
 
 export function obj(publicAPI = {}, model = {}) {
   const callbacks = [];
-  model.mtime = model.mtime || ++globalMTime;
+  model.mtime = (model.hasOwnProperty('mtime') ? model.mtime : ++globalMTime);
```
jourdain: I would do it like that instead

model.mtime = Number.isInteger(model.mtime) ? model.mtime : ++globalMTime;

jourdain:
Small fix in macro, otherwise it looks good to me

PR #111, opened 2017-02-03
Molecular rendering - Add vtkMolecule DataModel - MultipleBonds display - Remove elem from display

MayeulChassagnard:


Sources/IO/Misc/PDBReader/example/index.js
```diff
@@ -26,6 +26,7 @@ const sphereActor = vtkActor.newInstance();
 const stickActor = vtkActor.newInstance();
 
 filter.setInputConnection(reader.getOutputPort());
+filter.setHideElement('H');
```
jourdain: Does setHideElement() expect an array? Or does it convert the type internally?

jourdain:
Could it be interesting to have some generic methods inside vtkMolecule? Just wondering.

Sources/IO/Misc/PDBReader/example/index.js
MayeulChassagnard: it does expect an array of string which will be compared to the elements.json file to hide them

MayeulChassagnard:
@jourdain  What generic methods do you think about?

Sources/IO/Misc/PDBReader/example/index.js
jourdain: is that what's happening?

```
model.hideElement = [];
model.hideElement = 'H'; // after filter.setHideElement('H')
```

jourdain: maybe, it should be names `hideElements` and when you should call `setHideElements(['H'])` or `setHideElements('H', 'C', 'O')`.

But you need to use the proper macro for the `HideElements`.

Sources/Filters/General/MoleculeToRepresentation/index.js
```diff
@@ -190,6 +300,7 @@ export function extend(publicAPI, model, initialValues = {}) {
   macro.setGet(publicAPI, model, [
     'sphereScaleArrayName',
     'radiusType',
+    'hideElement',
```
jourdain: this is wrong if you don't do `setHideElement([...])`

jourdain:
I don't know, do you use something in your representation. 
Something to deal with some nesting aspect of the data structure.
It is possible that you don't need anything.

Sources/IO/Misc/PDBReader/example/index.js
MayeulChassagnard: mmh well, yes I guess you right, but it seems working.
The code which check `model.hideElement` is: 
```
if (model.hideElement.includes(ATOMS[atomicNumber[i]].id)) {
  continue;
}
```
And it does work with all following settings
```
filter.setHideElement('H');
filter.setHideElement(['H']);
filter.setHideElement(['H', 'O']);
```

Sources/Filters/General/MoleculeToRepresentation/index.js
jourdain: You should use

https://github.com/Kitware/vtk-js/blob/master/Sources/macro.js#L283-L316

But you may have to fix/improve it so you could provide a size of -1 in case you don't know it so it does not throw an exception.

Sources/IO/Misc/PDBReader/example/index.js
jourdain: Yes because `'H'[0] === 'H'`

jourdain: But it won't with 'Sn'

MayeulChassagnard: Good call indeed! I'm going to work on the version that you advise providing multiple string separated by a coma: `filter.setHideElements('H', 'O');` which makes more sense indeed

jourdain: The array macro is doing exactly that. So you just need to manage the `size` argument since you don't know how many you will have.

jourdain: Maybe, if `size == -1`, the macro could add `addXXX`, and `removeXXX` methods.

Sources/IO/Misc/PDBReader/index.js
```diff
@@ -238,7 +190,6 @@ export function extend(publicAPI, model, initialValues = {}) {
     'xSpacing',
```
jourdain: Do you use `xSpacing`, `ySpacing`, `zScaling`?

MayeulChassagnard: yes [here](https://github.com/Kitware/vtk-js/pull/111/files/eb548f89e24ec17c8ee4a072bd58c6cbee7dfd30#diff-0898b2950a2093dc15347d7b971be6d9R106) but actually we might remove them, not quite useful for now indeed

jourdain: Ok I thought they were coming from the copy paste of the elevation reader. 
To me it looks like scaling. Is it useful to have asymmetric scaling across axis?

MayeulChassagnard: Yes at the beginning it was copied from elevator reader and I thought it was maybe an important parameter but maybe not finally. We should ask @cryos.

cryos: I can't see a use for this in the molecular data reading/visualization.

jourdain:
@MayeulChassagnard you should rebase on master and cleanup your import if not absolute.

jourdain:
Should we merge it?

martinken:
Yes

jourdain:
Ok I'll merge it after my next push since I have only fixes...

PR #117, opened 2017-02-09
Use absolute import

jourdain:


Sources/Filters/General/Calculator/index.js
```diff
@@ -216,7 +208,7 @@ export function extend(publicAPI, model, initialValues = {}) {
 
 // ----------------------------------------------------------------------------
 
-export const newInstance = macro.newInstance(extend, 'vtkWarpScalar');
+export const newInstance = macro.newInstance(extend, 'vtkCalculator');
```
jourdain: FYI @vibraphone

jourdain:
FYI @martinken @MayeulChassagnard @fogleman @wschroed @vibraphone @scottwittenburg @aronhelser 

On top of fixing the import to be absolute, I fix some common mistake (me including) like:

1. Forgetting the registration of the constructor in the vtk factory by providing its name here:

```
export const newInstance = macro.newInstance(extend, 'vtkClassName__HERE');
``` 

2. Forgetting to register your class/module into and [index.js](https://github.com/Kitware/vtk-js/blob/c00d897b2c7c6fb1f0a563f62449db26aac3f297/Sources/Interaction/Manipulators/index.js)

Sources/Filters/General/Calculator/index.js
vibraphone: Oops, you have discovered my secret method for developing a new vtkjs object.

jourdain: Don't worry I do the same thing. I also fixed some comments too, but at least the comments don't break anything... ;-)

PR #118, opened 2017-02-09
fix(DynamicFloat32Array): Try to provide replacement for DynamicTypedArray

jourdain:


jourdain:
@martinken do you think it will better than the current code?

jourdain:
The test seems to pass but I'm not sure that `colorData` is for sure an UInt8Array?

jourdain:
@martinken , I think this looks good and was planning to merge it. Do you agree?

martinken:
Yes colorData should always be Uint8

jourdain:
perfect thanks!

PR #130, opened 2017-02-22
Fix function vtkRotate used inside JacobiN function

laurennlam:
The vtkRotate fnction takes 5 arguments. In VTK, the first arguments is set inside the function. Actually, it's the first attribute of the JacobiN function which is set. So the first argument of thevtkRotate function is never set.


jourdain:
@laurennlam you should use `npm run commit` to format your commit message so the appropriate change log and version will be generated. In that case, since you did not use the format, no new version will be generated.

PR #132, opened 2017-02-22
test(ColorTransferFunction): Add test for presets and correct color i…

jourdain:
…nterpolation


jourdain:
@martinken can you take over? I've commented the test but the code I added in ColorTransferFunction need to be fix and then the test should be enabled by default with the correct baseline.

jourdain:
Now it is truly all yours @martinken since I don't see what I'm missing...

PR #133, opened 2017-02-22
fix(Rendering): minor fix to ctfun plus onload for textures

jourdain:
Textures should handle incomplete images now
Two bad code segments in color transfer function


jourdain:
@martinken I've fixed what I could but the textures does not seems to load on the opengl side of things.

jourdain:
Probably a mtime issue in code invalidation

PR #135, opened 2017-02-23
fix(Rendering): fixes to color transfer function

martinken:
A couple bugs in the clor transfer function
when applying an existing table


martinken:
@jourdain  I'm pretty sure there is still an issue in the applyColorMap code as the LAB results look funny. But this fixes two issues with it that were problems.  I also changed the default unspecified size of a ctfun to be 4094 so that when the above and below are added it is 4096, a power of two.

PR #139, opened 2017-02-24
Add cellArray module to vtkjs (after compilation)

laurennlam:
When vtkjs is compiled, vtkCellArray wasn't added.


jourdain:
Could you just push the cellArray fix as the way you wrote the revert, it won't be taken in the change log and it will think there is a new feature.

jourdain:
Otherwise good catch for that.

jourdain:
Did it within on of my other PR...

PR #144, opened 2017-02-24
fix(Rendering): sort operation of color transfer function

martinken:
nodes were being sorted in an incorrect way


jourdain:
nice

PR #152, opened 2017-03-02
Remove console usage

jourdain:


jourdain:
@martinken do you mind testing it on edge?

Sources/macro.js
```diff
@@ -7,6 +7,52 @@ export function getCurrentGlobalMTime() {
...
+// Loggins function calls
+// ----------------------------------------------------------------------------
+
+const fakeConsole = {};
+function noOp() {}
+
+const consoleMethods = ['log', 'debug', 'info', 'warn', 'error', 'time', 'timeEnd', 'group', 'groupEnd'];
+consoleMethods.forEach(methodName => (fakeConsole[methodName] = noOp));
+
+global.console = window.console || fakeConsole;
```
aronhelser: I feel like we need to do this for ParaViewWeb (and ERS) - is there a good place?

jourdain: Agree but don't know where yet...

jourdain: Maybe in Common/Core/Console

jourdain: Then we just do a `import 'paraviewweb/src/Common/Core/Console' somewhere...

aronhelser: I don't get a crash on Edge in the ERS project when doing console.log - maybe babel or something is already doing a polyfill for us?

aronhelser: Yup, seems to be happening in autobahn\lib\polyfill\console.js

jourdain: Good to know but we should have our own as we don't always have autobahn.

PR #162, opened 2017-03-07
fix(landmarkTransform): Fix the call of constants that doesn't exist

laurennlam:
Bad call of one of a mode

### DIFF
--- a/Sources/Common/Transform/LandmarkTransform/index.js
+++ b/Sources/Common/Transform/LandmarkTransform/index.js
@@ -71,8 +71,10 @@ function vtkLandmarkTransform(publicAPI, model) {
 
     // -- build the 3x3 matrix M --
 
-    const M = mat3.fromValues(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0);
-    const AAT = mat3.fromValues(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0);
+    const M = mat3.create();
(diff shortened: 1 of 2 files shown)


jourdain:
Let's wait for your fix on the exposition of constants. Just keep pushing to the same topic. There is no need to create a new version each time unless you think it is worth it.

laurennlam:
The merge can wait. I'll add the feature you told me. But it seems to have some issues with glmatrix. I'll push all commits at the same time when all will be fixed. Thanks !

jourdain:
You can push everything as separate commits. We will just merge when everything is ready.

laurennlam:
@jourdain , I just discovered that the version of gl-matrix I used in vtkLandmarkTransform wasn't the same as the current one in package.json. May I change it from 2.3.1 to 2.3.2 ? Or would you prefer to me to update the code to fit with 2.3.1 ? Thanks.

jourdain:
We carefully picked that version for some reason. So you should use that version otherwise you may break some other code.

laurennlam:
ok. Thanks ! I'll update the code.

laurennlam:
All seems to be ok.

jourdain:
Do you know that because you made a feat(), you will move vtk version from 2.12.0 to 2.13.0

laurennlam:
No I didn't know...
I'll make a new pull request.

jourdain:
Not that it matters, but I wanted you to be aware... Look at https://github.com/semantic-release/semantic-release for more understanding.

PR #163, opened 2017-03-07
Constant issues landmark transform

laurennlam:
Add some fixes for constant and the using of gl-matrix


jourdain:
LGTM, feel free to merge yourself? (unless you don't have the rights?)

laurennlam:
I think I don't have right. But for now, i'd like to not have it.
It's the first time I contributed and as you can see, I'm not completely aware of how it works...

PR #166, opened 2017-03-09
feat(Rendering): Add support for volume rendering

martinken:
This topic has a number of changes. The largest is that
it adds support for Volume rendering of single component
data with opacity and color transfer functions. The data
can be of any native type and it will be encoded into
multiple unsigned char bytes per value if needed to
preserve the resolution of the original data. If the
original volume is too large to fit in one texture then
it will be automatically subsampled in X and Y so that
it fits.

The volume rendering supports settnig both sample
distance and image sample distance so that you can
control rendering rates and performance. When
image sample distance is not unity a framebuffer
is used for rendering and then stretched into the
resulting window.

Volume rendering includes support for nearest
neightbor, fast linear, and linear interpolation.
Fast linear interpolates in X and Y but not Z.

Future feature may include lighting, opacity gradient
functions, intermixed opaque geometry, cropping planes,
and multicomponent input data.

This topic includes a fix for animations driven by
mouse movements. In the past mouse movements would
queue up rendering irregardless of the performance
of the graphcis card. This could cause sluggishness
and lurching behavior. The new code makes use of the
requestAnimationFrame functionality to let the browser
drive the rendering rate. Due to this change the
interactors will no longer respond to mouse move events
but rather animation events. Upon a mouse down they will
request aniation frames, respond to them, and upon mouse
up, cancel them.

A variety of minor bug fixes are included as well.


martinken:
@jourdain  FYI

 I removed theeslint CR/LF line ending check as npm run example always complained making it hard to find real errors. 

Other than that a lot of changes so I'll wait for your +2

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -519,6 +519,9 @@ export function extend(publicAPI, model, initialValues = {}) {
   // Build VTK API
   macro.obj(publicAPI, model); // FIXME parent is not vtkObject
   macro.algo(publicAPI, model, 1, 0);
+
+  model.bounds = [1, -1, 1, -1, 1, -1];
```
jourdain: No check?

Sources/Rendering/Core/InteractorStyle/index.js
```diff
@@ -173,20 +174,36 @@ function vtkInteractorStyle(publicAPI, model) {
...
+  };
+
   publicAPI.startState = (state) => {
     model.state = state;
     if (model.animationState === States.IS_ANIM_OFF) {
-      const rwi = model.interactor;
-      rwi.getRenderWindow().setDesiredUpdateRate(rwi.getDesiredUpdateRate());
-      model.invokeStartInteractionEvent({ type: 'StartInteractionEvent' });
+      // const rwi = model.interactor;
+      // rwi.getRenderWindow().setDesiredUpdateRate(rwi.getDesiredUpdateRate());
```
jourdain: Should we remove those commented lines?

Sources/Rendering/Core/Camera/index.js
```diff
@@ -426,6 +426,38 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+  if (model.viewPlaneNormal === null) {
+    model.viewPlaneNormal = [0, 0, 1];
+  }
+  if (model.screenBottomLeft === null) {
+    model.screenBottomLeft = [-0.5, -0.5, -0.5];
+  }
+  if (model.screenBottomRight === null) {
+    model.screenBottomRight = [0.5, -0.5, -0.5];
+  }
+  if (model.screenTopRight === null) {
```
jourdain: instead of checking for exactly null, I'm wondering if it would be better to check more loosely.

```
if (!model.screenTopRight) {
```

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -284,7 +284,7 @@ function vtkRenderer(publicAPI, model) {
 
     if (nothingVisible) {
       vtkMath.uninitializeBounds(allBounds);
-      vtkDebugMacro('Can\'t compute bounds, no 3D props are visible');
+     // vtkDebugMacro('Can\'t compute bounds, no 3D props are visible');
```
jourdain: Do we want to keep them? Maybe we can remove the logging of debug by default instead?

Sources/Common/DataModel/ImageData/index.js
```diff
@@ -173,9 +173,9 @@ function vtkImageData(publicAPI, model) {
 // ----------------------------------------------------------------------------
 
 const DEFAULT_VALUES = {
-  spacing: [1.0, 1.0, 1.0],
-  origin: [0.0, 0.0, 0.0],
-  extent: [0, -1, 0, -1, 0, -1],
+  spacing: null,
+  origin: null,
+  extent: null,
   dataDescription: StructuredType.EMPTY,
```
jourdain: instead of setting them to null, I would comment them.

Sources/Rendering/Core/InteractorStyle/index.js
martinken: no as I am planning to implement interactive render rate support in the future

Sources/Rendering/Core/Mapper/index.js
martinken: Yeh, I'll add a check

Sources/Rendering/Core/Camera/index.js
martinken: By initializing it to null a check against null means it has not been set. I'm not sure what leaving it uninitialized and then testing via ! does.  Put another way, I'm pretty sure the way I am doing it works and I am not sure other ways would work. Maybe they would. I just am not that smart.

Sources/Rendering/Core/Renderer/index.js
martinken: Not sure why I commented those out

Sources/Rendering/Core/Camera/index.js
jourdain: I'll suggest some other way and we can see which one we prefer. But I kind of don't like the current one as it is very verbose.

Sources/Rendering/Core/InteractorStyle/index.js
```diff
@@ -173,20 +174,36 @@ function vtkInteractorStyle(publicAPI, model) {
...
+      publicAPI.invokeStartInteractionEvent({ type: 'StartInteractionEvent' });
     }
   };
 
   publicAPI.stopState = () => {
     model.state = States.IS_NONE;
     if (model.animationState === States.IS_ANIM_OFF) {
       const rwi = model.interactor;
-      rwi.getRenderWindow().setDesiredUpdateRate(rwi.getStillUpdateRate());
+      // rwi.getRenderWindow().setDesiredUpdateRate(rwi.getStillUpdateRate());
```
jourdain: should we remove that?

jourdain:
@martinken can you give me your thought on https://github.com/Kitware/vtk-js/pull/166/commits/14044d6144ed56d28f2ca8f21c182a860661d132

martinken:
I like it as long as it works (specifically will not get screwed up on Objects/Maps etc.).  I'm not sure what answers yes to isArray()

jourdain:
```
Array.isArray(new Float32Array())
false
Array.isArray([])
true
Array.isArray({})
false
```

jourdain:
The test works... ;-)

jourdain:
Beautiful +10

PR #171, opened 2017-03-10
perf(Rendering): add high res render as needed, multitouch

martinken:
This topic adds a final high res render after animation had ended.
It also uses the animation logic for multitouch movements.


jourdain:
Since you are in the event code section, do you want to handle the zoom by "scroll"?
I've added some code to capture those events, they just need to trigger the proper events, I'm guessing...

martinken:
OK added wheelies

jourdain:
thanks

jourdain:
This is SO BEAUTIFUL @martinken !

Works perfectly on my iPod... The only issue is that on my laptop I can not trigger the LOD, I'm always full resolution at full speed... ;-)

PR #178, opened 2017-03-22
Add create cube from raw in open gl texture

laurennlam:


jourdain:
@laurennlam is ready to merge or is it WIP? @martinken should be the one to review since it is on the GL side...

laurennlam:
It should be ready to merge.

martinken:
+2  looks good to me

at some  point it would be nice to have a test/example that uses a cube map texture.

laurennlam:
@martinken I keep in mind the test example. Thanks !

PR #179, opened 2017-03-23
feat(cubeSource): Add vtkCubeSource

luciemac:


Sources/Filters/Sources/CubeSource/index.js
```diff
@@ -0,0 +1,243 @@
...
+    polys[polyIndex++] = 4;
+    polys[polyIndex++] = 20;
+    polys[polyIndex++] = 21;
+    polys[polyIndex++] = 23;
+    polys[polyIndex] = 22;
+  }
+
+  publicAPI.setBounds = (xMin, xMax, yMin, yMax, zMin, zMax) => {
+    const bounds = [xMin, xMax, yMin, yMax, zMin, zMax];
+    model.setBounds(bounds);
```
jourdain: setBounds() should be called on publicAPI not on model.

jourdain:
Minor fix, but looks good otherwise. A test could be nice too.

Sources/Filters/Sources/CubeSource/index.js
luciemac: Actually, after double checking, this function cannot be reached. Only the second function setBounds(bounds) is. 
I think it is no longer necessary to keep it.

luciemac:
I will a test!

Sources/Filters/Sources/CubeSource/index.js
jourdain: You right, only one function can exist. But you can make it so both profiles can exist.

```
publicAPI.setBounds = (...bounds) => {
   let boundsArray = bounds;
   if (Arrays.isArray(bounds[0])) {
      boundsArray = bounds[0];
   }
   [...]
}
```

Sources/Filters/Sources/CubeSource/index.js
```diff
@@ -0,0 +1,249 @@
...
+
+    polys[polyIndex++] = 4;
+    polys[polyIndex++] = 20;
+    polys[polyIndex++] = 21;
+    polys[polyIndex++] = 23;
+    polys[polyIndex] = 22;
+  }
+
+  publicAPI.setBounds = (...bounds) => {
+    let boundsArray = [];
```
jourdain: Could be written like below which prevent the else case.

```
let boundsArray = [].concat(bounds);
```

PR #182, opened 2017-04-03
fix(Rendering): fix bugs in new texture resize code

martinken:
A few critical bugs in newly added texture code.


jourdain:
@laurennlam before requesting a merge, you should run `npm t` to trigger WebGL testing on your local machine as Travis does not have WebGL support and will only test the other part. You last merge broke volume rendering...

laurennlam:
Sorry for the trouble. I didn't know... I won't forget for the next time.

PR #195, opened 2017-04-12
Add shader replacement properties

luciemac:


martinken:
Love the tests and like the approach. 

My only thought is I think we should move the OpenGL code to the OpenGL/PolyDataMapper/api.md file instead of in the Core/Mapper/index itself.  Keep model.viewSpecificProperties/set/get as that is generic. 

So Core/Mapper/api we add text that says

if you want to provide specific properties for rendering engines you can use viewSpecificProperties etc  For a list of view specific properties see the underlying mapper for the rendering engine you are using. For example, for OpenGL/WebGL see OpenGL/PolyDataMapper/api.md

Then in the docs for OpenGLPolyDataMapper we list all the OpenGL extra view properties with examples

mapper.getViewSpecificProperties()['OpenGL'] = {
  vertexShader: ....
  shaderReplacements: [
    {foo:
     bar: }, 
   {foo:
    bar: }]
....
}

I would skip the trivial set/get methods.  No need for them. Only the add/clear have some meat to them, so maybe stick them in the documentation as examples of helper methods.

I'll mark up the code a bit to help you know what I'm talking about.

Sources/Rendering/Core/Mapper/api.md
```diff
@@ -359,3 +359,41 @@ Provide read access to the color texture coordinate array
 
 Provide read access to the color texture array
 
+### viewSpecificProperties
+
+Object that contains specific properties to be able to apply
+your own substitutions to the shader creation process.
+You can apply your own string replacements to add features you desire.
+
+PolyDataMapper is the only object that uses it.
```
martinken: I would add test here to say, go look in the rendering backend of your choice for details on specific properties.  For OpenGL/WebGL see ...

Sources/Rendering/Core/Mapper/api.md
```diff
@@ -359,3 +359,41 @@ Provide read access to the color texture coordinate array
...
+- replacementValue: string: replacement value
+- replaceAll: boolean: defines if need to replace only the first occurrence (false)
+                       or all of them (true)
+
+To remove a property on ShaderReplacements:
+mapper.getViewModelSpecificProperties().clearShaderReplacement(
+  shaderType: ...,
+  originalValue: ...,
+  replaceFirst: ...
+);
```
martinken: Then move all this to OpenGL/PolyDataMapper/api.md with a few changes

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+      }
+      if (indexToRemove > -1) {
+        shaderReplacement.splice(indexToRemove, 1);
+      }
+    },
+
+    getShaderReplacements() {
+      return this.OpenGL.ShaderReplacements;
+    },
+
```
martinken: Yank these little guys

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
   ], 2);
 
+  model.viewSpecificProperties = {
+    OpenGL: {
+      ShaderReplacements: [],
+      VertexShaderCode: '',
+      FragmentShaderCode: '',
+      GeometryShaderCode: '',
+    },
+
```
martinken: Move the OpenGL stuff to OpenGL/PDM/api.md as documentation

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+    addShaderReplacement(_shaderType, _originalValue, _replaceFirst, _replacementValue, _replaceAll) {
+      this.OpenGL.ShaderReplacements.push({
+        shaderType: _shaderType,
+        originalValue: _originalValue,
+        replaceFirst: _replaceFirst,
+        replacementValue: _replacementValue,
+        replaceAll: _replaceAll,
+      });
+    },
+
```
martinken: In the docs you could show adding this function as a property on the mapper. That way your examples can continue to use it.

luciemac: What do you mean by "as a property on the mapper"?
Do you mean to add this function as a method of the viewSpecificProperty object?

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+  model.viewSpecificProperties = {
+    OpenGL: {
+      ShaderReplacements: [],
+      VertexShaderCode: '',
+      FragmentShaderCode: '',
+      GeometryShaderCode: '',
+    },
+
+    addShaderReplacement(_shaderType, _originalValue, _replaceFirst, _replacementValue, _replaceAll) {
+      this.OpenGL.ShaderReplacements.push({
```
jourdain: The should be no `this` in vtk-js.

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+        shaderType: _shaderType,
+        originalValue: _originalValue,
+        replaceFirst: _replaceFirst,
+        replacementValue: _replacementValue,
+        replaceAll: _replaceAll,
+      });
+    },
+
+    clearShaderReplacement(_shaderType, _originalValue, _replaceFirst) {
+      const shaderReplacement = this.OpenGL.ShaderReplacements;
```
jourdain: no `this`

Sources/Rendering/Core/Mapper/index.js
martinken: I did, but you could probably just make it  a function in the example and not stick it in the properties. That is probably easier/cleaner.

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+          break;
+        }
+      }
+      if (indexToRemove > -1) {
+        shaderReplacement.splice(indexToRemove, 1);
+      }
+    },
+
+    getShaderReplacements() {
+      return this.OpenGL.ShaderReplacements;
```
jourdain: no `this`

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+    },
+
+    setVertexShaderCode(vertexShaderCode) {
+      this.OpenGL.VertexShaderCode = vertexShaderCode;
```
jourdain: no `this`

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+    },
+
+    getVertexShaderCode() {
+      return this.OpenGL.VertexShaderCode;
```
jourdain: no `this`

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+    },
+
+    setGeometryShaderCode(geometryShaderCode) {
+      this.OpenGL.GeometryShaderCode = geometryShaderCode;
```
jourdain: no `this`

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+    },
+
+    getGeometryShaderCode() {
+      return this.OpenGL.GeometryShaderCode;
```
jourdain: no `this`

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+    },
+
+    setFragmentShaderCode(fragmenShaderCode) {
+      this.OpenGL.FragmentShaderCode = fragmenShaderCode;
```
jourdain: no `this`

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -545,6 +552,70 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+    },
+
+    getFragmentShaderCode() {
+      return this.OpenGL.FragmentShaderCode;
```
jourdain: no `this`

martinken: The first eight comments you made left me confused Seb, I was like, "what is he trying to say?"  But finally on this one I got it ;-) You don't want us to use a grey background on our text.

jourdain: Exactly! But most importantly vtk-js should not use `this` as it can be undefined if we use method outside of their context.

jourdain: Basically I don't want `return (undefined).OpenGL.FragmentShaderCode;` to happen.

martinken:
Yup that's the idea. LGTM +2

jourdain:
Do we want to squash all the commits into one? Or should I merge it like that?

@martinken @luciemac ?

Issue #196, opened 2017-04-12
Rendering does not update when calling vtkTexture.setImage() with different image


jourdain:
The old image texture seems to stick around for ever.

PR #197, opened 2017-04-12
Updated texture

martinken:


martinken:
addresses https://github.com/Kitware/vtk-js/issues/196

PR #203, opened 2017-04-21
fix(OpenGLMapper): Add some test to avoid error when several mapper a…

laurennlam:
Add some test to avoid error when several mapper are instanciated

### DIFF
--- a/Sources/Rendering/OpenGL/PolyDataMapper/index.js
+++ b/Sources/Rendering/OpenGL/PolyDataMapper/index.js
@@ -78,11 +78,11 @@ export function vtkOpenGLPolyDataMapper(publicAPI, model) {
     // user specified pre replacements
     const openGLSpec = model.renderable.getViewSpecificProperties().OpenGL;
     let shaderReplacements = null;
-    if (openGLSpec !== undefined) {
+    if (openGLSpec) {
       shaderReplacements = openGLSpec.ShaderReplacements;
     }
 
-    if (shaderReplacements !== null) {
+    if (shaderReplacements) {
       for (let i = 0; i < shaderReplacements.length; i++) {
         const currReplacement = shaderReplacements[i];
         if (currReplacement.replaceFirst) {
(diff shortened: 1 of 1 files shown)


Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -78,11 +78,11 @@ export function vtkOpenGLPolyDataMapper(publicAPI, model) {
     // user specified pre replacements
     const openGLSpec = model.renderable.getViewSpecificProperties().OpenGL;
     let shaderReplacements = null;
-    if (openGLSpec !== undefined) {
+    if (openGLSpec !== undefined && openGLSpec !== null) {
```
jourdain: Why not do `if (openGLSpec)` instead?

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -78,11 +78,11 @@ export function vtkOpenGLPolyDataMapper(publicAPI, model) {
...
       shaderReplacements = openGLSpec.ShaderReplacements;
     }
 
-    if (shaderReplacements !== null) {
+    if (shaderReplacements !== undefined && shaderReplacements !== null) {
```
jourdain: Why not do `if (shaderReplacements)` instead?

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -97,7 +97,7 @@ export function vtkOpenGLPolyDataMapper(publicAPI, model) {
     publicAPI.replaceShaderValues(shaders, ren, actor);
 
     // user specified post replacements
-    if (shaderReplacements !== null) {
+    if (shaderReplacements !== undefined && shaderReplacements !== null) {
```
jourdain: Why not do `if (shaderReplacements)` instead?

laurennlam:
I'll change it. I didn't know that `if(blabla)` checked both 'undefined' and 'null'

aronhelser:
Well, it checks for any 'falsey' value, which includes null and undefined, but also includes 0 and '' (empty string). As long as you are sure your variable contains either an object or undefined/null, then if (blabla) is a good check. If it's a string or number, you have to watch out for '' and 0.

laurennlam:
Thanks for the explanations @aronhelser !

PR #204, opened 2017-04-21
feat(SynchronizableRenderWindow): Provide a new type of vtkRenderWindow

scottwittenburg:
The vtkSynchronizableRenderWindow is designed to be kept in sync with a ParaView/VTK view on the server side, or even with some other source of state.


scottwittenburg:
@jourdain I ran the tests locally and they all passed.

PR #208, opened 2017-04-24
feat(OpenGLTexture)Add view properties to texture

laurennlam:


Sources/Rendering/Core/Texture/index.js
```diff
@@ -69,6 +72,8 @@ export function extend(publicAPI, model, initialValues = {}) {
     'image',
   ]);
 
+  model.viewSpecificProperties = {};
```
jourdain: Don't you check first if that `viewSpecificProperties` was given to you in the constructor? Otherwise here, you will always reset it.

If would do that instead

```
if (!model.viewSpecificProperties) {
 viewSpecificProperties = {};
}
```

Sources/Rendering/Core/Texture/index.js
```diff
@@ -35,6 +35,8 @@ function vtkTexture(publicAPI, model) {
 
     publicAPI.modified();
   };
+
+  publicAPI.getViewSpecificProperties = (() => model.viewSpecificProperties);
```
jourdain: You can use the macro to create that getter.

jourdain:
I'm guessing you ran all the tests locally?

laurennlam:
I ran `npm t` from vtkjs repository. Does it mean it's launched locally ?

jourdain:
yes

laurennlam:
Is there an other way to launch test ? Did I do anything wrong ?

jourdain:
No, but you could have ran only yours without checking the previous ones... And since Travis can not check the GL test, I wanted to make sure you ran all the tests with success on your computer before merging your changes. ;-)

martinken:
This should not have been merged.

jourdain:
Oups why?

jourdain:
I can revert

Sources/Rendering/Core/Texture/index.js
jourdain: It should. We may need to validate the `executive` logic to make sure it does not break on a null when trying to update its input.

martinken: That is how it should be done.

martinken:
I'll fix it - Ken

PR #211, opened 2017-04-25
fix(Rendering): improve cube map support

martinken:
Added an option for CubeSource to generate cube map texture
coordinates.

Modified PolyDataMapper to handle cube textures and 3d
texture coordinates.


martinken:
@laurennlam  Here are the changes I made to add support for cubeMaps as a standard feature.  From the textures point of view it is just a texture that has six inputs. OpenGL specific code is all in the OpenGL classes.

Probably at some point need to update the setImage on texture to take six images as well to keep it consistent.

laurennlam:
@martinken It looks clearer than what I've done. Thanks

PR #212, opened 2017-04-25
fix(Rendering): fix issue with image mapper

martinken:
image mapper would not be rendered if it
was the only object in view as it was not
responding to the query pass


zachmullen:
I just confirmed that this fixes the rendering issue. 👍

martinken:
Thanks for testing it Zach!

PR #217, opened 2017-04-26
fix(Rendering): another case of using mipmaps on greyscale

martinken:
Can only use mipmaps on RGBA


MayeulChassagnard:
Works perfectly :+1:

PR #222, opened 2017-05-02
fix(SynchronizableRenderWindow): Fix infinitely deep recursion

scottwittenburg:


Sources/Rendering/Misc/SynchronizableRenderWindow/index.js
```diff
@@ -225,9 +225,11 @@ export function vtkSynchronizableRenderWindow(publicAPI, model) {
 
   const addOn = createSyncFunction(publicAPI, model.synchronizerContext);
 
-  Object.keys(addOn).forEach((methodName) => {
-    publicAPI[methodName] = addOn[methodName];
-  });
+  Object.keys(addOn)
+    .filter(methodName => methodName.indexOf('SynchronizedViewId') < 0)
```
scottwittenburg: @jourdain  If we add the methods `getSynchronizedViewId` and `setSynchronizedViewId` to the publicAPI, then in the use case where we don't provide a `synchronizedViewId` in the initial state, we end up recursing over and over again until the call stack gets too large (see line 148 of macro.js).

I'm not sure what the actual solution is, but this fix helped me get things working at least.

jourdain: I would expand the `set` method in [macro](https://github.com/Kitware/vtk-js/blob/master/Sources/macro.js#L141) like below:

```
publicAPI.set = (map = {}, noWarning = false, noFunction = false) => {
    let ret = false;
    Object.keys(map).forEach((name) => {
      const fn = noFunction ? null : publicAPI[`set${capitalize(name)}`];
      if (fn && Array.isArray(map[name])) {
        ret = fn(...map[name]) || ret;
      } else if (fn) {
        ret = fn(map[name]) || ret;
      } else {
        // Set data on model directly
        if (['mtime'].indexOf(name) === -1 && !noWarning) {
          vtkWarningMacro(`Warning: Set value to model directly ${name}, ${map[name]}`);
        }
        model[name] = map[name];
        ret = true;
      }
    });
    return ret;
  };
```

jourdain: And then I'll keep the addOn the way it was.

scottwittenburg:
@jourdain This could be good now, as long as you're ok with it.

PR #224, opened 2017-05-04
fix(SynchronizableRenderWindow): Handle renderer removal properly

scottwittenburg:


scottwittenburg:
@jourdain This solves the issue we have been discussing offline about removal of renderers from the renderwindow.  Thanks to @martinken for the suggestions on how best to handle this.

PR #225, opened 2017-05-05
fix(Rendering): volume rend went one step too far

martinken:
loop went one step too far


martinken:
@jourdain  this should fix the bulk of the issues you are seeing

PR #226, opened 2017-05-05
fix(Rendering): handle multiple renderers

martinken:
missing some key scissor and viewport calls


martinken:
@scottwittenburg  this should fix some of the issues you were seeing.

scottwittenburg:
Awesome, thank you!

PR #227, opened 2017-05-05
test(RenderWindow): Test adding multiple renderers

scottwittenburg:


scottwittenburg:
@martinken Here is the test that verifies your recent multiple renderer fix, thanks again!

When I copied the image out of the debug browser to make the baseline, I noticed some weird dashes around the top and right sides.  Do you have any idea what could be causing that?  Not sure if you can see them clearly in the github changes view or not.

scottwittenburg:
I have rebased this since the fixes @martinken made the coordinate conversion fix, and I generated a new baseline without the funky dashes.  @jourdain I think this should be good to merge once the checks complete.

PR #230, opened 2017-05-08
fix(exports): Remove unwanted exports

scottwittenburg:
Exporting these named functions prevents us from importing things the way we want (or eslint complains). 
 And for the most part, we shouldn't use those methods from outside the modules anyway, instead we should rely on the default export for its newInstance method.


scottwittenburg:
@jourdain PTAL

scottwittenburg:
Github also suggested @wschroed and @martinken may want to review :smile:

PR #231, opened 2017-05-09
fix(SynchronizableRenderWindow): Expose some missing functionality

scottwittenburg:


Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -290,8 +289,8 @@ function polydataUpdater(instance, state, context) {
       .then(
         (values) => {
           const array = vtkDataArray.newInstance(Object.assign({ values }, arrayMetadata));
-          // console.log('array', array.getName());
-          arraysToBind.push([instance.get(arrayMetadata.location)[arrayMetadata.location].addArray, [array]]);
+          const regMethod = arrayMetadata.registrationMethod ? arrayMetadata.registrationMethod : 'addArray';
```
jourdain: I don't know if we want the same name as [here](https://github.com/Kitware/vtk-js/blob/master/Sources/IO/Core/HttpDataSetReader/index.js#L62)

I mean, I like the explicit part of `registrationMethod` vs `registration`, but just wondering.

scottwittenburg:
@jourdain The most recent commit here, along with the VTK MR I just pinged you about, is resulting in the shading we expect in the client.

screen shot 2017-05-09 at 11 22 37 am

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
scottwittenburg: I'd almost rather change that dataset reader to say `registrationMethod`, but up to you...

jourdain: If you change the dataset reader, you will have to fix the data as well and the python script that `may` generate that attribute.

scottwittenburg:
@jourdain I think this could be good now.

PR #233, opened 2017-05-10
fix(vtkMath): Update the name of a function

laurennlam:
`createUninitializedBounds ` was called by ImageMapper but it crashed during execution. The real function was defined under the name `createUninitializedBouds`.


martinken:
+2

PR #244, opened 2017-05-17
Update vtk plane

luciemac:


jourdain:
Is it ready to merge? (I didn't look closer to the code but it looks good to me)

@wschroed you will be able to reuse that class directly now that it as been implemented.

luciemac:
It is ready to merge!

PR #246, opened 2017-05-17
feat(ShaderProgram): Add functions to set uniform2 and 4

jpouderoux:
Use similar code that what existed for uniform3.

### DIFF
--- a/Sources/Rendering/OpenGL/ShaderProgram/index.js
+++ b/Sources/Rendering/OpenGL/ShaderProgram/index.js
@@ -124,26 +124,6 @@ export function vtkShaderProgram(publicAPI, model) {
     return true;
   };
 
-  publicAPI.setUniformf = (name, v) => {
-    const location = publicAPI.findUniform(name);
-    if (location === -1) {
-      model.error = `Could not set uniform ${name} . No such uniform.`;
-      return false;
-    }
-    model.context.uniform1f(location, v);
-    return true;
-  };
-
-  publicAPI.setUniformi = (name, v) => {
-    const location = publicAPI.findUniform(name);
-    if (location === -1) {
-      model.error = `Could not set uniform ${name} . No such uniform.`;
-      return false;
-    }
-    model.context.uniform1i(location, v);
-    return true;
-  };
-
   publicAPI.setUniformMatrix = (name, v) => {
     const location = publicAPI.findUniform(name);
     if (location === -1) {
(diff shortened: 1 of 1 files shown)


jpouderoux:
@laurennlam @martinken

jourdain:
@martinken, I let you click the green button if you approve.

Sources/Rendering/OpenGL/ShaderProgram/index.js
```diff
@@ -184,6 +184,26 @@ export function vtkShaderProgram(publicAPI, model) {
...
+    return true;
+  };
+
+  publicAPI.setUniform4fv = (name, count, v) => {
+    const location = publicAPI.findUniform(name);
+    if (location === -1) {
+      model.error = `Could not set uniform ${name} . No such uniform.`;
+      return false;
+    }
+    model.context.uniform4fv(location, count, v);
```
luciemac: @jourdain  @jpouderoux 
I tried to use this function and it is not working because uniform4fv method needs 2 parameters instead of 3.
[](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/uniform)

I tried to put only `model.context.uniform4fv(location, v);` but it returns the following error:" Expected an array of length a multiple of 4, got an array of length 1."

A solution here is the following one:
```
for (let i = 0; i<v.length; i++) {
        model.context.uniform4fv(location, v[i]);
}
```
I tested it, and it works, I am not sure about the quality of it.

jpouderoux: I just copied the code of `setUniforms3fv()` but it looks like you are right, there is no overload with such a `count` parameter for any of them.

But then I am not sure that your fix is correct because the `count` parameter is not used but I am not sure what it is supposed to do as the webgl function is just `void gl.uniform[1|2|3|4]fv(location, value);`

jpouderoux: That said I think we have to add the `setUniform2*` functions here too!

jpouderoux: (and the `i` (int) version for all of them too)

jpouderoux: - see https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/uniform

martinken: model.context.uniform4fv(location, v);

is correct and that is what it needs to be. Your error was probably that you were not passing in the right value. My guess is you were passing an array of arrays as opposed to just an array.

luciemac: You are right, I was passing an array of arrays, and not just an array, now it works well with only `model.context.uniform4fv(location, v);`

martinken: Cool, if one of you can update the topic to reflect that we should be good. If you want to add the other signatures that would be fine as well.

jpouderoux: @martinken Sorry I am confused. Do you agree that `count` parameter must be removed here?

jpouderoux: Then I don't understand the difference in use between `setUniform3f(name, v)` and `setUniform3fv(name, v)`, as `v` is an array in both cases. Shouldn't we define `setUniform3f(name, v1, v2, v3)` instead?

martinken: Yes, count should be removed from the signatures.

Here is an example of what I think it should be, I did not test it but that should give you the idea.

 publicAPI.setUniform3f = (name, ...args) => {
    const location = publicAPI.findUniform(name);
    if (location === -1) {
      model.error = `Could not set uniform ${name} . No such uniform.`;
      return false;
    }

    let array = args;
    // allow an array passed as a single arg.
    if (array.length === 1 && Array.isArray(array[0])) {
      array = array[0];
    }

    if (array.length !== 3) {
      throw new RangeError('Invalid number of values for array');
    }

    model.context.uniform3f(location, array[0], array[1], array[2]);
    return true;
  };

  publicAPI.setUniform3fv = (name, v) => {
    const location = publicAPI.findUniform(name);
    if (location === -1) {
      model.error = `Could not set uniform ${name} . No such uniform.`;
      return false;
    }
    model.context.uniform3fv(location, v);
    return true;
  };

FYI the difference in the 4f versus 4fv is that 4fv will take any multiple of 4 args. So you can do

vec4 colors[5];

and then setUniform4fv(colors, array[20])

and it will set 5 (which was count)  vec4s at once.

martinken:
+2

martinken:
@jpouderoux   let me know if you want me to merge this. LGTM

jpouderoux:
Thanks @martinken . Is it OK for you @luciemac ? :)

luciemac:
It looks good to me as well!

PR #247, opened 2017-05-17
feat(Macro): Add function to get the InputConnection of an algorithm

jpouderoux:


jpouderoux:
@laurennlam

Sources/macro.js
```diff
@@ -451,6 +451,12 @@ export function algo(publicAPI, model, numberOfInputs, numberOfOutputs) {
     model.inputConnection[port] = outputPort;
   }
 
+  function getInputConnection(port = 0) {
+    if (model.inputConnection[port]) {
+      return model.inputConnection[port];
+    }
+  }
```
jourdain: Should you return something no matter what?
I would not even try to guard it at that point unless you want a `null` instead of an `undefined`

jpouderoux:
@jourdain My bad. As a c++ developer I prefer the null option but let me know if undefined is more suited with the usual way to deal with this in vtk-js.

jourdain:
Usually in JS we need to consider both therefore their is no benefit of using null.
And based on the behavior of arrays I will simply do

```
function getInputConnection(port = 0) {
     return model.inputConnection[port];
}
```

jourdain:
Hi @jpouderoux,

I forgot to mention that you should follow a specific formatting for your commit message so it could be reflected in the change log and version number. The format description is described [here](https://gist.github.com/stephenparish/9941e89d80e2bc58a153).

jpouderoux:
Thanks @jourdain , sorry for being such a newbie here :D

jourdain:
No worries... We all start there. ;-)

PR #248, opened 2017-05-19
fix(Macro): Fix GetInputConnection method

jpouderoux:
Function must always return something, the inpout or undefined.


jpouderoux:
@luciemac

PR #253, opened 2017-05-23
feat(HttpDataAccessHelper): Expose progress in API

zachmullen:
This also adds support in the OBJ Reader to make use of this new API.

We can add this to more methods within HttpDataAccessHelper, but I wanted to put this out there for comment to see if you think this is an OK approach.

### DIFF
--- a/Sources/IO/Misc/OBJReader/index.js
+++ b/Sources/IO/Misc/OBJReader/index.js
@@ -174,8 +174,8 @@ function vtkOBJReader(publicAPI, model) {
   }
 
   // Internal method to fetch Array
-  function fetchData(url) {
-    return model.dataAccessHelper.fetchText(publicAPI, url, model.compression);
+  function fetchData(url, option = {}) {
+    return model.dataAccessHelper.fetchText(publicAPI, url, model.compression, option.progressCallback);
   }
 
   // Set DataSet url
@@ -195,12 +195,12 @@ function vtkOBJReader(publicAPI, model) {
     model.compression = option.compression;
 
     // Fetch metadata
-    return publicAPI.loadData();
+    return publicAPI.loadData({progressCallback: option.progressCallback});
   };
 
   // Fetch the actual data arrays
-  publicAPI.loadData = () => {
-    const promise = fetchData(model.url);
+  publicAPI.loadData = (option = {}) => {
+    const promise = fetchData(model.url, option);
     promise.then(publicAPI.parse);
     return promise;
   };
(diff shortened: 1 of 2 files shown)


jourdain:
Hi Zach,

can you follow the commit message convention? That way the change log will be correctly build and the new release version will match your work.

https://gist.github.com/stephenparish/9941e89d80e2bc58a153

Sources/IO/Misc/OBJReader/index.js
```diff
@@ -195,12 +195,12 @@ function vtkOBJReader(publicAPI, model) {
...
     // Fetch metadata
-    return publicAPI.loadData();
+    return publicAPI.loadData({progressCallback: option.progressCallback});
   };
 
   // Fetch the actual data arrays
-  publicAPI.loadData = () => {
-    const promise = fetchData(model.url);
+  publicAPI.loadData = (option = {}) => {
+    const promise = fetchData(model.url, {progressCallback: option.progressCallback});
```
jourdain: Why not do that instead? 
`const promise = fetchData(model.url, option);`

zachmullen: Would work for now, but in the future the wrapped method may take different options or just a subset of the options from the outer method. If you prefer that way, though, I will change it.

jourdain: This open to other options without loss. Basically you deferred to the dataFetch layer to handle those options.

zachmullen: Sure, we can do that, it just changes the semantics of this parameter from "loadData options" to "options that will be passed through to fetchData"

zachmullen: Speaking of making parameter semantics clear, I'd love to be able to document them inline. I don't see any JSDoc or anything in the code, is there a preferred way to formally document function params?

jourdain: Based on your previous comment you right, `fetchText` should probably take some options.
So the reader just forward those options over.

For the doc, we usually ask the user to write a `api.md` next to the class definition.

jourdain: See https://kitware.github.io/vtk-js/docs/develop_class.html#API-documentation

jourdain: Although, looking at the code `compression` should be part of those options. I think, I'll fix that aspect later on. If you want I can merge the current state of your code. Thx

zachmullen: Ok, sounds good. Thanks for the review.

PR #258, opened 2017-05-25
perf(Rendering): make building the VBO faster

martinken:
Some improvements for cases where the data is changing
frequently. Remove the DynamicFloatArray as it was only
used for building the VBOs and is a performance issue.


Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1478,7 +1480,7 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
       + `C${(n ? n.getMTime() : 1)}D${(c ? c.getMTime() : 1)}`
       + `E${actor.getProperty().getEdgeVisibility()}`
       + `F${(tcoords ? tcoords.getMTime() : 1)}`;
-    if (model.VBOBuildString !== toString) {
+    if (true || model.VBOBuildString !== toString) {
```
martinken: I cannot confirm or deny that

PR #261, opened 2017-05-25
Fix view angle

jourdain:


Sources/Interaction/Misc/DeviceOrientationToCamera/index.js
```diff
@@ -26,7 +26,6 @@ function updateListeners() {
   listeners.forEach((listener) => {
     if (listener) {
       const { camera, render } = listener;
-      camera.setViewAngle(80.0);
```
jourdain: I'm guessing another debug test. ;-)

PR #262, opened 2017-05-26
Device orientation improvement

jourdain:


jourdain:
@martinken Do you mind double checking the logic. It is working now in my other project...

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -239,11 +239,11 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
     if (model.currentRenderer && model.currentManipulator) {
       // When an interaction is active, we should not change the renderer being
       // interacted with.
-    } else {
+    } else if (pos) {
       publicAPI.findPokedRenderer(pos.x, pos.y);
     }
```
martinken: Can you instead do

else {
  publicAPI.findPokedRenderer((pos ? pos.x : 0), (pos ? pos.y : 0));

It is important to have the posed renderer always be set

jourdain: What about the change that I made?

jourdain:
@martinken I actually reversed to my original code as those manipulator expect mouse events and if they are getting other events like device orientation, then we should not delegate any callback to the underlying manipulator as they won't be able to query button down or previous location...

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -239,11 +239,11 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
...
 
-    if (model.currentManipulator) {
+    if (model.currentManipulator && model.currentRenderer && pos) {
```
jourdain: I'm guarding that here also in case we are missing the event and the currentRenderer.

Sources/Interaction/Style/InteractorStyleManipulator/index.js
martinken: We need to call findpokenrenderer every time

martinken: even if we call it with 0,0

jourdain:
@scottwittenburg FYI regarding the manipulators

jourdain:
@martinken this should be good for review/merge...

jourdain:
@martinken I did not clean the `Interaction/Style/InteractorStyleTrackballCamera` render calls that are made on the `RenderWindowInteractor`.

jourdain:
@martinken merging it in an hour if I don't hear otherwise... 
(No pressure...)

PR #265, opened 2017-05-30
fix(RenderWindowInteractor): Bind event when mouse leave

laurennlam:
If user clicks on the render window, then mouves out of the render window, releases the button and go back to the render window, then the moving is like the mouse was never released.
With this code, if the user clicks on the render window and then moves outside and go back to the render window, there is no moving. If the mouse leave the renderwindow, it's like if the user releases it.

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -136,6 +136,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     canvas.addEventListener('click', preventDefault);
     canvas.addEventListener('mousewheel', publicAPI.handleWheel);
     canvas.addEventListener('DOMMouseScroll', publicAPI.handleWheel);
+    canvas.addEventListener('mouseout', publicAPI.handleMouseOut);
 
     canvas.addEventListener('mousedown', publicAPI.handleMouseDown);
     document.querySelector('body').addEventListener('keypress', publicAPI.handleKeyPress);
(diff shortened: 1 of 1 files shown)


jourdain:
@laurennlam and @martinken  do you think we should merge it?

laurennlam:
I think we should, but it's my personal point. The first time I 'played' with vtk.js to understand how it works, I was surprised about the movement of the scene when mouse was out of the render window. This movement doesn't seem natural for me. I think this behaviour is not in VTK (c++), isn't it ?

martinken:
I have no clue. I am not a user experience gui developer :-)   One example online I looked at did not cancel the mouse when exiting the canvas. PV does not cancel movement when exiting the window. So my take is

1) movement should continue outside the canvas, currently it does not

2) if the mouse is released outside the canvas, it should count as a release, currently it does not

3) mouse presses outside the canvas should be ignored, I think this is the case right now

jourdain:
For that we will have to create a mouse handler at the browser (body/document) level and delegate the various events based on the initial mouse press.

aronhelser:
Yes, missing the mouse release is extremely annoying, and hard to deal
with. I don't think the solution of release the mouse if you go outside the
window is a good one, because it's very easy to brush the edge of the
window while dragging, and suddenly your motion stops! It's unexpected from
user's experience with scroll bars and drag-n-drop, which continues outside
the containing window.

There might be a way to check if the button is still pressed in the mouse
move event?

jourdain:
The issue is that the mouse_move event is only attached to a given div, so once you go out of that div, you stop receiving any events... Hence my comment on a more higher level listener which could be responsible for the move/release. 

I think I start having a good idea on how to implement it. I guess the next step would be some time/funding for me to implement it.

felixveysseyre:
@jourdain What is the status here ?
We would need this functionality for a commercial project.

jourdain:
I guess, I'll try to tackle that this week.

felixveysseyre:
Can't we just merge-in the PR ? At least in a first time

jourdain:
Not that one... It is not a good solution.

jourdain:
Can you try this instead https://github.com/Kitware/vtk-js/pull/527

felixveysseyre:
@luciemac Could you test ?

PR #281, opened 2017-06-27
perf(PolyDataMapper): add method to update VBO instead of creating it…

luciemac:
… again when a modified is called only on pointData or points to improve execution time.

### DIFF
--- a/Sources/Rendering/OpenGL/PolyDataMapper/index.js
+++ b/Sources/Rendering/OpenGL/PolyDataMapper/index.js
@@ -1403,7 +1403,28 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
     // Rebuild buffers if needed
     if (publicAPI.getNeedToRebuildBufferObjects(ren, actor)) {
       publicAPI.buildBufferObjects(ren, actor);
+    } else {
+      const updateOptions = publicAPI.getNeedToUpdatePointsDataVBO();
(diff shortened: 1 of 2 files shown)


Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1403,9 +1403,30 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
     // Rebuild buffers if needed
     if (publicAPI.getNeedToRebuildBufferObjects(ren, actor)) {
       publicAPI.buildBufferObjects(ren, actor);
+    } else {
+      const updateOptions = publicAPI.getNeedToUpdatePointsDataVBO();
+      if (updateOptions.points || updateOptions.normals || updateOptions.tCoords) {
+        model.primitives[primTypes.Tris].getCABO().updatePointsDataVBO(model.currentInput, updateOptions);
+        model.VBOUpdatePointsTime.modified();
+      }
     }
```
martinken: I think this should be coded a bit differently. If the data has changed then getNeedToRebuild... should return true, I am guessing you are updating the data but not calling modified right now. Instead you should mark the specific array you modified as modified (but nothing else).  Then we need to make sure getNeedToRebuild... returns true when any array has changed.  Finally in buildBufferObjects we add smarts (much like what you wrote) so that we are more efficient in rebuilding the CABO when only one/few arrays have changed, etc.

PR #289, opened 2017-07-28
fix(DeviceOrientationToCamera): Add missing resetCameraClippingRange

jourdain:


jourdain:
@martinken can you check to see if it is what you were thinking. Going to test on my end before merging...

jourdain:
@martinken the callback approach seems easier and enable more. So merging it...

PR #290, opened 2017-07-28
fix(Renderer): allow a transparent clear color

aronhelser:
Expand the setArray macro to allow mulitple array sizes.
Allow a background color of 3 or 4. If it's length 3, assume opaque.

Also fix export-scene-macro.py to be Py3 compatible.

@martinken @scottwittenburg Here's the change that allows a background image/div to show through.

### DIFF
--- a/Sources/Rendering/Core/Renderer/index.js
+++ b/Sources/Rendering/Core/Renderer/index.js
@@ -20,6 +20,10 @@ function notImplemented(method) {
 function vtkRenderer(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkRenderer');
+  // make sure background has 4 entries. Default to opaque black
+  if (!model.background) model.background = [0, 0, 0, 1];
+  while (model.background.length < 3) model.background.push(0);
+  if (model.background.length === 3) model.background.push(1);
 
   publicAPI.updateCamera = () => {
     if (!model.activeCamera) {
@@ -597,7 +601,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'pass',
   ]);
   macro.getArray(publicAPI, model, ['actors', 'volumes', 'lights']);
-  macro.setGetArray(publicAPI, model, ['background'], [3, 4]);
+  macro.setGetArray(publicAPI, model, ['background'], 4, 1.0);
 
   // Object methods
   vtkRenderer(publicAPI, model);
--- a/Sources/Rendering/Core/Renderer/index.js
+++ b/Sources/Rendering/Core/Renderer/index.js
@@ -597,7 +597,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'pass',
   ]);
   macro.getArray(publicAPI, model, ['actors', 'volumes', 'lights']);
-  macro.setGetArray(publicAPI, model, ['background'], 3);
+  macro.setGetArray(publicAPI, model, ['background'], [3, 4]);
 
   // Object methods
   vtkRenderer(publicAPI, model);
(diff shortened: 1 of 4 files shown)


jourdain:
Conceptually, I would rather have a single size for arrays, but then if the one provided is smaller, I will pad the missing values with a given value provided to the macro.

That way, you won't have to worry about the size of the background. It will always be 4. But then providing the opacity become optional.

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -597,7 +597,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'pass',
   ]);
   macro.getArray(publicAPI, model, ['actors', 'volumes', 'lights']);
-  macro.setGetArray(publicAPI, model, ['background'], 3);
+  macro.setGetArray(publicAPI, model, ['background'], [3, 4]);
```
scottwittenburg: I could be wrong, but looking at `macro.js`, I don't think this is going to work.  It seems like you were thinking of supporting background setting which could take 3 or 4 element array?  Another way would be just creating a background opacity property that would default to 1.0, but could be set to something else.

Sources/macro.js
```diff
@@ -360,7 +360,11 @@ export function setArray(publicAPI, model, fieldNames, size) {
         array = array[0];
       }
 
-      if (array.length !== size) {
+      if (Array.isArray(size)) {
+        if (!size.includes(array.length)) {
+          throw new RangeError('Invalid number of values for array setter');
+        }
+      } else if (array.length !== size) {
```
scottwittenburg: Sorry, I hadn't got here yet.  I guess your stuff above could work, though I might have preferred a new renderer property "backgroundOpacity" instead of trying to support setters for arrays of variable length.  I wonder what @jourdain thinks?

Sources/Rendering/Core/Renderer/index.js
aronhelser: y, I updated macro.js to handle it, I think....

Adding a default value for shorter arrays makes sense to me, and could be useful in other situations. If a default value is not supplied, then I'll still throw an error if the array is too short.

Sources/macro.js
jourdain: I don't like the variable length nor the backgroundOpacity additional property. The opacity should go along with the background color.

aronhelser:
Updated.
I had to leave in the check in the renderer for a background of length 3, because many examples add the background using the model initialization (when newInstance() is called) - so setBackground is never called to ensure a length of 4.

Sources/Rendering/OpenGL/Renderer/index.js
```diff
@@ -136,7 +136,8 @@ function vtkOpenGLRenderer(publicAPI, model) {
 
     if (!model.renderable.getTransparent()) {
       const background = model.renderable.getBackground();
-      model.context.clearColor(background[0], background[1], background[2], 1.0);
+      if (background.length === 3) background.push(1.0);
```
jourdain: You could do the check in the vtkOpenGLRenderer(publicAPI, model) function. Only once at instantiation time.

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -597,7 +597,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'pass',
   ]);
   macro.getArray(publicAPI, model, ['actors', 'volumes', 'lights']);
-  macro.setGetArray(publicAPI, model, ['background'], 3);
+  macro.setGetArray(publicAPI, model, ['background'], 4, 1.0);
```
jourdain: +1

Sources/Rendering/OpenGL/Renderer/index.js
aronhelser: Good idea! I updated.

aronhelser:
OK, should be ready now. Do I need to update a dist?

jourdain:
You are good. Merging it. Thanks!

PR #296, opened 2017-08-07
Some fixes for loading scenes

scottwittenburg:


jourdain:
Should those be merged at some point?
Assuming they still work with the latest changes...

I just don't want them to get lost and forget...

scottwittenburg:
I will need to rewrite some of the history before merging, but mainly I was waiting until I had the zbuffer issues worked out.  Since I haven't had time to work on it, I was just letting it rest for a bit.  I expect we should be back on soon, at which point I'll rebase off the latest master, get the zbuffer stuff working, and then request review.  As long as that's good with you.

jourdain:
Yes perfect... I just have a tendency to merge sooner for the pieces that already works.
But if you want to tackle it later that's fine too.

scottwittenburg:
@jourdain  I have cleaned up the history and rebased off master.  The tests all pass locally, and I think this is generally ready for review.

PR #299, opened 2017-08-09
fix(IO): export actor position/origin/scale/rotation

martinken:
update pv export macro to also export actor pos/origin/scale/rot
update scene importer to import the values


martinken:
@jourdain

PR #300, opened 2017-08-09
refactor(ExportMacro): Allow ParaView to properly display names

jourdain:


Utilities/ParaView/export-scene-macro.py
```diff
@@ -253,6 +253,19 @@ def generateSceneName():
...
+
+def getComponentName(actor):
+  srcs = simple.GetSources()
+  for key, val in srcs.items():
+    actorRep = simple.GetRepresentation(val).GetClientSideObject().GetActiveRepresentation().GetActor()
+    if actor == actorRep:
+      return key[0]
+  return '%d' % componentIndex
+
+
```
scottwittenburg: It's cool you found a way to do this!

PR #301, opened 2017-08-09
fix(HttpSceneLoader): Add support for texture export/load

jourdain:
@martinken


jourdain:
I see the texture to be loaded with what seems to be valid but I might be missing something with the actor binding.

Maybe you can give it a try and figure out what could be the issue. I'll play a bit more in case I found something.

jourdain:
@martinken now I'm getting a perfectly black texture... ;-)

PR #303, opened 2017-08-10
fix(Rendering): texture was not honoring Repeat setting

martinken:
now if repeat is on the texture will be set to repeat


martinken:
@jourdain

PR #305, opened 2017-08-18
style(OpenGLImageMapper): Replace 'boounds' model attribute by 'bounds'

laurennlam:


martinken:
@laurennlam   Thanks!

PR #309, opened 2017-08-24
fix(BoundingBox): Correct initialization of bounds

laurennlam:


jourdain:
That is wrong you want the initial bounds to be something that you can grow. Not something that is so big that it will never grow.

laurennlam:
Got it ! I thought it has to be max value it could.

jourdain:
The point is also to start with an invalid bound that we can check too.

laurennlam:
Thanks for the explanations @jourdain

PR #310, opened 2017-08-24
feat(vtkCube): Add vtkCube class

laurennlam:
Add the vtkCube class + associated test

### DIFF
--- a/Sources/Common/DataModel/Box/index.js
+++ b/Sources/Common/DataModel/Box/index.js
@@ -1,0 +1,205 @@
+import macro   from 'vtk.js/Sources/macro';
+import vtkBoundingBox from 'vtk.js/Sources/Common/DataModel/BoundingBox';
+
+// ----------------------------------------------------------------------------
+// Global methods
+// ----------------------------------------------------------------------------
+
+// The method returns a non-zero value if the bounding box is hit.
+// Origin[3] starts the ray, dir[3] is the vector components of the ray in the x-y-z
+// directions, coord[3] is the location of hit, and t is the parametric
+// coordinate along line. (Notes: the intersection ray dir[3] is NOT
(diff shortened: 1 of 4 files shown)


martinken:
Are you adding vtkCubeSource or vtkBox? Just want to make sure the name is consistent and matches VTK

laurennlam:
It's the vtkBox, sorry for my mistake.

martinken:
@jourdain this look OK to you?

Sources/Common/DataModel/Box/index.js
```diff
@@ -0,0 +1,205 @@
...
+
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  // Object methods
+  macro.obj(publicAPI, model);
+
+  if (!model.bbox) {
```
jourdain: We could improve that also support when we are getting a serialized version of the bbox. But I could not find a good example where I've already done it.

So it could be good for now. We can always improve that later when we bump into it...

jourdain:
+2 merge it when you want..

martinken:
@laurennlam  Normally the author merges a topic so that a reviewer doesn't accidentally merge when the author is not ready. So feel free to merge this when you are ready.

laurennlam:
@martinken I never merge in vtk.js, but I think that I don't have the right to do this, am I right ?

jourdain:
If you see that big green button, that means you do. ;-)

jourdain:
But if you don't then I can click on it... ;-)

laurennlam:
@jourdain I don't see any green button to merge. So you can click on it ! Thanks !

PR #311, opened 2017-08-29
fix(Rendering): fix conversion to normalize display

martinken:
The conversion to normalized display coordinates was
incorrect.


martinken:
@laurennlam  if you are using a non 0.0 to 1.0 viewport then this fix should help with the coordinate conversions.

PR #312, opened 2017-08-30
fix(Rendering): fix world to view conversions

martinken:
The world to view conversions in Renderer were hardcoded to
a 1.0 aspect ratio and an incorrect zvalue range. This change
requires the aspect ratio to be passed in. As a convinience
when doing conversions that involve world -- view methods
have been added on the view (OpenGL/RenderWindow). These
methods take the renderer as an argument and include display -- world
and view -- world conversions.

### DIFF
--- a/Sources/Rendering/Core/InteractorObserver/index.js
+++ b/Sources/Rendering/Core/InteractorObserver/index.js
@@ -58,10 +58,8 @@ function vtkInteractorObserver(publicAPI, model) {
       return null;
     }
 
-    const ndp =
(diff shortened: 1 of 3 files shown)


martinken:
@laurennlam I added displayToWorld and worldToDisplay methods on the view that you can use. They just need the x y z and a renderer.  A bit easier.  Can you test it out and let me know if it works fo ryou?

laurennlam:
@martinken I'll test and let you know the result. Thanks.

laurennlam:
@martinken It works well !

PR #314, opened 2017-08-30
feat(Picker): Add a point picker

laurennlam:
Add the point picker to pick 3D point from mouse clicking.

### DIFF
--- a/Sources/Rendering/Core/AbstractPicker/index.js
+++ b/Sources/Rendering/Core/AbstractPicker/index.js
@@ -1,0 +1,72 @@
+import macro        from 'vtk.js/Sources/macro';
+
+// ----------------------------------------------------------------------------
+// vtkAbstractPicker methods
+// ----------------------------------------------------------------------------
+
+function vtkAbstractPicker(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkAbstractPicker');
+
+  publicAPI.initialize = () => {
+    model.renderer = null;
+
+    model.selectionPoint[0] = 0.0;
+    model.selectionPoint[1] = 0.0;
+    model.selectionPoint[2] = 0.0;
+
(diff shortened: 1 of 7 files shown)


Sources/Rendering/Core/AbstractPicker/index.js
```diff
@@ -0,0 +1,72 @@
...
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  // Build VTK API
+  macro.obj(publicAPI, model);
+
+  macro.get(publicAPI, model, ['renderer', 'pickList']);
+  macro.getArray(publicAPI, model, ['selectionPoint', 'pickPosition'], 3);
```
jourdain: I don't think the getArray macro needs to have the size. Only the set or set/get will need it.

jourdain:
I let @martinken review it.

martinken:
I love it. +2

jourdain:
@laurennlam if you can fix the `macro.getArray` call that could be nice. 
That can be done in a chore(...)

FYI
```
export function getArray(publicAPI, model, fieldNames)
// setXXX: add setter for object of type array
// if 'defaultVal' is supplied, shorter arrays will be padded to 'size' with 'defaultVal'
export function setArray(publicAPI, model, fieldNames, size, defaultVal = undefined)
export function setGetArray(publicAPI, model, fieldNames, size, defaultVal = undefined)
```

laurennlam:
@jourdain What do you mean by fixing the` macro.getArray` ?

jourdain:
No, I mean calling the function with the proper set of arguments

`macro.getArray(publicAPI, model, ['selectionPoint', 'pickPosition'], 3);` vs `macro.getArray(publicAPI, model, ['selectionPoint', 'pickPosition']);`

PR #319, opened 2017-09-06
fix(vtkMath): Determinant2x2 was called with 4 arguments but only 2 were consider

laurennlam:
The method determinant2x2 only take two arguments. But this method was called several times with 4 parameters.

### DIFF
--- a/Sources/Common/Core/Math/index.js
+++ b/Sources/Common/Core/Math/index.js
@@ -331,8 +331,14 @@ function normalize2D(x) {
   return den;
 }
 
-function determinant2x2(c1, c2) {
-  return ((c1[0] * c2[1]) - (c2[0] * c1[1]));
+function determinant2x2(a, b, c, d) {
+  if (arguments.length === 2) {
+    return ((a[0] * b[1]) - (b[0] * a[1]));
+  }
+  if (arguments.length === 4) {
+    return (a * d) - (b * c);
+  }
+  return Number.NaN;
 }
 
 function LUFactor3x3(mat_3x3, index_3) {


Sources/Common/Core/Math/index.js
```diff
@@ -331,8 +331,14 @@ function normalize2D(x) {
...
-function determinant2x2(c1, c2) {
-  return ((c1[0] * c2[1]) - (c2[0] * c1[1]));
+function determinant2x2(a, b, c, d) {
+  if (arguments.length === 2) {
+    return ((a[0] * b[1]) - (b[0] * a[1]));
+  }
+  if (arguments.length === 4) {
+    return (a * d) - (b * c);
+  }
+  return Number.Nan;
```
jourdain:
```
Number.NaN
NaN
Number.Nan
undefined
```

jourdain:
Where were the call with 4 args? Just wondering...

What do you think @martinken

laurennlam:
In vtkMath, this function is called in method : linearSolve3x3, invert3x3, solveLinearSystem with 4 arguments

martinken:
JS is so bad, seriously bad.   Love this change though +2

jourdain:
I would have write it that way though (not that it matters, just FYI)

```
function determinant2x2(...args) {
  if (args.length === 2) {
    return ((args[0][0] * args[1][1]) - (args[1][0] * args[0][1]));
  }
  if (args.length === 4) {
    return (args[0] * args[3]) - (args[1] * args[2]);
  }
  return Number.NaN;
}
```

laurennlam:
@jourdain I never use it. I'll know for the next time. Thanks.

PR #326, opened 2017-09-11
vtkCoordinates: Add getters on computed values

luciemac:


laurennlam:
@jourdain @martinken Is it possible to merge it ? I'll need it for the vtkHandleRepresentation
@luciemac reused tests from VTK (Rendering\Core\Testing\Cxx\otherCoordinates.cxx)

jourdain:
I'll merge it at the same time as the other one to move only the version only once. (thanks for the reminder)

PR #328, opened 2017-09-12
Add cell picker

laurennlam:
Add the notion of cell picker.
Need vtkCell notion merged before merging this PR.
Careful, only cells with three points can be processed for now as only vtkTriangle cell has been created.


laurennlam:
The build failed because vtkCell/VtkTriangle need to be merge.

laurennlam:
This PR has been merge into one in 'Add Cell picker and cell notion'

PR #329, opened 2017-09-12
fix(Rendering): add support for multiple directional lights

martinken:
THis topic adds support for multiple directional lights
in vtk-js.  Positional lights are still not implemented.

Simplified the light class to use a single color as that
makes more sense for lights and is what is implemented.

Updated the cone test to use multiple lights.


martinken:
@laurennlam  this should fix the issue for multiple directional lights. If you are using positional lights with setPositional(true) then that would take a bit more work to get going.

laurennlam:
@martinken Thanks a lot. Only setting the light position, color, ambient color,... is enough. I'll take a look on my code to understand what's wrong on my side.
Thanks again for your reactivity.

PR #330, opened 2017-09-12
Add cell picker and cell notion

laurennlam:
Implement cell picker which use vtkCell and particulary vtkTriangle.
The cellPicker can only pick triangles for now because other kind of cells are not implemented.
An example if the cell picker have been added : press shift key + click

### DIFF
--- a/Sources/Common/Core/Math/index.js
+++ b/Sources/Common/Core/Math/index.js
@@ -331,12 +331,12 @@ function normalize2D(x) {
   return den;
 }
 
-function determinant2x2(a, b, c, d) {
-  if (arguments.length === 2) {
-    return ((a[0] * b[1]) - (b[0] * a[1]));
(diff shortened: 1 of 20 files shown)


jourdain:
I'm guessing that's the one you want me to merge?

Is there anything you want specific review on?

laurennlam:
Yes, that's the one.
Maybe a review in the cell picker could be nice. I skip most of things and I don't know if I made it in the right way. But it seems to work so if you don't have time, I think it's not a problem to merge it.

jourdain:
@martinken do you want to review it? Or can I merge it and then we can always fix it later if need be. ;-)

martinken:
Merge it and we can address issues as they pop up. Having these classes is a ***lot*** better than nothing so definitely merge :-)

PR #333, opened 2017-09-15
chore(tools): Add Size limit to monitor size origin

jourdain:


jourdain:
We need to wait for the next version of size limit to support custom webpack config

PR #337, opened 2017-09-18
fix(OBJReader): Allow tab and several space

jourdain:


jourdain:
FYI @martinken and thanks for the test files...

PR #340, opened 2017-09-20
Add widget notion (+ widget representation)

laurennlam:
This PR add the possibility to create widget.
A handleWidget have been implemented (+ an using example)

Some modifications needs to be added and should be reviewed :
- [vtkCoordinate] the value should be get
- [vtkRenderWindowInteractor] A macro `get ` has been added to the `canvas` because some attributes of the `clientBoundingRect `are used to change 2D picking position. If we don't modify it, then there's a shift when mouse move (cf vtkHandleWidget when we get the event position)
- [vtkOpenGLRenderWindow/vtkViewNode] In VTK, a WidgetRepresentation can have several actors. To avoid adding each actor as actor in the renderer (`addActor`), we would like to only add the representation which will render it's child = actors. So we add a new method which get ViewProps and create a sceneGraph with the representation as a node, and its actors as subnode. (If it's not clear, please tell me)
- [SceneGraphViewNodeFactory/macro] : Instead of testing only the class name to create a node, we check the whole hierarchy. So we change the macro.js file to allow vtkObject to get its all hierarchy.
- A new generic file for the widget representation has been added in order to allow the view node creation of widgetRepresentation. (SceneGraphGenericWidgetRepresentation). Maybe an other name could be set to this file.

@martinken @jourdain We modified a lot of files because we need it and we didn't find an other solution. So don't hesitate to tell us if there's an other way to di it.

### DIFF
--- a/Sources/Rendering/OpenGL/Renderer/index.js
+++ b/Sources/Rendering/OpenGL/Renderer/index.js
@@ -28,9 +28,6 @@ function vtkOpenGLRenderer(publicAPI, model) {
       publicAPI.prepareNodes();
       publicAPI.addMissingNode(model.renderable.getActiveCamera());
       publicAPI.addMissingPropNodes(model.renderable.getViewProps());
-      publicAPI.addMissingNodes(model.renderable.getActors());
-      publicAPI.addMissingNodes(model.renderable.getActors2D());
-      publicAPI.addMissingNodes(model.renderable.getVolumes());
       publicAPI.removeUnusedNodes();
     }
   };
--- a/Sources/Rendering/OpenGL/Renderer/index.js
+++ b/Sources/Rendering/OpenGL/Renderer/index.js
@@ -27,6 +27,7 @@ function vtkOpenGLRenderer(publicAPI, model) {
       publicAPI.updateLights();
       publicAPI.prepareNodes();
       publicAPI.addMissingNode(model.renderable.getActiveCamera());
+      publicAPI.addMissingPropNodes(model.renderable.getViewProps());
       publicAPI.addMissingNodes(model.renderable.getActors());
       publicAPI.addMissingNodes(model.renderable.getActors2D());
       publicAPI.addMissingNodes(model.renderable.getVolumes());
--- a/Sources/Rendering/SceneGraph/ViewNodeFactory/index.js
+++ b/Sources/Rendering/SceneGraph/ViewNodeFactory/index.js
@@ -15,10 +15,24 @@ function vtkViewNodeFactory(publicAPI, model) {
     if (dataObject.isDeleted()) {
       return null;
     }
-    if (Object.keys(model.overrides).indexOf(dataObject.getClassName()) === -1) {
+    // if (Object.keys(model.overrides).indexOf(dataObject.getClassName()) === -1) {
+    //   return null;
+    // }
+
+    const hierarchy = dataObject.getHierarchy();
+    let isObject = false;
+    let className = '';
+    for (let i = hierarchy.length - 1; i > 0; i--) {
+      if (Object.keys(model.overrides).indexOf(hierarchy[i]) !== -1) {
+        isObject = true;
+        className = hierarchy[i];
+        break;
+      }
+    }
+    if (!isObject) {
       return null;
     }
-    const vn = model.overrides[dataObject.getClassName()]();
+    const vn = model.overrides[className]();
     vn.setMyFactory(publicAPI);
     return vn;
   };
--- a/Sources/Rendering/SceneGraph/ViewNodeFactory/index.js
+++ b/Sources/Rendering/SceneGraph/ViewNodeFactory/index.js
@@ -15,20 +15,19 @@ function vtkViewNodeFactory(publicAPI, model) {
     if (dataObject.isDeleted()) {
       return null;
     }
-    // if (Object.keys(model.overrides).indexOf(dataObject.getClassName()) === -1) {
-    //   return null;
-    // }
 
-    const hierarchy = dataObject.getHierarchy();
+    let cpt = 0;
+    let className = dataObject.getClassName(cpt++);
     let isObject = false;
-    let className = '';
-    for (let i = hierarchy.length - 1; i > 0; i--) {
-      if (Object.keys(model.overrides).indexOf(hierarchy[i]) !== -1) {
+    const keys = Object.keys(model.overrides);
+    while (className && !isObject) {
+      if (keys.indexOf(className) !== -1) {
         isObject = true;
-        className = hierarchy[i];
-        break;
+      } else {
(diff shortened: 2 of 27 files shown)


Sources/Rendering/SceneGraph/ViewNodeFactory/index.js
```diff
@@ -15,10 +15,24 @@ function vtkViewNodeFactory(publicAPI, model) {
     if (dataObject.isDeleted()) {
       return null;
     }
-    if (Object.keys(model.overrides).indexOf(dataObject.getClassName()) === -1) {
+    // if (Object.keys(model.overrides).indexOf(dataObject.getClassName()) === -1) {
+    //   return null;
+    // }
+
+    const hierarchy = dataObject.getHierarchy();
```
jourdain: I'm guessing the `.isA()` is not enough, and you need to check the class name hierarchy in reverse order?

That seems strange, and instead of providing a method to get the full hierarchy, I would maybe prefer to extend the existing method like follow:

`publicAPI.getClassName = (depth=0) => model.classHierarchy.slice(-1 - depth)[0];`

laurennlam: yes, that's the point.
I understand the extension, but how do you know the depth ? If I try getClassName(10) and the object is a vtkProp, what does it do ?

jourdain: If you go beyond the bounds of the classHierarchy array, it will always return the 1st element which is vtkObject.

```
const handledClasses = Object.keys(model.overrides);
let className = '';
let classNameDepth = 0;
while (className !== 'vtkObject' && handledClasses.indexOf(className) === -1) {
    className = dataObject.getClassName(classNameDepth++);
}

if (className === 'vtkObject') {
   return null;
}
```

laurennlam: I'm not sure to understand, do I have to add this last code ? Because I don't find it.

jourdain: The code I wrote is the replacement of your change below if you extend the `getClassName` the way I was previously suggesting.

```
 +    let isObject = false;
 +    let className = '';
 +    for (let i = hierarchy.length - 1; i > 0; i--) {
 +      if (Object.keys(model.overrides).indexOf(hierarchy[i]) !== -1) {
 +        isObject = true;
 +        className = hierarchy[i];
 +        break;
 +      }
 +    }
 +    if (!isObject) {
        return null;
      }
```

laurennlam: Ok thanks !

jourdain: A better solution would be 

`publicAPI.getClassName = (depth=0) => model.classHierarchy[model.classHierarchy.length - 1 - depth];`

This will prevent array allocation and you will get `undefined` if you go out of bound.

Sources/Rendering/OpenGL/Renderer/index.js
```diff
@@ -27,6 +27,7 @@ function vtkOpenGLRenderer(publicAPI, model) {
       publicAPI.updateLights();
       publicAPI.prepareNodes();
       publicAPI.addMissingNode(model.renderable.getActiveCamera());
+      publicAPI.addMissingPropNodes(model.renderable.getViewProps());
```
jourdain: @martinken does it make sense to have that new `addMissingPropNodes`?

Sources/Rendering/SceneGraph/ViewNodeFactory/index.js
```diff
@@ -15,20 +15,18 @@ function vtkViewNodeFactory(publicAPI, model) {
...
 
-    const hierarchy = dataObject.getHierarchy();
+    let cpt = 0;
+    let className = dataObject.getClassName(cpt++);
     let isObject = false;
-    let className = '';
-    for (let i = hierarchy.length - 1; i > 0; i--) {
-      if (Object.keys(model.overrides).indexOf(hierarchy[i]) !== -1) {
+    while (className) {
+      if (Object.keys(model.overrides).indexOf(className) !== -1) {
```
jourdain: Try to cache `Object.keys(model.overrides)` instead of recomputing each time inside a loop.

jourdain: Since you have a while, you can put your break statement in that `while()`

laurennlam:
@jourdain Would it be possible to merge it ?

jourdain:
You didn't put a comment saying that it was ready...

@martinken are you ok with the new method `addMissingPropNodes`?

Sources/Rendering/OpenGL/Renderer/index.js
martinken: Let me look at it when I get a sec. Sorry I missed this.

laurennlam:
@jourdain Sorry, I thought I did it.

Sources/Rendering/OpenGL/Renderer/index.js
jourdain: @martinken any thoughts?

martinken: You should yank the next few lines as they are useless and just consume time. Actors/Volumes etc are all viewprops so you are trying to add them twice from what I can see.

jourdain:
@laurennlam try to rebase from master instead of merging master into your branch that makes the review easier.

laurennlam:
@jourdain Sorry, I made some bad command. It should be better now

jourdain:
Thanks! Much better ;-)

jourdain:
Is it ready to be merged?

laurennlam:
I think so, if you or Ken have no more requirement :)

Sources/Rendering/OpenGL/Renderer/index.js
```diff
@@ -27,9 +27,10 @@ function vtkOpenGLRenderer(publicAPI, model) {
       publicAPI.updateLights();
       publicAPI.prepareNodes();
       publicAPI.addMissingNode(model.renderable.getActiveCamera());
-      publicAPI.addMissingNodes(model.renderable.getActors());
-      publicAPI.addMissingNodes(model.renderable.getActors2D());
-      publicAPI.addMissingNodes(model.renderable.getVolumes());
+      publicAPI.addMissingPropNodes(model.renderable.getViewProps());
+      // publicAPI.addMissingNodes(model.renderable.getActors());
```
jourdain: @martinken is that ok?

Should we remove those commented line @laurennlam

martinken:
Thanks for yanking the extra calls. My other suggestions I can implement when I add the WebVR support as it will be creating more complex events.

martinken:
Yeh sure

PR #341, opened 2017-09-21
Add volume slicing on each direction (X, Y, Z)

laurennlam:


laurennlam:
Ready to Merge

laurennlam:
@jourdain, Do you know why the example doesn't work in  https://kitware.github.io/vtk-js/examples/ImageMapper.html ? It works well why I run the example on my own machine.

PR #345, opened 2017-09-26
Fix example

jourdain:


jourdain:
@laurennlam The path to the data is wrong. `Data` only exist on your local file system but not on the github.io endpoint. If you look any example that load data they all use the lower case path.

laurennlam:
@jourdain Okay thanks !

jourdain:
@laurennlam is there any setting that can be adjust to make the slice less bright? Just so we can get a better viz.
Maybe add a slider to control the `brightness` like the drag interaction you have here: https://kitware.github.io/vtk-js/examples/ImageMapper.html

Thanks!

laurennlam:
Sure

jourdain:
thx

PR #346, opened 2017-09-26
chore(ImageMapper): Add sliders in the example (brightness)

laurennlam:
Add sliders in the example to change level and window of slices


laurennlam:
@jourdain I click on PR before seing that there's conflicts. Sorry. I think I forgot to pull the master branch...

jourdain:
@laurennlam you need to rebase since I moved your example in `vtk-js/Examples/Volume/MultiSliceImageMapper/index.js`

jourdain:
as it was conflicting with the ImageMapper example in Rendering/Core

jourdain:
The conflict reason is because all examples are gathered in the same directory so you can not have two example with the same name...

laurennlam:
@jourdain thanks. I'll change it.

PR #349, opened 2017-09-29
ImageMarchingCubes working

wschroed:
Added Plane and Sphere implicit functions.

Added SampleFunction to sample implicit functions on a volume.

Added ImageMarchingCubes to create isosurface on a volume. Also defined a MC case table as part of the work.


jourdain:
Thanks Will! I'll give it a look...

Sources/Common/DataModel/Plane/index.js
```diff
@@ -42,7 +60,13 @@ export function extend(publicAPI, model, initialValues = {}) {
 
   // Object methods
   macro.obj(publicAPI, model);
-  macro.setGet(publicAPI, model, ['bounds']);
+  macro.setGetArray(publicAPI, model, [
+    'origin',
+  ], 3);
+  macro.setGetArray(publicAPI, model, [
+    'normal',
+  ], 3);
```
jourdain: You could have gathered `normal` and `origin` inside that list just to have a single `macro.setGetArray` call.

Sources/Filters/General/ImageMarchingCubes/index.js
```diff
@@ -0,0 +1,184 @@
+import macro       from 'vtk.js/Sources/macro';
+import vtkPolyData from 'vtk.js/Sources/Common/DataModel/PolyData';
+
+var vtkCaseTable = require('vtk.js/Sources/Filters/General/ImageMarchingCubes/caseTable');
```
jourdain: const would be more appropriate than `var`

Sources/Filters/General/ImageMarchingCubes/caseTable.js
```diff
@@ -0,0 +1,298 @@
+// ----------------------------------------------------------------------------
+// Marching cubes case functions (using triangles to complete tessellation).
+// For each case, a list of edge ids that form the triangles. A -1 marks the
+// end of the list of edges. Edges are taken three at a time to generate
+// triangle points.
+// ----------------------------------------------------------------------------
+export function getCase(index) {
+  const cases =
```
jourdain: That `const` should be global to the module (same as static).

Sources/Filters/General/ImageMarchingCubes/caseTable.js
```diff
@@ -0,0 +1,298 @@
...
+      0, 8, 3, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, /* 254 1 */
+      -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, /* 255 0 */
+    ];
+
+  return cases.slice(index * 16, (index + 1) * 16);
+}
+
+// Define the twelve edges of the voxel by the following pairs of vertices
+export function getEdge(eid) {
+  const edges =
```
jourdain: That `const` should be global to the module (same as static).

Sources/Imaging/Hybrid/SampleFunction/index.js
```diff
@@ -0,0 +1,118 @@
...
+
+function vtkSampleFunction(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkSampleFunction');
+
+  // Capture "parentClass" api for internal use
+  const superClass = Object.assign({}, publicAPI);
+
+  publicAPI.getMTime = () => {
+    if (model.implicitFunction === undefined) {
```
jourdain: `if (!model.implicitFunction)` will be more generic as it will handle `null` and `undefined`.

And better could be

`if (!(model.implicitFunction && model.implicitFunction.getMTime))`

wschroed:
Seb I believe I made all your changes

jourdain:
Already merge via another branch

PR #350, opened 2017-09-29
Implicit function framework seb

jourdain:


jourdain:
@wschroed FYI.
Did a couple formatting fix and added an example.

PR #358, opened 2017-10-10
perf(Cell): Don't recreate vtkPoints when vtkCell is initialize

laurennlam:
We don't want to recreate a vtkPoints when we call initialize on a vtkCell.
With the cellpicker and a mesh with 120000 cells, we pick on 4 sec. With this modification, we pick on 0.5sec.


jourdain:
ready to merge?

laurennlam:
Yes

PR #359, opened 2017-10-10
fix(XMLImageDataReader): Add new image data reader

jourdain:


jourdain:
@martinken you can load vti file that are ascii or binary with no compression.

But the volume rendering example that comes along is really strange. I'm not sure if it is because of invalid Volume pipeline setup or else. I may ping you tomorrow about it.

PR #365, opened 2017-10-13
fix an issue where vol ren did not handle non zero min ext

martinken:
If the minimum extnt for an image data was not 0,0,0
the rendering would be off. This fixes that.


martinken:
@jourdain  this might fix the extent issue you were seeing

martinken:
And I forgot to use the right commit style oops

jourdain:
No worries. I may push something later which will trigger a new release with your change

PR #367, opened 2017-10-16
fix(Rendering): make sure scalar visibility works properly

martinken:
For Stick and Sphere mappers turing scalar visibility on and
then off would leave the rendering in the on state.


jourdain:
Feel free to merge and include the test from Scott @martinken

PR #373, opened 2017-10-20
Rendering fixes

jourdain:


jourdain:
@martinken I'm merging but you should double check just in case...

PR #375, opened 2017-10-20
Fix picker

laurennlam:
The user matrix need to be transpose before being used to project points

### DIFF
--- a/Sources/Rendering/Core/Picker/index.js
+++ b/Sources/Rendering/Core/Picker/index.js
@@ -2,7 +2,7 @@ import macro        from 'vtk.js/Sources/macro';
 import vtkAbstractPicker from 'vtk.js/Sources/Rendering/Core/AbstractPicker';
 import vtkBox from 'vtk.js/Sources/Common/DataModel/Box';
 import vtkMath from 'vtk.js/Sources/Common/Core/Math';
-import { mat4, vec3 } from 'gl-matrix';
+import { mat4, vec4 } from 'gl-matrix';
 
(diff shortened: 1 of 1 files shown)


jourdain:
@martinken ?
@laurennlam Ready to merge?

laurennlam:
Yes, Ready to merge !
It's a fix if the user transform of an actor has been changed.

Sources/Rendering/Core/Picker/index.js
```diff
@@ -209,8 +214,8 @@ function vtkPicker(publicAPI, model) {
         scale[1] = vtkMath.norm(col2);
         scale[2] = vtkMath.norm(col3);
 
-        vec3.transformMat4(p1Mapper, p1World, model.transformMatrix);
-        vec3.transformMat4(p2Mapper, p2World, model.transformMatrix);
+        vec4.transformMat4(p1Mapper, p1World, model.transformMatrix);
+        vec4.transformMat4(p2Mapper, p2World, model.transformMatrix);
```
martinken: Do you need divide p*Mapper by p*Mapper[3] here? I think you should to be safe.

martinken: Ping

laurennlam: Yes, sorry I just made the correction after your response. But yes, to be more efficient and understandable, it should be safe do divide pMapper by pMapper[3].

laurennlam:
Is it okay now ?

PR #378, opened 2017-10-23
fix(Rendering): some fixes and performance improvements

martinken:
Clean up some of the rendering interaction code to better
manage animating and to ignore render calls when in the
middle of interacting. Also remove some calls that were
rendering the still image twice.

Also some code cleanup to improve performance some. Still a ways
to go in cases with hundreds of actors.

### DIFF
--- a/Sources/Rendering/Core/InteractorStyle/index.js
+++ b/Sources/Rendering/Core/InteractorStyle/index.js
@@ -194,8 +194,6 @@ function vtkInteractorStyle(publicAPI, model) {
   publicAPI.startState = (state) => {
     model.state = state;
     if (model.animationState === States.IS_ANIM_OFF) {
-      // const rwi = model.interactor;
-      // rwi.getRenderWindow().setDesiredUpdateRate(rwi.getDesiredUpdateRate());
       publicAPI.invokeStartInteractionEvent({ type: 'StartInteractionEvent' });
     }
   };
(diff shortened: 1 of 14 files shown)


martinken:
@jourdain There were enough changes I wanted to get this merged before continuing

Sources/Rendering/Core/InteractorStyle/index.js
```diff
@@ -194,8 +194,6 @@ function vtkInteractorStyle(publicAPI, model) {
   publicAPI.startState = (state) => {
     model.state = state;
     if (model.animationState === States.IS_ANIM_OFF) {
-      // const rwi = model.interactor;
-      // rwi.getRenderWindow().setDesiredUpdateRate(rwi.getDesiredUpdateRate());
       publicAPI.invokeStartInteractionEvent({ type: 'StartInteractionEvent' });
```
jourdain: @martinken we could also have that StartInteractionEvent/EndInterationEvent be static object defined at the top of the module.

Sources/Rendering/Core/Prop/index.js
```diff
@@ -17,6 +17,7 @@ function vtkProp(publicAPI, model) {
       Math.max(val.getMTime(), acc)
     , 0));
 
+  publicAPI.getProps = () => publicAPI;
```
jourdain: There must be a typo here?

martinken: By default a prop is a prop right? So getProps should by default return itself. Or are you saying it should be [publicAPI] ?

jourdain: You are right, it is correct. I was expecting another method call and thought you forgot the method.

PR #380, opened 2017-10-23
fix(Rendering): more performance improvements

martinken:
A fair number of calls were being done that didn't
need to be done. Also some VAO rebuilding was always
happening when it should not have.

### DIFF
--- a/Sources/Common/DataModel/DataSetAttributes/index.js
+++ b/Sources/Common/DataModel/DataSetAttributes/index.js
@@ -131,7 +131,8 @@ function vtkDataSetAttributes(publicAPI, model) {
   };
 
   attrTypes.forEach((value) => {
-    publicAPI[`get${value}`] = () => publicAPI.getArrayByIndex(model[`active${value}`]);
+    const activeVal = `active${value}`;
+    publicAPI[`get${value}`] = () => publicAPI.getArrayByIndex(model[activeVal]);
     publicAPI[`set${value}`] = da => publicAPI.setAttribute(da, value);
     publicAPI[`setActive${value}`] =
       arrayName => publicAPI.setActiveAttributeByIndex(
--- a/Sources/Rendering/Core/Renderer/index.js
+++ b/Sources/Rendering/Core/Renderer/index.js
@@ -57,8 +57,8 @@ function vtkRenderer(publicAPI, model) {
         // vtkLight::SetLightTypeToSceneLight()
       } else if (light.lightTypeIsHeadLight()) {
         // update position and orientation of light to match camera.
-        light.setPosition(camera.getPosition());
-        light.setFocalPoint(camera.getFocalPoint());
+        light.setPosition(camera.getPosition(true));
+        light.setFocalPoint(camera.getFocalPoint(true));
       } else if (light.lightTypeIsCameraLight()) {
         light.setTransformMatrix(lightMatrix);
       } else {
@@ -257,12 +257,12 @@ function vtkRenderer(publicAPI, model) {
     const allBounds = [].concat(INIT_BOUNDS);
     let nothingVisible = true;
 
-    publicAPI.invokeEvent({ type: 'ComputeVisiblePropBoundsEvent', renderer: publicAPI });
+    publicAPI.invokeEvent(model.computeVisiblePropBoundsEventData);
 
     // loop through all props
-    model.props
-      .filter(prop => prop.getVisibility() && prop.getUseBounds())
-      .forEach((prop) => {
+    for (let index = 0; index < model.props.length; ++index) {
+      const prop = model.props[index];
+      if (prop.getVisibility() && prop.getUseBounds()) {
         const bounds = prop.getBounds();
         if (bounds && vtkMath.areBoundsInitialized(bounds)) {
           nothingVisible = false;
@@ -286,7 +286,8 @@ function vtkRenderer(publicAPI, model) {
             allBounds[5] = bounds[5];
           }
         }
-      });
+      }
+    }
 
     if (nothingVisible) {
       vtkMath.uninitializeBounds(allBounds);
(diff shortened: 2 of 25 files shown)


martinken:
@jourdain  You might want to take a look at this one. Lots of crazy here. At some point I want to change the getArray(true) stuff to a different name. Maybe getArrayByReference to make it more clear and clean.

jourdain:
I like the getReference approach as a fast path for everything. I guess we can talk about it tomorrow if you want.

jourdain:
BTW, if I think it's a good first step should I merge?

Sources/Common/DataModel/DataSetAttributes/index.js
```diff
@@ -131,7 +131,8 @@ function vtkDataSetAttributes(publicAPI, model) {
   };
 
   attrTypes.forEach((value) => {
-    publicAPI[`get${value}`] = () => publicAPI.getArrayByIndex(model[`active${value}`]);
+    const activeVal = `active${value}`;
+    publicAPI[`get${value}`] = () => publicAPI.getArrayByIndex(model[activeVal]);
```
jourdain: Love that closure optimization...

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -57,8 +57,8 @@ function vtkRenderer(publicAPI, model) {
         // vtkLight::SetLightTypeToSceneLight()
       } else if (light.lightTypeIsHeadLight()) {
         // update position and orientation of light to match camera.
-        light.setPosition(camera.getPosition());
-        light.setFocalPoint(camera.getFocalPoint());
+        light.setPosition(camera.getPosition(true));
```
jourdain: Should we get a fast path for both the set and get?

Sources/Rendering/Core/Viewport/index.js
```diff
@@ -36,9 +36,11 @@ function vtkViewport(publicAPI, model) {
 
   // this method get all the props including any nested props
   publicAPI.getViewPropsWithNestedProps = () => {
-    model.allprops = [];
+    model.allprops = model.props;
     model.props.forEach((prop) => {
-      model.allprops = model.allprops.concat(prop.getProps());
+      if (prop.getNestedProps()) {
+        model.allprops = model.allprops.concat(prop.getProps());
+      }
```
jourdain: May need some talk...

Sources/Rendering/OpenGL/Actor/index.js
```diff
@@ -93,11 +107,10 @@ function vtkOpenGLActor(publicAPI, model) {
...
       publicAPI.activateTextures();
-    } else {
-      // deactivate textures
-      model.activeTextures.forEach((child) => {
-        child.deactivate();
-      });
+    } else if (model.activeTextures) {
+      for (let index = 0; index < model.activeTextures.length; index++) {
+        model.activeTextures[index].deactivate();
+      }
```
jourdain: Did you noticed that the `forEach` were bad compare to a simple for loop?

Another approach for fast looping is that:

```
let count = model.activeTextures.length;
while(count--) {
  model.activeTextures[count].deactivate();
}
```

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1102,30 +1111,31 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
     // bind some light settings
     let numberOfLights = 0;
 
-    const lights = ren.getLights();
-    Object.keys(lights).map(key => lights[key]).forEach((light) => {
+    const lights = ren.getLights(true);
+    for (let index = 0; index < lights.length; ++index) {
```
jourdain: Are Lights now an array instead of an object?

Sources/macro.js
```diff
@@ -169,6 +169,8 @@ export function obj(publicAPI = {}, model = {}) {
     return subset;
   };
 
+  publicAPI.getOneProperty = val => model[val];
```
jourdain: Could it be replaced with the getReference macro? (I've added it while reviewing your code and reading your comment)

jourdain:
@martinken you may want to remove my last commit if you've tested and noticed that the forEach are slower or allocate more memory for the GC. Otherwise, theoretically they should be optimized by the JIT.

Another approach to speedup loops is the `while(count--){}` option.

jourdain:
Moreover, we should talk about `getOneProperty`. ;-)

jourdain:
My last commit is to support something like that:

```
light.setPositionFrom(camera.getPositionByReference());
light.setFocalPointFrom(camera.getFocalPointByReference());
light.modified(camera.getMTime());
```

jourdain:
Hum... The tests are failing on my machine.

Sources/macro.js
jourdain: I guess in that case it will just be a naming thing.

Sources/Rendering/Core/Renderer/index.js
jourdain: See my last comment at the bottom of that chat...

Sources/macro.js
martinken: Yes I think the implementation of getReference would work for this case.

martinken:
OK can you give that a shot and see if it passes for you?

jourdain:
I've tried on my `day` Mac and it is working but I have a build error/warning:

```
/Users/seb/Documents/code/Web/vtk-js/Sources/Rendering/OpenGL/VertexArrayObject/index.js
  229:11  error  'gl' is assigned a value but never used  no-unused-vars
```

I'll checkout the old code to make sure that I can reproduce the issue on my `day` Mac. Otherwise you will have to wait for me to try on my `evening` Mac. ;-)

PR #387, opened 2017-10-26
Mc speed improvements

jourdain:


jourdain:
That remind me a conversation at the hackathon where I didn't wanted to create any helper function in data array to force the user to directly used the typed array...

PR #409, opened 2017-11-02
doc(PiecewiseGaussianWidget): Update exampe to be mobile friendly

jourdain:


jourdain:
FYI @zachmullen 
But you will need Travis to finish in order to get the update version online which usually takes around 20 minutes...

PR #416, opened 2017-11-05
perf(Rendering): update values for volumeViewer

martinken:
Provide better range of edge enhancement values
and spacing values.


jourdain:
Thanks @martinken !

PR #419, opened 2017-11-08
perf(Rendering): more systems default to webgl2

martinken:
added linux and Netscape to webgl2 defaults


jourdain:
Do you know that perf() does not trigger a new release?

jourdain:
Do you know that perf() does not trigger a new release?

PR #421, opened 2017-11-08
 feat(ImageMapper): Add ability to specify scalar to color mapping

thewtex:

We want a rendered image to have the same color mapping as an associated volume
rendering.

Todos:

- [x]  getTable is not available in `vtkLookupTable`: this is in
  `vtkColorTransferFunction` -- use a function in `vtkScalarsToColors` in both
  the volume and image mappers?
- [x] Create an identity model.colorTexture to use in the OpenGLImageMapper
  when no color map is specified
- [x] Test and fix up the shaders in the OpenGLImageMapper to use
  model.colorTexture in all relevant cases
- [x] Re-enable all tests


thewtex:
CC: @martinken

laurennlam:
I think you didn't want to add test.js file with only your test :)

thewtex:
@laurennlam yep -- will re-enable them after the WIP is complete. I added an item to the todo list to keep track of this. 🔢

thewtex:
@martinken I updated the implementation so `getTable` is available. `vtkImageProperty` now has `setRGBTransferFunction`, which is nicely more consistent with `vtkVolumeProperty`.

Another consideration: `vtkVolumeProperty` has `setGrayTransferFunction`. However, this is not used by the OpenGL renderer yet, as far as I can see. Instead, we could use window / level like `vtkImageProperty`...

thewtex:
All set after a fantastic hacking session with @martinken!

PR #422, opened 2017-11-09
Fix min value

felixveysseyre:


felixveysseyre:
@jourdain I am asking myself another question, why is it better to use `Number.MAX_VALUE` rather than `Infinity` constant ?

martinken:
+1

felixveysseyre:
Also, I did not replaced `Number.MAX_VALUE` in `distanceToLine` and `intersectWithLine` functions as it seemed correct to me.

Can somebody can confirm me that ?

felixveysseyre:
@jourdain

- `distanceToLine`: `vtk-js/Sources/Common/DataModel/Line/index.js`, line 12
- `intersectWithLine`: `vtk-js/Sources/Common/DataModel/Line/index.js`, line 136

jourdain:
I think it would be nice to change them so we don't get tempted to wrongly used `Number.MIN_VALUE`, but based on the way the code is written that value is not actually used.

PR #427, opened 2017-11-21
feat(ImageMarchingCubes): Added 2D image or slice contouring

wschroed:
A new isocontouring algorithm for 2D images or a zslice from a
volume has been added. Multiple contours are supported, and merging
intersection points along edges may be enabled or disabled.

### DIFF
--- a/Sources/Common/DataModel/ImplicitBoolean/index.js
+++ b/Sources/Common/DataModel/ImplicitBoolean/index.js
@@ -19,6 +19,18 @@ function vtkImplicitBoolean(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkImplicitBoolean');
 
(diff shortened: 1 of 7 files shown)


Sources/Common/DataModel/ImplicitBoolean/index.js
```diff
@@ -19,6 +19,18 @@ function vtkImplicitBoolean(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkImplicitBoolean');
 
+  publicAPI.getMTime = () => {
+    let mTime = model.superClass.getMTime();
```
jourdain: Did you bound the superClass on the model? 
It should be a closure/private variable not something shared across all the hierarchy via the model.

PR #428, opened 2017-11-21
fix(Rendering): add a glyph3d mapper

martinken:
This is a first implementaiton of a glyph3d mapper
designed to more quickly render colored, scaled,
positioned, and oriented glyphs. This is a first
cut and will be followed quickly by an update that
adds a faster rendering path for implementations
that support instancing like webgl 2


martinken:
@jourdain  here is a first cut that implements the API and the webgl 1 path. I'll add the webgl  2 fast path soon. Minimal testing but I wanted to get something to you.

jourdain:
Thanks! Still working on the 2D rendering, but getting really close... I will start working on the 3D side on Monday...

PR #431, opened 2017-11-28
test(Rendering): add test for using vectorComponent coloring

martinken:
Added a test testing setVectorComponent on the lut.


jourdain:
You should uncomment that line

https://github.com/Kitware/vtk-js/blob/master/Examples/Applications/GeometryViewer/index.js#L238

martinken:
OK I did my +19-6 uncommenting of that line

PR #434, opened 2017-11-29
fix(Glyph3DMapper): Add support for color on glyph

jourdain:


jourdain:
@martinken all yours to merge...

PR #439, opened 2017-11-30
fix(Rendering): fix two issues with glyph scaling

martinken:
1) the bounds when using scale by magnitude were too small
when there were mure than one component.

2) The scaleFactor was use din the bounds but not in the actual
rendering.


martinken:
@jourdain any idea on the failures?

jourdain:
Travis seems to be in a bad state, complaining on:

```
Cannot start ChromeHeadless
The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/google/chrome/chrome-sandbox [...]
```

Looking into it...

jourdain:
You can rebase on master or we can merge it... ;-)

PR #441, opened 2017-12-01
fix(Rendering): tighter glyph bounds

martinken:
Now computes a more accurate and tight bounding
box for the glyph.


martinken:
@jourdain check out my changes to INIT_BOUNDS and let me know if that is the right direction.

jourdain:
perfect that's exactly what I would have done...

martinken:
You can see from the new valid image that the bounds are a bit better. But this may have nothing to do with the issue you were seeing which sounds like an Actor bounds issue.

jourdain:
Yes, I still don't know as I tried to reproduce the issue with the cone example but couldn't get something as wrong as I was seeing with the glyphs... So not really sure of anything. ;-)

PR #446, opened 2017-12-05
Add vti/vtp appended data format support

floryst:
WIP

This adds initial capabilities to read data arrays out of a VTI/VTP appended data file format, as generated from VTK.

### DIFF
--- a/Sources/IO/XML/XMLReader/index.js
+++ b/Sources/IO/XML/XMLReader/index.js
@@ -152,12 +152,18 @@ function processDataArray(size, dataArrayElem, compressor, byteOrder, headerType
       }
     }
   } else if (format === 'appended') {
-    const offset = dataArrayElem.getAttribute('offset');
+    const offset = Number(dataArrayElem.getAttribute('offset'));
     // read header
     // NOTE: this will incorrectly read the size if headerType is (U)Int64 and
     // the value requires (U)Int64.
     const header = new TYPED_ARRAY[headerType](binaryBuffer, offset, 1);
-    const arraySize = header[0] / TYPED_ARRAY_BYTES[dataType];
+    let arraySize = header[0] / TYPED_ARRAY_BYTES[dataType];
+
+    // if we are dealing with Uint64, we need to get double the values since
+    // TYPED_ARRAY[Uint64] is Uint32.
+    if (dataType.indexOf('Int64') !== -1) {
+      arraySize *= 2;
+    }
 
     // read values
     values = new TYPED_ARRAY[dataType](binaryBuffer, offset + TYPED_ARRAY_BYTES[headerType], arraySize);
@@ -174,7 +180,7 @@ function processDataArray(size, dataArrayElem, compressor, byteOrder, headerType
 
 // ----------------------------------------------------------------------------
 
-function processCells(size, containerElem, compressor, byteOrder, headerType) {
+function processCells(size, containerElem, compressor, byteOrder, headerType, binaryBuffer) {
   const arrayElems = {};
   const dataArrayElems = containerElem.getElementsByTagName('DataArray');
   for (let elIdx = 0; elIdx < dataArrayElems.length; elIdx++) {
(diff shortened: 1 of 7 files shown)


jourdain:
Your commit message need to be reformatted.
enh() does not exist, use fix() instead...

Sources/IO/XML/XMLReader/index.js
```diff
@@ -18,6 +18,48 @@ function stringToXML(xmlStr) {
   return (new DOMParser()).parseFromString(xmlStr, 'application/xml');
 }
 
+/**
+ * Extracts binary data out of a file bytearray given a prefix/suffix.
+ */
+function extractBinary(arrayBuffer, prefixRegex, suffixRegex = null) {
```
jourdain: That function can be used for other parser, therefore, we may want it in a more generic spot or we need to export it in some way...

Sources/IO/XML/XMLReader/index.js
```diff
@@ -151,6 +193,13 @@ function processDataArray(size, dataArrayElem, compressor, byteOrder, headerType
         values = integer64to32(values);
       }
     }
+  } else if (format === 'appended') {
+    const offset = dataArrayElem.getAttribute('offset');
+    // read header
+    const header = binaryBuffer.slice(offset, offset + TYPED_ARRAY_BYTES[headerType]);
```
jourdain: You don't want to call slice on that array....

jourdain:
```
const header = new TYPED_ARRAY[headerType](offset, 1);
```

jourdain: I missed the ArrayBuffer arg...

jourdain:
Looks good... Couple of edits but really close...

Sources/IO/XML/XMLReader/index.js
floryst: I was thinking somewhere in vtk-js/Utilities since it is generic enough to come out of the XML directory. Not sure if it'll be used outside of IO though, so maybe some IO utilities?

jourdain: IO/Core/BinaryHelper

jourdain: Maybe with a set of functions like:

```
uint64ToUint32(...)
binaryToString(arrayBuffer) => str
...
```

floryst:
would feat(IO) be a better commit message prefix?

jourdain:
stick with fix()... Unless it is a "big" thing...

floryst:
Sounds good.

Sources/IO/XML/XMLReader/index.js
floryst: I've added an initial IO/Core/BinaryHelper.

Sources/IO/XML/XMLReader/index.js
```diff
@@ -307,7 +319,117 @@ function vtkXMLReader(publicAPI, model) {
...
+          const offset = Number(arrays[i].getAttribute('offset'));
+          let nextOffset = 0;
+          if (i === arrays.length - 1) {
+            nextOffset = model.binaryBuffer.byteLength;
+          } else {
+            nextOffset = Number(arrays[i + 1].getAttribute('offset'));
+          }
+
+          // need to slice here otherwise readerHeader breaks
+          const uint8 = new Uint8Array(model.binaryBuffer.slice(offset, nextOffset));
```
jourdain: can't you do a view instead of slice?

Sources/IO/XML/XMLReader/index.js
```diff
@@ -290,11 +341,27 @@ function vtkXMLReader(publicAPI, model) {
       return;
     }
 
+    // appended format
+    if (rootElem.querySelector('AppendedData')) {
```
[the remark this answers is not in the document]
floryst: Done (works with all the zlib test cases I have)

jourdain:
Are we ready to merge?

floryst:
Yeah, no more edits on my side.

PR #450, opened 2017-12-07
feat(TextureMapToSphere): Add the vtkTextureMapToSphere filter

laurennlam:


jourdain:
LGTM, should we merge? I guess that code come from the C++?

@martinken what do you think...

martinken:
Sure +1 Nice filter to have.

laurennlam:
It's ready to merge @jourdain . Indeed, the code comes from C++ filter.

PR #451, opened 2017-12-08
fix(OpenGLRenderWindow): Prevent invalid size

jourdain:


jourdain:
@floryst can you test if that fix the crash?

@martinken do you think we should add a guard like I did? Should that run on each `modified` or should we do it in a custom setSize() method?

floryst:
This didn't fix my weird crash :( webgl still crashes.

PR #456, opened 2017-12-11
fix(vtkTextureMapToSphere): keep points and cells data

laurennlam:
Set the right number of components when points are set.
Make a deep copy of points and cells data to not loose it.


jourdain:
Why do you deep copy... VTK filters always do shallow copy to prevent unnecessary memory usage...

PR #457, opened 2017-12-11
perf(Rendering): support ANGLE_instanced_arrays

martinken:
Support instacing on Webgl1 when the extension is
present. In spite of being named ANGLE many non
angle based browsers support this extension.


martinken:
@jourdain Did this really fast, but it seems to work on chrome windows (which is webgl1) and I think it may work on safari as well.

PR #458, opened 2017-12-11
Add vr  manipulator

jourdain:


martinken:
@jourdain  BTW this ate part of my MR. IO'll push a fix

PR #460, opened 2017-12-12
feat(Sources) Added new source for Cylinder

sankhesh:
Implemented faceted polygons for sides and caps, normals and texture
coordinates for cylinder source. Added API documentation, test and example.


sankhesh:
@martinken @jourdain This is the first step for an arrow source.

jourdain:
LGTM, how was your first impression with vtk.js?

Should we merge?

jourdain:
Thanks for the feedback... I would be curious about the convoluted part. But we can talk about it offline.

PR #470, opened 2017-12-28
test(testUtils): Show a difference image in the test log

sankhesh:
When an image comparison test fails, the recorder will now also record
the difference image in Documentation/content/coverage/tests.md

### DIFF
--- a/Sources/Testing/testUtils.js
+++ b/Sources/Testing/testUtils.js
@@ -12,6 +12,7 @@ function compareImages(
 ) {
   const nbBaselines = baselines.length;
   let minDelta = 100;
+  let minDiff = '';
   let isSameDimensions = false;
 
   function done() {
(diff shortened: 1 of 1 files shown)


sankhesh:
@jourdain Please review

Sources/Testing/testUtils.js
```diff
@@ -5,14 +5,15 @@ let REMOVE_DOM_ELEMENTS = true;
 function compareImages(image, baselines, testName, tapeContext, threshold = 0.5, nextCallback = null) {
   const nbBaselines = baselines.length;
   let minDelta = 100;
+  let minDiff = new Image();
```
jourdain: minDiff is a string that represent an encoded image, therefore, you can not initialize it as an image. ;-)

`let minDiff = '';`

jourdain:
This should not be a feat() as it only affect testing not the code. Every fix() and feat() will trigger a new release.

```
fix() => +0.0.1
feat() => +0.1.0
```

jourdain:
Since that affect only tests and not the code, this should not trigger any new release. Therefore is can not be a feat() or a fix()...

sankhesh:
I see. I thought you wanted me to change the `feat` to `fix`. 

So looking at: 

commitizen

Should this be a `chore` or `refactor` (or can it be anything but `feat` or `fix`)?

jourdain:
Seems that `test()` is better fit.

PR #478, opened 2018-01-03
fix(Rendering): fix volume visibility

martinken:
The visibility ivar was not being honored for volumes

Also update image for glyph3d, not sure what the diff was
as I could see nothing different.

Also remove unecc import from append filter


martinken:
@jourdain Any idea what is up with travis?

PR #485, opened 2018-01-05
fix(OpenGLImageMapper): shader modification to create a vec4

laurennlam:
When trying to add an actor with an image with 4 components, the shader can't build. It seemed that the vec4 constructor wasn't good. 
Does @martinken can confirm ?


martinken:
+1

PR #493, opened 2018-01-09
Fix incorrect interpretation of last block size in zLib+appended vt* formats

floryst:
This only affects data files whose data arrays' uncompressed sizes align with 32768 (or whatever the uncompressed block size is specified in the compression header).


floryst:
This should fix your VTI issues @agirault

agirault:
👍

PR #494, opened 2018-01-09
Priority event listeners and event abort

floryst:
This adds priorities to event listeners, as well as capability for event listeners to abort the continuation/propagation of a vtkjs event.

### DIFF
--- a/Sources/macro.js
+++ b/Sources/macro.js
@@ -757,7 +757,7 @@ export function event(publicAPI, model, eventName) {
 
     const callbackID = curCallbackID++;
     callbacks.push([callbackID, callback, priority]);
-    callbacks.sort(([, cb1, pri1], [, cb2, pri2]) => pri2 - pri1);
+    callbacks.sort((cb1, cb2) => cb2[2] - cb1[2]);
     return on(callbackID);
   };
 
--- a/Sources/macro.js
+++ b/Sources/macro.js
@@ -755,7 +755,7 @@ export function event(publicAPI, model, eventName) {
 
   publicAPI.delete = () => {
     previousDelete();
-    callbacks.forEach((el, index) => off(index));
+    callbacks.forEach(([cb]) => off(cb));
   };
 }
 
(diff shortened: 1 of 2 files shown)


Sources/macro.js
```diff
@@ -722,30 +733,37 @@ export function event(publicAPI, model, eventName) {
...
       vtkErrorMacro('instance deleted - cannot call any method');
       return null;
     }
 
-    const index = callbacks.length;
-    callbacks.push(callback);
-    return on(index);
+    const callbackID = curCallbackID++;
+    callbacks.push([callbackID, callback, priority]);
+    callbacks.sort(([, cb1, pri1], [, cb2, pri2]) => pri2 - pri1);
```
jourdain: your sort does not use cb1 and cb2...

jourdain: what about doing it like that

```
callbacks.sort((a, b) => b[2] - a[2]);
```

With maybe the function be only defined once so it does not get created at each call?

floryst: The comparator is small enough do not need a separate function. Will there be a measurable performance hit? Maybe if on* gets called a lot...

jourdain: I agree, and I'm not expecting perf hit.

floryst: :+1:

floryst:
Are there any docs for macro.js that I can update?

jourdain:
No we have the big comment on each macro function. But it would be nice to have a markdown for it.

jourdain:
Maybe a full markdown could make sense next to https://kitware.github.io/vtk-js/docs/develop_class.html which can be mentioned in that class doc as well.

floryst:
Hm, I was thinking an `api.md` sort of file, but the `develop_class.md` could link to it. How does that sound?

jourdain:
api.md would be good, but macro is special in the sense that's not a class. Hence I was thinking `./Documentation/content/docs/develop_macro.md` instead that we can put in the menu of `Docs` as well as referencing it inside the `develop_class.md`.

jourdain:
Should we merge it? Or do you want to push the doc at the same time?

PR #496, opened 2018-01-10
Add crop widget

floryst:
Interactive crop widget.


Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
```diff
@@ -0,0 +1,384 @@
...
+      publicAPI.modified();
+    }
+  };
+
+  publicAPI.setProperty = (property) => {
+    model.actor.setProperty(property);
+  };
+
+  // invoke the constructor after setting up public methods
+  constructor();
```
jourdain: Why don't you inline the constructor content right there?

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
```diff
@@ -0,0 +1,384 @@
...
+    model.actor.getProperty().setOpacity(opacity);
+    updateCenterOpacity();
+  };
+
+  publicAPI.setEdgeColor = (...edgeColor) => {
+    model.edgeColor = edgeColor;
+    model.actor.getProperty().setEdgeColor(...model.edgeColor);
+  };
+
+  publicAPI.setPlanePositions = (...positions) => {
```
jourdain: Should set methods call `publicAPI.modified()`?

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
```diff
@@ -0,0 +1,384 @@
...
+  };
+
+  publicAPI.setOpacity = (opacityValue) => {
+    const opacity = Math.max(0, Math.min(1, opacityValue));
+    model.opacity = opacity;
+    model.actor.getProperty().setOpacity(opacity);
+    updateCenterOpacity();
+  };
+
+  publicAPI.setEdgeColor = (...edgeColor) => {
```
jourdain: Should set methods call `publicAPI.modified()`?

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
```diff
@@ -0,0 +1,384 @@
...
+  };
+
+  publicAPI.setSliceOrientation = (sliceOrientation) => {
+    if (sliceOrientation !== model.sliceOrientation) {
+      model.sliceOrientation = sliceOrientation;
+      publicAPI.updateGeometry();
+    }
+  };
+
+  publicAPI.setOpacity = (opacityValue) => {
```
jourdain: Should set methods call `publicAPI.modified()`?

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
```diff
@@ -0,0 +1,384 @@
...
+  };
+
+  publicAPI.setSlice = (slice) => {
+    if (slice !== model.slice) {
+      model.slice = slice;
+      publicAPI.updateGeometry();
+    }
+  };
+
+  publicAPI.setSliceOrientation = (sliceOrientation) => {
```
jourdain: Should set methods call `publicAPI.modified()`?

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
```diff
@@ -0,0 +1,384 @@
...
+      (newBounds[1] - newBounds[0]) * (newBounds[1] - newBounds[0]) +
+        (newBounds[3] - newBounds[2]) * (newBounds[3] - newBounds[2]) +
+        (newBounds[5] - newBounds[4]) * (newBounds[5] - newBounds[4])
+    );
+
+    // set plane positions
+    publicAPI.setPlanePositions(...model.initialBounds);
+  };
+
+  publicAPI.setSlice = (slice) => {
```
jourdain: Should set methods call `publicAPI.modified()`?

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
floryst: Having the constructor code at the top was just a preference.

jourdain: What about having a `initialize(publicAPI, model)` outside of the function, at the very top of the file?

jourdain: either way, I'm fine, I was just curious regarding that closure.

floryst: I've been treating closures inside the main method as private methods that can access private variables in the upper scope of the closures. I don't have any private variables in this class, but I still wrote the code like that for consistency. :P

floryst:
Re: modified() calls: Thanks, forgot about modified(). I've added modified() calls such that all relevant public methods call modified().

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
jourdain: only if `model.opacity !== opacity`...

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
jourdain: A function inside a module is private too...

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
floryst: fixed

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
```diff
@@ -0,0 +1,390 @@
...
+      model.actor.getProperty().setOpacity(opacity);
+      updateCenterOpacity();
+      publicAPI.modified();
+    }
+  };
+
+  publicAPI.setEdgeColor = (...edgeColor) => {
+    model.edgeColor = edgeColor;
+    model.actor.getProperty().setEdgeColor(...model.edgeColor);
+    publicAPI.modified();
```
jourdain: only if different...

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
```diff
@@ -0,0 +1,390 @@
...
+
+    // This requires updating whenever geometry updates
+    updateCenterOpacity();
+
+    model.regionPolyData.modified();
+    model.centerRegionPolyData.modified();
+
+    publicAPI.invokePlanesPositionChanged();
+    publicAPI.modified();
+  };
```
jourdain: Should we call modified() if the geometry did not actually changed?

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
```diff
@@ -0,0 +1,390 @@
...
+
+  publicAPI.getBounds = () => model.initialBounds;
+
+  publicAPI.buildRepresentation = () => {
+    if (model.renderer) {
+      if (!model.placed) {
+        model.validPick = 1;
+        model.placed = 1;
+      }
+      publicAPI.modified();
```
jourdain: always?

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
floryst: Agreed. I put those private methods alongside private variables (like in ImageCropRegionsWidget) in the main function, since I opted to not put the variables in the module scope.

I can still go ahead and make changes if you'd like, since I'm not partial to any one approach and it's just what convention we want to go with at this point.

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
floryst: A call to updateGeometry always assumes the geometry has changed, since I didn't think it would matter if the geometry actually didn't change.

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
jourdain: I think we use helper functions at the module scope sometime when needed as private function that only needs to be define once.

Since I like you comment ` invoke after setting up public methods`, I can see the initialization process done in a function. But I would put it at the module level with a name like `initialize(publicAPI, model)` or `setup(publicAPI, model)` or something else. But I don't like `constructor()` as it is like an ES6 class which we are not and that can be confusing.

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
floryst: I'm not too sure with buildRepresentation, since that's called by AbstractWidget. Other buildRepresentations call modified(), so I went along with it, even though I build the representation in the constructor rather than here. Maybe I can move that code to buildRepresentation now that I think about it...

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
floryst: Ah, agreed with the ES6 class naming clash there. I'm thinking of moving the representation construction to buildRepresentation as per the comment I made [here](https://github.com/Kitware/vtk-js/pull/496#discussion_r160814080), which would remove the need to have a separate `initialize(publicAPI, model)` function.

Sources/Interaction/Widgets/ImageCroppingRegionsRepresentation/index.js
jourdain: You know the code better than I do. I let you do another path to clean it up. I was just mentioning thing that might be important such the modified only when truly required.

floryst: I'll do another pass. I saw LineRepresentation constructed the model inside extend(), which I found odd so I opted to put the construction elsewhere.

jourdain: Yes model should not be filled inside extend() unless it is 3 lines...

floryst:
I've made some reorganization changes.

PR #500, opened 2018-01-12
fix(Rendering): fix imagemapper depth pass

martinken:
The imagemapper was not handling the depth pass so
implement that. At the same time improve the polydatamapper
depth pass to not rebuild shaders every frame when
depth passes are requested. Instead it rebuilds the shader
the first time a depth pass is requested and then uses
a uniform from that point forward.


jourdain:
```
/home/travis/build/Kitware/vtk-js/Sources/tests.js
  46:8  error  Unable to resolve path to module './Rendering/OpenGL/VolumeMapper/test/testIntermixedImage'  import/no-unresolved
  46:8  error  Missing file extension for "./Rendering/OpenGL/VolumeMapper/test/testIntermixedImage"        import/extensions
✖ 2 problems (2 errors, 0 warnings)
```

PR #502, opened 2018-01-15
feat(cameraManipulator): expose pinch/scroll event to manipulators

agirault:
- [x] allow zoom on pinch in manipulator
- [x] add flag to `CameraManipulator` to only call `onPinch` when wanted
- [x] update camera manipulator example to show scroll capabilities

screen shot 2018-01-15 at 6 39 06 pm


agirault:
@jourdain I currently have added a flag for pinch, but maybe it would make sense to use button instead (4), so that we can also use `findManipulator` and have different behaviors by mixing scroll with Shift, Control and/or Alt

PR #505, opened 2018-01-17
SliceManipulator and ImageMapper improvements

agirault:
- add slicing manipulator: vtkSliceManipulator
- improve ImageMapper to:
  - clamp slicing to the volume bounds
  - allow slicing from X, Y and Z axis from the camera

### DIFF
--- a/Sources/Interaction/Manipulators/SliceManipulator/index.js
+++ b/Sources/Interaction/Manipulators/SliceManipulator/index.js
@@ -1,0 +1,109 @@
+import macro from 'vtk.js/Sources/macro';
+import vtkCameraManipulator from 'vtk.js/Sources/Interaction/Manipulators/CameraManipulator';
+import vtkMath from 'vtk.js/Sources/Common/Core/Math';
+
+// ----------------------------------------------------------------------------
+// vtkSliceManipulator methods
+// ----------------------------------------------------------------------------
+
+function vtkSliceManipulator(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkSliceManipulator');
+
+  publicAPI.onAnimation = (interactor, renderer) => {
+    const lastPtr = interactor.getPointerIndex();
+    const pos = interactor.getAnimationEventPosition(lastPtr);
+    const lastPos = interactor.getLastAnimationEventPosition(lastPtr);
+
+    if (!pos || !lastPos || !renderer) {
+      return;
+    }
+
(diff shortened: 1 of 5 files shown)


jourdain:
Already pushed and merged your fix regarding the `lights geometry to follow camera`

Sources/Interaction/Manipulators/SliceManipulator/index.js
```diff
@@ -0,0 +1,111 @@
...
+    distance += delta;
+
+    // clamp the distance to the clipping range
+    if (distance < range[0]) {
+      distance = range[0];
+    }
+    if (distance > range[1]) {
+      distance = range[1];
+    }
+    camera.setDistance(distance);
```
agirault: @jourdain @floryst @martinken what are your thoughts on this?

This is the same as seen in [`vtkInteractorStyleImage`](https://github.com/Kitware/vtk-js/blob/b807950244e1ed7db1ae6858c6829ce33bc32786/Sources/Interaction/Style/InteractorStyleImage/index.js#L261-L268): this ends up calling [`vtkImageMapper.setZSliceFromCamera`](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ImageMapper/index.js#L19-L25) where the focal point (defined here by `distance`) is converted into a slice index to display.

The issue I see is that this distance is clamped by the clipping range of the camera, which I assume should match the bounds of the volume. However, when testing in the `imageMapper` example ([here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ImageMapper/example/index.js#L36-L38), either with that interactorStyle or with the slice manipulator), I see that the z slice index goes between [-18; 216], even though the bounds of that image are [0; 200].

Thoughts? Also, does that only work for the z axis? We could reclamp the camera focalpoint by checking if the index is out of bounds in `setZSliceFromCamera`, but that doesn't seem clean to me.

agirault:
> Already pushed and merged a fix regarding the lights geometry to follow camera

Beat me to it! Rebasing my branch.

jourdain:
Like your changes to support any axis.

Sources/Interaction/Manipulators/SliceManipulator/index.js
agirault: After discussion with @jourdain , we will leave this as is, and implement a `rangeManipulator` that will be generic enough to control scalar values (like slice, window, level), from horizontal drag, vertical drag, or pinch.

PR #509, opened 2018-01-18
fix(imageMapper): fix example by setting slicing mode

agirault:
Default slice mode was `None`. Since d0bec193, the slicing from
camera focal point is based on the slicing mode. We therefore
use the slicing mode in this example as well as adjust the camera
orientation to face the slicing axis.


agirault:
do not merge: error in viewup

jourdain:
Since you are only fixing an example your commit should not trigger a new version as the code that will be bundle will be the same as before.

Therefore, your commit message should be:

`doc(ImageMapperExample): Fix slice mode issue`

agirault:
@jourdain let me know if something else's missing.

PR #510, opened 2018-01-18
Orientation marker: better viewport

floryst:
This implements a bounded square viewport for the orientation marker.

### DIFF
--- a/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
+++ b/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
@@ -19,10 +19,12 @@ function vtkOrientationMarkerWidget(publicAPI, model) {
 
   publicAPI.computeViewport = () => {
     const [viewXSize, viewYSize] = model.interactor.getView().getSize();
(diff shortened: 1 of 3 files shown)


Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
```diff
@@ -81,20 +86,25 @@ function vtkOrientationMarkerWidget(publicAPI, model) {
...
+      publicAPI.updateMarkerOrientation();
+
       model.enabled = true;
     } else {
       if (!model.enabled) {
         return;
       }
       model.enabled = false;
 
+      window.removeEventListener('resize', publicAPI.updateViewport);
```
jourdain: I don't think we want to bind listener here...

jourdain: wait... I see... It might be fine, but better we may be able to listen to vtkXXXObject modified or if we have a size change event. That way, we don't rely on the window...

If the user update the size by hand for example...

jourdain:
Looks good, but I don't know about my comment if we can do better with a more abstract way to listen to window size change...

jourdain:
LGTM

Issue #515, opened 2018-01-19
Glitchy volume rendering with multiple renderers and an image slice


floryst:
Volume rendering is all glitchy when rendered alongside an image slice and a RenderWindow with more than one renderer. The video below demonstrates this issue, where there is one image slice, one volume, and the black rectangle in the lower left is another Renderer.

![image-volume-multirenderer-bug](https://user-images.githubusercontent.com/1812167/35165489-79b0380e-fd1c-11e7-8c9d-55fd28d5f758.gif)

This was tested on the ImageMapper example, modified to add another Renderer to the RenderWindow and to add a Volume to the scene.

PR #519, opened 2018-01-22
fix(Renderer): Update gl viewport on zbuffer pass

floryst:
This fixes issue #515. Thanks @martinken for debugging this.


martinken:
+1

PR #520, opened 2018-01-23
Tube filter

sankhesh:


jourdain:
Can you rebase on master and fix the conflicts?

Sources/Filters/General/AppendPolyData/index.js
```diff
@@ -4,9 +4,9 @@ import vtkPoints        from 'vtk.js/Sources/Common/Core/Points';
 import vtkPolyData      from 'vtk.js/Sources/Common/DataModel/PolyData';
 import { VtkDataTypes } from 'vtk.js/Sources/Common/Core/DataArray/Constants';
 
-import Constants        from 'vtk.js/Sources/Filters/General/AppendPolyData/Constants';
+import Constants        from 'vtk.js/Sources/Filters/General/Constants';
 
-const { PointPrecision } = Constants;
+const { VtkPointPrecision } = Constants;
```
jourdain: Your constant should not start with `Vtk` [see](https://github.com/Kitware/vtk-js/blob/master/Sources/Common/DataModel/DataSet/Constants.js)

Sources/Filters/General/AppendPolyData/index.js
```diff
@@ -4,9 +4,9 @@ import vtkPoints        from 'vtk.js/Sources/Common/Core/Points';
 import vtkPolyData      from 'vtk.js/Sources/Common/DataModel/PolyData';
 import { VtkDataTypes } from 'vtk.js/Sources/Common/Core/DataArray/Constants';
 
-import Constants        from 'vtk.js/Sources/Filters/General/AppendPolyData/Constants';
+import Constants        from 'vtk.js/Sources/Filters/General/Constants';
```
jourdain: Is that constant supposed to be at the module level `Filters/General`?
I think they should belong to a class like `PointSets`?

sankhesh: Well, technically it is supposed to be an algorithm enum like [vtkAlgorithm::DesiredOutputPrecision](https://www.vtk.org/doc/nightly/html/classvtkAlgorithm.html#ac2bf3c779aa60908da5bc3df9b066db6) 

That's why I put it at the module level so that all ***general filters*** have access to it.

Sources/Filters/General/AppendPolyData/index.js
jourdain: The VtkDataTypes list the naming for the VTK Types that an vtkDataArray can contain, hence the VTK in it. But it you look in that same file, the other constants don't have Vtk as prefix.

sankhesh: I see. Cool, I'll fix it when rebasing.

Sources/Filters/General/AppendPolyData/index.js
jourdain: I see, but we don't have a base class for vtkAlgorithm... ;-)

Then maybe we want to add a class vtkAlgorithm which only expose its constants. Is there any other constants in that class/enum? If that's the only one, I'm wondering if we should find a better place...

PR #534, opened 2018-01-25
Expose orientation in ViewProxy

agirault:
- Move it from View2DProxy to ViewProxy.
- Use superUpdateOrientation from View2DProxy then update slicing mode


jourdain:
Can you try that? https://github.com/Kitware/vtk-js/pull/535

PR #542, opened 2018-01-29
fix(BinaryHelper): Use TextDecoder for arraybuffer->string

floryst:
This also removes some old code that likely doesn't persist arraybuffers in memory, but might as well remove them just in case.


jourdain:
LGTM

PR #549, opened 2018-01-30
Abstract widget events handling

agirault:
- [x] rename `eventsWeHandle` to `handledEvents` and expose them from the interactor
- [x] consolidate `setInteractor` and `listenEvents` in AbstractWidget
- [x] abort events when needed in HandleWidget
- [x] fix callback infinite loop in `macro.invoke`
- [x] ensure callbacks are not stored twice

### DIFF
--- a/Sources/Interaction/Widgets/AbstractWidget/index.js
+++ b/Sources/Interaction/Widgets/AbstractWidget/index.js
@@ -20,6 +20,13 @@ function vtkAbstractWidget(publicAPI, model) {
       vtkErrorMacro('The interactor must be set before listening events');
       return;
     }
+
+    // Remove current events
+    while (model.unsubscribes.length) {
+      model.unsubscribes.pop().unsubscribe();
+    }
+
+    // Check what events we can handle and register callbacks
     vtkRenderWindowInteractor.handledEvents.forEach((eventName) => {
       if (publicAPI[`handle${eventName}`]) {
         model.unsubscribes.push(
@@ -36,17 +43,9 @@ function vtkAbstractWidget(publicAPI, model) {
     if (i === model.interactor) {
       return;
     }
-
-    // if we already have an Interactor then stop observing it
-    if (model.interactor) {
-      while (model.unsubscribes.length) {
-        model.unsubscribes.pop().unsubscribe();
-      }
-    }
-
     model.interactor = i;
 
-    if (i) {
+    if (i && model.enabled) {
       publicAPI.listenEvents();
     }
 
@@ -60,11 +59,11 @@ function vtkAbstractWidget(publicAPI, model) {
   };
 
   publicAPI.setEnable = (enable) => {
+    if (enable === model.enabled) {
+      return;
+    }
+
     if (enable) {
-      if (model.enabled) {
-        // widget already enabled
-        return;
-      }
       if (!model.interactor) {
         vtkErrorMacro(
           'The interactor must be set prior to enabling the widget'
@@ -97,13 +96,9 @@ function vtkAbstractWidget(publicAPI, model) {
       model.widgetRep.buildRepresentation();
       model.currentRenderer.addViewProp(model.widgetRep);
     } else {
-      if (!model.enabled) {
-        // already
-        return;
-      }
       model.enabled = 0;
 
-      // Don't listen events
+      // Don't listen to events
       while (model.unsubscribes.length) {
         model.unsubscribes.pop().unsubscribe();
       }
--- a/Sources/Interaction/Widgets/AbstractWidget/index.js
+++ b/Sources/Interaction/Widgets/AbstractWidget/index.js
@@ -1,5 +1,6 @@
 import macro from 'vtk.js/Sources/macro';
 import vtkInteractorObserver from 'vtk.js/Sources/Rendering/Core/InteractorObserver';
+import vtkRenderWindowInteractor from 'vtk.js/Sources/Rendering/Core/RenderWindowInteractor';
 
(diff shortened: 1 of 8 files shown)


agirault:
wip: issue with infinite loop when adding a widget when in a callback (example: add a handlewidget when right mouse press)
@floryst

agirault:
@floryst @martinken @jourdain

Sources/Interaction/Widgets/LineWidget/index.js
```diff
@@ -245,11 +238,13 @@ function vtkLineWidget(publicAPI, model) {
 
     publicAPI.invokeInteractionEvent();
     publicAPI.render();
+
+    return EVENT_ABORT;
```
agirault: @floryst @jourdain would you know if i'm missing a `return VOID` somewhere in `LineWidget.moveAction` ?

agirault:
@jourdain like seen with @martinken, it seems all callbacks are pushed twice to the callback array, so I'll investigate that, could put it in this PR.

jourdain:
LGTM assuming that works.. ;-)

Sources/Interaction/Widgets/LineWidget/index.js
jourdain: The VOID is just for the linter in case of multiple return path. As long as you don't intent to ABORT, you are good regardless of what you return.

Sources/Interaction/Widgets/AbstractWidget/index.js
```diff
@@ -14,8 +15,43 @@ function vtkAbstractWidget(publicAPI, model) {
...
+      }
+    }
+
+    model.interactor = i;
+
+    if (i) {
+      publicAPI.listenEvents();
+    }
+
+    publicAPI.modified();
```
agirault: @floryst need modified here?

Sources/Interaction/Widgets/LineWidget/index.js
agirault: That's what I'm wondering: shouldn't there be a case in LineWidget.moveAction where we should not abort? Like [here](https://github.com/Kitware/vtk-js/pull/549/files#diff-99b12f1ff9fb5fe232106e92d9054025R159) for HandleWidget.

jourdain: @martinken

Sources/Interaction/Widgets/AbstractWidget/index.js
```diff
@@ -14,8 +15,43 @@ function vtkAbstractWidget(publicAPI, model) {
...
+    // if we already have an Interactor then stop observing it
+    if (model.interactor) {
+      while (model.unsubscribes.length) {
+        model.unsubscribes.pop().unsubscribe();
+      }
+    }
+
+    model.interactor = i;
+
+    if (i) {
```
floryst: Also check if model.enabled is true here

Sources/Interaction/Widgets/AbstractWidget/index.js
floryst: I think it's better to have it in case an onModified() listener cares that the interactor changed.

agirault:
Found the source of the duplicated callbacks: see explanations and fix at https://github.com/Kitware/vtk-js/pull/549/commits/63554fbfe28c7877c28a2e6a8dfaa97ab239b528

I just need to ensure the LineWidget is only aborting event forwarding when needed, and it should be good to go.

Sources/Interaction/Widgets/LineWidget/index.js
floryst: I think we should only abort when the widgetState is not START.

agirault: Should we still call `invokeInteractionEvent` and `render` if START? We only do render for the handlewidget: see https://github.com/agirault/vtk-js/blob/63554fbfe28c7877c28a2e6a8dfaa97ab239b528/Sources/Interaction/Widgets/HandleWidget/index.js#L150-L159

agirault: Testing things now.

Sources/Interaction/Widgets/LineWidget/index.js
```diff
@@ -141,37 +95,23 @@ function vtkLineWidget(publicAPI, model) {
...
-  };
+  publicAPI.handleRightButtonPress = () => publicAPI.scaleAction();
 
-  publicAPI.handleRightButtonRelease = () => {
-    publicAPI.endSelectAction();
-  };
+  publicAPI.handleRightButtonRelease = () => publicAPI.endSelectAction();
 
   publicAPI.selectAction = () => {
     if (model.widgetRep.getInteractionState === InteractionState.OUTSIDE) {
```
floryst: These are supposed to be calls to getInteractionState, otherwise this will always evaluate to false. The same goes with `publicAPI.translateAction` and `publicAPI.scaleAction`. This isn't related to the PR, but I'm making a note of this as this affects discussion of event abortion in `moveAction`.

floryst: And actually, fixing this breaks LineWidget.

agirault:
@jourdain ready to go. LineWidget has other inherent issues that I could address separately, I'll open an issue for it on github.

Sources/Interaction/Widgets/LineWidget/index.js
agirault: Let's create an issue where we can list problems with the line widget. Thanks @floryst.

PR #556, opened 2018-02-02
Refactor interactor

agirault:
- [x] Clean up unused functions in InteractorObserver
- [x] Improve switch to animation states in interactorStyle
- [x] Remove events not handled in Interactor
- [x] Consolidate setInteractor and setEnabled in InteractorObserver
- [x] Improve event state recording in the Interactor
- [x] Correctly store event positions in the interactor based on canvas
- [x] Fix issues in LineWidget and its representations
- [x] Pass event states as call data to interactor observers
- [x] Fix touch interaction

BREAKING CHANGES:
https://github.com/agirault/vtk-js/blob/0763417b649f9a5f5704d34a818f4831f5c9b39a/BREAKING_CHANGES.md#from-5x-to-6

### DIFF
--- a/Sources/Interaction/Manipulators/MouseManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseManipulator/index.js
@@ -10,11 +10,14 @@ function vtkMouseManipulator(publicAPI, model) {
 
   publicAPI.startInteraction = () => {};
   publicAPI.endInteraction = () => {};
-  publicAPI.onButtonDown = (interactor) => {};
+  publicAPI.onButtonDown = (interactor, renderer, position) => {};
   publicAPI.onButtonUp = (interactor) => {};
-  publicAPI.onAnimation = (interactor, renderer) => {};
-  publicAPI.onKeyUp = (interactor) => {};
-  publicAPI.onKeyDown = (interactor) => {};
+  publicAPI.onMouseMove = (interactor, renderer, position) => {};
+  publicAPI.onKeyUp = (interactor, key) => {};
+  publicAPI.onKeyDown = (interactor, key) => {};
+  publicAPI.onStartScroll = (interactor, renderer, delta) => {};
+  publicAPI.onScroll = (interactor, renderer, delta) => {};
+  publicAPI.onEndScroll = (interactor) => {};
 }
 
 // ----------------------------------------------------------------------------
@@ -28,7 +31,7 @@ const DEFAULT_VALUES = {
   shift: false,
   control: false,
   alt: false,
-  pinch: false,
+  scroll: false,
 };
 
 // ----------------------------------------------------------------------------
@@ -46,7 +49,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'shift',
     'control',
     'alt',
-    'pinch',
+    'scroll',
   ]);
 
   // Object specific methods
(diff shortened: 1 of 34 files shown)


agirault:
@jourdain @floryst @ken FYI, making some good progress on this.

The next step, now that the observers (styles + widget) and the interactor are better setup and organized, would be to organize data from events in state structures, and pass them down to the observers as calldata. Could do this in a separate PR, but that might create other breaking changes (maybe?) so if that was an issue I could keep working on that branch/PR.

Sources/Interaction/Manipulators/MouseManipulator/index.js
```diff
@@ -28,7 +29,7 @@ const DEFAULT_VALUES = {
   shift: false,
   control: false,
   alt: false,
-  pinch: false,
+  scroll: false,
 };
```
jourdain: As I told you, since you are refactoring it, pinch or scroll are not states to test against. They are more a mode that you want to be enabled against. Therefore, you can not keep the same naming. Maybe `enableMouseDrag`, `enableScroll` and `enablePinch`? Where by default all are false except `enableMouseDrag`...

jourdain:
It is hard to review and I trust your judgment on that. I've made a little comment to try to clarify some conflicting concept. 

Also make sure in your commit messages are properly tagged with each breaking changes so we will see them reported on the changelog. For that you can use `npm run commit`

Sources/Interaction/Widgets/AbstractWidget/index.js
```diff
@@ -11,105 +8,41 @@ const { vtkErrorMacro } = macro;
...
-    } else {
-      model.enabled = 0;
-
-      // Don't listen to events
-      while (model.unsubscribes.length) {
-        model.unsubscribes.pop().unsubscribe();
-      }
-
-      if (model.currentRenderer) {
-        model.currentRenderer.removeViewProp(model.widgetRep);
```
agirault: Should we still hide it when disabling?

jourdain: don't know what does VTK do?

floryst: AFAIK VTK removes the prop like above.

Interestingly, LineWidget uses setEnabled(false) to turn off the widget, and then re-adds the representation into the scene. This has the effect of disabling widget interaction while keeping the widget in-scene. Maybe we can add a method like `setInteractionEnabled` so widgets can disable interactions, but keep the representation in the renderer?

Sources/Interaction/Manipulators/MouseManipulator/index.js
agirault: this PR is already huge, can we keep that for later? So that the behavior just doesn't change?

jourdain: ok, do you want to merge to a branch `next` and create new PR. Then when we feel that `next` is ready we can merge it to master. Or do you rather merge that PR into master before doing any other work/PR?

Sources/Interaction/Widgets/AbstractWidget/index.js
agirault: For the next round of things :)

agirault:
@martinken @floryst @jourdain Ready to go!

About 600 lines of code removed altogether, with:
- refactored/simplified observers
- event data passed through callbacks
- fixed position in interactor relative to the canvas
- fixed line widget
- fixed touch interactions
- all examples working
- passing tests

What I did not test was VR.

I listed the breaking changes in each commit (sometimes multiple per commits, hard to separate), and in the description of this PR above. @jourdain I also need to add them to a README, correct?

jourdain:
Just update https://github.com/Kitware/vtk-js/blob/master/BREAKING_CHANGES.md

I'll test the VR tomorrow.

PR #557, opened 2018-02-02
fix(macro): Add keystore extension to vtk objects

floryst:


floryst:
The keystore is basically an ES6 Map. (ES5? I forget) Do we want to just forward all Map methods?

Sources/macro.js
```diff
@@ -1200,6 +1200,37 @@ export function proxyPropertyState(
...
+//
+// Generated API
+//  setKey(key, value) : undefined
+//  getKey(key) : mixed
+//  getAllKeys() : [mixed]
+//  deleteKey(key) : Boolean
+// ----------------------------------------------------------------------------
+
+export function keystore(publicAPI, model, initialKeystore = {}) {
+  const store = new Map();
```
jourdain: Do we really need a Map? Would it be good enough to use `{}` instead?

Do we want it stored on the `model`?

jourdain:
Then we should add a call to it inside the macro.proxy

Sources/macro.js
floryst: I tend to use Map for key/value stores. Plus, if someone really wants to use an object as a key, they can with Map. :P Is there a reason to use `{}` instead?

Yes, storing on the model would be nice. I'll do that.

jourdain: I understand, but then you break the current pattern that enable serialization on any vtkObject:

```
const str = obj.getState();
const newObj = vtk(str);
```

Also you would have to do something like

```
if (!model.keystore) {
   model.keystore = {};
}
```

floryst: Oh, didn't know about object serialization. In that case yes, I'll switch to using an object.

PR #558, opened 2018-02-03
Lut data range rescale

jourdain:


jourdain:
FYI @thewtex

Sources/Interaction/UI/VolumeController/index.js
```diff
@@ -251,6 +259,15 @@ function vtkVolumeController(publicAPI, model) {
     }
   };
 
+  const rescaleColorMap = publicAPI.setRescaleColorMap;
+  publicAPI.setSescaleColorMap = (value) => {
```
thewtex: `setSescaleColorMap`?  What is this function doing?

jourdain: oups, should be `setRescaleColorMap`

jourdain: true/false so either the full data range is used for the color map or only the opacity range.

thewtex: Ah, thanks for the explanation :+1:

jourdain: Do you mind fixing it? thx!

thewtex: Yes, will do :-)

PR #565, opened 2018-02-07
fix(ProxyManager): Add proxy registration event

floryst:
Adds a `proxyRegistrationChange` event to the Proxy manager.


jourdain:
can you add in those event the proxy name and group?

floryst:
Added

PR #571, opened 2018-02-08
fix(InteractorStyleTrackballCamera): Fix import of vtkInteractorStyle…

thewtex:
… States

Addresses:

itkVtkImageViewer.js:61588 Uncaught TypeError: Cannot read property 'States' of undefined
   at Object.defineProperty.value (itkVtkImageViewer.js:61588)
var States = _InteractorStyle2.default.States;


thewtex:
CC: @jourdain

PR #575, opened 2018-02-09
Improve animation + clipping

agirault:


Sources/Proxy/Core/AbstractRepresentationProxy/index.js
```diff
@@ -302,7 +302,7 @@ function extend(publicAPI, model, initialValues = {}) {
   Object.assign(model, DEFAULT_VALUES, initialValues);
 
   macro.obj(publicAPI, model);
-  macro.get(publicAPI, model, ['input', 'actors', 'volumes']);
+  macro.get(publicAPI, model, ['input', 'mapper', 'actors', 'volumes']);
```
jourdain: But if there is no mapper?

jourdain: Meaning `model.mapper` is `undefined` as it is not used or set?

agirault: Can a representation not have a mapper?? and if it is undefined won't that just return undefined?

Sources/Proxy/Core/ViewProxy/index.js
```diff
@@ -253,9 +253,9 @@ function vtkViewProxy(publicAPI, model) {
 
   publicAPI.setAnimation = (enable) => {
     if (enable) {
-      model.renderWindow.getInteractor().requestAnimation('proxy');
+      model.renderWindow.getInteractor().requestAnimation(publicAPI);
     } else {
-      model.renderWindow.getInteractor().cancelAnimation('proxy');
+      model.renderWindow.getInteractor().cancelAnimation(publicAPI);
```
agirault: @jourdain is that enough to make distinction between different viewproxys?

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -113,6 +113,30 @@ function vtkImageMapper(publicAPI, model) {
     return image.extentToBounds(ex);
   };
 
+  publicAPI.getBoundsForSlice = (slice, thickness = 0) => {
```
agirault: @martinken @jourdain I need this on the ImageSlice also to my understanding, to apply the necessary transform

Sources/Proxy/Core/AbstractRepresentationProxy/index.js
jourdain: yes a rep can be anything. But I guess that's ok to return undefined. 
Just wanted to mention that you should not expect to always have a mapper in a rep.

agirault: don't representations always have a mapper? And even if it didn't, wouldn't it just return `undefined` and then whoever called this need to handle it?

Sources/Proxy/Core/ViewProxy/index.js
jourdain: yes I guess.

Sources/Rendering/Core/ImageMapper/index.js
jourdain: ok

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -604,7 +632,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
   // do not want extra renders as the make the apparent interaction
   // rate slower.
   publicAPI.render = () => {
-    if (model.requestAnimationCount === 0) {
+    if (model.animationRequest === null) {
```
jourdain: are you sure you want to compare against null?

agirault: :man_shrugging:

agirault: Could also do `animationRequesters.size === 0`

jourdain: I don't know but just wanted to mention that an object can be `null`, `undefined` and `undefined === null => false`

agirault: I see. Then should be all good:
- init: https://github.com/Kitware/vtk-js/pull/575/files/3b786ece3a26aec42bba6ede6ee8768b6754a21f#diff-f698a8e6250bde65d1ae6422124e9a84R870
- deleted: https://github.com/Kitware/vtk-js/pull/575/files/3b786ece3a26aec42bba6ede6ee8768b6754a21f#diff-f698a8e6250bde65d1ae6422124e9a84R291

Sources/Rendering/Core/ImageMapper/index.js
agirault: I don't quite understand how that is working here... https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ImageSlice/index.js#L56-L102

agirault: done, ready

Sources/Proxy/Core/AbstractRepresentationProxy/index.js
agirault: your call, i don't need it anymore for my use case (and could access it by reference instead than get)

PR #577, opened 2018-02-12
perf(Math): Avoid Math.min, Math.max in vtkMath.arrayMin, vtkMath.arr…

thewtex:
…ayMax

Timings:

  Existing Math.min: ~210ms for arrayMin, ~2712ms for entire volume rendering initialization
  Comparison operator: ~35ms for arrayMin, ~2070ms for entire volume rendering initialization
  Ternary operator: ~37ms for arrayMin, ~2125ms for entire volume rendering initialization

### DIFF
--- a/Sources/Common/Core/Math/index.js
+++ b/Sources/Common/Core/Math/index.js
@@ -56,7 +56,7 @@ function arrayMin(arr) {
 function arrayMax(arr) {
   let maxValue = -Infinity;
   for (let i = 0, len = arr.length; i < len; ++i) {
-    if (arr[i] > maxValue) {
+    if (maxValue < arr[i]) {
       maxValue = arr[i];
     }
   }
--- a/Sources/Common/Core/Math/index.js
+++ b/Sources/Common/Core/Math/index.js
@@ -12,7 +12,6 @@ const { vtkErrorMacro, vtkWarningMacro } = macro;
 let randomSeedValue = 0;
 const VTK_MAX_ROTATIONS = 20;
 const VTK_SMALL_NUMBER = 1.0e-12;
-const MAX_FUNCTION_ARGUMENTS = 32768;
 
 function notImplemented(method) {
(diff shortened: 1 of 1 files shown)


martinken:
This is a great change. Best change ever. All my people like this change. +1

Sources/Common/Core/Math/index.js
```diff
@@ -45,25 +44,21 @@ const { round, floor, ceil, min, max } = Math;
...
 function arrayMax(arr) {
   let maxValue = -Infinity;
-  for (let i = 0, len = arr.length; i < len; i += MAX_FUNCTION_ARGUMENTS) {
-    const submax = Math.max.apply(
-      null,
-      arr.slice(i, Math.min(i + MAX_FUNCTION_ARGUMENTS, len))
-    );
-    maxValue = Math.max(submax, maxValue);
+  for (let i = 0, len = arr.length; i < len; ++i) {
+    if (arr[i] > maxValue) {
```
jourdain: you should use `<` as `>` is interpreted differently in JS.

martinken: I'll bite. How is > interpreted?

jourdain:
```
Note: The algorithm is only defined for a < b. So, a > b is handled as b < a.
```

https://github.com/getify/You-Dont-Know-JS/blob/master/types%20%26%20grammar/ch4.md

I guess it doesn't matter since the Browser will reverse it...
But in case the reverse action take time? ;-)

thewtex: Thanks for the tip @jourdain . JavaScript is sometimes `<` optimal, but we'll go with it :-P. Adding in a follow-up commit.

agirault:
Will the `perf` trigger a new release? Because I want a new release :heart:

jourdain:
No but since I merged it with other fixes, you will get you new release...

jourdain:
Actually looking at the ChangeLog, perf() does create a new release... ;-)

PR #578, opened 2018-02-12
fix(ITKHelper): Add ITK to VTK js image converter

floryst:
@thewtex @agirault

### DIFF
--- a/Sources/Common/DataModel/ITKHelper/index.js
+++ b/Sources/Common/DataModel/ITKHelper/index.js
@@ -1,0 +1,30 @@
+import vtkImageData from 'vtk.js/Sources/Common/DataModel/ImageData';
+import vtkDataArray from 'vtk.js/Sources/Common/Core/DataArray';
+
+/**
+ * Converts an itk.js image to a vtk.js image.
+ *
+ * Requires an itk.js image as input.
+ */
+function convertItkToVtkImage(itkImage) {
+  // create VTK image data
+  const imageData = vtkImageData.newInstance({
+    origin: itkImage.origin.slice(),
+    spacing: itkImage.spacing.slice(),
+  });
+  const scalars = vtkDataArray.newInstance({
+    name: 'Scalars',
+    values: itkImage.data,
+    numberOfComponents: itkImage.imageType.components,
+  });
+
+  imageData.setDirection(itkImage.direction.data);
+  imageData.setDimensions(...itkImage.size);
+  imageData.getPointData().setScalars(scalars);
+
+  return imageData;
+}
+
+export default {
+  convertItkToVtkImage,
+};
--- a/Sources/Common/DataModel/ITKHelper/index.js
+++ b/Sources/Common/DataModel/ITKHelper/index.js
@@ -6,14 +6,14 @@ import vtkDataArray from 'vtk.js/Sources/Common/Core/DataArray';
  *
  * Requires an itk.js image as input.
  */
-function convertItkToVtkImage(itkImage) {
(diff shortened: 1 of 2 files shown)


Sources/Common/DataModel/ITKHelper/index.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+function convertItkToVtkImage(itkImage) {
+  // create VTK image data
+  const imageData = vtkImageData.newInstance({
+    origin: itkImage.origin.slice(),
+    spacing: itkImage.spacing.slice(),
+  });
+  const scalars = vtkDataArray.newInstance({
+    values: itkImage.data,
+    numberOfComponents: itkImage.imageType.components,
```
jourdain: Do you have a name for that array?

floryst: No....."Scalars"?

jourdain: Then add `name: 'Scalars',1

jourdain: Maybe ask for it as 2nd arg with a default to 'Scalars'

floryst: Seems reasonable. Would it be preferable to accept that through options object as the 2nd arg in the event we ever decide to add more configurability to the conversion function?

jourdain:
LGTM

agirault:
Awesome! Should we handle the flipping here? So we don't have a legacy state in vtk.js with wrong orientation?
@thewtex

Sources/Common/DataModel/ITKHelper/index.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+function convertItkToVtkImage(itkImage, options) {
+  // create VTK image data
+  const imageData = vtkImageData.newInstance({
+    origin: itkImage.origin.slice(),
+    spacing: itkImage.spacing.slice(),
+  });
+  const scalars = vtkDataArray.newInstance({
+    name: options.scalarArrayName || 'Scalars',
+    values: itkImage.data,
```
agirault: no need to `slice()` data?

floryst: I'd rather not make another copy of the image data, and ideally nobody else tries to modify it anyways.

thewtex:
Thanks @floryst !

@agirault We could handle the data layout change later as a `fix` in my humble opinion ;-).

Sources/Common/DataModel/ITKHelper/index.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Requires an itk.js image as input.
+ */
+function convertItkToVtkImage(itkImage, options) {
+  // create VTK image data
+  const imageData = vtkImageData.newInstance({
+    origin: itkImage.origin.slice(),
+    spacing: itkImage.spacing.slice(),
+  });
+  const scalars = vtkDataArray.newInstance({
```
thewtex: It would be good to add a `switch` on `itkImage.imageType.pixelType` per

  https://github.com/InsightSoftwareConsortium/itk-js/blob/master/src/PixelTypes.js

where `Scalar`, `RGB`, and `RGBA` we do this, and `default` to throwing an error for not implemented yet. Per:

  https://www.vtk.org/doc/nightly/html/classvtkDataSetAttributes.html#a156a8842bffd9139fa81193cca0d5bcb

floryst: Moved to #582

PR #579, opened 2018-02-12
fix(RenderWindowInteractor): do not share animation requesters

agirault:
The `Set` was shared between every instance of RenderWindowInteractor.

Co-Authored-by: @floryst


agirault:
FYI @thewtex

PR #585, opened 2018-02-14
fix(InteractorStyleManipulator): properly register animation requesters

agirault:
Do not use same key for animations due to mouse drag and due to mouse wheel
or the second event will not be registered as an animation requester, and
the animation will stop when the first event terminates.

@jourdain no rush to integrate it on my side, can be merged with anything else that comes up later.


agirault:
@jourdain no rush for that on my side, can be merged with anything else that comes up later.

PR #587, opened 2018-02-14
Improve linked properties

agirault:
1. Remove hardcoded property links to rely on dynamic property links instead: https://github.com/kitware/vtk-js/commit/28f40b2
2. Add some properties on geometry representation


jourdain:
LGTM, should I merge?

PR #591, opened 2018-02-15
Fix orientations + XYZ slicing

agirault:
- [x] Use proper orientation matrix in ITKHelper
  - Needed to transpose it from ITK to VTK.
  - Tested with numerous datasets with orientation matrices different than identity.
- [x] Add X,Y,Z slicing (and rename previous IJK slicing for data space slicing)
  - Only supports it if X, Y, Z are orthogonal to I, J, K axis for now (aka flips and permutations). Will still do it if one of those component is > 0.99 (so with a small tilt). All world axis that do not perfectly align one of IJK axis should require a reformat to compute the slice, but this will be added in future work (@martinken promised :clinking_glasses: )
  - Still use indexes for X, Y, Z for now (instead of world coordinates) since they currently still map to IJK. In the future, X, Y, Z or any world axis slicing (ex: normal to the camera or to a plane widget...) should be in world coordinates (based on imagedata spacing/bounds).

### DIFF
--- a/Sources/Proxy/Representations/SliceRepresentationProxy/index.js
+++ b/Sources/Proxy/Representations/SliceRepresentationProxy/index.js
@@ -89,9 +89,20 @@ function vtkSliceRepresentationProxy(publicAPI, model) {
 
     // Init slice location
     const extent = inputDataset.getExtent();
-    publicAPI.setXSlice(Math.floor(mean(extent[0], extent[1])));
-    publicAPI.setYSlice(Math.floor(mean(extent[2], extent[3])));
-    publicAPI.setZSlice(Math.floor(mean(extent[4], extent[5])));
+    const { ijkMode } = model.mapper.getClosestIJKSlice();
+    switch (ijkMode) {
+      case vtkImageMapper.SlicingMode.I:
+        publicAPI.setSlice(Math.floor(mean(extent[0], extent[1])));
+        break;
+      case vtkImageMapper.SlicingMode.J:
+        publicAPI.setSlice(Math.floor(mean(extent[2], extent[3])));
+        break;
+      case vtkImageMapper.SlicingMode.K:
+        publicAPI.setSlice(Math.floor(mean(extent[4], extent[5])));
+        break;
+      default:
+        break;
+    }
   }
 
   // Keep things updated
@@ -99,12 +110,6 @@ function vtkSliceRepresentationProxy(publicAPI, model) {
   model.sourceDependencies.push({ setInputData });
 
   // API ----------------------------------------------------------------------
-
-  publicAPI.setSliceIndex = (index) =>
-    model.mapper[`set${model.slicingMode}Slice`](index);
-
-  publicAPI.getSliceIndex = () =>
-    model.slicingMode ? model.mapper[`get${model.slicingMode}Slice`]() : 0;
 
   publicAPI.setSlicingMode = (mode) => {
     if (!mode) {
@@ -172,9 +177,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     visibility: { modelKey: 'actor', property: 'visibility' },
     colorWindow: { modelKey: 'property', property: 'colorWindow' },
     colorLevel: { modelKey: 'property', property: 'colorLevel' },
-    xSlice: { modelKey: 'mapper', property: 'xSlice' },
-    ySlice: { modelKey: 'mapper', property: 'ySlice' },
-    zSlice: { modelKey: 'mapper', property: 'zSlice' },
+    sliceIndex: { modelKey: 'mapper', property: 'slice' },
   });
 }
 
--- a/Sources/Proxy/Representations/SliceRepresentationProxy/index.js
+++ b/Sources/Proxy/Representations/SliceRepresentationProxy/index.js
@@ -122,7 +122,7 @@ function vtkSliceRepresentationProxy(publicAPI, model) {
     // Init slice location
     const bounds = inputDataset.getBounds();
     const extent = inputDataset.getExtent();
-    switch (model.mapper.getCurrentSlicingMode()) {
+    switch (model.mapper.getSlicingMode()) {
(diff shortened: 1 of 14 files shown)


floryst:
LGTM

agirault:
I'll be working on updating the image mapper to properly do X,Y,Z slicing. Could merge those together or separately.

thewtex:
All wee needed was a


You get a 10.0, Alexis.

agirault:
Almost ready!
- @jourdain: there are some API changes here (removing xSlice, ySlice, zSlice from mapper and slicerep). Enough to be considered breaking changes? Knowing that, when we add reformatting, we'll need to update the API so that X, Y, Z is not index based but world based, so that will change again.
- @martinken: currently failing tests with textures:
  - [ ]  vtkOpenGLImageMapper testImage not ok 413
  - [ ]  Image match resolution not ok 415 
  - [ ] vtkOpenGLImageMapper testImage not ok 417
  - [ ] mage match resolution not ok 419

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -570,7 +575,7 @@ function vtkOpenGLImageMapper(publicAPI, model) {
         ptsArray[9] = ext[1];
         ptsArray[10] = nSlice;
         ptsArray[11] = ext[5];
-      } else {
+      } else if (ijkMode === SlicingMode.K) {
```
agirault: @martinken this seems to be the root of the tests failing: should this code run if we are not doing slicing? (`currentSlicingMode = SlicingMode.None`), and not just if we are doing K slicing? I thought this was only for what we used to call Z slicing.

agirault:
Tests fixed, I was thinking of switching to using world coordinates and not indexed anymore for X, Y, Z.

agirault:
tests and example pass, with X, Y, Z slicing using world coordinates and not indexed coordinates!
Ready for review @jourdain @martinken @floryst. Thank you :pray:

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
```diff
@@ -20,14 +20,14 @@ function mean(...array) {
...
   const axisIndex = 'XYZ'.indexOf(slicingMode);
 
   const propToUpdate = {
     sliceIndex: {
       domain: {
-        min: extent[axisIndex * 2],
-        max: extent[axisIndex * 2 + 1],
+        min: bounds[axisIndex * 2],
+        max: bounds[axisIndex * 2 + 1],
         step: 1,
```
jourdain: Should this be updated as well to match the correct axis spacing?

agirault: done

jourdain:
Looks good enough for me... But what about you @martinken ?

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -16,80 +17,191 @@ function vtkImageMapper(publicAPI, model) {
...
 
-    if (changeDetected && mtime === publicAPI.getMTime()) {
-      publicAPI.modified();
+    // Project vec3 onto direction cosines
+    const out = [0, 0, 0];
+    const a = publicAPI.getInputData().getDirection();
+    const mat3 = [[a[0], a[1], a[2]], [a[3], a[4], a[5]], [a[6], a[7], a[8]]];
+    const mat3I = [[0, 0, 0], [0, 0, 0], [0, 0, 0]];
+    vtkMath.invertMatrix(mat3, mat3I, 3);
+    vtkMath.multiply3x3_vect3(mat3I, inVec3, out);
```
martinken: Maybe just reorder indicies as rotation inverse is just a transpose I think

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -101,38 +213,80 @@ function vtkImageMapper(publicAPI, model) {
...
+      case SlicingMode.Y:
+        nSlice = publicAPI.getSliceAtPosition(model.slice);
+        extent[2] = nSlice - thickness;
+        extent[3] = nSlice + thickness;
+        break;
+      case SlicingMode.Z:
+        nSlice = publicAPI.getSliceAtPosition(model.slice);
+        extent[4] = nSlice - thickness;
+        extent[5] = nSlice + thickness;
+        break;
```
martinken: Do not know that X==i here etc

Sources/Rendering/Core/ImageMapper/index.js
agirault: https://en.wikipedia.org/wiki/Rotation_matrix#Multiplication :raised_hands:

agirault: Done: https://github.com/Kitware/vtk-js/pull/591/commits/cbdae0dd024f42e5bd75a4b9316d092a8bd38cf0#diff-9762803aa1d1e7c8613c71d155ed4e0dR89

Sources/Rendering/Core/ImageMapper/index.js
agirault: Done:
- https://github.com/Kitware/vtk-js/pull/591/commits/8b80c6ff33c104f6be3c450f7cad186a26afc26b#diff-9762803aa1d1e7c8613c71d155ed4e0dR215
- https://github.com/Kitware/vtk-js/pull/591/commits/8b80c6ff33c104f6be3c450f7cad186a26afc26b#diff-9762803aa1d1e7c8613c71d155ed4e0dR247

agirault:
Last changes discussed with @martinken made. Also renamed `currentSlicingMode` to `slicingMode` and updated `croppingregionswidget` example to use IJK. I believe it's ready @jourdain !

agirault:
@floryst could you give a look at the last commit? it updates the imageCroppingRegionsWidget example to use IJK.

floryst:
LGTM

PR #592, opened 2018-02-16
fix(ProxyManager): Optionally set proxy to active, if possible

floryst:
setActive is default true, so code must explicitly disable this.

### DIFF
--- a/Sources/Proxy/Core/ProxyManager/core.js
+++ b/Sources/Proxy/Core/ProxyManager/core.js
@@ -1,6 +1,6 @@
 import macro from 'vtk.js/Sources/macro';
 
-const { capitalize, vtkErrorMacro } = macro;
+const { vtkErrorMacro } = macro;
 
 // ----------------------------------------------------------------------------
 // Proxy Registration Handling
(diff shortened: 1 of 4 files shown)


floryst:
Do we want `setActive` default to true or false?

jourdain:
I was thinking it could be nice to have a `activate()` method on any proxy.
Then in the config we could choose the behavior (auto activate or not). And by default they become active. 
That way the method call won't have the extra arg.
Basically only the source need to be 'patch' to become valid and maybe Representation but Views should be fine.

What do you think? Maybe the default should not be active...

floryst:
I like the idea of having an `activate()` method. It also reduces code duplication with `setActive*` methods in ProxyManager/core.js. Would it make sense to have `activate` on all proxies, or just ones that make sense (e.g. Views and Sources, but not Representations)?

I think the default should be not active, so applications should clearly specify they are activating a proxy.

jourdain:
I think all proxy should have a `activate()` (inside macro.proxy).

Then the default should be not to activate a new proxy.

jourdain:
But it could be nice to also specify when a new should automatically activate the created proxy in the config. With the default being to not do it.

floryst:
Ok, I'll add an activate. And agreed, having an `activateOnCreate` config option for a proxy is useful.

floryst:
I've pushed a working implementation.

jourdain:
LGTM

PR #593, opened 2018-02-16
Update cubesource

jourdain:


Sources/Filters/Sources/CubeSource/index.js
```diff
@@ -172,6 +173,21 @@ function vtkCubeSource(publicAPI, model) {
       n[2] += 2.0;
     }
 
+    // Apply rotation to the points coordinates and normals
+    vtkMatrixBuilder
+      .buildFromRadian()
```
jourdain: Don't you want degree? It is easier for a user to input.

Sources/Filters/Sources/CubeSource/api.md
```diff
@@ -23,5 +23,9 @@ Floating value representing cube length along z axis.
 
 Float array of size 3 representing the center of the cube.
 
+### rotations (set/get)
+
+Float array of size 3 representing the angles of rotation for the cube.
```
jourdain: specify units. Degree would be better than radian.

jourdain:
Example change should not trigger a new release as the published/shared code does not change.

Also since those changes are small, maybe we want to stick with fix() instead of feat()?

PR #594, opened 2018-02-18
perf(Texture): Only scale cube texture to power of two when not WebGL2

thewtex:

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -724,7 +724,11 @@ function vtkOpenGLTexture(publicAPI, model) {
     publicAPI.bind();
 
     const pixData = updateArrayDataType(dataType, data);
-    const scaledData = scaleTextureToHighestPowerOfTwo(pixData);
+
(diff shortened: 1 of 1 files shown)


thewtex:
This saves about 0.14 sec on load.

thewtex:

martinken:
+1 Thanks!

thewtex:
@jourdain can this please have some merge love? :heart_eyes_cat:

jourdain:
Sure, can I bundle them with others?

thewtex:
Yes

agirault:
@martinken @thewtex could this be called here also?
https://github.com/Kitware/vtk-js/pull/594/files#diff-80916a397e2ba7d72b3769b22d96fcecL678

Or instead - if we should always skip for gl2 - could we return the input right away if that's the case to avoid forgetting that check in the future?

Seems like I'm getting a lot of slowliness because of this.

martinken:
Yes,  we should skip for WebGL2 there as well.

agirault:
thanks, i'll make a patch.

PR #595, opened 2018-02-19
Webworker

thewtex:
This improves volume rendering initialization from 5.5 seconds to 2.1 seconds! :zap: 


martinken:
Can you change the test such that the new valid image is not needed? I would rather not add that valid image as it really isn't valid. It matches what I would expect the image to look like if the lighting/GO failed.

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -904,7 +913,20 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
+      } else {
+        // We wanted to render, but the gradients have not finished computing.
+        // So, re-render later.
+        if (computedGradientsRenderTimeout !== null) {
+          clearTimeout(computedGradientsRenderTimeout);
+        }
+        computedGradientsRenderTimeout = setTimeout(
+          model.openGLRenderWindow.modified,
+          20
+        );
```
martinken: I'm not sure I get what this is doing in the timeout. Marking the renderWindow modified shouldn't change anything last I knew.

package.json
```diff
@@ -1,83 +1,85 @@
...
+    "base64-js": "1.2.1",
+    "blueimp-md5": "2.10.0",
+    "commander": "2.11.0",
+    "gl-matrix": "2.3.1",
+    "jszip": "3.1.4",
+    "pako": "1.0.6",
+    "seedrandom": "2.4.3",
+    "shelljs": "0.7.8",
+    "webworker-promise": "^0.4.1",
+    "worker-loader": "^1.1.0"
```
jourdain: This is already provided inside kw-web-suite.

Sources/Interaction/Widgets/PiecewiseGaussianWidget/ComputeHistogram.worker.js
```diff
@@ -0,0 +1,17 @@
+import registerWebworker from 'webworker-promise/lib/register';
+
+registerWebworker(async ({ array, min, max, numberOfBins }, emit) => {
+  const delta = max - min;
+  const histogram = new Float32Array(numberOfBins);
```
jourdain: Any reason for float32 instead of uint32?

Sources/Rendering/OpenGL/VolumeMapper/index.js
thewtex: I should make a comment here, but this triggers a render after the gradients are ready.

package.json
thewtex: Good to know -- I will see if we can build without it.

Sources/Interaction/Widgets/PiecewiseGaussianWidget/ComputeHistogram.worker.js
thewtex: Later, we normalize.

thewtex:
> Can you change the test such that the new valid image is not needed? I would rather not add that valid image as it really isn't valid. It matches what I would expect the image to look like if the lighting/GO failed.

I will see if capturing the rendering snapshot after a timeout delay is possible and does the trick.

thewtex:
> I will see if capturing the rendering snapshot after a timeout delay is possible and does the trick.

This does not work given the way that JavaScript executes code. The timeout is never reached.

@martinken Now, the volume mapper asynchronously re-renders with lighting effects enabled after the gradients have finished computing. If we really want to test this, I could try adding a `.onLightingActivate` or similarly named function to `vtkVolumeMapper`. This would be passed a callback that gets called when lighting is activated in `vtkOpenGLVolumeMapper`. In addition to re-rendering, this could grab a screen shot and perform the baseline comparison in the test. However, I am not sure how this should be exposed / propagated in `vtkVolumeMapper` and `vtkOpenGLVolumeMapper`. Thoughts?

martinken:
Thinking about this more generally. We need to know when "stuff" is done. Imagine if we did image flips in a worker. For FDA approval we would really need to be able to tell the doctor, yeh the image flips have all been done and what you are looking at is correct. Sort of a pending-work-is-all-done check. Not sure how JS handles this normally.

One thought, maybe when situations like this arise we instead request animation from the interactor and when done we cancel animation.  

For tests we setup a callback on render that checks some state and if ready captures the test image and returns. @jourdain  thoughts?

jourdain:
We need to have some kind of progress handler which does not necessarily tell you how much progress has been made but if there is some work going on and when a given work is completed.

Those async handling are usually handle via promises and in lot of my web apps I have a wrapper around those promise to capture/monitor the busy state. We just need to add similar infrastructure for vtk.js.

thewtex:
@martinken @jourdain Yes, that sounds great. I am not sure how to implement it :confused: .

In the meantime, I looked into the callback approach, which could serve as a substitute until a promise-based render initialization infrastructure is in place. And, maybe this is used to inform the promise in the implementation? The test takes a bit longer to run, but it passes, and actually checks that lighting occurred :-). I will push the commit to the tip of this branch -- it could be omitted if desired.

Sources/Rendering/Core/VolumeMapper/index.js
```diff
@@ -53,6 +54,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'imageSampleDistance',
     'maximumSamplesPerRay',
     'autoAdjustSampleDistances',
+    'onLightingActivated',
```
jourdain: Use `macro.event(publicAPI, model, 'lightingActivated')` instead.

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -916,16 +916,19 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
       // Only activate once volumeInfo has been populated.
       if (model.lightingTexture.getComputedGradients()) {
         model.lightingTexture.activate();
+        const onLightingActivated = model.renderable.getOnLightingActivated();
+        if (typeof onLightingActivated === 'function') {
+          onLightingActivated();
+        }
```
jourdain: With the event stuff, you will be able to call `model.renderable.invokeLightingActivated()`.

Sources/Rendering/OpenGL/VolumeMapper/test/testLighting.js
```diff
@@ -91,16 +90,18 @@ test.onlyIfWebGL('Test Lighted Volume Rendering', (t) => {
       renderer.getActiveCamera().zoom(1.4);
       renderer.resetCameraClippingRange();
       renderer.updateLightsGeometryToFollowCamera();
+      mapper.setOnLightingActivated(() => {
```
jourdain: this will become `mapper.onLightingActivated(() => {`

Sources/Rendering/Core/VolumeMapper/index.js
thewtex: Nice. :candy: Done.

Sources/Rendering/OpenGL/VolumeMapper/index.js
thewtex: Done.

Sources/Rendering/OpenGL/VolumeMapper/test/testLighting.js
thewtex: Done.

jourdain:
I'm guessing you validate the code with the event stuff?

jourdain:
I'm also guessing that you haven't pushed yet...

thewtex:
Not yet, now I am debugging a timeout...

thewtex:
Ran into some interesting bugs, including a bug in babel:

  https://github.com/babel/babel/issues/5085

<a href="https://imgflip.com/i/24zhax"></a>

All good now :cake:

thewtex:
Merci, @jourdain ! :man_dancing:

PR #596, opened 2018-02-19
test(CubeSource): Add an additional baseline for the CubeSource test

thewtex:
Obtained on Linux/Chrome. The cubes are not zoomed in as far.


thewtex:
@jourdain From testing on

> Version 64.0.3282.119 (Developer Build) built on Debian 9.3, running on Debian 9.3 (64-bit)

PR #603, opened 2018-02-21
fix(CubeSource): Do not translate twice

thewtex:
This is follow-up to Kitware/vtk-js@12681ad21d4893d454eff87165aeb1d421120008. A translation was
added there even though the center translation already occurs in the preceeding code.

CC: @jourdain @scottwittenburg


jourdain:
We may actually apply the rotation before the translate. Which require the use of 0 instead of model.center[...] when computing the original points coordinates.

What do you think @thewtex ?

FYI @patrickoleary

thewtex:
@jourdain Yes, good point. I'll try that...

thewtex:
@jourdain done.

PR #606, opened 2018-02-22
Animation + Domain + texture perfs + STL orientation + ImageMapper slicing

agirault:


Sources/macro.js
```diff
@@ -1150,6 +1152,10 @@ export function proxy(publicAPI, model) {
   publicAPI.getPropertyByName = (name) =>
     getProperties().find((p) => p.name === name);
 
+  publicAPI.getPropertyDomainByName = (name) => {
+    return propertyMap[name].domain;
+  };
+
```
jourdain: need to be reformated

`publicAPI.getPropertyDomainByName = (name) => propertyMap[name].domain;`

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -77,6 +93,15 @@ function vtkSTLReader(publicAPI, model) {
     model.parseData = content;
 
     // Binary parsing
+    // Header
+    const headerData = new Uint8Array(content, 0, 80);
+    let headerStr = '';
+    for (let i = 0; i < 80; ++i) {
+      headerStr += String.fromCharCode(headerData[i]);
+    }
+    const header = parseHeader(headerStr);
```
jourdain: There is a better way to do it. See the XML readers.

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -110,6 +135,49 @@ function vtkSTLReader(publicAPI, model) {
       cellDataValues[faceIdx] = dataView.getUint16(offset + 48, true);
     }
 
+    // Rotate points
+    const orientationField = 'SPACE';
+    if (orientationField in header && header[orientationField] !== 'LPS') {
+      const XYZ = header[orientationField];
+      const mat4 = [...new Array(16)].map((x) => 0);
```
jourdain: new Float32Array(16); will already be filled with 0.

Sources/IO/Geometry/STLReader/index.js
agirault: this?
https://github.com/Kitware/vtk-js/blob/master/Sources/IO/XML/XMLReader/index.js#L68-L95

jourdain: https://github.com/Kitware/vtk-js/blob/master/Sources/IO/Core/BinaryHelper/index.js#L7-L19

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -77,6 +96,12 @@ function vtkSTLReader(publicAPI, model) {
     model.parseData = content;
 
     // Binary parsing
+    // Header
+    const headerData = new Uint8Array(content, 0, 80);
+    const headerStr = BinaryHelper.arrayBufferToString(headerData);
```
agirault: @jourdain this works, but should I initialize `headerData` as something else than a `Uint8Array`?

jourdain: Check with @floryst

jourdain:
You have linter/prettier issue

Sources/IO/Geometry/STLReader/index.js
floryst: `arrayBufferToString` should take in an ArrayBuffer, but it'll work with UInt8Arrays.

Note the fallback code in `arrayBufferToString` will make a copy of your input if it's UInt8Array.

agirault: used `slice()` to provide ArrayBuffer of proper offset and length.

Sources/IO/Geometry/STLReader/index.js
agirault: done

agirault:
@jourdain fixed.

@martinken the patch to skip scaling textures (ef2cb1f) makes the ImageMapper tests fail. Is that expected? See below:
screen shot 2018-02-22 at 3 19 37 pm

martinken:
Pending the other fix (webgl error) we talked about this morning those images look fine to me. But I suspect we may need to have two sets of images or just a larger threshold. To me they look the same but I get that they are different because on one we are interpolating to a power of two while on the other opengl is using the npot image

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -538,12 +538,13 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
       if (ijkMode === SlicingMode.I) {
-        scalars = [];
+        scalars = new basicScalars.constructor(dims[2] * dims[1]);
+        let id = 0;
         for (let k = 0; k < dims[2]; k++) {
           for (let j = 0; j < dims[1]; j++) {
-            scalars.push(
-              basicScalars[sliceOffset + j * dims[0] + k * dims[0] * dims[1]]
-            );
+            id = j * dims[2] + k;
```
floryst: Should this be `id = k * dims[1] + j`?

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -561,12 +562,13 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
       } else if (ijkMode === SlicingMode.J) {
-        scalars = [];
+        scalars = new basicScalars.constructor(dims[2] * dims[0]);
+        let id = 0;
         for (let k = 0; k < dims[2]; k++) {
           for (let i = 0; i < dims[0]; i++) {
-            scalars.push(
-              basicScalars[i + sliceOffset * dims[0] + k * dims[0] * dims[1]]
-            );
+            id = i * dims[2] + k;
```
floryst: Same as https://github.com/Kitware/vtk-js/pull/606/commits/aa1d5dd138ca04b51892c42241a5ecf436678fad#r170292540

agirault:
Ok everything is ready. Who dares review is welcome to.

@martinken the deltas are 5.76% and 5% for those two failing tests. What would you recommend I do?

jourdain:
Did you fix the test (baseline/threshold/?) ?

Should we merge?

agirault:
@jourdain now yes :)

martinken:
+1

jourdain:
Once green, I'll merge

agirault:
Rebased on https://github.com/Kitware/vtk-js/pull/607 from @thewtex to ensure release build passes.

Sources/IO/Geometry/STLReader/index.js
agirault: @jourdain when mat4 is a `Float32Array`, `Array.isArray(mat4x4)` returns `false` in MatrixBuilder (but no warning or error message)
https://github.com/Kitware/vtk-js/commit/d1fb6679af8edfeffd4644c46e3246e09ee2abe8#diff-915eb56ed796b4179f838589e323b110R111

Which part should I modify? `MatrixBuilder.setMatrix` or this?

PR #611, opened 2018-02-26
Various fix

jourdain:


Sources/Proxy/Representations/SliceRepresentationProxy/index.js
```diff
@@ -201,7 +247,11 @@ function vtkSliceRepresentationProxy(publicAPI, model) {
 // Object factory
 // ----------------------------------------------------------------------------
 
-const DEFAULT_VALUES = {};
+const DEFAULT_VALUES = {
+  xSlice: 0,
+  ySlice: 0,
+  zSlice: 0,
```
agirault: Why? A slice representation should have only one slice value.

Sources/Common/Core/MatrixBuilder/index.js
```diff
@@ -128,6 +130,11 @@ class Transform {
     }
     return this;
   }
+
+  identity() {
+    mat4.identity(this.matrix);
```
agirault: Is this an opengl matrix (column-major) ?
We discussed with @martinken and @thewtex, and we'll probably want to switch matrix library to consolidate our math implementation in vtk-js (itk, vtk, itk.js, vtk.js Math are row-major, but opengl is column-major, which lead to a lot of confusions).

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
jourdain: I kind of agree but needed a backward compatible implementation so latest VTK could be used in Glance. The correct fix for Glance is to do something similar to what you've done with the various configuration but I could not do it with the time I could spend on it yesterday.

So I brought back the previous API at the proxy level so we can upgrade vtk.js for glance for backward compatibility. Once we cleanup Glance with proper configuration management, we could remove those add-on on that slice representation.

agirault: I managed to do it in my projects by creating multiple name for the distinct representations:
```js
      SliceX: {
        class: vtkSliceRepresentationProxy,
        options: {
          links: [
            { link: 'ColorWindow', property: 'colorWindow' },
            { link: 'ColorLevel', property: 'colorLevel' },
            { link: 'SliceX', property: 'slice' },
          ],
        },
        props: {
          slicingMode: 'X',
        },
      },
      SliceY: {
        class: vtkSliceRepresentationProxy,
        options: {
          links: [
            { link: 'ColorWindow', property: 'colorWindow' },
            { link: 'ColorLevel', property: 'colorLevel' },
            { link: 'SliceY', property: 'slice' },
          ],
        },
        props: {
          slicingMode: 'Y',
        },
      },
      SliceZ: {
        class: vtkSliceRepresentationProxy,
        options: {
          links: [
            { link: 'ColorWindow', property: 'colorWindow' },
            { link: 'ColorLevel', property: 'colorLevel' },
            { link: 'SliceZ', property: 'slice' },
          ],
        },
        props: {
          slicingMode: 'Z',
        },
      },
      Volume: {
        class: vtkVolumeRepresentationProxy,
        options: {
          links: [
            { link: 'ColorWindow', property: 'colorWindow' },
            { link: 'ColorLevel', property: 'colorLevel' },
            { link: 'SliceX', property: 'xSlice' },
            { link: 'SliceY', property: 'ySlice' },
            { link: 'SliceZ', property: 'zSlice' },
          ],
        },
      },
```

Sources/Common/Core/MatrixBuilder/index.js
agirault: Correct, that was more a global comment regarding the matrix in vtkMatrixBuilder. How should I fill it up in setMatrix?

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
jourdain: I know, I told you to do so, but the Glance support several layouts and usage which makes things slightly more complex. Specially with views that can change orientation dynamically.
As I said, not difficult or impossible, but that would require more time (~2 hours).

Sources/Common/Core/MatrixBuilder/index.js
jourdain: Based on your comment, I would have 2 methods which encode the matrix format in their name or via an argument flag.

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
agirault: ok

PR #613, opened 2018-02-26
fix(MatrixBuilder): allow using typed array in setMatrix

agirault:


Sources/Common/Core/MatrixBuilder/index.js
```diff
@@ -110,7 +110,13 @@ class Transform {
   }
 
   setMatrix(mat4x4) {
-    if (Array.isArray(mat4x4) && mat4x4.length === 16) {
+    if (!mat4x4) {
+      return this;
+    }
+    if (
+      Object.prototype.hasOwnProperty.call(mat4x4, 'length') &&
```
jourdain: Do you really need that line? (117)

agirault: @jourdain still returning false :thinking:

jourdain: Hence my comment..

jourdain:
LGTM

jourdain:
Do you have other fix for today? Or should I merge it ?

agirault:
no other fix so far but it's not urgent either.

jourdain:
ok, will wait till the end of the day in case we want to tackle other things too...

PR #616, opened 2018-02-27
Misc fixes

floryst:
`getNumberOf*` methods on polydata were incorrectly capitalized as `getNumberOfverts`, when it should be `getNumberOfVerts`, and so forth.

Other minor changes and spelling changes.


jourdain:
Thanks!

PR #622, opened 2018-03-02
fix(PiecewiseGaussianWidget): Enable API adjustment of Gaussians

thewtex:
…Gaussian

CC: @agirault @jourdain

### DIFF
--- a/Sources/Interaction/Widgets/PiecewiseGaussianWidget/index.js
+++ b/Sources/Interaction/Widgets/PiecewiseGaussianWidget/index.js
@@ -1050,7 +1050,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'colorTransferFunction',
     'backgroundImage',
   ]);
-  macro.get(publicAPI, model, ['size', 'canvas']);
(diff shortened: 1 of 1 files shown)


Sources/Interaction/Widgets/PiecewiseGaussianWidget/index.js
```diff
@@ -1050,7 +1057,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'colorTransferFunction',
     'backgroundImage',
   ]);
-  macro.get(publicAPI, model, ['size', 'canvas']);
+  macro.get(publicAPI, model, ['size', 'canvas', 'selectedGaussian']);
```
jourdain: You will override your `getSelectedGaussian` method if you do that so the resulting `getSelectedGaussian()` will return the index.

thewtex: Good call. Removing...

thewtex:
Thanks, @jourdain !

PR #623, opened 2018-03-05
fix(VolumeMapperExample): Update rendering after lighting activated

thewtex:
Otherwise, a re-render does not occur until the user clicks on the
scene.

### DIFF
--- a/Sources/Rendering/Core/VolumeMapper/example/index.js
+++ b/Sources/Rendering/Core/VolumeMapper/example/index.js
@@ -60,11 +60,16 @@ mapper.setInputConnection(reader.getOutputPort());
 reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
   reader.loadData().then(() => {
     renderer.addVolume(actor);
-    renderWindow.getInteractor().setDesiredUpdateRate(15.0);
+    const interactor = renderWindow.getInteractor();
+    interactor.setDesiredUpdateRate(15.0);
     renderer.resetCamera();
     renderer.getActiveCamera().zoom(1.5);
     renderer.getActiveCamera().elevation(70);
     renderWindow.render();
+    interactor.requestAnimation('lighting');
+    mapper.onLightingActivated(() => {
+      interactor.cancelAnimation('lighting');
+    });
   });
 });
 


Sources/Rendering/Core/VolumeMapper/example/index.js
```diff
@@ -60,11 +60,16 @@ mapper.setInputConnection(reader.getOutputPort());
...
+    const interactor = renderWindow.getInteractor();
+    interactor.setDesiredUpdateRate(15.0);
     renderer.resetCamera();
     renderer.getActiveCamera().zoom(1.5);
     renderer.getActiveCamera().elevation(70);
     renderWindow.render();
+    interactor.requestAnimation('lighting');
+    mapper.onLightingActivated(() => {
+      interactor.cancelAnimation('lighting');
+    });
```
agirault: Instead of animating, could you just call `render()` in the callback? 
` mapper.onLightingActivated(renderWindow.render);`

martinken: No No No, no mappers calling render IMO

martinken: Well I guess in this case the mapper is not calling render, it is the callback so that would be fine in this one case

thewtex: `render()` in the callback did not do the trick in this case.

agirault: That makes sense: I did not understand what your code was doing at first. What is happening is that you constantly render (using animations), until the lighting is activated, then you can cancel the animation. `lightingActivated` is only invoked during a rendering pass, so you need to trigger a render yourself (here done by the animation).
I think it's a little intricate: anyone who has a volume mapper and wants to see a correct volume rendering will have to do something like this, which IMO is not super intuitive. I don't have another solution, other than invoking an event as soon as the gradients have been set up, 
and call render in its callback, but @martinken you said that could trigger too many renders?

martinken: Yeh, if I have 10,000 mappers all doing some async load I do not want 10,000 forced renders that would take 10,000/60 seconds. Using the built in animation support makes more sense.

PR #628, opened 2018-03-06
fix(CubeMapReader): Add reader to create cube textures

jourdain:


jourdain:
@martinken I can wait for your changes to properly use the skybox in the example/app.

PR #630, opened 2018-03-07
add vtkSkybox class and fix depth comparison issue

martinken:
also add ability to render a texture from another
node without having it be in the scene graph.


jourdain:
It seems we are missing the skybox files...

martinken:
working as intended ;-)

martinken:
Do: merge

jourdain:
Sorry... The robot took some time to decode your `Do: merge` and click on the green button. ;-)

martinken:
Fricking lazy robots ;-)

PR #638, opened 2018-03-09
Performance improvements (& windows fixes)

agirault:
- [x] Reduce CPU memory footprint for gradients computation (~50%)
- [x] Avoid crash in webgl1 by limiting textures size
- [x] Transfer magnitudes sub array from worker to main thread (instead of copy)
- [x] Terminate web workers adequately

cc: @martinken, @thewtex

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/ComputeGradients.worker.js
+++ b/Sources/Rendering/OpenGL/Texture/ComputeGradients.worker.js
@@ -10,6 +10,8 @@ registerWebworker(
     message,
     emit
   ) {
+    // have to compute the gradient to get the normal
+    // and magnitude
     var width = message.width;
     var height = message.height;
     var depth = message.depth;
@@ -19,10 +21,13 @@ registerWebworker(
     var depthStart = message.depthStart;
     var depthEnd = message.depthEnd;
 
-    // have to compute the gradient to get the normal
-    // and magnitude
+    // only compute gradient for last slice of the web worker
+    // if it is the last slice of the volume data
+    if (depthEnd !== depth - 1) {
+      depthEnd--;
+    }
     var depthLength = depthEnd - depthStart + 1;
-    var gradients = new Float32Array(width * height * depthLength * 4);
+    var gradients = new Uint8Array(width * height * depthLength * 3);
     var gradientMagnitudes = new Float32Array(width * height * depthLength);
 
     var sliceSize = width * height;
@@ -37,7 +42,7 @@ registerWebworker(
     );
     var minMag = vec3.length(grad);
     var maxMag = -1.0;
-    for (var z = depthStart; z < depthEnd + 1; ++z) {
+    for (var z = depthStart; z <= depthEnd; ++z) {
       var zedge = 0;
       if (z === depth - 1) {
         zedge = -sliceSize;
@@ -61,12 +66,10 @@ registerWebworker(
 
           var mag = vec3.length(grad);
           vec3.normalize(grad, grad);
-          gradients[outPtr++] = grad[0];
-          gradients[outPtr++] = grad[1];
-          gradients[outPtr++] = grad[2];
-          gradients[outPtr++] = mag;
-          gradientMagnitudes[inPtr] = mag;
-          inPtr++;
+          gradients[outPtr++] = 127.5 + 127.5 * grad[0];
+          gradients[outPtr++] = 127.5 + 127.5 * grad[1];
+          gradients[outPtr++] = 127.5 + 127.5 * grad[2];
+          gradientMagnitudes[inPtr++] = mag;
         }
       }
     }
@@ -77,9 +80,11 @@ registerWebworker(
 
     var result = {
       subGradients: gradients,
+      subMagnitudes: gradientMagnitudes,
       subMinMag: minMag,
       subMaxMag: maxMag,
       subDepthStart: depthStart,
+      subDepthEnd: depthEnd,
     };
     return Promise.resolve(
       new registerWebworker.TransferableResponse(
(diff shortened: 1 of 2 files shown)


Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1103,6 +1118,12 @@ function vtkOpenGLTexture(publicAPI, model) {
       newArray
     );
 
+    model.context.finish();
+    const err = model.context.getError();
+    if (err !== model.context.NO_ERROR) {
+      vtkErrorMacro('bad stuff');
```
zachmullen: Is this debugging code or is this meant to be here?

agirault: not meant to be there

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1120,17 +1135,149 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+          4,
+          VtkDataTypes.UNSIGNED_CHAR,
+          reformattedGradients
+        );
+      };
+    } else {
+      // Now determine the texture parameters using the arguments.
+      publicAPI.getOpenGLDataType(VtkDataTypes.UNSIGNED_CHAR);
+      publicAPI.getInternalFormat(VtkDataTypes.UNSIGNED_CHAR, 4);
+      publicAPI.getFormat(VtkDataTypes.UNSIGNED_CHAR, 4);
```
zachmullen: These functions all are called `get...` but they mutate internal state, which is a bit confusing. Might consider renaming them in the next major version.

Sources/Rendering/OpenGL/Texture/ComputeGradients.worker.js
```diff
@@ -61,12 +66,10 @@ registerWebworker(
 
           var mag = vec3.length(grad);
           vec3.normalize(grad, grad);
-          gradients[outPtr++] = grad[0];
-          gradients[outPtr++] = grad[1];
-          gradients[outPtr++] = grad[2];
-          gradients[outPtr++] = mag;
-          gradientMagnitudes[inPtr] = mag;
-          inPtr++;
+          gradients[outPtr++] = 127.5 + 127.5 * grad[0];
```
zachmullen: A good rule of thumb for me is that when strange numbers like this appear in code, it's worthy of an inline comment explaining them.

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1011,15 +1013,28 @@ function vtkOpenGLTexture(publicAPI, model) {
...
     model.numberOfDimensions = 2;
 
-    // have to pack this 3D texture into pot 2D texture
-    const maxTexDim = model.context.getParameter(
-      model.context.MAX_TEXTURE_SIZE
-    );
+    // MAX_TEXTURE_SIZE gives the max dimensions that can be supported by the GPU,
+    // but it doesn't mean it will fit in memory. If we have to use a float data type
+    // or 4 components, there are good chances that the texture size will blow up
+    // and coud not fit in the GPU memory. Use a smaller texture size in that case,
```
zachmullen: could*

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1120,17 +1135,149 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+      };
+    } else {
+      // Now determine the texture parameters using the arguments.
+      publicAPI.getOpenGLDataType(VtkDataTypes.UNSIGNED_CHAR);
+      publicAPI.getInternalFormat(VtkDataTypes.UNSIGNED_CHAR, 4);
+      publicAPI.getFormat(VtkDataTypes.UNSIGNED_CHAR, 4);
+
+      if (!model.internalFormat || !model.format || !model.openGLDataType) {
+        vtkErrorMacro('Failed to determine texture parameters.');
+        return;
```
zachmullen: What are the general design principles about exception handling in vtk.js? Would it be better to throw an exception instead of just logging and returning?

Sources/Rendering/OpenGL/Texture/ComputeGradients.worker.js
agirault: yep, I can add one

Sources/Rendering/OpenGL/Texture/index.js
agirault: I don't really know, this sure looks very C++y. @jourdain @martinken ?

Sources/Rendering/OpenGL/Texture/ComputeGradients.worker.js
martinken: Just say normal encoding, it is pretty standard in shaders

Sources/Rendering/OpenGL/Texture/index.js
jourdain: Use the macro for now. We can make those macro throw an exception if we want to go that path.

Sources/Rendering/OpenGL/Texture/index.js
agirault: maybe that was a convention imported from c++? 
cc: @martinken @jourdain

jourdain: I'm not sure what those methods do, but I agree `get` should not mutate the state...

agirault: Referenced here: https://github.com/Kitware/vtk-js/issues/639

Sources/Rendering/OpenGL/Texture/index.js
agirault: Referenced in this issue: https://github.com/Kitware/vtk-js/issues/640

Sources/Rendering/OpenGL/Texture/index.js
martinken: It is exactly because someone wrote them that way in C++. I am not a fan of how it is done either.

PR #641, opened 2018-03-12
fix(VolumeRepresentationProxy): do not override initial properties

agirault:
SampleDistance and EdgeGradient could be defined in a config file,
but would then be overridden when `setInputData`.

### DIFF
--- a/Sources/Proxy/Representations/VolumeRepresentationProxy/index.js
+++ b/Sources/Proxy/Representations/VolumeRepresentationProxy/index.js
@@ -204,8 +204,12 @@ function vtkVolumeRepresentationProxy(publicAPI, model) {
     model.property.setScalarOpacity(0, pwfProxy.getPiecewiseFunction());
 
     updateConfiguration(inputDataset, publicAPI.getDataArray(), model);
-    publicAPI.setSampleDistance();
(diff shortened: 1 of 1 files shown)


Sources/Proxy/Representations/VolumeRepresentationProxy/index.js
```diff
@@ -236,7 +240,7 @@ function vtkVolumeRepresentationProxy(publicAPI, model) {
   );
   publicAPI.getSliceVisibility = model.actorX.getVisibility;
 
-  publicAPI.setSampleDistance = (distance = 0.4) => {
```
jourdain: So you are ok that `model.sampleDistance = undefined`?

(I'm not...)

Sources/Proxy/Representations/VolumeRepresentationProxy/index.js
```diff
@@ -256,7 +260,7 @@ function vtkVolumeRepresentationProxy(publicAPI, model) {
     }
   };
 
-  publicAPI.setEdgeGradient = (edgeGradient = 0.2) => {
```
jourdain: So you are ok that `model.edgeGradient = undefined`?

PR #642, opened 2018-03-13
try changing how the vol ren shader works

martinken:
Try using a fixed for loop instead of float
variable  loop.


martinken:
@agirault  Give this topic a try and see if it works better. I am seeing significant speed improvement. It is a WIP but only a minor change or two are required for it to be ready to go.

agirault:
Rebased on master here: https://github.com/agirault/vtk-js/commit/9dde7be722408252f34003cda348d165be953612

agirault:
Rebased on master here: https://github.com/Kitware/vtk-js/pull/643

PR #643, opened 2018-03-13
webgl2 change

agirault:
cc: @martinken 
- Use a fixed for loop instead of float variable loop.

### DIFF
--- a/Sources/Rendering/OpenGL/RenderWindow/index.js
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.js
@@ -221,6 +221,7 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
       }
     }
     if (!result) {
+      vtkDebugMacro('using webgl1');
       result =
         model.canvas.getContext('webgl', options) ||
         model.canvas.getContext('experimental-webgl', options);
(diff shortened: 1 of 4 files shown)


Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -543,6 +542,10 @@ export function extend(publicAPI, model, initialValues = {}) {
   if (typeof InstallTrigger !== 'undefined') {
     model.defaultToWebgl2 = true;
   }
+  model.defaultToWebgl2 = true;
```
agirault: @martinken should I remove `defaultToWebgl2` altogether? So we always use webgl2 if available?

Sources/Rendering/OpenGL/glsl/vtkVolumeFS2.glsl
```diff
@@ -165,17 +165,18 @@ void main()
...
     vpos = vpos + vdelta*(0.01 + 0.98*jitter);
     vec4 color = vec4(0.0, 0.0, 0.0, 0.0);
 
+    int count = int(numSteps - 0.2); // end slightly inside
+
     vec3 ijk = vpos * vVCToIJK;
     vdelta = vdelta * vVCToIJK;
-    float i = 0.0;
-    while (i < numSteps - 0.2)
+    for (int i = 0; i < //VTK::MaximumSamplesValue ; ++i)
```
agirault: @martinken how is `i` incrementing with that comment?

Sources/Rendering/OpenGL/RenderWindow/index.js
martinken: I would leave it for now and see how it goes. After a while we can yank it if we find no cases where we need to fall back to webgl 1 even when webgl2 is available.

agirault:
All tests passing

PR #647, opened 2018-03-14
fix(proxy): allow links to be on the application level

agirault:
use `type` to define where the link lives: `source` (default) or `application`.

### DIFF
--- a/Sources/macro.js
+++ b/Sources/macro.js
@@ -1055,14 +1055,18 @@ export function proxy(publicAPI, model) {
   };
 
   // property link
-  model.propertyLinkSubscribers = [];
-  publicAPI.registerPropertyLinkForGC = (otherLink) => {
-    model.propertyLinkSubscribers.push(otherLink);
+  model.propertyLinkSubscribers = {};
+  publicAPI.registerPropertyLinkForGC = (otherLink, type) => {
+    if (!(type in model.propertyLinkSubscribers)) {
+      model.propertyLinkSubscribers[type] = [];
+    }
+    model.propertyLinkSubscribers[type].push(otherLink);
   };
 
-  publicAPI.gcPropertyLinks = () => {
-    while (model.propertyLinkSubscribers.length) {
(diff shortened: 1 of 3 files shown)


Sources/macro.js
```diff
@@ -1219,12 +1223,23 @@ export function proxy(publicAPI, model) {
...
       model.propertyLinkMap[list[count]].unsubscribe();
     }
-    count = model.propertyLinkSubscribers.length;
-    while (count--) {
-      model.propertyLinkSubscribers[count].unbind(publicAPI);
-    }
+    model.propertyLinkSubscribers.keys.forEach((key, index) => {
+      publicAPI.gcPropertyLinks(key);
+    });
     parentDelete();
```
jourdain:
```
Object.keys(model.propertyLinkSubscribers).forEach(publicAPI.gcPropertyLinks);
```

Sources/macro.js
```diff
@@ -1219,12 +1223,23 @@ export function proxy(publicAPI, model) {
...
+  if (model.links) {
+    for (let i = 0; i < model.links.length; i++) {
+      const { link, property, persistent, updateOnBind, type } = model.links[i];
+      if (type === 'application') {
+        const sLink = model.proxyManager.getPropertyLink(link, persistent);
+        publicAPI.registerPropertyLinkForGC(sLink, 'application');
+        sLink.bind(publicAPI, property, updateOnBind);
+      }
+    }
+  }
```
jourdain:
```
function registerLinks() {
  // Allow dynamic registration of links at the application level
  if (model.links) {
    for (let i = 0; i < model.links.length; i++) {
      const { link, property, persistent, updateOnBind, type } = model.links[i];
      if (type === 'application') {
        const sLink = model.proxyManager.getPropertyLink(link, persistent);
        publicAPI.registerPropertyLinkForGC(sLink, 'application');
        sLink.bind(publicAPI, property, updateOnBind);
      }
    }
  }
}
setImmediate(registerLinks);
```

PR #648, opened 2018-03-14
Speed improvements

agirault:
speed up `dataArray.getTuple` and `Plane/Box.evaluateFunction`.

### DIFF
--- a/Sources/Common/DataModel/Box/index.js
+++ b/Sources/Common/DataModel/Box/index.js
@@ -118,15 +118,8 @@ function vtkBox(publicAPI, model) {
 
   publicAPI.getBounds = () => model.bbox.getBounds();
 
-  publicAPI.evaluateFunction = (...x) => {
-    let point = [].concat(x);
-    if (Array.isArray(x[0])) {
-      point = x[0];
-    }
-
-    if (point.length !== 3) {
-      return Number.MAX_VALUE;
-    }
+  publicAPI.evaluateFunction = (x, y, z) => {
+    const point = Array.isArray(x) ? x : [x, y, z];
 
     let diff;
     let dist;
--- a/Sources/Common/DataModel/Plane/test/testPlane.js
+++ b/Sources/Common/DataModel/Plane/test/testPlane.js
@@ -141,14 +141,15 @@ test('Test vtkPlane evaluateFunction', (t) => {
   plane.setOrigin(0.0, 0.0, 0.0);
   plane.setNormal(1.0, 1.0, 1.0);
 
-  let res = plane.evaluateFunction([1.0, 1.0, 1.0]);
+  const point = [1.0, 1.0, 1.0];
+  let res = plane.evaluateFunction(point);
   t.equal(res, 3);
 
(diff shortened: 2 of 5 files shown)


Sources/Common/DataModel/Box/index.js
```diff
@@ -118,13 +120,11 @@ function vtkBox(publicAPI, model) {
...
   publicAPI.getBounds = () => model.bbox.getBounds();
 
-  publicAPI.evaluateFunction = (...x) => {
-    let point = [].concat(x);
-    if (Array.isArray(x[0])) {
-      point = x[0];
-    }
-
-    if (point.length !== 3) {
+  publicAPI.evaluateFunction = (xyz) => {
```
jourdain: so now we can't use that function like `evaluateFunction (1,2,3)` and `evaluateFunction ([1,2,3])`.

agirault: - `evaluateFunction ([1,2,3])`: yes, that's the only thing it can, just like for Cone, Cylinder, Sphere, ImplicitBoolean..
- `evaluateFunction (1,2,3)`: no, that's the point: doing the following is super slow:
```js
publicAPI.evaluateFunction = (...x) => {
  let point = [].concat(x);
  if (Array.isArray(x[0])) {
    point = x[0];
  }
```

Sources/Common/DataModel/Plane/test/testPlane.js
```diff
@@ -144,8 +144,8 @@ test('Test vtkPlane evaluateFunction', (t) => {
   let res = plane.evaluateFunction([1.0, 1.0, 1.0]);
   t.equal(res, 3);
 
-  res = plane.evaluateFunction(1.0, 1.0, 1.0);
-  t.equal(res, 3);
+  res = plane.evaluateFunction(...[1.0, 1.0, 1.0]);
```
jourdain: this is wrong...

Sources/Common/DataModel/Box/index.js
jourdain: The way it was handle was slow. But that does not mean it has to be slow for handling both use case.

jourdain: Just FYI, any setArray macro will support both method profile and having a limitation on that will be bad as you will never know which one is valid.

jourdain: try that and let me know if that's slow

```
publicAPI.evaluateFunction = (x, y, z) => {
  const point = y === undefined ? x : [x,y,z];

Comment (agirault) [reply Sources/Common/DataModel/Box/index.js]: if you have a better suggestion please let me know. It's better to support one fast and warn that the second is not supported (so users just update their code), than supporting both and having one silently slow.

Comment (jourdain) [reply Sources/Common/DataModel/Box/index.js]: You can add warning for user that it pick the slow path via an extra

```
if (point !== x) {
   vtkDebug('slow pass');
}

jourdain:
LGTM

agirault:
I'll have a PR up soon to fix the cutter and i think it would be good to go.

PR #654, opened 2018-03-16
fix(RenderWindowInteractor): Disable mouse/touch event listener

laurennlam:
Fix the problem in mobile device when user touches renderWindow, release touches and then use GUI component. The mouse/touch event weren't removed.


jourdain:
I'll merge it along with @agirault fix for the cutter.

PR #655, opened 2018-03-16
Adding vtk cutter

agirault:
cc: @laurennlam 
- [x] add vtkcutter
- [x] add test
- [x] add example
- [x] add sliced geometry representation proxy

### DIFF
--- a/Sources/Filters/Core/Cutter/index.js
+++ b/Sources/Filters/Core/Cutter/index.js
@@ -1,0 +1,272 @@
+import macro from 'vtk.js/Sources/macro';
+import vtkPolyData from 'vtk.js/Sources/Common/DataModel/PolyData';
+
+const { vtkErrorMacro } = macro;
+
+// ----------------------------------------------------------------------------
+// vtkCutter methods
+// ----------------------------------------------------------------------------
+
+function vtkCutter(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkCutter');
+
+  function dataSetCutter(input, output) {
+    const numCells = input.getNumberOfCells();
+    const points = input.getPoints();
+    const pointsData = points.getData();
+    const numPts = points.getNumberOfPoints();
+    const newPointsData = [];
+    const newLinesData = [];
+    const newPolysData = [];
+
+    if (!model.cutScalars || model.cutScalars.length < numPts) {
+      model.cutScalars = new Float32Array(numPts);
+    }
+
+    // Loop over all points evaluating scalar function at each point
+    let inOffset = 0;
+    let outOffset = 0;
+    while (inOffset < pointsData.length) {
+      model.cutScalars[outOffset++] = model.cutFunction.evaluateFunction(
+        pointsData[inOffset++],
+        pointsData[inOffset++],
+        pointsData[inOffset++]
+      );
+    }
+
+    const dataCell = input.getPolys().getData();
+    const crossedEdges = [];
+    const x1 = new Array(3);
+    const x2 = new Array(3);
+    // Loop over all cells; get scalar values for all cell points
+    // and process each cell.
+    /* eslint-disable no-continue */
+    for (let cellId = 0; cellId < numCells; cellId++) {
+      const nbPointsInCell = dataCell[0];
+      // Check that cells have at least 3 points
+      if (nbPointsInCell <= 2) {
+        continue;
+      }
+
+      // Get associated scalar of points that constitute the current cell
+      const cellPointsScalars = [];
+      const valuesInCell = nbPointsInCell + 1; // first value is size
+      let pointIndex;
+      for (
+        let i = valuesInCell * cellId + 1;
+        i < valuesInCell * (cellId + 1);
+        i++
+      ) {
+        pointIndex = dataCell[i];
+        cellPointsScalars.push(model.cutScalars[pointIndex]);
+      }
+
+      // Check if all cell points are on same side (same side == cell not crossed by cut function)
+      // TODO: won't work if one point scalar is = 0 ?
+      const sideFirstPoint = cellPointsScalars[0] > 0;
+      let allPointsSameSide = true;
+      for (let i = 1; i < cellPointsScalars.length; i++) {
+        const sideCurrentPoint = cellPointsScalars[i] > 0;
+        if (sideCurrentPoint !== sideFirstPoint) {
+          allPointsSameSide = false;
+          break;
+        }
+      }
+
+      // Go to next cell if cell is not crossed by cut function
+      if (allPointsSameSide) {
+        continue;
+      }
+
+      // Get id of points that constitute the current cell
+      const cellPointsID = [];
+      for (
+        let i = valuesInCell * cellId + 1;
+        i < valuesInCell * (cellId + 1);
+        i++
+      ) {
+        cellPointsID.push(dataCell[i]);
+      }
+
+      // Find and compute edges which intersect cells
+      const intersectedEdgesList = [];
+      for (let i = 0; i < cellPointsID.length; i++) {
+        const idNext = i + 1 === cellPointsID.length ? 0 : i + 1;
+
+        // Go to next edge if edge is not crossed
+        // TODO: in most come cases, (numberOfPointsInCell - 1) or 0 edges of the cell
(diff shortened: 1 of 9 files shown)


agirault:
@jourdain I just have one thing to check with the sliced geometry and it should be good to go (proxy doesn't initialize properly in 2d views)

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
```diff
@@ -0,0 +1,142 @@
...
+
+  publicAPI.setOffset = (offset) => {
+    if (offset === model.offset || offset === undefined) {
+      return;
+    }
+    updateOffset(offset);
+    publicAPI.modified();
+  };
+
+  publicAPI.setSlicingMode = (mode) => {
```
jourdain: Where is the get?

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
```diff
@@ -0,0 +1,142 @@
...
+
+  // Internals
+  model.plane = vtkPlane.newInstance();
+  model.cutter = vtkCutter.newInstance();
+  model.cutter.setCutFunction(model.plane);
+  model.mapper = vtkMapper.newInstance({
+    interpolateScalarsBeforeMapping: false,
+    useLookupTableScalarRange: false,
+    scalarVisibility: false,
+  });
```
jourdain: Are those good defaults for everyone?

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
```diff
@@ -0,0 +1,142 @@
...
+  model.actor = vtkActor.newInstance();
+  model.property = model.actor.getProperty();
+  model.property.setLighting(false);
```
jourdain: Good default? Why?

agirault: would have been needed if I was using tubes (to get the thickness but not the depth). When only using lines, I'm not sure that matters, I'll test without it. I think it can be a good default anyway: when you look at a sliced geometry representation, it's usually just a contour, and you do not really care for lighting.

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
agirault: Great question, I do not know what those really are for. I'll take any suggestions on that.

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
agirault: is it ok if I store it in model.slicingMode and just use the get macro?

jourdain: yes definitely that would be the way I would have done it.

jourdain: Also I would guard that method if the sliceMode remain the same. To prevent unnecessary modified/computation.
I know it seems useless, but VTK rely a lot on modified time and lot of code assume it will not changed if we keep setting the same value again and again. So to not get bitten by it later, you should ALWAYS do it.

jourdain:
Looking at the code, you don't have a very good strategy to handle modified(). Do you know that calling `setXXX()` will return true if something has changed and false otherwise. That way you could easily chain or use those to actually call modified() on yourself when it matters. That is given when using the macro. But when you write your own `setXXX` method, you should do the same so that pattern could still exist through out the code...

agirault:
> Do you know that calling setXXX() will return true if something has changed and false otherwise.

I did not. Do you mean setXXX() from the macro? Or the ones I should be writing?

jourdain:
The method `setXXX` from the macro would do it. Therefore, it would be nice if your `setXXX` would do the same as well as using this behavior to detect when calling `modified()` is needed.

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
agirault: no it makes sense. I did not do it because i did not store model.slicingMode. Now that I do, I will.

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
```diff
@@ -0,0 +1,142 @@
...
+    publicAPI.modified();
+  };
+
+  publicAPI.setOffset = (offset) => {
+    if (offset === model.offset || offset === undefined) {
+      return;
+    }
+    updateOffset(offset);
+    publicAPI.modified();
+  };
```
jourdain: I would write that method that way

```
  function updateOffset(offset) {
    let changeDetected =  model.offset !== offset;
    model.offset = offset;
    const normal = model.plane.getNormal();
    return model.actor.setPosition(
      offset * normal[0],
      offset * normal[1],
      offset * normal[2]
    ) || changeDetected;
  }
```

```
publicAPI.setOffset = (offset = 0) => {
    if (updateOffset(offset)) {
       publicAPI.modified();
       return true;
    }
    return false;
  };
```

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
```diff
@@ -36,39 +36,42 @@ function vtkSlicedGeometryRepresentationProxy(publicAPI, model) {
...
-    model.plane.setOrigin(n[0] * slice, n[1] * slice, n[2] * slice);
-    model.cutter.modified();
-  }
+    const planeModified = model.plane.setOrigin(
+      n[0] * slice,
+      n[1] * slice,
+      n[2] * slice
+    );
+    if (planeModified) {
+      model.cutter.modified();
```
jourdain: Should the cutter rely on the mtime of the function instead?

Sources/Filters/Core/Cutter/index.js
```diff
@@ -0,0 +1,272 @@
...
+      return;
+    }
+
+    const output = vtkPolyData.newInstance();
+
+    dataSetCutter(input, output);
+
+    outData[0] = output;
+  };
+}
```
jourdain: I think you are missing the getMTime() that's why you call `modified()` on it when changing the implicit function.

jourdain: see https://github.com/Kitware/vtk-js/blob/master/Sources/Common/DataModel/ImplicitBoolean/index.js#L25-L35

Sources/macro.js
```diff
@@ -185,7 +187,7 @@ export function obj(publicAPI = {}, model = {}) {
       return;
     }
 
-    if (otherMTime && otherMTime < model.mtime) {
+    if (otherMTime && otherMTime < publicAPI.getMTime()) {
```
jourdain: not good keep the old one

jourdain: Maybe not... It is fine.

agirault: we should always read mtime using getMTime, shouldn't we?

jourdain: yes

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
jourdain: I would not set them if they are not chosen on purpose.

Sources/Filters/Core/Cutter/index.js
agirault: fixed

Sources/Proxy/Representations/SlicedGeometryRepresentationProxy/index.js
```diff
@@ -0,0 +1,140 @@
...
+      offset * normal[2]
+    );
+    if (actorModified || stateModified) {
+      publicAPI.modified();
+      return true;
+    }
+    return false;
+  };
+
+  publicAPI.setSlicingMode = (mode) => {
```
jourdain: missing return true/false

agirault: 🙆‍♂️

PR #658, opened 2018-03-20
Fix scroll values from mouse wheel event

agirault:
References related to mouse wheel delta problem:
- info (wheelDelta): https://developer.mozilla.org/en-US/docs/Web/Events/mousewheel
- issue: https://stackoverflow.com/questions/5527601/normalizing-mousewheel-speed-across-browsers
- solution: https://github.com/facebookarchive/fixed-data-table/blob/master/src/vendor_upstream/dom/normalizeWheel.js
- package: https://www.npmjs.com/package/normalize-wheel

### DIFF
--- a/Sources/Interaction/Manipulators/MouseCameraTrackballZoomManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseCameraTrackballZoomManipulator/index.js
@@ -74,7 +74,7 @@ function vtkMouseCameraTrackballZoomManipulator(publicAPI, model) {
 
     const camera = renderer.getActiveCamera();
 
-    const dyf = delta;
+    const dyf = 1 - delta / 10;
 
     if (camera.getParallelProjection()) {
       camera.setParallelScale(camera.getParallelScale() / dyf);
--- a/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
@@ -9,11 +9,22 @@ function vtkMouseRangeManipulator(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkMouseRangeManipulator');
 
+  // Keep track of delta that is below the value
+  // of one step to progressively increment it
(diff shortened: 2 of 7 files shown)


Sources/Interaction/Manipulators/MouseCameraTrackballZoomManipulator/index.js
```diff
@@ -74,7 +74,7 @@ function vtkMouseCameraTrackballZoomManipulator(publicAPI, model) {
 
     const camera = renderer.getActiveCamera();
 
-    const dyf = delta;
+    const dyf = 1 - delta / 10;
```
thewtex: can we use a more descriptive name for `dyf` here? And add a comment that explains why `10`?

agirault: So I'm trying to guess what `dyf` is... maybe `deltaYFactor`? :thinking: 

`delta` used to be the scroll value given from the mouse wheel event, and it was pretty arbitrary: https://github.com/Kitware/vtk-js/blob/051f85534bb6053f17f8fffb5f7f6c97e121eb60/Sources/Rendering/Core/RenderWindowInteractor/index.js#L410-L419

`1 - delta / 10` is just as arbitrary, but it is done so that the "speed" of zooming during the scroll stays consistent now that we track the scroll value differently (see below):
https://github.com/Kitware/vtk-js/blob/2b68fe93b400afd6737f985a1879c564040fc59d/Sources/Rendering/Core/RenderWindowInteractor/index.js#L425

agirault:
@jourdain I added the normalization code to macro instead of using the package.

agirault:
Thanks for the reviews everyone!

Sources/Interaction/Manipulators/MouseCameraTrackballZoomManipulator/index.js
timthirion: There's a comment in there somewhere. ;) You can say `const scale = 10; // to keep delta Y around 1`.

PR #659, opened 2018-03-20
fix(Rendering): skybox and cube map was borked

martinken:
A number of fixes to cube maps and skyboxes. They are
slightly different in that one looks at a cube from outside
while the other looks at a cube from inside. This changes
how the texture coordinates are setup. Add to that the fact
that opengl uses an off negative Y image format for cubemaps
left over from Renderman standards.


Sources/Rendering/OpenGL/Texture/test/testCreateCubeFromRawTexture.js
```diff
@@ -121,12 +103,12 @@ test.onlyIfWebGL('Test vtkOpenGLTexture Rendering', (t) => {
...
-  texturePathList.push(`${path}py.vti`);
-  texturePathList.push(`${path}ny.vti`);
-  texturePathList.push(`${path}pz.vti`);
-  texturePathList.push(`${path}nz.vti`);
+  texturePathList.push(`${path}right.jpg`);
+  texturePathList.push(`${path}left.jpg`);
+  texturePathList.push(`${path}top.jpg`);
+  texturePathList.push(`${path}bottom.jpg`);
+  texturePathList.push(`${path}front.jpg`); // front is +z on a cube
+  texturePathList.push(`${path}back.jpg`);
```
agirault: @martinken need to remove the vti's and add the jpg's from the source?

agirault: also, no biggie, but typo in commit title

martinken: Yup

PR #661, opened 2018-03-21
fix(Rendering): fix skybox with shared actors

martinken:
The texture could get flipped in memory twice


martinken:
@jourdain needed for the VR view as the textures are shared

PR #662, opened 2018-03-21
fix(DeviceOrientationToCamera): Prevent laptop to set invalid orientation

jourdain:


jourdain:
FYI @martinken

martinken:
Thanks, seems to work on desktop now.

PR #664, opened 2018-03-22
fix(Picker): Deep copy matrix to not modify original matrix

laurennlam:
getMatrix() function returns a reference to the matrix. As we revert the matrix, it modifies the original which is not requested.


laurennlam:
I think it's ready to merge

PR #674, opened 2018-03-29
fix(RenderWindowInteractor): handle keydown event

luciemac:
When an arrow key is pressed 'keyPressEvent' is not emited. 
Listening 'keyDownEvent' fix this.

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -33,6 +33,7 @@ const handledEvents = [
   'RightButtonPress',
   'RightButtonRelease',
   'KeyPress',
+  'KeyDown',
   'KeyUp',
   'StartMouseWheel',
   'MouseWheel',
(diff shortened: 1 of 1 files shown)


agirault:
Looking at the documentation:
* https://developer.mozilla.org/en-US/docs/Web/Events/keydown
* https://developer.mozilla.org/en-US/docs/Web/Events/keypress

I remember removing keydown because it was almost equivalent to keypress. It seems that keydown can handle any keys, while keypress only handles characters. Should we therefore get rid of keypress to avoid almost-duplicate APIs ?

Also @luciemac, I see you use `getKeysFor`. Is `getModifierKeysFor` valid for the keydown event? What if you press on CTRL, ALT or SHIFT?

Same question for `charCode`: is that key valid for a keydown event that is not a character?

agirault:
Looking at the documentation:
* https://developer.mozilla.org/en-US/docs/Web/Events/keydown
* https://developer.mozilla.org/en-US/docs/Web/Events/keypress

I remember removing keydown because it was almost equivalent to keypress after discussion with Ken. It seems that keydown can handle any keys, while keypress only handles characters. Should we therefore get rid of keypress to avoid almost-duplicate APIs ?

Also @luciemac, I see you use `getKeysFor`. Is `getModifierKeysFor` valid for the keydown event? What if you press on CTRL, ALT or SHIFT?

Same question for `charCode`: is that key valid for a keydown event that is not a character?

luciemac:
@agirault `getModifierKeysFor  ` seems to work fine. 

If you press CTRL for instance, it returns:
```
{
    altKey:false
    controlKey:true
    shiftKey:false
}
```

Issue #545, opened 2018-01-30
Multiple Picker issues


agirault:
- [x] (1) In `CellPicker`, we should retrieve `pickData.point` and not `pickData.x`: https://github.com/Kitware/vtk-js/commit/5ea58fc6989502442f3de48f9b468a676d557a7a#diff-6290e5c29feaaaa4a4ebc10bc1942fb5R176
- [ ] (2) When picking an ImageMapper with a CellPicker, the picked position is currently stored in three different places:
  - In [AbstractPicker.pickPosition](https://github.com/Kitware/vtk-js/blob/v5.15.10/Sources/Rendering/Core/AbstractPicker/index.js#L60): always [0,0,0] for imageMapper
  - In [Picker.pickedPositions](https://github.com/Kitware/vtk-js/blob/v5.15.10/Sources/Rendering/Core/Picker/index.js#L360): populated [here](https://github.com/Kitware/vtk-js/blob/v5.15.10/Sources/Rendering/Core/Picker/index.js#L292-L314) but has a wrong depth when picking imageMapper
  - In [CellPicker.pCoords](https://github.com/Kitware/vtk-js/blob/v5.15.10/Sources/Rendering/Core/CellPicker/index.js#L415) which is populated [here](https://github.com/Kitware/vtk-js/blob/v5.15.10/Sources/Rendering/Core/CellPicker/index.js#L187) and is correct since (1) was fixed.
- [x] (3) Picking with an ImageMapper does not work when a render window's origin is not at the top left corner of the canvas  ([this check fails](https://github.com/Kitware/vtk-js/blob/v5.15.10/Sources/Rendering/Core/Picker/index.js#L291))

agirault:
FYI @laurennlam @jourdain @floryst

floryst:
Also, CellPicker.getCelIJK seems to work, but is in IJK space.

agirault:
(3) will be fixed when the commit https://github.com/Kitware/vtk-js/pull/556/commits/b651049b3745520605234bdf4c2e693f1d023517 as part of https://github.com/Kitware/vtk-js/pull/556 and was due to picking in render windows that were not at the top left corner of the canvas

agirault:
(1) fixed in https://github.com/Kitware/vtk-js/pull/566

agirault:
(2) fixes in #676:
- AbstractPicker.pickPosition is the screen position where we pick (input)
-  Picker.pickedPositions is now properly populated with the picked position
[...]

PR #676, opened 2018-03-29
Some fixes (ImageMapper picking and mouse manipulator)

agirault:


Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -37,8 +37,15 @@ function vtkMouseRangeManipulator(publicAPI, model) {
...
       listener.setValue(value);
       model.incrementalDelta[listener] = 0;
-    } else if (value !== listener.min && value !== listener.max) {
+    } else {
       model.incrementalDelta[listener] += delta;
+      // Do not allow incremental delta to go past range
+      if (
+        (value === listener.min && model.incrementalDelta[listener] < 0) ||
+        (value === listener.max && model.incrementalDelta[listener] > 0)
+      ) {
```
jourdain: Does it need to be that strict ? `===` vs `<` ?

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -339,85 +343,23 @@ function vtkImageMapper(publicAPI, model) {
       const absoluteIJK = [0, 0, 0];
       imageData.worldToIndex(point, absoluteIJK);
 
-      // Are we outside our actual extent/bounds
```
jourdain: I don't know about that. @floryst should have an application on which it can test it. I defer the review of that part to him.

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
agirault: could be `<=` to be safe, but not `<` (which should never happen since value is clamped between min and max): this is to check if we are AT (===) an extremity (min or max), if we handle incrementing the delta or not:
* if at min and decreasing delta, ignore delta.
* if at min and increasing delta, increment delta.
* if at max and increasing delta, ignore delta.
* if at max and decreasing delta, increment delta.

Sources/Rendering/Core/ImageMapper/index.js
agirault: Sure. Explanations in the commit message of https://github.com/Kitware/vtk-js/pull/676/commits/43cb0be9e30e7e53ebf5de2aefd26e2bf2ffd968.

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
jourdain: Sure I just wanted to make sure that equality has a chance to happen.

Sources/Rendering/Core/ImageMapper/index.js
floryst: With regards to (2), `intersectWithLineForPointPicking` duplicated the code from `intersectWithLineForCellPicking`, so this change effectively is a refactor with no functionality changes. The only difference between the two was `...PointPicking` used Math.round, whereas `...CellPicking` used Math.floor when going from absoluteIJK to ijk.

Also, I couldn't find a single usage of `intersectWithLineForPointPicking`, so AFAIK this doesn't break anything at least.

jourdain: It is true that to pick the right cell in regards to ijk, it needs to be `Math.floor` while points needs `Math.round`.
Looking at just the diff, I'm not sure how things are behaving now.

floryst: I'm testing it now

agirault: > It is true that to pick the right cell in regards to ijk, it needs to be Math.floor while points needs Math.round when using the point picking information of the intersection location.

No. Why would it need `floor`? worldToIndex matrix and the worldPosition input precision limitations can result in non-integer resulting absoluteIndex output, which makes perfect sense. That means you could have an absoluteIndex output value of 4.99999 for example, and you would clearly want the resulting ijk value (which is supposed to be the closest cell at the picked position) to be 5, not 4.

Right now, with `floor`, I have multiple use cases that show the picking failing due to this.

jourdain: So a point ijk of (0.999, 0.999, 0.999) correspond to a cell ijk of (0,0,0)

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -339,85 +343,23 @@ function vtkImageMapper(publicAPI, model) {
...
-      imageData.worldToIndex(point, absoluteIJK);
-
       // Are we outside our actual extent/bounds
       if (
-        absoluteIJK[0] < extent[0] ||
-        absoluteIJK[0] > extent[1] ||
-        absoluteIJK[1] < extent[2] ||
-        absoluteIJK[1] > extent[3] ||
-        absoluteIJK[2] < extent[4] ||
-        absoluteIJK[2] > extent[5]
```
jourdain: You should add a snap to the corresponding extent for absoluteIJK when:

`extent[0] - absoluteIJK[0] < 0.25` ...

Sources/Rendering/Core/ImageMapper/index.js
floryst: From what I've seen, it looks like Alexis needs `Math.round` to pick the discrete ijk cell closest to absoluteIJK, whereas the `intersectWithLineForCellPicking` code uses `floor` to get the containing ijk cell for absoluteIJK (from my understanding of vtkjs cells). So we have two different use case: one to get containing cell, another to get closest cell to a continuous ijk index.

jourdain: I think what @agirault is saying is that when he is getting close to the edge, the picking is not working because of precision issue while the system thinks it is outside of the domain while in fact it should snap to the current extent of the data which will then allow to pick that first cell.

```
+--------+--------+
|       x|        |
|        |        |
+--------+--------+
```
In that example the first cell is picked not the second one...

But in the following one, the first one should be picked too

```
 +--------+--------+
x|        |        |
 |        |        |
 +--------+--------+
```

floryst: Yeah I see that with his erroneous test case. The worldToIndex transform is returning values just slightly under integral values. I think it's a question of where in the pipeline should the precision issue be handled.

agirault: Ok, I understand the distinction between cell vs point thanks for taking the time to explain it... however, when doing the picking intersection, the picking will always be AT a slice, so perfectly between two cells (along the slicing axis):

```
+--------+--------+
|        x        |
|        |        |
+--------+--------+
```

So how would you choose a cell id here?

Also, does that mean that to get an ijk slice index I should not use a cell picker?

Sources/Rendering/Core/ImageMapper/index.js
agirault: Or just do the check after the flooring/rounding.

Also, according to what you said, we shouldn't compare to the extent when doing cell picking, correct?

Sources/Rendering/Core/ImageMapper/index.js
jourdain: > So how would you choose a cell id here?
 
Math.floor

> Also, does that mean that to get an ijk slice index I should not use a cell picker?

You should use the picker of the location of your data. 
- PointData => PointPicker
- CellData => CellPicker

agirault:
@jourdain @floryst I reverted the discussed change, to have floor used for cell picking and round use for point picking. I also updated the extent for cell picking to limit it to [extentMin, extentMax - 1].
Also, added image mapper point picking to the point picker class, and fixed an issue with pCoords and pickedPositions not properly set (fixes #545).
cc: @martinken @thewtex

jourdain:
Scanning through the changes, it seems ok.

agirault:
By @thewtex :

thewtex:
Some desirables, in my humble opinion,

- The commit messages mention fixes to things like the picked position, but there are not any tests that I see that demonstrate what was fixed and ensure it stays fixed ;-).
- One letter variables are hard to search for and do not convey meaning well. If `t` and `p` are used, especially in contexts outside of there they are defined, a comment they explains what they mean would go a long way. Otherwise, a semantic name would be helpful.

agirault:
> The commit messages mention fixes to things like the picked position, but there are not any tests that I see that demonstrate what was fixed and ensure it stays fixed ;-).

Thanks, I'll look at adding a test. By returning a valid `t` instead of 0 [here](https://github.com/Kitware/vtk-js/blob/16e6c71cc0106d367cc9ffe53f566f5fd3225c5d/Sources/Rendering/Core/CellPicker/index.js#L185), a correct position will be computed and stored in pickedPositions [here](https://github.com/Kitware/vtk-js/blob/16e6c71cc0106d367cc9ffe53f566f5fd3225c5d/Sources/Rendering/Core/Picker/index.js#L292-L314).

> One letter variables are hard to search for and do not convey meaning well. If `t` and `p` are used, especially in contexts outside of there they are defined, a comment they explains what they mean would go a long way. Otherwise, a semantic name would be helpful.

Which `p` are you referring to?

For `t`:
- Defined: https://github.com/Kitware/vtk-js/blob/16e6c71/Sources/Common/DataModel/Plane/index.js#L101
- Used: https://github.com/Kitware/vtk-js/blob/16e6c71/Sources/Rendering/Core/Picker/index.js#L292-L302

thewtex:
> Thanks a bunch @thewtex, I'll look at adding tests.

Merci! :bowing_man: 

> Which p are you referring to?

```
function doPicking(p1, p2) {
[...]
publicAPI.intersectWithLineForPointPicking = (p1, p2) 
[...]
publicAPI.intersectWithLine = (p1, p2, tol, mapper) 
```

Are the p's here physical positions? Parametric cell positions? `point` or `parametricValue` would convey meaning.

> Where do you suggest I add a comment?

In functions where `t` is used, a comment like

```
// `t` is the parametric position along the line, defined in Plane.intersetWithLine
```

Sources/Rendering/Core/CellPicker/test/testCellPicker.js
```diff
@@ -0,0 +1,79 @@
...
+  const actors = picker.getActors();
+  t.equal(actors.length, 1);
+  t.equal(actors[0], actor);
+
+  const positions = picker.getPickedPositions();
+  t.equal(positions.length, 1);
+  const xyz = positions[0];
+  t.equal(xyz[0], 64.49344014125458);
+  t.equal(xyz[1], 75.65264146187872);
+  t.equal(xyz[2], 12.000000145434939);
```
thewtex: :+1:

Sources/Rendering/Core/PointPicker/test/testPointPicker.js
```diff
@@ -0,0 +1,79 @@
...
+  const actors = picker.getActors();
+  t.equal(actors.length, 1);
+  t.equal(actors[0], actor);
+
+  const positions = picker.getPickedPositions();
+  t.equal(positions.length, 1);
+  const xyz = positions[0];
+  t.equal(xyz[0], 64.49344014125458);
+  t.equal(xyz[1], 75.65264146187872);
+  t.equal(xyz[2], 12.000000145434939);
```
thewtex: :1st_place_medal:

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -178,9 +178,9 @@ function vtkCellPicker(publicAPI, model) {
     if (mapper.isA('vtkImageMapper')) {
       const pickData = mapper.intersectWithLineForCellPicking(p1, p2);
       if (pickData) {
-        tMin = 0;
+        tMin = pickData.t;
         model.cellIJK = pickData.ijk;
-        model.pCoords = pickData.point;
+        model.pCoords = pickData.pCoords;
```
thewtex: Hurray for tests! :fireworks: :champagne:

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -338,9 +338,11 @@ function vtkImageMapper(publicAPI, model) {
       const point = intersect.x;
       const absoluteIJK = [0, 0, 0];
       imageData.worldToIndex(point, absoluteIJK);
+      // `t` is the parametric position along the line
+      // defined in Plane.intersectWithLine
```
thewtex: Nice, thanks

PR #677, opened 2018-03-29
fix(AnnotatedCubeActor): Vertically flip face textures

floryst:
Textures got inverted in 43ed0fb, so as a workaround we Y-flip the faces.  @agirault


agirault:
@martinken is that how we should do things?

martinken:
Yes, if taking images from a canvas you have to flip them in Y

PR #679, opened 2018-04-01
nbSteps shoud be non negative otherwise values is never populated

AndreFrelicot:


jourdain:
Is there any reason why you would provide a negative number of steps?

jourdain:
Also could you format your commit message like described here: https://kitware.github.io/vtk-js/docs/misc_tools.html

AndreFrelicot:
In Glance generateValues is called like this :

```
if (this.activeRepresentation && this.activeRepresentation.getSlice) {
      const {
        min,
        max,
        step,
      } = this.activeRepresentation.getPropertyDomainByName('slice');
      this.slider.generateValues(min, max, 1 + (max - min) / step);

```

And I tested on some DICOM datasets in which le Z min and max stays in the negative range (ex -1000 --> -500) so a negative nbSteps is provided.

jourdain:
The issue should be fixed in domain computation to make sure the nbStep is positive.

AndreFrelicot:
I confirm I get negative step with for example this image : 
image

I've finally fixed it at paraview glance level (Layout2D) : 
`this.slider.generateValues(min, max, 1 + (max - min) / Math.abs(step));`

I don't know how to do it in "domain computation". Maybe at the UI level is the best way (?)

AndreFrelicot:
I'm on the Z orientation, the step = the Z spacing (-3.8666...). I think it's better to fix this at the UI level in pv-web-viewer and cancel this pull request.

PR #685, opened 2018-04-09
fix(Rendering): add timer for skybox app

martinken:
If the timer url argument is provided then the skybox
viewer will automatically move to the next position
every Y seconds where Y is the provided timer argument.


martinken:
@jourdain Look OK to you?

jourdain:
Yes except that `let currPos = parseInt(mapReader.getPosition(), 10);` assume the name of the position is an number that map to its index. It is not necessary true.

Maybe consider with an issue where `allPositions` are not ordered the way you want...
=> try to sort them in the reader...

```
if (autoIncrementTimer !== 0) {
  setInterval(() => {
    const currentPosition = mapReader.getPosition();
    const allPositions = mapReader.getPositions();
    const nextIdx = (allPositions.indexOf(currentPosition) + 1) % allPositions.length;
    updateSkybox(allPositions[nextIdx]);
  }, autoIncrementTimer * 1000);
}
```

PR #688, opened 2018-04-09
Support adding and removing colormap presets

floryst:
Haven't tested addPreset and removePresetByName, but they look reasonable thus far.

### DIFF
--- a/Sources/Rendering/Core/ColorTransferFunction/ColorMaps.js
+++ b/Sources/Rendering/Core/ColorTransferFunction/ColorMaps.js
@@ -1,29 +1,55 @@
 import vtkColorMaps from 'vtk.js/Sources/Rendering/Core/ColorTransferFunction/ColorMaps.json';
 
-const uniqueNames = {};
+const presetMap = Object.create(null);
 
 vtkColorMaps
   .filter((p) => p.RGBPoints)
   .filter((p) => p.ColorSpace !== 'CIELAB')
-  .map((p) => p.Name)
-  .forEach((name) => {
-    uniqueNames[name] = true;
+  .forEach((p) => {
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/ColorTransferFunction/ColorMaps.js
```diff
@@ -1,29 +1,55 @@
 import vtkColorMaps from 'vtk.js/Sources/Rendering/Core/ColorTransferFunction/ColorMaps.json';
 
-const uniqueNames = {};
+const presetMap = Object.create(null);
```
jourdain: Why `Object.create(null)` vs `{}` ? ;-)

floryst: Just a sane default. Realistically I don't think anyone would try to get a preset with the names `__proto__` or `toString` and get the function, but this removes that possibility.

aronhelser: I think this is a bad idea. Then presetMap is an object without any of the expected default methods. It would definitely surprise me. For instance, `${presetMap}` will now fail with a TypeError (not that the string "[object Object]" is all that useful, but you get the idea).

jourdain: Except that it is a totally private instance. We could also use `new Map()` without worrying of breaking anything.

floryst: Good point. However, presetMap doesn't really get exposed to the user and is only used internally to store presets. The only time I can see this as an issue is if you were adding console.logs to ColorMaps.js to debug presetMap in some way.

PR #694, opened 2018-04-12
refactor(RenderWindowInteractor): Avoid stopping onClick event propag…

luciemac:
…ation


jourdain:
@floryst I think it is fine as we don't do anything with click. But wanted to run it by you if you were thinking of a use case that I forgot about.

floryst:
I can't think of anything that uses click right now since we pretty much use mousedown/move. This shouldn't affect things in the future should we happen to use onclick, so LGTM.

PR #696, opened 2018-04-13
fix(SkyboxReader): Improve image processing

jourdain:


jourdain:
@martinken this allow to process the images before all of them are loaded. 
That's still not enough to solved the crash we were seeing this morning. But it is working...

If you feel like it, we can merge it.

PR #698, opened 2018-04-16
PiecewiseFunctionProxy modes

floryst:
Adds Points and Nodes modes to piecewise function proxy. 

Adds `arrayEquals` and `objEquals` to test utils. Adds PiecewiseFunctionProxy test.

Fixes naming error in PiecewiseFunction (Midpoint -> midpoint, Sharpness -> sharpness).

### DIFF
--- a/Sources/Common/DataModel/PiecewiseFunction/index.js
+++ b/Sources/Common/DataModel/PiecewiseFunction/index.js
@@ -149,8 +149,8 @@ function vtkPiecewiseFunction(publicAPI, model) {
 
     val[0] = model.nodes[index].x;
     val[1] = model.nodes[index].y;
-    val[2] = model.nodes[index].Midpoint;
-    val[3] = model.nodes[index].Sharpness;
+    val[2] = model.nodes[index].midpoint;
+    val[3] = model.nodes[index].sharpness;
 
     return 1;
   };
@@ -167,8 +167,8 @@ function vtkPiecewiseFunction(publicAPI, model) {
     const oldX = model.nodes[index].x;
     model.nodes[index].x = val[0];
     model.nodes[index].y = val[1];
-    model.nodes[index].Midpoint = val[2];
-    model.nodes[index].Sharpness = val[3];
+    model.nodes[index].midpoint = val[2];
+    model.nodes[index].sharpness = val[3];
 
     if (oldX !== val[0]) {
       // The point has been moved, the order of points or the range might have
--- a/Sources/Proxy/Core/PiecewiseFunctionProxy/index.js
+++ b/Sources/Proxy/Core/PiecewiseFunctionProxy/index.js
@@ -3,6 +3,34 @@ import macro from 'vtk.js/Sources/macro';
 import vtkPiecewiseFunction from 'vtk.js/Sources/Common/DataModel/PiecewiseFunction';
 import vtkPiecewiseGaussianWidget from 'vtk.js/Sources/Interaction/Widgets/PiecewiseGaussianWidget';
 
+import Constants from './Constants';
+
+const { Mode, Defaults } = Constants;
+
+// ----------------------------------------------------------------------------
+
+function applyPointsToPiecewiseFunction(points, range, pwf) {
+  const width = range[1] - range[0];
+  const rescaled = points.map(([x, y]) => [x * width + range[0], y]);
+
+  pwf.removeAllPoints();
+  rescaled.forEach(([x, y]) => pwf.addPoint(x, y));
+  pwf.sortAndUpdateRange();
+}
+
+// ----------------------------------------------------------------------------
+
+function applyNodesToPiecewiseFunction(nodes, range, pwf) {
+  const width = range[1] - range[0];
+  const rescaled = nodes.map((n) =>
+    Object.assign({}, n, { x: n.x * width + range[0] })
+  );
+
+  pwf.removeAllPoints();
+  pwf.set({ nodes: rescaled }, true);
+  pwf.sortAndUpdateRange();
+}
+
 // ----------------------------------------------------------------------------
 // vtkPiecewiseFunctionProxy methods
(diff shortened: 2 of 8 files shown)


Sources/Common/DataModel/PiecewiseFunction/index.js
```diff
@@ -167,8 +167,8 @@ function vtkPiecewiseFunction(publicAPI, model) {
     const oldX = model.nodes[index].x;
     model.nodes[index].x = val[0];
     model.nodes[index].y = val[1];
-    model.nodes[index].Midpoint = val[2];
-    model.nodes[index].Sharpness = val[3];
+    model.nodes[index].midpoint = val[2];
+    model.nodes[index].sharpness = val[3];
```
jourdain: Are you sure that nowhere else Midpoint/Sharpness is used? It should have been lower case but since it is a breaking change. We need to make sure vtk.js is consistent internally...

Sources/Proxy/Core/PiecewiseFunctionProxy/index.js
```diff
@@ -11,17 +15,83 @@ function vtkPiecewiseFunctionProxy(publicAPI, model) {
...
+        publicAPI.modified();
+        break;
+
+      case Mode.Points:
+        model.piecewiseFunction.removeAllPoints();
+        model.points.forEach((point) =>
+          model.piecewiseFunction.addPoint(...point)
+        );
+        model.piecewiseFunction.sortAndUpdateRange();
+
```
jourdain: you have a scaling issue

Sources/Proxy/Core/PiecewiseFunctionProxy/index.js
```diff
@@ -11,17 +15,83 @@ function vtkPiecewiseFunctionProxy(publicAPI, model) {
...
+        publicAPI.modified();
+        break;
+
+      case Mode.Nodes:
+        model.piecewiseFunction.removeAllPoints();
+        model.piecewiseFunction.set({ nodes: model.nodes }, true);
+        model.piecewiseFunction.sortAndUpdateRange();
+
```
jourdain: you have a scaling issue

Sources/Common/DataModel/PiecewiseFunction/index.js
floryst: I grepped the source code and found no usage of `Midpoint` and `Sharpness` in capitalized form. For that matter, all tests pass as well.

floryst:
Scaling issues addressed.

floryst:
Does this look good?

Sources/Proxy/Core/PiecewiseFunctionProxy/index.js
```diff
@@ -3,6 +3,34 @@ import macro from 'vtk.js/Sources/macro';
...
+// ----------------------------------------------------------------------------
+
+function applyNodesToPiecewiseFunction(nodes, range, pwf) {
+  const width = range[1] - range[0];
+  const rescaled = nodes.map((n) =>
+    Object.assign({}, n, { x: n.x * width + range[0] })
+  );
+
+  pwf.removeAllPoints();
+  pwf.set({ nodes: rescaled }, true);
```
jourdain: There is not setNodes? Are we sure we don't need to do additional actions like modified(), build()... ?

jourdain: I think you can call set points with sharpness and so on...

Sources/Proxy/Core/PiecewiseFunctionProxy/index.js
```diff
@@ -11,17 +39,86 @@ function vtkPiecewiseFunctionProxy(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkPiecewiseFunctionProxy');
 
+  // Copy to not keep references
+  model.gaussians = model.gaussians.slice();
+  model.points = model.points.slice();
+  model.nodes = model.nodes.slice();
+  model.dataRange = model.dataRange.slice();
+
```
jourdain: This is already done inside the macro.safeArray called by macro.obj (I think)

jourdain:
LGTM except the nodes handling (if we can do better).
Also check the safeArray part.

Sources/Proxy/Core/PiecewiseFunctionProxy/index.js
floryst: There's `setNodeValue`, but no `setNodes`. I modeled this after `applyGaussianToPiecewiseFunction`, which uses `set()` directly. `sortAndUpdateRange` is the PiecewiseFunction equivalent of build+modified as far as I can tell.

Sources/Proxy/Core/PiecewiseFunctionProxy/index.js
floryst: Cool! Checked it out, and it's there. That's good to know.

Sources/Proxy/Core/PiecewiseFunctionProxy/index.js
floryst: Should we add `setNodes`, which will implicitly call `sortAndUpdateRange`, to PiecewiseFunction?

jourdain: ok, I guess this is fine for now. 
Still we should have a single method inside the piecewiseFunction to not externalize the responsibility like I did in the VolumeRepProxy with the default...

jourdain: Yes I would think so...

floryst: I'll make that a separate PR then.

jourdain: ok

PR #703, opened 2018-04-17
fix(LookupTableProxy): Support different data input modes

floryst:
This is similar to the PiecewiseFunctionProxy modes except applied to the LookupTableProxy, since I needed to be able to set custom colormap points (e.g. shifting points) without actually registering a new colormap each time.

Does this seem reasonable enough? 

Also edited ColorTransferFunction to support setting nodes directly, similar to PiecewiseFunction.setNodes.


floryst:
I know the ColorTransferFunction supports adding HSV points as well, so I can also add that...

Sources/Proxy/Core/LookupTableProxy/index.js
```diff
@@ -3,7 +3,45 @@ import macro from 'vtk.js/Sources/macro';
...
+
+// ----------------------------------------------------------------------------
+
+function applyRGBPointsToLookupTable(rgbPoints, dataRange, lut) {
+  const width = dataRange[1] - dataRange[0];
+  const rescaled = rgbPoints.map(([x, ...rgb]) => [
+    x * width + dataRange[0],
+    ...rgb,
+  ]);
+
```
jourdain: setMappingRange is doing the rescale. So no need to do it.

Sources/Proxy/Core/LookupTableProxy/index.js
```diff
@@ -3,7 +3,45 @@ import macro from 'vtk.js/Sources/macro';
...
+  lut.updateRange();
+}
+
+// ----------------------------------------------------------------------------
+
+function applyNodesToLookupTable(nodes, dataRange, lut) {
+  const width = dataRange[1] - dataRange[0];
+  const rescaled = nodes.map((n) =>
+    Object.assign({}, n, { x: n.x * width + dataRange[0] })
+  );
```
jourdain: no scaling needed.

Sources/Proxy/Core/LookupTableProxy/index.js
floryst: Oh haha I thought it was going to be the same as with PiecewiseFunction...

floryst:
Removed scaling and added HSVPoints.

jourdain:
shifting is similar to change data range...

floryst:
Yeah, looks like setting the data range for the LookupTableProxy does the same thing I need for colormaps. I assumed it wouldn't work since it didn't work for me with the PiecewiseFunction and the companion Proxy.

jourdain:
It was a something we needed in LookupTable hence we added it within the core of the class while the piecewiseFunction was left behind. That's why in one case it is just a method call and in the other we need to rework the data externally... We should probably revisit the pwf to do something similar...

jourdain:
One of the reason why we needed it sooner on the LUT was that the preset had some random range definitions.

floryst:
Gotcha, yeah adding some mappingRange to the pwf will be useful in the future.

The changes in this PR aren't relevant to me anymore, but I suppose it still has use and can be merged?

jourdain:
Sure but we might want to bundle it along with the other PR as they are ready too...

PR #711, opened 2018-04-20
fix(ITKImageReader): Support itk.js 9.0.0

floryst:
itk.js 9.0.0 contains breaking changes. This changeset keeps
ITKImageReader backwards compatible with pre-9.0.0 itk.js installs.

This just terminates all webworkers after a single use. Do we want this configurable?

@thewtex


thewtex:
:+1: Nice, @floryst 

> Do we want this configurable?

I think it is good as-is, at least for now.

Issue #682, opened 2018-04-04
Widgets refactoring


agirault:
Following-up from discussion notes here:
https://docs.google.com/document/d/1lbFj4N68PhuFC8NElz_7wVW655mMdxRdP8Qkn17wwA4

cc: @lassoan  @floryst @jcfr @martinken @thewtex @aylward @timthirion

## Background
Widgets in vtk-js are at a very early stage. We brought up some issues with the current architecture, and believe now is the time to address some of them since the number of widgets in vtk.js is still very limited.

## Identified Issues
* Confusions between roles of widget and widgetRepresentation leads to unnecessary complexities
* Lack of separation between representation and source limits data sharing between multiple representations and forces duplications and syncing of state
* Very costly to have a lot of handle widgets

## Proposed solutions
- [ ] Separate widget architecture into 3 classes for:
  - interaction (listening currently done in widget, and handling currently done in widgetRep),
  - representation (currently actor and mapper in widgetRep)
  - state/data (currently source in widgetRep).
- [ ] Base handlewidget on glyphs to optimize the creation of a high number of similar widgets (?)

*Please edit with suggestions...*

lassoan:
Design could be somewhat similar to model-view-controller pattern, so I will use these terms here, but the final class names may be different:

- Model: Stores point positions, curve interpolation type, etc. There is one instance shared between all renderers. Maybe it could store some state information, such as selected or highlighted point, so that a point that is being selected or dragged can be highlighted in all views.
- View: Contains actors. Stores display properties (color, opacity, etc). There must be a separate instance for each renderer. There are different views, for example one for 3D and another one for 2D views. Probably the view needs to observe GUI events (mouse move, click, etc) and generate high-level events (point clicked, point moved, key pressed) that the controller observes.
- Controller: Observes view events and updates the model based on that. Probably we need a controller instance for each renderer, but it might be shared between renderers.

Handle widget: I don't think we can afford to have a HandleWidget object per each handle. This is what killing the performance today. Instead a HandleWidget object must manage a set of points (up to at least a few thousands): rendering, picking, highlighting, selecting, moving

agirault:
@lassoan what about:
* model/state/source: what you said, with API to update state (such as translate, scale...)
* view/representation: storing actor, mapper, and display properties, with API to check if it is picked at a certain location (but no event handling, it's almost the state of the rendering)
* controller: listen to events from interactor (and not representation), check if interacting with the representation (using representation API), and update model/state accordingly.

[...]

PR #715, opened 2018-04-26
fix(LineWidget): Do not unnecessary refresh view on mouse move

finetjul:
In Manipulate mode, when hovering the render window, the LineWidget could trigger an unnecessary
InteractionEvent and refresh

### DIFF
--- a/Sources/Interaction/Widgets/LineWidget/index.js
+++ b/Sources/Interaction/Widgets/LineWidget/index.js
@@ -156,23 +156,24 @@ function vtkLineWidget(publicAPI, model) {
 
   publicAPI.moveAction = (callData) => {
     const position = [callData.position.x, callData.position.y];
+    let modified = false;
 
     if (model.widgetState === WidgetState.MANIPULATE) {
       // In MANIPULATE, we are hovering above the widget
       // Check if above a sphere and enable/disable if needed
       const state = model.widgetRep.computeInteractionState(position);
       setCursor(state);
-      if (state !== State.OUTSIDE) {
-        if (state === State.ONP1) {
-          model.point1Widget.setEnabled(1);
-          model.point2Widget.setEnabled(0);
-        } else if (state === State.ONP2) {
-          model.point1Widget.setEnabled(0);
-          model.point2Widget.setEnabled(1);
-        }
-      } else {
-        model.point1Widget.setEnabled(0);
-        model.point2Widget.setEnabled(0);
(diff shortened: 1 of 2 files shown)


agirault:
Nice, thanks!

One thing I wonder, instead of doing:
```js
const enablePoint1Widget = state === State.ONP1;
if (enablePoint1Widget !== model.point1Widget.getEnabled()) {
  model.point1Widget.setEnabled(enablePoint1Widget);
  modified = true;
}
```
it would be nice to do:
```js
modified = modified || model.point1Widget.setEnabled(state === State.ONP1);
```

`AbstractWidget.setEnabled` and `InteractorObserver.setEnabled` already do the check internally not to call `modified()` if nothing changed. We could return `true/false` if it has been modified, my understanding is that it's the convention for `set` methods in VTK and vtk.js. @jourdain could you confirm?

Then, the same thing goes for `setPoint2WorldPosition` and the other methods in `LineRepresentation`, it would be easy too just return the result of the internal `set*` call... but now that I look at it, there are a LOT of calls, so I understand if we don't want to do that for that PR.

jourdain:
Yes the set should return true if something has changed. The macro generated method will do it but it is good practice to do it inside your own methods...

finetjul:
I like your suggestion of having setEnabled returning true/false if modified. 
However there is a risk in setXXX() methods to always perform the check if the new value is truely different from the current value. 
There might be some cases where the check would be too costly (large arrays) or not well defined (precision of the check for floating numbers)...

jourdain:
If costly then go with the "always mark modified" to be on the safe side.

finetjul:
For the costly case, how about adding an extra `force` parameter (default to false). e.g. 
```
setMyParamWithExpensiveCheck(newValue, true); // where true means to "force" the new value without checking
```
For the precision case, how about adding a precision parameter ? e.g.
```
setPoint2WorldPosition(newPos, 0.001); // where 0.001 is the precision used to compare each X,Y and Z values
```

jourdain:
Sure, if it is your own implementation, do what you want. Just know that by convention the set method should `return true`, if the call resulted with a change and false otherwise.

agirault:
Indeed :wink: thanks. I edited my first comment to avoid confusion.

That pr is merged though, I think there is no need to worry too much about using good conventions for the widgets, we plan to do some refactoring when we get some funding for a project that requires widgets: #682

PR #732, opened 2018-05-17
feat(SphereHandleRepresentation): Expose property and selectProperty

finetjul:
The selection color of a handle widget must be editable

### DIFF
--- a/Sources/Interaction/Widgets/SphereHandleRepresentation/index.js
+++ b/Sources/Interaction/Widgets/SphereHandleRepresentation/index.js
@@ -298,9 +298,9 @@ function vtkSphereHandleRepresentation(publicAPI, model) {
 
   publicAPI.highlight = (highlight) => {
     if (highlight) {
-      model.actor.setProperty(model.selectProperty);
+      publicAPI.applyProperty(model.selectProperty);
     } else {
-      model.actor.setProperty(model.property);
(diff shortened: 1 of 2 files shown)


Sources/Interaction/Widgets/SphereHandleRepresentation/index.js
```diff
@@ -350,7 +350,11 @@ export function extend(publicAPI, model, initialValues = {}) {
   // Inheritance
   vtkHandleRepresentation.extend(publicAPI, model, initialValues);
 
-  macro.setGet(publicAPI, model, ['translationMode']);
+  macro.setGet(publicAPI, model, [
+    'translationMode',
+    'property',
```
jourdain: Is there a `property` in the model?

finetjul: Yes, line 338

jourdain: I don't see it...

laurennlam: @jourdain , the property of SphereHandleRepresentation is defined here : https://github.com/finetjul/vtk-js/blob/e706894165f7afbd45fc57978b902226ba2c92ea/Sources/Interaction/Widgets/SphereHandleRepresentation/index.js#L338

PR #743, opened 2018-05-28
fix(ImageMapper): Render InterpolationType changes

thewtex:
Make the system recognize the state change and respond
when vtkImageProperty.setInterpolationType is called.


martinken:
+1

PR #747, opened 2018-06-04
Add colors to the sticks representing bonds between atoms

### LINKED ISSUE #544: Bonds should use atom colors in molecule rendering
Adding this to track the issue, I don't have any time to do it right now. Virtually all molecular rendering uses the connected atom colors in the bonds - vtk.js currently uses a [single color](https://kitware.github.io/vtk-js/examples/PDBReader.html) for all bonds (white). See the ParaView/VTK rendering shown a [blog post](https://blog.kitware.com/new-features-paraview-chemistry-pettt/) about new features for example, and also [PDB examples](http://www.rcsb.org/3d-view/5NNK).

The rendering code should be modified to either use two solid colors as in the VTK example, or to linearly blend from the first atom color to the second along the tube's length. This should become the default once implemented.

alesgenova:
The sticks representing covalent bonds in molecules are now colored in
the same way as the atoms involved in the bond.

If the bond involves two atoms of different species, the stick is
actually split into two sticks, so that we can color each independently.

Fixes #544 

screenshot from 2018-06-04 13-54-26


jourdain:
Oups, I forgot to mention to @alesgenova that the formatting of the commit message was important so new release get triggered automatically. [See](https://kitware.github.io/vtk-js/docs/misc_contributing.html#Workflow) for more details.

alesgenova:
@jourdain 
Sorry, I actually read the instructions when I started developing but forgot about it when commit time came. Will keep it in mind for next time!

jourdain:
No worries, I'm sure someone will push something new to vtk.js in the coming days and a new version with your fix will then be available to anyone.

alesgenova:
@jourdain I can look into it!
Where is the structure coming from? Steps to reproduce the bug?

jourdain:
That happened once I start filtering some atoms.

cryos:
What kind of filtering? I didn't know vtk-js could filter atoms. If there is example code we could use that as a starting point to diagnose what is wrong. It is odd that the start/end position is correct, but I would guess the atomic number seen is wrong.

jourdain:
https://github.com/Kitware/vtk-js/blob/master/Sources/Filters/General/MoleculeToRepresentation/index.js#L451

alesgenova:
It should be fixed in #753

jourdain:
Still see the issue...

jourdain:
screen shot 2018-06-12 at 2 03 04 pm
screen shot 2018-06-12 at 2 03 12 pm

alesgenova:
re-fixed in #755 
Also tested in Paraview Glance

jourdain:
Validated as well and working. Thanks!

PR #756, opened 2018-06-13
refactor(Camera): Allow setting key matrices directly

sankhesh:
This change provides the ability to set the camera key matrices directly
from an external application. This ability opens up the posibility of
integrating vtk-js within a larger WebGL application.

### DIFF
--- a/Sources/Rendering/Core/Camera/index.js
+++ b/Sources/Rendering/Core/Camera/index.js
@@ -417,6 +417,10 @@ function vtkCamera(publicAPI, model) {
     // invert to get hmd -> world
     mat4.invert(viewMatrix, viewMatrix);
 
+    publicAPI.computeViewParametersFromViewMatrix();
+  };
+
+  publicAPI.computeViewParametersFromViewMatrix = () => {
     // then extract the params position, orientation
     // push 0,0,0 through to get a translation
     vec3.transformMat4(tmpvec1, origin, viewMatrix);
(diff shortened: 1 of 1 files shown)


sankhesh:
@martinken @aashish24 Please review

Sources/Rendering/Core/Camera/index.js
```diff
@@ -417,6 +417,10 @@ function vtkCamera(publicAPI, model) {
     // invert to get hmd -> world
     mat4.invert(viewMatrix, viewMatrix);
 
+    publicAPI.computeViewParametersFromViewMatrix();
+  };
+
+  publicAPI.computeViewParametersFromViewMatrix = () => {
     // then extract the params position, orientation
     // push 0,0,0 through to get a translation
     vec3.transformMat4(tmpvec1, origin, viewMatrix);
```
jourdain: viewMatrix is not defined here? You may need to store / retrieve it from the model...

Sources/Rendering/Core/Camera/index.js
```diff
@@ -417,6 +417,10 @@ function vtkCamera(publicAPI, model) {
     // invert to get hmd -> world
     mat4.invert(viewMatrix, viewMatrix);
```
jourdain: need to save viewMatrix to model...

Sources/Rendering/Core/Camera/index.js
sankhesh: Same as above.

The idea is to have a function scope viewMatrix as well as a model scope one. When an external application sets the viewMatrix explicitly, it is assigned to the model scope and used whenever `getViewMatrix` is called.

Sources/Rendering/Core/Camera/index.js
sankhesh: Nope, that's intentional.
`viewMatrix` is defined in the vtkCamera function. 
See https://github.com/sankhesh/vtk-js/blob/00c21d6f77aa9c82cd97e5053fa2508c1eb9912e/Sources/Rendering/Core/Camera/index.js#L29

jourdain:
I see thanks for the explanation. Glancing from the diff, it was not clear where that viewMatrix was defined.

Sources/Rendering/Core/Camera/index.js
jourdain: Should the publicAPI.computeViewParametersFromViewMatrix function put on the public API?

martinken:
Seems reasonable +1

Sources/Rendering/Core/Camera/index.js
jourdain: Just don't bind that function on publicAPI and that will be fine. that way the function will be private inside that scope.

Sources/Rendering/Core/Camera/index.js
```diff
@@ -438,24 +424,37 @@ function vtkCamera(publicAPI, model) {
...
   publicAPI.getViewMatrix = () => {
-    const result = mat4.create();
     if (model.viewMatrix) {
-      mat4.copy(result, model.viewMatrix);
-      mat4.transpose(result, result);
-      return result;
+      return model.viewMatrix;
     }
 
+    const result = mat4.create();
```
jourdain: maybe not needed anymore?

sankhesh:
@jourdain Thanks for reviewing and merging 👍

PR #775, opened 2018-06-23
fix(STLReader): add ascii support

jourdain:


agirault:
FYI, I think the stl reader already existed in itk.js (@thewtex)

jourdain:
That's weird, because anything that itk list as supported file, we simply expose and use.

jourdain:
But I don't think vtk.js has a mapping between ITK/mesh and VTK/mesh, we only have an Image handler. That could be the reason too...

thewtex:
Added an issue here for a bridge to itk.js mesh file formats:

  https://github.com/Kitware/vtk-js/issues/789

PR #784, opened 2018-07-03
feat(ImageReslice): Add vtkImageReslice filter

finetjul:
This is a limited implementation of vtkImageReslice.
It does not support:
 - LINEAR or CUBIC interpolation.
 - Slab mode
 - Permute optimization
 - Convenient reslice axes methods


finetjul:
@agirault can you please review ?

laurennlam:
@jourdain Would it be possible to merge it as there are only new files which don't affect the actual vtk.js version ? Because we really need it for a customer.

Sources/Imaging/Core/AbstractImageInterpolator/index.js
```diff
@@ -0,0 +1,240 @@
...
+const DEFAULT_VALUES = {
+  outValue: 0,
+  tolerance: Number.EPSILON,
+  componentOffset: 0,
+  componentCount: -1,
+  borderMode: ImageBorderMode.CLAMP,
+  slidingWindow: false,
+
+  scalars: null,
+  interpolationInfo: Object.assign({}, vtkInterpolationInfo),
```
jourdain: I'm guessing that you know that all the arrays inside are just reference to a single instance of those array?

finetjul: Yes, I use `subarray` as an address pointer. I don't want to copy the array only to point at different places in the array.

jourdain: subarray will recreate a new array...

finetjul: It's a new array but it points on the same array buffer. So it is fast (no copy)

jourdain: Ok I was talking about the array you define by [] inside vtkInterpolationInfo.

PR #790, opened 2018-07-05
Python 3.6 compatible conversion script

cewee:
minor changes , print (), and buffer => memoryview


cewee:
Will do that. Reopening a different MR.

PR #792, opened 2018-07-09
Update vtk-data-converter.py

cewee:
for vtk image data, the setScalars are missing.
not sure if this is the appropriate way to add it, but this change fixes the problem for me.
see commit:  https://github.com/Kitware/vtk-js/commit/96d7965fe738c06e084bf5e87e22f937d5e7f5c0


jourdain:
That's good enough for now. Thanks for your contribution

PR #794, opened 2018-07-10
docs(typo) Patch a typo in the "Creating a new example" page

swederik:
Do I need to do anything other than edit the Markdown for a docs change? I am assuming the CI will build the docs.

Thanks!

### DIFF
--- a/Documentation/content/docs/develop_example.md
+++ b/Documentation/content/docs/develop_example.md
@@ -50,7 +50,7 @@ To list the examples you can run the following command:
 $ npm run example
 ```
 
-Then base on the list of examples you get, you can run a given one by providing its name as argument like follow:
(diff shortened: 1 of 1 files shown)


Documentation/content/docs/develop_example.md
```diff
@@ -50,7 +50,7 @@ To list the examples you can run the following command:
 $ npm run example
 ```
 
-Then base on the list of examples you get, you can run a given one by providing its name as argument like follow:
+Then, based on the list of examples you get, you can run a given one by providing its name as argument like follow:
```
aronhelser: Can you change "like follow" to "like the following"? Looks good! CI will build the docs, like you suspect.

swederik:
There you are!

aronhelser:
Thanks!

PR #798, opened 2018-07-12
Add label widget

luciemac:


luciemac:
@agirault can you review this?

jourdain:
Is it ready to merge? I mean widget architecture may change down the road, but if it is working now, we can probably merge it (if ready)

What do you think @martinken @floryst

luciemac:
@jourdain yes, it is ready to merge.

PR #804, opened 2018-07-18
fix(ViewProxy): Use unique animation requestor id in updateOrientation

thewtex:
This prevents multiple calls to requestAnimation with the same ViewProxy
publicAPI as the requestor, for example, during pipeline initiation
prior to rendering when ViewProxy.updateOrientation may be called
multiple times.


jourdain:
was there any issue related to multiple call with the same requestor?

PR #813, opened 2018-07-20
fix(ProxyManager): Reject promise on save/load error

floryst:
Errors that occur inside the promise should be handled accordingly.


jourdain:
LGTM, the formatting was hard to read

floryst:
Yeah, formatting is a wonky thing for this. Just adds `catch(reject)` to the end of each `Promise.all()`.

jourdain:
yes I managed to extract that... ;-)

jourdain:
The good thing is that glance will have the touch support along with your fix.

PR #819, opened 2018-08-02
fix(ITKImageReader): dynamic scalars name by default

jourdain:


jourdain:
If we merge that we will have an issue with Glance state since the array name / lookup table / piecewise function setup won't match...
We need to fix the array name inside the state so the reader use that specific array name.

@floryst @faustomilletari

Sources/IO/Misc/ITKImageReader/index.js
```diff
@@ -4,6 +4,13 @@ import ITKHelper from 'vtk.js/Sources/Common/DataModel/ITKHelper';
 const { convertItkToVtkImage } = ITKHelper;
 let readImageArrayBuffer = null;
 let resultPreprocessor = (result) => result;
+let uniqueArryCounter = 0;
+
+function getUniqueArrayName() {
+  uniqueArryCounter++;
+  console.log('generate unique scalars', `Scalars_${uniqueArryCounter}`);
```
floryst: Remove console.log.

Did you want to have the name be "Scalars_${num}" or "Scalars ${num}"? Just checking because the log doesn't match the return value.

jourdain: I started with _ but them putting a space looks better

faustomilletari:
will these changes require work on the paraview-glance side?

jourdain:
no only for the state that we have as sample. But @floryst  already have them ready.

faustomilletari:
Weird stuff!!! 

This fixed perfectly the app on decent browsers, such as Chrome et al but it doesn't fix it for safari.
Ugly thing is that safari is also what is being used by iPhone WebView UI component as backend, which makes difficult to build something that looks as an iOS app and works similar to chrome.

Great work anyway! your fix is helping solving the issue! 

(I guess the problem is that the name is not being different for different images in safari!)

jourdain:
Or a cache issue ?

faustomilletari:
yes. I apologize for my mistake. I forgot to empty the cache relative to the IP address of the server. Works 100% now

great work! truly!

PR #825, opened 2018-08-06
fix(Rendering): Update maxLevel of texture mipmap

mcmigliore:
Base and max levels should be sent to WebGL in order to be able to access all levels through `textureLod`.  
Remove `generateMipMap` for cubemaps (all levels are manually uploaded).

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -288,10 +288,20 @@ function vtkOpenGLTexture(publicAPI, model) {
       publicAPI.getOpenGLFilterMode(model.magnificationFilter)
     );
 
+    model.context.texParameteri(
+      model.target,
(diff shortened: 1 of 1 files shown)


mcmigliore:
@laurennlam @martinken please review

martinken:
Are you sure this doesn't remove mipmap support from 2D images? It looks like it does to me as it is setting both base and max level to 0.

mcmigliore:
It seems that you're right. Should I set `maxLevel` parameter to `1000` (default OpenGL value) in `DEFAULT_VALUES` object ?

martinken:
I think so. I suspect webgl is smart enough that it will only use the levels available, so defaulting it to 1000 works with genmipmaps automatic generation.

mcmigliore:
Thanks for pointing this out. Fixed.

martinken:
+1

PR #828, opened 2018-08-07
Glyph fixes

lavaux:
This is a proposed fix for Rendering/Core/Glyph3DMapper. Two issues were found:
- orientation vector is supposed to hold the angles in radian. There was a spurious division by a constant number that looks like Pi. This division has been removed.
- 'orientationMode' attribute was missing from the model in publicAPI.


jourdain:
Usually VTK expect angles to be in Degree, so I'm not sure that your change is correct.

@martinken should know for sure.

Also the OrientationMode was probably expected to be only changed by   `setOrientationModeToDirection()` or `setOrientationModeToRotation()`, but I would be fine adding the set/get on it...

jourdain:
I see that make sense the `set...` should exist. ;-)

I agree that the rotation api expect `Radian` but since VTK consistently require `Degree`, maybe we just need to fix the conversion. But let see what @martinken says...

jourdain:
@martinken what about that one?

martinken:
Yup this is good!

PR #833, opened 2018-08-09
fix(MoleculeToRepresentation): Let bonds start from the center of the atoms

alesgenova:
It looks like after #831 the workaround of having the sticks start from the ball surface is no longer needed


martinken:
I do not know this code very well, but I suspect we should merge it :-) @jourdain ?

jourdain:
I guess

PR #834, opened 2018-08-09
fix(Rendering): improve the termination of vol ren rays

martinken:
Better handling of the end condition of rays where the
opacity peaks right where the ray ends. Solve by taking
a last residual step on webgl2. Improve for both webgl1
and webgl2 with better starting loaction and step calculation.

Fix naming of openGLRenderWindow in GenericRenderWindow

### DIFF
--- a/Sources/Rendering/OpenGL/glsl/vtkVolumeFS2.glsl
+++ b/Sources/Rendering/OpenGL/glsl/vtkVolumeFS2.glsl
@@ -163,18 +163,22 @@ void main()
       dot(endVC, vPlaneNormal0),
       dot(endVC, vPlaneNormal2),
       dot(endVC, vPlaneNormal4));
-    vec3 vdelta = endvpos - vpos;
-    float numSteps = length(vdelta) / sampleDistance;
-    vdelta = vdelta / float(numSteps);
 
     // start slightly inside and apply some jitter
     float jitter = texture2D(jtexture, gl_FragCoord.xy/32.0).r;
-    vpos = vpos + vdelta*(0.01 + 0.98*jitter);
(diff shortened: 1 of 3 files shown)


Sources/Rendering/OpenGL/glsl/vtkVolumeFS2.glsl
```diff
@@ -163,18 +163,22 @@ void main()
...
 
-    // start slightly inside and apply some jitter
-    float jitter = texture2D(jtexture, gl_FragCoord.xy/32.0).r;
-    vpos = vpos + vdelta*(0.01 + 0.98*jitter);
     vec4 color = vec4(0.0, 0.0, 0.0, 0.0);
 
     vec3 ijk = vpos * vVCToIJK;
     vdelta = vdelta * vVCToIJK;
-    int count = int(numSteps - 0.2); // end slightly inside
+    int count = int(numSteps - 0.05); // end slightly inside
```
aashish24: @martinken perhaps I am missing a lot of details here but why not check on opacity saturation or boundary of the box for termination? Is that because we want fix number of steps?

martinken:
Webgl1 does not support while loops and break does not always work unfortunately. So we have to use fixed for loops and test inside for the stop condition and work around the fact that the loop may continue to run. Sort of crazy. 

Regardless count basically translates to iterations before loop end, so it is essentially a test on the boundary condition. e.g. we pre compute the ray start and end locations and compute count to traverse that distance. So count is the test if the ray has reached it's end.

martinken:
You will see down below a opacity test as well. We have it, we just do not count on the break actually working.

Issue #840, opened 2018-08-16
CellData coloring with InterpolateScalarsBeforeMapping


jourdain:
When rendering data with a field on its cells while the flag InterpolateScalarsBeforeMapping is turned ON leads to invalid rendering. That flag should be ignore when coloring by cell data.

Issue #842, opened 2018-08-21
model.rGBTransferFunction.getMTime is not a function


tomhaitao:
const actor = vtkImageSlice.newInstance();
const lookupTable = vtkColorTransferFunction.newInstance();
const preset = vtkColorMaps.getPresetByName('erdc_rainbow_bright');
lookupTable.applyColorMap(preset);
lookupTable.setMappingRange(...dataRange);
lookupTable.updateRange();
actor.getProperty().setRGBTransferFunction(0, lookupTable);

I use this code but vtkImageProperty class line 60 is error:
         model.rGBTransferFunction.getMTime is not a function

tomhaitao:
import vtkColorTransferFunction from 'vtk.js/Sources/Rendering/Core/ColorTransferFunction';

tomhaitao:
How to set color for vtkImageSlice?

jourdain:
@martinken

martinken:
actor.getProperty().setRGBTransferFunction(0, lookupTable);

should be

actor.getProperty().setRGBTransferFunction(lookupTable);

I'll add a fix so that it doesn't try calling getMTime on 0, the current test is !== null

PR #844, opened 2018-08-21
fix(Rendering): fix rendering of data with cell scalars

martinken:
Cell scalar fix
https://github.com/Kitware/vtk-js/issues/840

Also improve test output to indicate which image is the
test image and which is the baseline.

Also fix for ImageSlice issue
https://github.com/Kitware/vtk-js/issues/842


martinken:
@jourdain this might fix the cell scalars issue

Issue #803, opened 2018-07-17
InterpolateScalarsBeforeMapping


jourdain:
- [ ] Invalid rendering when color by cell array with `InterpolateScalarsBeforeMapping = true`
- [ ] Color does not match when `InterpolateScalarsBeforeMapping = true` vs `false` when they should match (i.e: computed contours or point cloud).

PR #845, opened 2018-08-21
fix(Rendering): fix for interpolate scalars before mapping

martinken:
The old code used the square of the color because it
mulitplied tghe scalar color twice.

Should fix https://github.com/Kitware/vtk-js/issues/803


jourdain:
I'll test it in visualizer that way I can test all

jourdain:
It does... Thanks!

PR #851, opened 2018-08-27
fix(Webgl): Add webgl 1 extension for texture and mipmap

laurennlam:
Some texture methods have not the same name with webgl 1
@martinken @mcmigliore 
Written in collaboration with Lucas Gandel

### DIFF
--- a/Sources/Rendering/OpenGL/ShaderCache/index.js
+++ b/Sources/Rendering/OpenGL/ShaderCache/index.js
@@ -38,11 +38,19 @@ function vtkShaderCache(publicAPI, model) {
         '#version 300 es\n' +
         '#define attribute in\n' +
         '#define textureCube texture\n' +
-        '#define texture2D texture\n';
+        '#define texture2D texture\n' +
+        '#define textureCubeLod textureLod\n' +
+        '#define texture2DLod textureLod\n';
     } else {
       model.context.getExtension('OES_standard_derivatives');
       if (model.context.getExtension('EXT_frag_depth')) {
         fragDepthString = '#extension GL_EXT_frag_depth : enable\n';
+      }
+      if (model.context.getExtension('EXT_shader_texture_lod')) {
+        fragDepthString +=
+          '#extension GL_EXT_shader_texture_lod : enable\n' +
+          '#define textureCubeLod textureCubeLodEXT\n' +
+          '#define texture2DLod texture2DLodEXT';
       }
     }
 
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -288,17 +288,19 @@ function vtkOpenGLTexture(publicAPI, model) {
       publicAPI.getOpenGLFilterMode(model.magnificationFilter)
     );
 
-    model.context.texParameteri(
-      model.target,
-      model.context.TEXTURE_BASE_LEVEL,
-      model.baseLevel
-    );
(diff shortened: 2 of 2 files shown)


Sources/Rendering/OpenGL/ShaderCache/index.js
```diff
@@ -38,12 +38,19 @@ function vtkShaderCache(publicAPI, model) {
...
+        '#define texture2DLod textureLod\n';
     } else {
       model.context.getExtension('OES_standard_derivatives');
       if (model.context.getExtension('EXT_frag_depth')) {
         fragDepthString = '#extension GL_EXT_frag_depth : enable\n';
       }
+      if (model.context.getExtension('EXT_shader_texture_lod')) {
+        fragDepthString +=
+          '#extension GL_EXT_shader_texture_lod : enable\n' +
+          '#define textureCubeLod textureCubeLodEXT\n';
```
mcmigliore: You should add `#define texture2DLod texture2DLodEXT` too.

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -288,17 +288,19 @@ function vtkOpenGLTexture(publicAPI, model) {
       publicAPI.getOpenGLFilterMode(model.magnificationFilter)
     );
 
-    model.context.texParameteri(
-      model.target,
-      model.context.TEXTURE_BASE_LEVEL,
-      model.baseLevel
-    );
+    if (model.openGLRenderWindow.getWebgl2()) {
```
mcmigliore: Can we do something here to set base and max level if we are in a WebGL1 context? Without proper levels, LOD interpolation will fail.

laurennlam:
@martinken @mcmigliore Do you valid the PR ? If so, is it possible to merge as soon as possible ? It's requested from one of our customer. Thanks.

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -776,22 +778,31 @@ function vtkOpenGLTexture(publicAPI, model) {
...
-            model.internalFormat,
-            w,
-            h,
-            0,
-            model.format,
-            model.openGLDataType,
-            invertedData[6 * j + i]
-          );
+      let j = 0;
+      let w = model.width / 2 ** j;
```
mcmigliore: `/ 2 ** j` is useless as j is equal to zero.

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -776,22 +778,31 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+      let h = model.height / 2 ** j;
+      while (w >= 1 && h >= 1) {
```
mcmigliore: Should we exit if we are in WebGL2 and j reached maxLevel in order to avoid unnecessarily `texImage2D` calls?

jourdain:
@martinken I let you merge assuming this is ready.

Sources/Rendering/OpenGL/Texture/index.js
laurennlam: In webgl1, you can't set base and max level. There are automatically computed by using input textures. All the missing textures are set to null. Then, it's inside your shader where you have to manage the level that will be use. Hope it's clear enough.

laurennlam:
@mcmigliore @martinken , do you think that anything is missing ? Else, can you merge it please. It becomes very urgent for our customer. Thanks.

mcmigliore:
Looks good to me

PR #854, opened 2018-08-28
test(HardwareSelector): Add test for performance

jourdain:


jourdain:
That's the one for you @martinken

PR #855, opened 2018-08-28
fix(Rendering): honor the pickable flag during selection

martinken:
The old code did not consider the state of vtkProp.getPickable


martinken:
@jourdain

jourdain:
Works perfectly ! Thanks... Will merge it from our branch...

PR #861, opened 2018-08-31
feat(Rendering): add a remote pass

jourdain:
Add remote pass support outline


jourdain:
@martinken if you want to see what it looks like so far.

PR #862, opened 2018-09-01
fix(Mapper): Typo for color mapping

jourdain:


jourdain:
FYI @martinken : that was the issue I was seeing

PR #868, opened 2018-09-10
feat(OpenGLRenderWindow): Allow screenshot to also capture labels

laurennlam:
When distance of label widget is added, then add the text into screenshot.
More generally, when a vtkProp contains a container with canvas, then they will be added into the screenshot.


laurennlam:
@luciemac @finetjul @felixveysseyre

PR #874, opened 2018-09-12
fix: return [] rather than null

cjh1:
Calling code is expecting to be able to concat the results of these functions. For example see the implementation of vtkRenderer::getActors() implementation. The current implementation
ends up with nulls in the array and iteration issues.

### DIFF
--- a/Sources/Rendering/Core/Prop/index.js
+++ b/Sources/Rendering/Core/Prop/index.js
@@ -24,9 +24,9 @@ function vtkProp(publicAPI, model) {
   };
 
   publicAPI.getNestedProps = () => null;
-  publicAPI.getActors = () => null;
-  publicAPI.getActors2D = () => null;
-  publicAPI.getVolumes = () => null;
+  publicAPI.getActors = () => [];
+  publicAPI.getActors2D = () => [];
+  publicAPI.getVolumes = () => [];
 
   publicAPI.pick = notImplemented('pick');
   publicAPI.hasKey = notImplemented('hasKey');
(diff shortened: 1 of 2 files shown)


jourdain:
I think we have a pass that expect null to take a faster pass. We may want to make use of that choice otherwise we can have a pretty big overhead for each rendered frame.

cjh1:
@jourdain OK, we are seeing errors in the key pressed event handling code. After a little digging I found that the list of actors returned from vtkRenderer::getActors() contains nulls. I can added null checks in that code if you are using null as a faster pass. However, I would have thought checking for null and iterating a zero length range would boil down to similar instructions.

jourdain:
Yes but vtkProp can be a tree of vtkProp and if by default they all say they have children, then we keep iterating over all of those empty array.

jourdain:
The other option is to make sure that if the array are empty we don't take the long processing pass.
That's the part to review

https://github.com/Kitware/vtk-js/commit/1b4ff6b93a9b6ef052da729ff596336fbf2b73ba

cjh1:
@jourdain Ok, so can this be merged?

jourdain:
Once the code I've linked is fixed as well with a condition like `if (children && children.length) {`

jourdain:
line 42

cjh1:
Done

jourdain:
Once it is green can you merge? Or do I need to click? Meaning do you have the rights?

cjh1:
@jourdain You will have to click the button.

PR #875, opened 2018-09-13
Add public API methods to expand the volume controller.

DominionSoftware:
The issue is actually caused by setupContent() where:
 const widgetContainer = model.el.querySelector('.js-pwf');

has a size. To allow the content to be opened with a small size would I believe require reworking this area. However, the screen update on the computer seems fast enough not to cause a "flash" of the size.


DominionSoftware:
Not sure why my checkin failed? Please advise. Thanks

PR #876, opened 2018-09-15
Add interaction events

DominionSoftware:
For interactor observers, the interation events need to be defined with the others.


jourdain:
Looks good but can you rewrite the commit so the proper changelog can be generated?

See https://gist.github.com/stephenparish/9941e89d80e2bc58a153 from https://kitware.github.io/vtk-js/docs/misc_tools.html

Thanks!

DominionSoftware:
I had to add the commit comments by hand as commitizen doesn't seem to support --amend

PR #884, opened 2018-09-24
Svg overlay widget

jourdain:
DO NOT MERGE

To make sure I don't forget so I could finish it someday

### DIFF
--- a/Sources/Rendering/Core/PixelSpaceCallbackMapper/index.js
+++ b/Sources/Rendering/Core/PixelSpaceCallbackMapper/index.js
@@ -56,7 +56,7 @@ function vtkPixelSpaceCallbackMapper(publicAPI, model) {
       coords.push(coord);
     }
 
-    model.callback(coords, camera, aspect, depthValues);
+    model.callback(coords, camera, aspect, depthValues, windowSize);
   };
 }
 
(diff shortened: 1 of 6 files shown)


jourdain:
@floryst you may want to take over.

The issue I was running into (that I didn't finish) was that a svg can not be dynamically edited (add/remove node inside svg). Basically you need to rebuild it when you add/remove nodes while editing attributes are fine. I guess my travel back from KRS was too short for me to finish it.

aronhelser:
I'm surprised - D3 adds and removes SVG nodes all the time. It's different
somehow?

jourdain:
No, but D3 is most likely rebuilding the SVG node when it does it like Vue.js and React. We just don't know it.

floryst:
Playing around with the svg overlay. Example here: http://forrestli.com/glance/PolyLineWidget/

finetjul:
Hi @floryst , there is a crash when you place a point at the exact same position than previous point. (-> double clic)

floryst:
Yeah I saw that too. I'll push the workaround (I'm pretty sure it's because the vtkTubeFilter doesn't properly handle two points that are the same location).

floryst:
@jourdain I can't seem to request a review from you, so here is a manual request for a review from you. :)

Notes:
- I had to modify the PixelSpaceCallbackMapper to return the window width/height.
- SVG representations don't use requestData, since they don't return a vtk.js object. Instead, they implement `render()` and return a Promise that should resolve to the contents of the SVG node.
- I've modified PolyLineWidget to be the showcase of SVGLandmarkRepresentation

Any comments are welcome.

jourdain:
Looking at the code, it seems reasonable. You should be able to merge unless there is a point you would like to discuss.

floryst:
I'll test it a bit more on my side and then merge. I think it's a reasonable first start.

jourdain:
I think so too.

PR #885, opened 2018-09-24
Travel contrib

jourdain:
DO NOT MERGE

Need to finish it up first but don't want to forget about it...


floryst:
I've made some minor naming fixes and refactored to use our copy of base64-js. This should be good to go now.

jourdain:
feel free to merge

PR #890, opened 2018-09-25
Volume controller colors

### LINKED ISSUE #889: volumeController does not allow user color tables.
currently, the volume controller initializes a set of lookupTables from a file. Ideally, allow the user to provide a set of lookup tables at creation of the volume controller.

DominionSoftware:
This fixes #889 to allow user color tables.

### DIFF
--- a/Sources/Interaction/UI/VolumeController/index.js
+++ b/Sources/Interaction/UI/VolumeController/index.js
@@ -12,9 +12,7 @@ import style from 'vtk.js/Sources/Interaction/UI/VolumeController/VolumeControll
 // Global structures
 // ----------------------------------------------------------------------------
 
-const PRESETS_OPTIONS = vtkColorMaps.rgbPresetNames.map(
-  (name) => `<option value="${name}">${name}</option>`
-);
+let PRESETS_OPTIONS;
 
 // ----------------------------------------------------------------------------
 // vtkVolumeController methods
@@ -23,6 +21,10 @@ const PRESETS_OPTIONS = vtkColorMaps.rgbPresetNames.map(
 function vtkVolumeController(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkVolumeController');
+
+  PRESETS_OPTIONS = vtkColorMaps.rgbPresetNames.map(
+    (name) => `<option value="${name}">${name}</option>`
+  );
 
   model.el = document.createElement('div');
   model.el.setAttribute('class', style.container);
@@ -294,6 +296,12 @@ const DEFAULT_VALUES = {
 };
 
 // ----------------------------------------------------------------------------
+function addCustomPresets(presets) {
+  vtkColorMaps.removeAllPresets();
+  vtkColorMaps.addPresets(presets);
+}
+
+// ----------------------------------------------------------------------------
 
 export function extend(publicAPI, model, initialValues = {}) {
   Object.assign(model, DEFAULT_VALUES, initialValues);
@@ -303,6 +311,10 @@ export function extend(publicAPI, model, initialValues = {}) {
   macro.setGet(publicAPI, model, ['actor', 'renderWindow', 'rescaleColorMap']);
   macro.get(publicAPI, model, ['widget']);
 
+  // Add user custom presets if they exist.
+  if (Object.prototype.hasOwnProperty.call(initialValues, 'userLookupTable')) {
+    addCustomPresets(initialValues.userLookupTable);
+  }
   // Object specific methods
   vtkVolumeController(publicAPI, model);
 }
--- a/Sources/Rendering/Core/ColorTransferFunction/ColorMaps.js
+++ b/Sources/Rendering/Core/ColorTransferFunction/ColorMaps.js
@@ -46,10 +46,22 @@ function removePresetByName(name) {
 }
 
 // ----------------------------------------------------------------------------
+function addPresets(presets) {
+  presets.map(addPreset);
+}
+// ----------------------------------------------------------------------------
+function removeAllPresets() {
+  rgbPresetNames.length = 0;
+  presetMap.length = 0;
+}
+
+// ----------------------------------------------------------------------------
 
 export default {
   addPreset,
   removePresetByName,
   getPresetByName,
   rgbPresetNames,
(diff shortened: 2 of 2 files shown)


Sources/Interaction/UI/VolumeController/index.js
```diff
@@ -303,6 +311,10 @@ export function extend(publicAPI, model, initialValues = {}) {
   macro.setGet(publicAPI, model, ['actor', 'renderWindow', 'rescaleColorMap']);
   macro.get(publicAPI, model, ['widget']);
 
+  // Add user custom presets if they exist.
+  if (Object.prototype.hasOwnProperty.call(initialValues, 'userLookupTable')) {
```
jourdain: I don't think as UI widget should modify the vtkColorMaps. This should be managed at the application level.

Sources/Interaction/UI/VolumeController/index.js
```diff
@@ -24,6 +22,10 @@ function vtkVolumeController(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkVolumeController');
 
+  PRESETS_OPTIONS = vtkColorMaps.rgbPresetNames.map(
```
jourdain: Instead of doing that, you should allow the user to either provide a list of names it wants to see in each of his VolumeController.

So I will drop the global PRESETS_OPTIONS that get's define at the module lever.

DominionSoftware: We (perhaps others) would like to be able to change them on the fly. When a different modality is loaded in (MR, vs CT) a different default would then automatically be displayed to the user. Can we make that happen?

jourdain: definitely that was a consequence of my suggestion.

DominionSoftware: Currently I have created a .json file which is similar/ same to yours. This contains the user visible names of the transfer functions. So, it's easier for use to just have the table passed in, similar to what you're doing.

DominionSoftware: e.g {
    "ColorSpace": "RGB",
    "Name": "CT-AAA2",
    "NanColor": [ 0, 0, 0 ],
    "RGBPoints": [
    -3024,
    0,
    0,
    0,
    129.542,
    0.54902,
    0.25098,
    0.14902,
    145.244,
    0.6,
    0.627451,
    0.843137,
    157.02,
    0.890196,
    0.47451,
    0.6,
    169.918,
    0.992157,
    0.870588,
    0.392157,
    395.575,
    1,
    0.886275,
    0.658824,
    1578.73,
    1,
    0.829256,
    0.957922,
    3071,
    0.827451,
    0.658824,
    1
]
},

jourdain: Well this is a widget, therefore its concern should be limited to a list of names that you want to see in the drop down.

That imply those names needs to be available on the colorMap helper. But that way the widget is just a view of what is available.

DominionSoftware: OK so get rid of the PRESET_OPTIONS variable, make that part of the model, and expose that via set/get of presets by the user? User could therefore just add one preset, or remove one, or replace them all? Thx

DominionSoftware: Do you want something similar for Opacity functions?

jourdain: This should be only a list of string and should not affect the `ColorTransferFunction/ColorMaps.js` and, yes just making that list part of the model.

I don't know for the opacity part.

DominionSoftware: I'm not following what you mean. We need different color-maps from what you have in ColorMaps.js. They are not tuned for the data we have. We have to have the option of replacing the default color maps.

DominionSoftware: We will also need custom Opacity functions (designed by us) at some point in the not so distant future....

jourdain: You already have a fix for adding custom presets in the proper component. Those should be independent from the changes in the widget. I don't want to link them as they should not.

jourdain: For the opacity stuff, you can design your own widget, it does not need to be that one. What matters is that vtk.js provides the pieces that are hard to build like the transfer function editor.
The rest should be specific to your application. It is true that we've created that UI but mostly for our examples code so people can learn how to do things. But I don't see that specific widget evolve too much other than just exposing basic controls for volume rendering. 

To me it still make sense that we can control the list in that drop down, but that's about it...

jourdain: vtk.js aims to be framework agnostic (Angular, React, Vue...) and therefore making UI in plain JavaScript does not simplify that task. That's why I don't see that widget evolve too much.

DominionSoftware: Transfer Function editor is very similar to opacity widget. Not as complex (one channel) but still not trivial. 
 I've written them more than I want to in C++ usually Qt. It would be nice to have one in the toolkit rather than reinvent that every time. 

Anyways, I still do not follow what I'm supposed to do.

 From the OHIF plugin architecture b/c the plugin is dynamically loaded I cannot do any import statements.

I address things such as  controllerWidget = vtk.Interaction.UI.vtkVolumeController.newInstance({
            size: [400, 150],
            rescaleColorMap: true,
        });

if I want to modify the presets prior to instantiation (so the volumeController gets my additional one) and try  vtk.Rendering.Core.ColorTransferFunction.ColorMaps.addPreset(myTables[0]);

I cannot access the ColorTranferFunctions for some reason....

jourdain: This is [the core component](https://github.com/Kitware/vtk-js/tree/master/Sources/Interaction/Widgets/PiecewiseGaussianWidget) that can be reused the way you want. You just need to provide the surrounding like the VolumeController is doing or ParaView Glance, ParaView Visualizer and so on.

Regarding the issue where the preset is missing after adding it. I don't know, it will need some investigation.

PR #901, opened 2018-10-04
Export constants from LookupTableProxy

brianhelba:
This matches what PiecewiseFunctionProxy does.

### DIFF
--- a/Sources/Proxy/Core/LookupTableProxy/index.js
+++ b/Sources/Proxy/Core/LookupTableProxy/index.js
@@ -157,4 +157,4 @@ export const newInstance = macro.newInstance(extend, 'vtkLookupTableProxy');
 
 // ----------------------------------------------------------------------------
 
-export default { newInstance, extend };
+export default { newInstance, extend, Mode, Defaults };


jourdain:
Your commit message is not valid. Please check the documentation for more details.

brianhelba:
@jourdain Fixed.

brianhelba:
@jourdain Commitizen does not generate log messages of this form (with a module name in parentheses). How should these commit messages be created?

jourdain:
That's the way it was doing it so far. Don't know what is going on.

jourdain:
That's how such changelog get generated

https://github.com/Kitware/vtk-js/releases/tag/v7.10.0

brianhelba:
@jourdain I found the problem here. I was using the wrong version of Commitizen before. I've since installed the correct one, and now the scope option is being offered to me.

I'll ensure that all my future commits will include the correct scope, but since this is now already merged, is there anything else I should do to ensure this makes it into the release notes?

jourdain:
Looking at the changelog that was generated, it is not bad.
I was just worried since it was different than what I was used to. 
It is just a matter of making sure everyone updates and that we end-up with some consistency.

I guess we can use your BREAKING changes to bundle all the new build tools as well. That way, vtk.js 8+ will use the new changelog syntax.

PR #904, opened 2018-10-04
fix: 🐛 Fix a bug in vtkProxyManager.render

brianhelba:


jourdain:
Let see how that new syntax affect the changelog

PR #909, opened 2018-10-05
Update dev tools

jourdain:


jourdain:
@brianhelba this needs to go first

PR #911, opened 2018-10-09
fix(GeometryViewer/index.js, Mapper/index.js, XMLReader/index.js): Ge…

jlee1549:
GeometryViewer - scalar ranges applied per component when relevant
Mapper - allow specific component of multicomponent cell data to be visualized 
XMLReader - es5 compatibility


Sources/Rendering/Core/Mapper/index.js
```diff
@@ -229,7 +229,7 @@ function vtkMapper(publicAPI, model) {
     if (lut) {
       // Ensure that the lookup table is built
       lut.build();
-      model.colorMapColors = lut.mapScalars(scalars, model.colorMode, 0);
+      model.colorMapColors = lut.mapScalars(scalars, model.colorMode, -1);
```
jourdain: @martinken do you agree with that change in vtkMapper?

Sources/IO/XML/XMLReader/index.js
```diff
@@ -285,10 +285,13 @@ function processFieldData(
       }
     });
 
-    const arrays = fieldElem.children;
-    const nbArrays = arrays.length;
+    const dataArrayElems = fieldElem.getElementsByTagName('DataArray');
+
+    // const arrays = fieldElem.children;
+    console.log('arrays = ', dataArrayElems, ' ', fieldElem);
```
jourdain: We may want to remove the console.log

Sources/Rendering/Core/Mapper/index.js
jourdain:
```
  // Description:
  // Internal methods that map a data array into a 4-component,
  // unsigned char RGBA array. The color mode determines the behavior
  // of mapping. If ColorMode.DEFAULT is set, then unsigned char
  // data arrays are treated as colors (and converted to RGBA if
  // necessary); If ColorMode.DIRECT_SCALARS is set, then all arrays
  // are treated as colors (integer types are clamped in the range 0-255,
  // floating point arrays are clamped in the range 0.0-1.0. Note 'char' does
  // not have enough values to represent a color so mapping this type is
  // considered an error);
  // otherwise, the data is mapped through this instance
  // of ScalarsToColors. The component argument is used for data
  // arrays with more than one component; it indicates which component
  // to use to do the blending.  When the component argument is -1,
  // then the this object uses its own selected technique to change a
  // vector into a scalar to map.
  publicAPI.mapScalars = (scalars, colorMode, componentIn) => {
```

jourdain:
You can not change package.json and lock.
Feel free to force push a new history.

jlee1549:
I'm trying to figure out how to remove that file from my commit...

jourdain:
Thanks and sorry to make it difficult to push changes...

jourdain:
Since you do have a nice split between commits, I can rework the history for you if you want?

jourdain:
I'll rework the history once I'm done with what I'm working on. Should be within a day or two...

jourdain:
See https://github.com/Kitware/vtk-js/pull/912

PR #915, opened 2018-10-12
fix(LegacyAsciiParser): Support VTK files with Windows EOL style

finetjul:


Sources/IO/Legacy/LegacyAsciiParser/index.js
```diff
@@ -257,7 +257,10 @@ function getParser(line, dataModel) {
 
 function parseLegacyASCII(content, dataModel = {}) {
   let parser = null;
-  content.split('\n').forEach((line, index) => {
+  const separatorRegExp = /\r?\n/;
+  const separatorRes = separatorRegExp.exec(content);
+  const separator = separatorRes !== null ? separatorRes[0] : null;
+  content.split(separator).forEach((line, index) => {
```
aronhelser: You can actually just split using the regular expression - this might be better, as it would handle mixed line-ending files.

aronhelser: i.e. content.split(separatorRegExp).forEach ....

finetjul: I don't think it is possible to have mixed line-ending for .vtk files. Or said differently, I can't see .VTK file writer doing mixed EOL

Therefore I was prefering my solution that should be much faster :-)

Let me know if you still want me to change to support mixed line-endings.

Thanks!

aronhelser:
OK, fine with me then!

PR #919, opened 2018-10-17
fix(Texture): Work around a Chrome canvas-webgl bug

manthey:
See https://bugs.chromium.org/p/chromium/issues/detail?id=896307 for details on this bug and a minimal reproducing example.


martinken:
LGTM +1

jourdain:
LGTM except the commit message that needs to be properly formatted otherwise that fix will skip the change-log and a release won't be made.

The following commit message can do the work:
`fix(Texture): Work around a Chrome canvas-webgl bug`

jourdain:
This needs to be in the commit, not the title of the PR.

manthey:
I changed the commit message.

PR #926, opened 2018-10-24
Allow for non-integer steps in MouseRangeManipulator

### LINKED ISSUE #808: Incorrect usage of an object as a key
Offending line: https://github.com/Kitware/vtk-js/blob/master/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js#L27

Line reproduced here:
```javascript
 let value = oldValue + delta + model.incrementalDelta[listener];
```

The issue is that `listener` is an object, so when used as a key for incrementalDelta it's cast to the string "[object Object]". This means that multiple listeners will all have the same incrementalDelta.

This hasn't been an issue yet, but this should be fixed.

@agirault

floryst:
Modifieds View2DProxy and MouseRangeManipulator to support non-integer steps when performing window width/level operations, i.e. in the range [0,1].

Seems to fix https://github.com/Kitware/paraview-glance/issues/217, and fixes #808 as well.

### DIFF
--- a/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
@@ -17,7 +17,7 @@ function vtkMouseRangeManipulator(publicAPI, model) {
   //-------------------------------------------------------------------------
   function scaleDeltaToRange(listener, normalizedDelta) {
     return (
-      normalizedDelta * ((listener.max - listener.min) / listener.step + 1)
+      normalizedDelta * ((listener.max - listener.min) / (listener.step + 1))
     );
   }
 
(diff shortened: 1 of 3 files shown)


Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -11,23 +11,23 @@ function vtkMouseRangeManipulator(publicAPI, model) {
 
   // Keep track of delta that is below the value
   // of one step to progressively increment it
-  model.incrementalDelta = {};
+  model.incrementalDelta = new Map();
```
jourdain: Can we make the Map as private so it is not stored on the model? That way that won't break the state serialization process.

floryst:
I'll poke around to see if things that use MouseRangeManipulator are broken, and then merge if okay.

floryst:
Update: Since MouseRangeManipulator now does not return integral values, code that expects integral values from this manipulator should be updated to Math.round the incoming values.

I've updated the MouseRangeManipulator example to reflect this.

Issue #930, opened 2018-10-26
Upgrade build tools


jourdain:
- [x] kw-web-suite with babel 7
- [x] better polyfill and IE11 support
- [x] use `*.module.css` instead of `*.mcss`
- [x] make sure examples works for IE 11
- [x] rework webpack config
- [x] improve build process (dashboard...)
- [x] review bundle size

CC: @floryst

jourdain:
see #931

PR #931, opened 2018-10-26
Update tool chain

jourdain:


jourdain:
I think this is ready but wanted to synch up before merging it since it is going to bump vtk.js to 8.0.0.
It is more a FYI, but if you want to say something, please do as I plan to merge it by the end of this week.

See https://github.com/Kitware/vtk-js/issues/930 for some details

CC: @floryst @martinken @luciemac @laurennlam @jjomier @finetjul @brianhelba @thewtex @agirault @aylward

jourdain:
Not yet, I was looking for the VR for Glance. But now, that I'm playing with the Windows machine, I'll try it out...

jourdain:
I did test on IE11 and tested the run of examples on Windows

PR #933, opened 2018-10-28
Update babel loader presets

jackyko1991:
update vtkjs dependency to latest kw web suite (babel 7 compatible) 

Associated update may found [here](https://github.com/Kitware/kw-web-suite/commit/43c7a116524c6df3eb2b9009256a3ba4098c1f34#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)


jourdain:
Thanks! I'll bundle it with the PR that will bring the new kw-web-suite...

jourdain:
Thanks! I'll bundle it with the PR that will bring the new kw-web-suite...

jourdain:
See #931

PR #948, opened 2018-11-08
feat(ImageStream): Add ViewStream inheritance support

finetjul:
Make it possible to create your own ViewStream and register it
in the ImageStream:
```
function createMyViewStream(imageStream, viewId = '-1', size = [400, 400]){
  const { setServerAnimationFPS,
          getServerAnimationFPS,
          unregisterViewStream } = imageStream;
  const viewStream = MyViewStream.newInstance({
    protocol: imageStream.getProtocol(),
    unregisterViewStream,
    sharedAPI: {
      setServerAnimationFPS,
      getServerAnimationFPS,
    },
  });
  viewStream.setViewId(viewId);
  viewStream.setSize(size[0], size[1]);

  return viewStream;
};

const viewStream = createMyViewStream(imageStream, viewId);
view.imageStream.registerViewStream(viewStream);
```

### DIFF
--- a/Sources/IO/Core/ImageStream/index.js
+++ b/Sources/IO/Core/ImageStream/index.js
@@ -74,14 +74,24 @@ function vtkImageStream(publicAPI, model) {
 
   // --------------------------------------------------------------------------
 
-  function unregisterViewStream(view) {
+  publicAPI.registerViewStream = (view) => {
+    model.viewStreams.push(view);
+  };
+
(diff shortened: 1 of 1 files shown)


jourdain:
Comment out the `protocol: null` for the default value. It is good to know, that field exist, but we don't init then to null.

finetjul:
Thanks @jourdain, it's good to know. 
I just pushed a revised version.

PR #957, opened 2018-11-26
fix(TubeFilter): Register filter to parent index

dspring2:
fix(TubeFilter): Register filter to parent index


jourdain:
Perfect but can you rename your commit message as follow so a proper change-log and a new version can be auto-generated?

```
fix(TubeFilter): Register filter to parent index
```  

Thanks...

PR #959, opened 2018-11-27
Feature tube filter coloring

dspring2:


jourdain:
1) [cd4bbbd](https://github.com/Kitware/vtk-js/pull/959/commits/cd4bbbd3538de0c9a6c6de48f70505d475a862c4) is more a big hack than anything else as this is not the way vtk aim to work. Moreover, why port 3 instead of 1? Also I don't see how that array get properly filled...

2) [2c0b736](https://github.com/Kitware/vtk-js/pull/959/commits/2c0b736498560776ee9b11531a8c45e4d1568221) Adding the test is awesome, thanks!

3) [adda6ca](https://github.com/Kitware/vtk-js/pull/959/commits/adda6ca9d35f643c2e2b73cd99519f04bda784d2) I don't agree with it. Maybe we can flex the usage of addInput but the check should remain. But I'm not sure what was the goal behind that change.

 4) Any commit that you make need to be specially formatted so that our CI could generate the appropriate release tag and change log. To learn more on that topic you can look [here](https://gist.github.com/stephenparish/9941e89d80e2bc58a153) or just follow the [step 5](https://kitware.github.io/vtk-js/docs/misc_contributing.html).

Thanks for taking the time to improve vtk.js...

dspring2:
https://github.com/Kitware/vtk-js/pull/959/commits/adda6ca9d35f643c2e2b73cd99519f04bda784d2 I was basing use of port 3 off of this example 
https://www.vtk.org/Wiki/VTK/Examples/Python/Visualization/ClampGlyphSizes 
It seems like port 0 was used correctly for scalars (line 694), and port 1 was used correctly for vectors (line 715), however nothing had been implemented for ports 2 and 3. In the above example, port 3 was being called out for colors, but I could have misjudged what the example was motivating. 

https://github.com/Kitware/vtk-js/pull/959/commits/adda6ca9d35f643c2e2b73cd99519f04bda784d2 The intention behind this is based on my understanding above. Without this change, the user was required to setInputArrayToProcess for all 4 ports, even if they only wanted to use ports 0 and 3, for scalars and colors. Again, this was based on my interpretation of the example above, which may not have been correct.

jourdain:
I see, but in that specific use case, the Glyph filter only have 2 input ports (pointset + glyph) while they are using the SetInputArrayToProcess with different port number to capture several arrays for several goals (orientation, scale, color...) while the filter only have truly 2 inputs. They are using that existing API to capture which array from input 0 they should use for what. 

To use the same technique (assuming we want), we can fix the get/lookup to fallback to port 0 when there is no corresponding input.

Anyhow, even with that change, I don't see where the array get properly filled with the values for the expended set of colors. It is true that we get the right size, but I don't think the values get properly created.

PR #964, opened 2018-12-03
Modifiers for scroll interaction

llangholz89:


jourdain:
Glancing through the change, it seems fine. I'm guessing you tested it?

llangholz89:
The commit messages seemed to have gotten cut off, maybe there's a character limit that I didn't see.  Let me know if you'd like me to fix

llangholz89:
Yes, tested it to the extent of the example.  Seems to work on my browsers.  We can wait to merge if you think some more research should be done to ensure those modifiers will always be present in all browsers

jourdain:
Yes the commit message should be rewritten like:

```
fix(InteractorStyleManipulator): Allow modifier to filter out scroll
doc(InteractorStyleManipulator): Update example to use scroll modifiers
```

I think we should be able to merge once the commit message are updated and travis get back with success.

Also could you run `npm t` to be sure as well.

Thanks!

jourdain:
?

llangholz89:
Ha, fatal issues on this end with resetting the commit history.  Still working on it...

PR #965, opened 2018-12-03
Modifiers for scroll interaction - take 2

llangholz89:
Re-worked from closed pr #964


llangholz89:
fyi @jourdain I made a mistake when changing the names of the old commits, and in attempting to fix it I just ran into more and more trouble.  So I decided to scrap it and redo.

The tests pass, however I did notice when playing with the example some more, when you switch quickly between scroll and click events you can get some warnings in the console about a mismatched requestor for the animation

Issue #970, opened 2018-12-05
SceneLight type doesn't work with vtkLight


laurennlam:
I tried a simple example to understand how the vtkLight works. And it seems that there are some issues, especially with the SceneLight type.
The default value of the light type is : 'SceneLight', but with this example, it seems that it's more a CameraLight than a SceneLight :

`this.renderer.removeAllLights();
		const position = [2, 2, 0];
		const focalPoint = [0, 0, 0];

		const light = vtkLight.newInstance();
		light.setPosition(...position);
		light.setColor(1, 0, 0);
		light.setFocalPoint(...focalPoint);
		this.renderer.addLight(light);

		const sphereSource = vtkSphereSource.newInstance();
		sphereSource.setCenter(...position);
		sphereSource.setRadius(0.05);
		const mapperSphere = vtkMapper.newInstance();
		mapperSphere.setInputConnection(sphereSource.getOutputPort());
		const actorSphere = vtkActor.newInstance();
		actorSphere.setMapper(mapperSphere);

		const lineSource = vtkLineSource.newInstance();
		lineSource.setPoint1(...position);
		lineSource.setPoint2(...focalPoint);
		const mapperLine = vtkMapper.newInstance();
		mapperLine.setInputConnection(lineSource.getOutputPort());
		const actorLine = vtkActor.newInstance();
		actorLine.setMapper(mapperLine);

		this.renderer.addActor(actorSphere);
		this.renderer.addActor(actorLine);
		this.updateRendering();`

The second point is that, when we are on 'HeadLight' type, then the color is not displayed anymore. Only white color is considered.
@mcmigliore took a quick look but maybe @martinken or @jourdain have some ideas on what's happening here.
Just tag @finetjul to keep him update.

jourdain:
The various light types are not yet implemented...

laurennlam:
There's an error threw when we explicitly define light type as CameraLight. But it seems that the default behavior looks like a Camera light instead of a scene light.

mcmigliore:
Yep, the light direction in the shader in expressed in View Coordinate.

laurennlam:
@jourdain , @mcmigliore found a way to use the three different types. I'll try to make a clean PR of it.

laurennlam:
#971

PR #971, opened 2018-12-06
fix(light): Restore Scene and camera light type

laurennlam:
#970


mcmigliore:
Looking good to me

jourdain:
@martinken it is up to you...

aronhelser:
Seems like this is directional lights only? No positional lights?

laurennlam:
positional lights seems to work too.

PR #972, opened 2018-12-07
Cursor 3d bugs fixed and example.

wuzhuobin:
- Fixing some known bugs for vtkCursor3D
- As request in #961, an example was created. 
The following picture is for reference. 
example


jourdain:
Looks good but feat() bump the middle number in the version. In your case, examples should be taken as documentation which does not affect the released code. Could you rename `feat(Cursor3D example): Adding an example for Cursor3D.` to `doc(Cursor3D): Adding an example for Cursor3D`

wuzhuobin:
> Looks good but feat() bump the middle number in the version. In your case, examples should be taken as documentation which does not affect the released code. Could you rename `feat(Cursor3D example): Adding an example for Cursor3D.` to `doc(Cursor3D): Adding an example for Cursor3D`

@jourdain  The commit has been changed as required.

PR #974, opened 2018-12-09
fix(Mapper): Fix vtkMapper.getResolveCoincidentTopology()

### LINKED ISSUE #973: Odd behavior with resolveCoincidentTopology
According to documentation, vtkMapper.resolveCoincidentTopology is static.
However, it is possible to call setResolveCoincidentTopology() and getResolveCoincidentTopology() on a non static mapper (e.g. the mapper of an actor).

Problem is that `getResolveCoincidentTopology()` does not return the variable set by `setResolveCoincidentTopology()`:
  In one case, `setResolveCoincidentTopology()` sets the static `resolveCoincidentTopology` variable, on the other hand `getResolveCoincidentTopology()` returns the variable of the vtkMapper model.

Could it be because `getResolveCoincidentTopology()` is not default exported in Mapper/Static.js ?

This becomes an issue in PolyDataMapper because `model.renderable.getResolveCoincidentTopology()` always returns false.

finetjul:
fix #973

### DIFF
--- a/Sources/Rendering/Core/Mapper/Static.js
+++ b/Sources/Rendering/Core/Mapper/Static.js
@@ -41,6 +41,7 @@ export function getResolveCoincidentTopologyAsString() {
 export default {
   getResolveCoincidentTopologyAsString,
   getResolveCoincidentTopologyPolygonOffsetFaces,
+  getResolveCoincidentTopology,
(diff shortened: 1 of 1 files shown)


finetjul:
@martinken @jourdain

martinken:
+1

PR #995, opened 2019-01-30
fix(ProxyManager): Use setContainer() to gain ability to check if unbind is needed

matthewma7:
Check if view proxy has container before try to unbind when deleting view proxy.
This is a bug fix for https://github.com/Kitware/vtk-js/pull/994

### DIFF
--- a/Sources/Proxy/Core/ProxyManager/core.js
+++ b/Sources/Proxy/Core/ProxyManager/core.js
@@ -213,7 +213,7 @@ export default function addRegistrationAPI(publicAPI, model) {
       proxy.getRepresentations().forEach((repProxy) => {
         publicAPI.deleteProxy(repProxy);
       });
-      proxy.getInteractor().unbindEvents();
+      proxy.setContainer(null);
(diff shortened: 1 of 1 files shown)


jourdain:
The interactor should not break if such method get called. Fix the issue on the interactor instead.

jourdain:
Or instead of doing `proxy.getInteractor().unbindEvents();` call that instead `proxy.setContainer(null);`

matthewma7:
I used unbindEvents() when I didn't like the behavior that ViewProxy.setContainer(null) actually remove stuff from the DOM, but I think use setContainer(null) makes more sense because it is setContainer calling the bind() and unbind() in this case. I'll change it to use setContainer(null)
Thanks for your quick feedback. @jourdain

jourdain:
Your branch is failing that's why we can't merge it.

matthewma7:
It seems was a transient error. I tried to restart it but I think I don't have permission to, so I pushed again to trigger another one. Should be good now. 🙂

PR #1009, opened 2019-02-24
Fixing Side-Effect CanvasVR chrome issue

aneeshikmat:
On chrome, canvas not updated to be with Vr view, but in firefox its work prefect, so that I suggest to re-request animation to update scene if chrome and its work perfect dependence of test.

Note: Tested on Latest Version from Chrome & Firefox for 64 bit.


aneeshikmat:
352:1   error  Delete `········`                                                                   prettier/prettier
354:13  error  Replace `(model.isChrome)` with `·(model.isChrome)·`                                prettier/prettier
387:7   error  Replace `(!model.vrDisplay.isPresenting)` with `·(!model.vrDisplay.isPresenting)·`

I don't understand these error! LOL!

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -376,6 +382,10 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
   };
 
   publicAPI.vrRender = () => {
+    // If no presneting for any resone we most stop submiting frame
```
jourdain:
```
// If not presenting for any reason, we do not submit frame
```

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -349,6 +349,12 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
           model.vrSceneFrame = model.vrDisplay.requestAnimationFrame(
             publicAPI.vrRender
           );
+          // If Broswer is chrome we need to request animation again to canvas update
+          if (model.isChrome) {
```
jourdain: Let's use a global variable in that module (at the top of the file).

```
const IS_CHROME = navigator.userAgent.indexOf('Chrome') !== -1;
```

jourdain:
Can you amend all your commits and format them to follow what semantic-release expect so the proper ChangeLog and version number could be generated from your contribution?

Thanks!

Sources/Rendering/OpenGL/RenderWindow/index.js
aneeshikmat: Resolved, Thanks!

Sources/Rendering/OpenGL/RenderWindow/index.js
aneeshikmat: Resolved, Thanks!

aneeshikmat:
> Can you amend all your commits and format them to follow what semantic-release expect so the proper ChangeLog and version number could be generated from your contribution?
> 
> Thanks!

Sure, Thank you!

jourdain:
The content is great, but we need only one commit (not 6) with the proper formatting for semantic release.

aneeshikmat:
I think I do it wrong, I will create a new branch for that! LOL

aneeshikmat:
#1013

PR #1010, opened 2019-02-24
fix(OpenGL/RenderWindow): Stop adding bgImage to DOM

### LINKED ISSUE #1007: OpenGL RenderWindow: Make background image optional?
Hello!

I've been using the GenericRenderWindow while integrating vtk-js for image display in the OHIF Viewer. It seems to be working fine, but I noticed it is creating an `` element which spans the whole page and ends up interfering with our drag/drop implementation.

It looks like it's being created here:

https://github.com/Kitware/vtk-js/blob/5ceb2af2bf9e3b98c241d6cfc33497a32f524cd7/Sources/Rendering/OpenGL/RenderWindow/index.js#L1079

and it seems like it's meant to display an optional background image. (or is there some other use for it?)

Would you accept a PR to make the addition of this HTML element to the DOM optional? i.e. if `setBackgroundImage` is never called, bgImage is never created.

Thanks!

swederik:
Stop adding bgImage to DOM unless setBackgroundImage has been called.

Background images are necessary for remote rendering, but the `` tag is currently being added to
the DOM unnecessarily, even if the user is not using remote rendering. This change prevents this
from occurring.

fix #1007 

This fix seems to be working for me.

### DIFF
--- a/Sources/Rendering/OpenGL/RenderWindow/index.js
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.js
@@ -135,7 +135,7 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
 
       // If the renderer is set to use a background
       // image, attach it to the DOM.
-      if (model.useBackgroundImage === true) {
+      if (model.useBackgroundImage) {
         model.el.appendChild(model.bgImage);
       }
 
@@ -518,15 +518,9 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
 
     // Add or remove the background image from the
     // DOM as specified.
-    if (
-      model.useBackgroundImage === true &&
-      model.el.contains(model.bgImage) === false
-    ) {
+    if (model.useBackgroundImage && !model.el.contains(model.bgImage)) {
       model.el.appendChild(model.bgImage);
-    } else if (
-      model.useBackgroundImage === false &&
-      model.el.contains(model.bgImage) === true
-    ) {
+    } else if (!model.useBackgroundImage && model.el.contains(model.bgImage)) {
       model.el.removeChild(model.bgImage);
     }
   };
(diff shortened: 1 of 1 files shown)


jourdain:
So that means that if we add the image after binding the container to the RW, that image will never be added to the DOM.

I would be fine to have a specific flag `useBackgroundImage` that can be used to determine if the window should add that element or not. And its default can be false.

swederik:
Ah whoops, I didn't notice that. Let me fix that.

swederik:
Okay so here's the logic now:

`setContainer`:
- If bgImage exists in the previous container, remove it.
- If `useBackgroundImage` is true, add it to the new container

New publicAPI method: `setUseBackgroundImage`:
- Sets the value of `useBackgroundImage`.
- If it is true, and the bgImage does not already exist in the container, it is added.
- If it is false, and the bgImage exists in the container, it is removed.

So now you should be able to call `setUseBackgroundImage`, `setContainer`, and `setBackgroundImage` in any order.

An alternative to this would be to move the logic that is inside `setUseBackgroundImage` to `setBackgroundImage`. I figured that since the `setBackgroundImage` function is used for streaming in remote-rendered images, it would be called frequently. Adding an if statement to set the `useBackgroundImage` flag and conditionally add the DOM element that would not be important apart from the first time `setBackgroundImage` is called seemed wasteful. 

Let me know if you would prefer to ditch `setUseBackgroundImage` and move this inside `setBackgroundImage` instead.

Should I also add `useBackgroundImage` to macro.setGet call? Would that overwrite the definition of `setUseBackgroundImage`?

jourdain:
That's perfect. Maybe just adding the get method via the macro.

Also I would rather not do ```=== true / false``` but just use the actual variable with or without ```!``` in front of it.

swederik:
Wow I somehow didn't notice that `macro.get` existed. Added `useBackgroundImage` to `macro.get` and changed the if statements as suggested. PR should be good now.

Thanks for reviewing so quickly!

jourdain:
You are welcome! Thanks for doing the work and contributing it back...

PR #1013, opened 2019-02-26
Fixing Side-Effect CanvasInVR chrome issue

aneeshikmat:
On chrome, canvas not updated to be with Vr view, but in firefox its work prefect, so that I suggest to re-request animation to update scene if chrome and its work perfect dependence of test.

Note: Tested on Latest Version from Chrome & Firefox for 64 bit.


jourdain:
Going to merge it, but because the format of the commit is not following semantic-release, your contribution won't appear in the change-log.

aneeshikmat:
Thank you!

Issue #1011, opened 2019-02-26
Unable to change color of object with actor.getProperty().setColor


longhuy322000:
I am unable to change the color of the object with actor.getProperty().setColor. It only works with pure red, blue or green, but not with the combination of red, green and blue. I tried to set scalarVisibility in mapper to False as well, but it still didn't work. Could you please tell me what I did wrong?

jourdain:
I'm guessing that you figure it out?

longhuyn:
Yeah I got it. Thank you! I have two questions, but not related to this problem though.
1. Does VTK js have an API that draws a string in the render window?
2. Does VTK support time? For example, I want to move an object from one coordinate to another coordinate respects to time. Does VTK have any mechanism to do this?
Thank you again!

jourdain:
1. Not yet but you can do that in plain HTML easily.
2. Not really but time can be handled at the application level.

longhuyn:
1. I agree. But if I move the object around, the label of the object doesn't move with the object at the same time

jourdain:
We have a mapper that will give you callback with (x,y) of a given point in 3D. 

https://kitware.github.io/vtk-js/examples/SpheresAndLabels.html

Something is broken with the example, but that's the idea.

SSPollock:
If you figured it out, do share.

a3VonG:
In case it's useful for other people that run into the issue:
 I solved it by setting the [scalarVisibility](url) to false on the mapper.

PR #1015, opened 2019-02-26
fix(Examples): Add interactors to DepthTest and SpheresAndLabels

### LINKED ISSUE #981: Interactor style missing in example
https://kitware.github.io/vtk-js/examples/DepthTest.html
https://kitware.github.io/vtk-js/examples/SpheresAndLabels.html

floryst:
Adds interactor styles to two examples that didn't have them.

Fixes #981. Pinging #1011 because it references this issue.


jourdain:
They could have been just `doc(Examples): ...` as it does not affect the code of vtk.js and therefore does not require a new release.

floryst:
Oh yeah, I realized that I should have switched to using `doc()` on my way home, but forgot to change the messages...

PR #1019, opened 2019-03-01
CI testing

thewtex:


thewtex:
We can merge with more confidence now and iterate faster :-)

Will merge and enable, some cleanup may be required...

PR #1021, opened 2019-03-03
Azure configuration tweaks

thewtex:
The `build` target starts a development webpack configuration, which
time's out on CI.


thewtex:
:green_heart:

PR #1030, opened 2019-03-12
feat(XMLWriter): Add XML Data Writer

finetjul:
Generates a VTI file from a vtkImageData

### DIFF
--- a/Sources/IO/XML/XMLWriter/index.js
+++ b/Sources/IO/XML/XMLWriter/index.js
@@ -3,7 +3,10 @@ import { fromByteArray } from 'base64-js';
 import pako from 'pako';
 
 import macro from 'vtk.js/Sources/macro';
-import { FormatTypes } from 'vtk.js/Sources/IO/XML/XMLWriter/Constants';
+import {
+  FormatTypes,
+  TYPED_ARRAY,
+} from 'vtk.js/Sources/IO/XML/XMLWriter/Constants';
 
 // ----------------------------------------------------------------------------
 // Global methods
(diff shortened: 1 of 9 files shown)


Sources/IO/XML/XMLWriter/index.js
```diff
@@ -0,0 +1,138 @@
+import builder from 'xmlbuilder';
```
floryst: Best to explicitly add `xmlbuilder` to package.json rather than relying on it being a dependency of xml2js.

Sources/IO/XML/XMLWriter/index.js
```diff
@@ -0,0 +1,138 @@
...
+      const headerUint8 = new Uint8Array(header.buffer);
+      uint8.set(headerUint8, uint8Offset);
+      uint8Offset += headerUint8Size;
+      for (let blockId = 0; blockId < nbBlocks; ++blockId) {
+        uint8.set(blocks[blockId], uint8Offset);
+        uint8Offset += header[3 + blockId];
+      }
+      return fromByteArray(uint8);
+    }
+    return 'Only vtkZLibDataCompressor is supported';
```
floryst: Better to `throw new Error('...message...')` than return your error message.

Same with line 78 and 80.

Sources/IO/XML/XMLWriter/index.js
```diff
@@ -0,0 +1,138 @@
...
+function vtkXMLWriter(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkXMLWriter');
+
+  // Can be overriden in subclass
+  publicAPI.create = (dataObject) =>
+    builder
+      .create('VTKFile')
+      .att('type', model.dataType)
+      .att('version', '0.1')
```
floryst: @jourdain Would it be better to use version 1.0 of the vti/vtp format? If generating version 0.1 is ok moving forward, then this is fine.

PR #1031, opened 2019-03-12
Add vtk windowed sinc poly data filter

finetjul:


Sources/Common/DataModel/CellTypes/index.js
```diff
@@ -0,0 +1,203 @@
+import macro from 'vtk.js/Sources/macro';
+// import vtkDataArray from 'vtk.js/Sources/Common/Core/DataArray';
+// import { VtkDataTypes } from 'vtk.js/Sources/Common/Core/DataArray/Constants';
```
thewtex: commented

PR #1032, opened 2019-03-14
fix(PaintFilter): Record history when using voxelFunc

### LINKED ISSUE #1028: fix undo/redo in PaintWidget example
Can undo but redo has no effect.

floryst:
Correctly record history when using voxelFunc to threshold voxels.

Fixes #1028 

@agirault


floryst:
Note: I plan on merging this when #1033 is finished.

PR #1033, opened 2019-03-14
fix: Add ImageCropFilter and ImageCroppingWidget to exports

### LINKED ISSUE #1029: export vtkImageCropFilter and vtkImageCroppingWidget in default

floryst:
Fixes #1029

### DIFF
--- a/Sources/Filters/General/index.js
+++ b/Sources/Filters/General/index.js
@@ -1,5 +1,6 @@
 import vtkAppendPolyData from './AppendPolyData';
 import vtkCalculator from './Calculator';
+import vtkImageCropFilter from './ImageCropFilter';
 import vtkImageMarchingCubes from './ImageMarchingCubes';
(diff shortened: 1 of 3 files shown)


agirault:
Thanks! Since you're adding more exports, could you maybe export the annotatedCubeActor [presets](https://github.com/Kitware/vtk-js/blob/9063bb4db9da420f5e127d72a9cba60802f6624b/Sources/Rendering/Core/AnnotatedCubeActor/Presets.js) in [here](https://github.com/Kitware/vtk-js/blob/9063bb4db9da420f5e127d72a9cba60802f6624b/Sources/Rendering/Core/AnnotatedCubeActor/index.js#L191)?

agirault:
Also, the first commit message is wrong, you are exporting ImageCropFilter not PaintFilter

floryst:
Oh thanks, good catch. Was working on the paint filter right before this...

PR #1036, opened 2019-03-15
fix(MouseRangeManipulator): Round stepsToDifference

### LINKED ISSUE #1035: MouseRangeManipulator stepping issues
Revert change made here: https://github.com/Kitware/vtk-js/commit/2fca89d1b749f1a429e563dfa0e8aef677679342#diff-c2337aa992718fde42b591ce8f0017d0R30

While some other changes in the commit allowed to set a non-integer step to the ViewProxy, removing the rounding in `const stepsToDifference = Math.round(difference / listener.step)` lead to a non-integer _number_ of steps being applied (we can do 1 step of 0.2, but shouldn't do 0.2 steps of 1).

The problem can be seen in the MouseRangeManipulator example when removing the rounding before setting setSlice, since it will set intermediary slices which isn't valid. The rounding also currently leads to an issue where slow iterative slicing isn't possible since the scrolling doesn't accumulate. Once the source of the issue is fixed, the example can be updated to remove the rounding of the slice value.

floryst:
Reverts part of 2fca89d.

Fixes #1035 

To test:
```
git checkout patch-mouse-range-manipulator
npm run example -- ImageMapper
```

### DIFF
--- a/Sources/Interaction/Manipulators/MouseRangeManipulator/example/index.js
+++ b/Sources/Interaction/Manipulators/MouseRangeManipulator/example/index.js
@@ -53,7 +53,7 @@ const extent = data.getExtent();
 const kMin = extent[4];
 const kMax = extent[5];
 const kGet = mapper.getSlice;
-const kSet = (slice) => mapper.setSlice(Math.round(slice));
+const kSet = mapper.setSlice;
 
 const rangeManipulator = Manipulators.vtkMouseRangeManipulator.newInstance({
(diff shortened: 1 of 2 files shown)


agirault:
@floryst travis says it hasn't completed, but if you click on the details, the build passed.

floryst:
Cool, I was wondering what was going on...I'll merge when I do one last pass at all other PRs I plan on merging.

PR #1040, opened 2019-04-01
fix(paintFilter): Fix setLabelMap when labelMap is null

luciemac:


luciemac:
@floryst can you review please ?

floryst:
LGTM

PR #1046, opened 2019-04-12
fix(WidgetManager): Support turning off picking before setRenderer()

scottwittenburg:
Similar to when we update our subscriptions with the interactor "onEndAnimation" event (a little
higher up in setRenderer()), we should first check if picking is enabled before calling the method
to enable it.


jourdain:
LGTM

PR #1067, opened 2019-04-29
fix(vtkSlider): invokeValueChange only when value changes

llangholz89:


jourdain:
I like the add of the return true/false. But I'm not sure why you removed the first section?

llangholz89:
Both original `if` conditions resulted in the same execution so I combined them and removed the `model.values.indexOf(v) !== -1` condition assuming it would be covered by `model.values[0] <= v &&
      v <= model.values.slice(-1)[0]`.  I can add it back though if that's a bad assumption

llangholz89:
The reason I even looked into this was because I was having issues where my `slider.onValueChange` callback was always invoked even if it was the same value being set (I guess the value was not in the `model.values` array)

jourdain:
You are right, if the `else if` was valid before, then your change totally make sense.

Thx for your contrib

PR #1073, opened 2019-05-07
perf(vtkOpenGLVolumeMapper): Remove mat allocs from render loop

llangholz89:
Per discussion in #1071 with @martinken


martinken:
LGTM +1

PR #1084, opened 2019-05-22
fix(ForwardPass): Render layered renderers in correct order

zmanvortex:
Modified ForwardPass to render layered renderers in correct order.

Rendering order was always based on the order that renderers were added to the renderwindow regardless of the layer number setting of the renderer. Modified ForwardPass to render by layer number order.

### DIFF
--- a/Sources/Rendering/OpenGL/ForwardPass/index.js
+++ b/Sources/Rendering/OpenGL/ForwardPass/index.js
@@ -24,55 +24,65 @@ function vtkForwardPass(publicAPI, model) {
     publicAPI.setCurrentOperation('buildPass');
     viewNode.traverse(publicAPI);
 
+    const numlayers = viewNode.getRenderable().getNumberOfLayers();
+
     // iterate over renderers
     const renderers = viewNode.getChildren();
-    for (let index = 0; index < renderers.length; index++) {
-      const renNode = renderers[index];
+    for (let i = 0; i < numlayers; i++) {
+      for (let index = 0; index < renderers.length; index++) {
+        const renNode = renderers[index];
+        const ren = viewNode.getRenderable().getRenderers()[index];
+        if (ren.getLayer() === i) {
+          // check for both opaque and volume actors
+          model.opaqueActorCount = 0;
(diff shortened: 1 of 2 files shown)


jourdain:
Could you rename your commit as follow so the generated change log will be easier to follow.

`fix(ForwardPass): Render layered renderers in correct order`

Also the tests seems to fail. Did you run them locally by calling `npm run test`?

jourdain:
There is 2 issue to fix in the code

```
/home/vsts/work/1/s/Sources/Rendering/OpenGL/ForwardPass/index.js
35:28  error  Expected '===' and instead saw '=='
57:17  error  Replace `fbSize·===·null·||·fbSize[0]·!==·size[0]·||·fbSize[1]·!==·size[1]` with `⏎··············fbSize·===·null·||⏎··············fbSize[0]·!==·size[0]·||⏎··············fbSize[1]·!==·size[1]⏎············`  prettier/prettier
   
```

You can fix one of them by calling `npm run reformat` then for the other one you need to replace your `==` by `===`.

zmanvortex:
When running 

> npm run test

It just hangs on 

> Chrome 74.0.3729 (Windows 10.0.0) LOG: 'ok 642 vtkOpenGLVolumeMapper Lighting'.
I had to set the following in karma.conf:

> captureTimeout: 210000,
> browserDisconnectTolerance: 3, 
> browserNoActivityTimeout: 600000,
> browserDisconnectTimeout: 600000,
otherwise it would error on timeout after 30000ms. Any ideas?
I have corrected the errors that you mentioned and am not sure if I should issue another pull request. I am confused whether it will create a duplicate pull request from this one?

jourdain:
Just force push to the current one (branch). That will retrigger the tests and we should be fine. You need to amend your commit with the correct text.

jourdain:
Can you squash those commits so only one appear?

jourdain:
Otherwise we will have 2 change log entry.

zmanvortex:
I had to push another branch after amending the commit name to the correct format.

jourdain:
thx!

PR #1085, opened 2019-05-22
fix(Sources/Interaction/Widgets/LabelRepresentation/index.js): Lines …

RemiCecchinato:
Lines are now properly splitted.


finetjul:
fix(LabelWidget) should be enough.

jourdain:
The commit message should be updated to follow what @finetjul suggest in order to get the proper changelog generation.

Thanks for your contribution.

martinken:
@jourdain look good? If so merge?

PR #1087, opened 2019-05-23
fix(Rendering): set default rendering layer to 0

martinken:
The prior default of 1 caused nothing to be drawn


martinken:
@jourdain

jourdain:
LGTM

PR #1089, opened 2019-05-27
WIP: Add Blend Modes to VolumeMapper

swederik:
Hello!

We would like to display volumes with maximum intensity projection and would like to help contribute this back to vtk-js. I know that ImageReslice has a SlabMode which can be used to do this in the CPU, but we would like to do it in the shader inside the VolumeMapper, if possible.

I took a stab at adding it to the VolumeMapper, and I was wondering if you could give me some feedback about whether or not this is the right approach before I went any further. My idea is to add [SetBlendModeToMaximumIntensity](https://vtk.org/doc/nightly/html/classvtkVolumeMapper.html#a8397e8507a80e81bfbf733f130fefde3) which is present in the C++ version.

All I've added now is a blend mode condition, which defaults to Composite, and a pathway in the WebGL2 shader to compute the maximum value along the ray.

Does this look like the right approach? Would you accept a PR to add additional blend modes to the mapper?

As an aside, it would be helpful if you could provide the LIDC2.vti file in .vtk format. Maybe I missed something, but I couldn't find an easy way to read the file. I know that in Utilities you provide a way to go from .vtk to the json .vti format, but I didn't see a way to go back, so for my testing I had to create a new file.

Here's a visualization of how it looks right now. On the left is the Python version, on the right is the VTK.js version.

VTK-MIP

The example can be run with
```
npm run example -- VolumeMapperBlendModes
```

but you will need to download the data linked at the bottom of the post, unzip it, and put it in the Data folder.

### Known issues:
- Visualization looks strange, there is black speckling when compared to the Python version
- Need to implement the same pathway in WebGL 1
- Need to add the set/get macro for the API to VolumeMapper
- Should decide which dataset we want to use for an Example, or if this should just be rolled into the normal VolumeMapper example. Seems unnecessary to include yet another large dataset for just this functionality, but I had no choice for now.

#### Etc:
- Fixed a couple of typos ('bounnds' and 'colortextureString') in the volumeMapper. I don't think they really had any impact though.

Python code I've been using for comparison:
```python
#!/usr/bin/env python
from vtk import *

filename = "QIN-HEADNECK-01-0024-PET-WB-0.vtk"

reader = vtkGenericDataObjectReader()
reader.SetFileName(filename)
reader.Update()
imageData = reader.GetOutput()

# Create transfer mapping scalar value to opacity.
opacityFunction = vtkPiecewiseFunction()
opacityFunction.AddPoint(0, 0.0)
opacityFunction.AddPoint(5000, 0.3)
opacityFunction.AddPoint(30000, 1.0)

volumeProperty = vtkVolumeProperty()
volumeProperty.SetScalarOpacity(opacityFunction)
volumeProperty.ShadeOn()
volumeProperty.SetInterpolationTypeToLinear()
volumeProperty.SetScalarOpacityUnitDistance(3)
volumeProperty.SetAmbient(0.2);
volumeProperty.SetDiffuse(0.7);
volumeProperty.SetSpecular(0.3);
volumeProperty.SetSpecularPower(8.0);

volumeMapper = vtkGPUVolumeRayCastMapper()
volumeMapper.SetBlendModeToMaximumIntensity()
volumeMapper.SetInputData(imageData)
volumeMapper.SetSampleDistance(1.1)

volume = vtkVolume()
volume.SetMapper(volumeMapper)
volume.SetProperty(volumeProperty)

# create a rendering window and renderer
renderer = vtkRenderer()
renderer.SetBackground(0.1,0.1,0.1)

window = vtkRenderWindow()
window.SetSize(1024,1024)
window.AddRenderer(renderer)

interactor = vtkRenderWindowInteractor()
interactor.SetRenderWindow(window)

style = vtkInteractorStyleTrackballCamera();
interactor.SetInteractorStyle(style);

renderer.AddVolume(volume)

renderer.ResetCameraClippingRange()
renderer.ResetCamera()

 # Start
interactor.Initialize()
window.Render()
interactor.Start()
```

**Here's the VTK file:**
*.vtk format:* https://www.dropbox.com/s/pppvgdi9wlcogoz/QIN-HEADNECK-01-0024-PET-WB-0.vtk?dl=0

*zipped .vti format for vtk-js:* https://www.dropbox.com/s/tr6kfambzcj1jcu/QIN-HEADNECK-01-0024-PET-WB-0.vti.zip?dl=0

### DIFF
--- a/Examples/Volume/VolumeMapperBlendModes/index.js
+++ b/Examples/Volume/VolumeMapperBlendModes/index.js
@@ -1,0 +1,73 @@
+import 'vtk.js/Sources/favicon';
+
+import vtkFullScreenRenderWindow from 'vtk.js/Sources/Rendering/Misc/FullScreenRenderWindow';
+import vtkHttpDataSetReader from 'vtk.js/Sources/IO/Core/HttpDataSetReader';
+import vtkPiecewiseFunction from 'vtk.js/Sources/Common/DataModel/PiecewiseFunction';
+import vtkVolume from 'vtk.js/Sources/Rendering/Core/Volume';
+import vtkVolumeMapper from 'vtk.js/Sources/Rendering/Core/VolumeMapper';
+
+// ----------------------------------------------------------------------------
+// Standard rendering code setup
+// ----------------------------------------------------------------------------
+
+const fullScreenRenderer = vtkFullScreenRenderWindow.newInstance({
+  background: [0.1, 0.1, 0.1],
+});
+const renderer = fullScreenRenderer.getRenderer();
+const renderWindow = fullScreenRenderer.getRenderWindow();
+
+// ----------------------------------------------------------------------------
+// Example code
+// ----------------------------------------------------------------------------
+// Server is not sending the .gz and whith the compress header
+// Need to fetch the true file name and uncompress it locally
+// ----------------------------------------------------------------------------
+
+const reader = vtkHttpDataSetReader.newInstance({ fetchGzip: true });
+
+const actor = vtkVolume.newInstance();
+const mapper = vtkVolumeMapper.newInstance();
+mapper.setSampleDistance(1.1);
+actor.setMapper(mapper);
+
+// create color and opacity transfer functions
+const ofun = vtkPiecewiseFunction.newInstance();
+ofun.addPoint(0.0, 0.0);
+ofun.addPoint(5000, 0.3);
+ofun.addPoint(30000, 1.0);
+
+actor.getProperty().setScalarOpacity(0, ofun);
+actor.getProperty().setScalarOpacityUnitDistance(0, 3.0);
+actor.getProperty().setInterpolationTypeToLinear();
+actor.getProperty().setShade(true);
+actor.getProperty().setAmbient(0.2);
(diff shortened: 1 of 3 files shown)


finetjul:
Regular VTIs can be read through XMLImageDataReader, you don't necessary need to use vti+json

martinken:
FYI I am gutting both the webgl1 and wbgl2 pathways for volume rendering right now in order to support multiple component rendering. The result is quite a bit cleaner but will conflict with the MR in a huge way.

martinken:
We definitely do want MIP though.

swederik:
Fair enough! So then I guess it's best to just wait and try to make a new PR with the updated shaders.

martinken:
If all goes well I should have it merged in the next day or two. It also collapses the webgl1 and webgl2 into one shader so that should make it easier for your changes to work on both paths.

swederik:
Sounds good! I'll keep an eye out for the changes and revisit MIP when they are ready. Thanks for the quick response.

martinken:
The new shader code has been merged now.

swederik:
Awesome, I'll try to update and recreate the PR. Thanks!

PR #1090, opened 2019-05-28
refactor(Rendering): compute volume lighting in the shader

martinken:
Significantly change the volume mapper to no longer compute
the normal and gradient volume, but to instead generate the
values as needed on the fly in the fragment shader. This
will result in a smaller GPU memory footprint and faster
volume load times. It may result in slower render times as there
are now 4 texture lookups per voxel where previously there were 2.
But it may be a tad faster as the lookups are all from the same
volume whereas before they were from two different volumes.

Note that the onLightingActivated callback no longer exists as lighting
is always activated.

Lots of code was cleaned up and made common between the webgl
and webgl2 cases. This should make future changes a bit easier.


martinken:
@jourdain  @thewtex   Note the removal of onLightingActivated

thewtex:
@martinken awesome!!

CC: @floryst

PR #1093, opened 2019-05-31
feat(VolumeMapper): Add MIP, MinIp, and AverageIP Blend Modes

swederik:
Had to open a new PR because I force pushed to the branch on https://github.com/Kitware/vtk-js/pull/1089

### TODO:
- [x] Average Intensity Projection isn't implemented as defined in the VTK C++ version: ["Average intensity blend mode works similar to the additive blend mode where the scalar values are multiplied by opacity calculated from the opacity transfer function and then added."](https://vtk.org/doc/nightly/html/classvtkVolumeMapper.html#aac00c48c3211f5dba0ca98c7a028e409). It is currently just averaging the scalar values along the ray.
- [x] Switch to the head.vti dataset for the example
- [x] Add test cases for the new blend modes

Happy for any comments!


martinken:
I just merged another big change that may cause some conflicts. But I'm happy to help get this in if you run into issues, just holler.

swederik:
Will do! I just need to find time to get back to this and finish it.

Would you happen to know what is causing the black speckling noise pattern in the VR images? Visible in the GIF here: https://github.com/Kitware/vtk-js/pull/1089

I've tried to play with sampling distance and some other parameters but couldn't see any effect on the noise.

martinken:
No clue, but when you get further along I can try testing it out and track it down. I suspect it is something simple.

swederik:
It looks like the black speckling is caused by the jitter used in the shader. If we set this value (https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl#L581) to zero, the speckling disappears in the example for the MPR Interactor Style (`npm run example -- InteractorStyleMPRSlice`). If we turn off the jitter for the MIP, we get results full of Moiré artifacts as expected, but the speckling is gone.

MoireArtifact

martinken:
You definately want the jitter, based on the non-jitter images it seems like your step size (sample distance) may be too large. Jitter just makes the banding you see be more randomly distributed. But if you have dark bands (like you do) then jitter will just make the dark spots be more randomly distributed. I think the issue in this case is that you are undersampling the volume enough that it is causing artifacts.  There is a trade off between sample distance and performance, small sample distance means slower performance, but better rendering. Large sample distance means better performance but potentially undersampling the volume. Generally you want your sample distance to be around one voxel in length.  Try a couple different sample distances and let me know what you see.

swederik:
Thanks for the quick response. We've been playing with sample distances but it doesn't seem to be improving the noise.

Volume spacing is `[3.5378181999999994, 3.5378181999999994, 3.375]`

| Sample distance: 5.0  | 3.5 | 0.5 |
| ------------- | ------------- | ------------- |
| Screenshot 2019-06-12 17 49 43 | Screenshot 2019-06-12 17 50 54 | Screenshot 2019-06-12 17 50 00 |

#### InteractorStyleMPRSlice
| Jitter On | Jitter Off |
| ------------- | ------------- |
| Screenshot 2019-06-12 17 57 12 |  Screenshot 2019-06-12 17 57 42 |


For the MPR case, would you accept a separate PR to add the following to the API? UseJitteringOn / UseJitteringOff / GetUseJittering? (like they are defined here: https://vtk.org/doc/nightly/html/classvtkGPUVolumeRayCastMapper.html#a9f58bdf2f81625edfbcb1fa7e9c26bd6)

martinken:
Let me look into your code first. Jitter should be fine so I want to dig into what is going on because I really think there is a coding issue somewhere in there.

swederik:
That would be great! You can also look here: https://kitware.github.io/vtk-js/examples/InteractorStyleMPRSlice.html because the artifact is also visible in master.

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -375,9 +382,150 @@ void main()
...
+      value /= (i_float + 1.0);
+    } else {
+      value /= i_float;
+    }
+
+    // Note that the resulting image is always grayscale i.e. aggregated values are not passed through the color transfer function. This is because the final value is a derived value and not a real data value along the sampling ray.
+    // TODO: Not sure this is correct
+    vec4 tcolor = vec4(value, value, value, value);
+    tcolor.a = residual*texture2D(otexture, vec2(value * oscale + oshift, 0.5)).r;
+
```
martinken: You should not multiply by residual here. It should be
```
tcolor.a = texture2D(otexture, vec2(value * oscale + oshift, 0.5)).r;
```

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -375,9 +382,150 @@ void main()
...
+      scalar = getTextureValue(posIS).r;
+
+      // Update the minimum/maximum value if necessary
+      value = min(scalar, value);
+    }
+
+    // now map through opacity and color
+    vec4 tcolor = texture2D(ctexture, vec2(value * cscale + cshift, 0.5));
+    tcolor.a = residual*texture2D(otexture, vec2(value * oscale + oshift, 0.5)).r;
+
```
martinken: Ditto here, do not mult by residual

martinken:
The issue in master is a bug related to jitter but not related to your issue oddly enough :-)

martinken:
I think most of this topic is good. I would want to rework the fragment shader a bit (once again I changed the one in master a ton so I am happy to do this) and I want to look a tad more at the code for the average case.  So a couple options. 

- If you want to, feel free to move to a more recent master which now has support for multicomponent volume rendering (big change to the shaders) and you can go from there. 

- or update this MR to just add the non fragment shader stuff (all the settings etc) without the actual changes to the fragment shader. We merge that. Then I'm happy to make the needed changes to the fragment shader in a subsequent MR in the next few days.

martinken:
My thoughts on the fragment shader are mainly that I would rather us use #if blend == 0 for the blend cases instead of if (blend ==) . Not sure if it makes a different once compiled but I feel the #if is safer as you know only the case you want will be compiled.

martinken:
And finally I *think* the issue with the thin slice is that the jitter residual is only handled on the end, but there is also a jitter residual at the start. Normally that makes little difference as the edges of the volume are transparent, but in this case, with such a thin slab, the missing jitter value from the beginning I think makes a difference. The fix will be to do a start residual, loop, then end residual. Right now we just have loop and end residual.

swederik:
Aha that makes some sense I guess. Glad to hear there is a sensible explanation. Thanks for the review. I've rebased the branch and I'll switch it to #if as soon as I can.

swederik:
I've switched it to use #if and rebased on the latest master branch.

One thing is that it doesn't use multi-component values for MIP / MinIP / AverageIP. It just uses `oscale0` / `oshift0` / etc. Is this something you would want fixed for this PR? Is that type of blending something someone would actually use for multi-component images?

Travis is failing but I do not understand why: https://travis-ci.org/Kitware/vtk-js/builds/549198893?utm_source=github_status&utm_medium=notification

Let me know if you'd like me to change anything else!

aronhelser:
@swederik Travis is failing on a 'prettier' command, which means it wants specific formatting.
You should be able do `npm run reformat` to automatically fix the formatting issue, then commit/push as needed.

swederik:
Thanks @aronhelser! I didn't see that because it looks like the fix is in an unrelated component's example (ResliceCursor). I'm not sure why Travis didn't catch it on that PR.

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -507,6 +510,206 @@ void getRayPointIntersectionBounds(
...
+    // operator = min or operator = max instead of duplicating
+    // the code?
+    if (blendMode == 1) { // MAXIMUM_INTENSITY_BLEND
+      // Sample along the ray until MaximumSamplesValue,
+      // ending slightly inside the total distance
+      for (int i = 0; i < //VTK::MaximumSamplesValue ; ++i)
+      {
+        // compute the scalar
+        scalar = getTextureValue(posIS).r;
+
```
martinken: Why not use tValue and tColor here? I think that would allow multiple component min/max

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -507,6 +510,206 @@ void getRayPointIntersectionBounds(
...
+      scalar = getTextureValue(posIS).r;
+
+      // Update the minimum/maximum value if necessary
+      value = min(scalar, value);
+    }
+
+    // now map through opacity and color
+    vec4 tcolor = texture2D(ctexture, vec2(value * cscale0 + cshift0, 0.5));
+    tcolor.a = texture2D(otexture, vec2(value * oscale0 + oshift0, 0.5)).r;
+
```
martinken: Then call getColorForTexture here

martinken:
This is looking good. I think with minor changes you could make it support multiple component min/max/average mainly by using/passing vec4 instead of float. The min/max functions work just as well on vec4 as float and the value to color function takes a vec4.

swederik:
Great! I'll give that a shot then. Thanks!

swederik:
Jul-02-2019 14-49-47
(left: python, right: javascript)

Everything looks like it's working now for MIP and MinIP. I'll try to do some more testing with RGB MIP/MinIP/AverageIP to make sure that works.

Notes:
- AverageIP no longer functions the way it does in VTK C++. Right now I've switched it to act the same was as Max/MinIP
- AverageIP still has some very weird boundary issues. I'm looking into these.
- The tests appeared to be running and passing even when I completely broke the shader. I don't know how that could have happened.

Screenshot 2019-07-02 14 55 21

martinken:
@swederik  you ready for this to be merged?

swederik:
I wanted to double check the tests since they seem to be passing even though I changed a number of things. If they appear to be passing now I suppose we can merge it and look at that after. I didn't get a chance to actually test it with RGB data, but it works fine with single component using the vec4 code, so I have no reason to assume it won't work.

I also wanted to ask what you thought about AverageIP. With this PR the way it is, AverageIP acts just like MinIP / MaxIP, not like the VTK C++ implementation. This is their description:

> Average intensity blend mode works similar to the additive blend mode where the scalar values are multiplied by opacity calculated from the opacity transfer function and then added. The additional step here is to divide the sum by the number of samples taken through the volume.

> One can control the scalar range by setting the AverageIPScalarRange ivar to disregard scalar values, not in the range of interest, from the average computation. As is the case with the additive intensity projection, the final image will always be grayscale i.e. the aggregated values are not passed through the color transfer function. This is because the resultant value is a derived value and not a real data value along the sampling ray.

With the switch to `getColorForValue` it seemed like a lot of extra code to add this behaviour though, so right now this just averages the vec4 for the component data within the volume (and within AverageIPScalarRangeMin/Max) and then calls `getColorForValue` to map it through opacity / color transfer functions. It sounds somehow more useful to me this way anyway.

If we keep the discrepant behaviour it would be good to document it somewhere (api.md?).

martinken:
I've never heard of using average projection so I'm not sure how folks would want it done.  I assumed it would be done like you implemented it, as opposed to the C++ version. My gut feeling is to merge this topic as it has been a while in development. Then go from there. Let me know if that works for you. I did download it and tested it on my machine and it passed all the tests as well.

swederik:
Awesome, then in that case I say merge away!

PR #1099, opened 2019-06-06
fix(ImageStream): typo in disconnect

aronhelser:
The subscription variable name was incorrect, and
the connected flag needed to be turned off.


aronhelser:
First time prettier has behaved differently on windows vs linux/mac........

jourdain:
lgtm

Issue #982, opened 2018-12-14
RGBA not supported for polydata?


jourdain:
Test example in codepen https://codepen.io/raimonds-vilums/pen/OrMrEz?editors=0010

```
let vtkFile = `
# vtk DataFile Version 2.0
geometry
ASCII
DATASET POLYDATA
POINTS 60 float
20 0 0
19.94 0.5197 0
19.78 1.0168 0
19.52 1.4694 0
19.17 1.8578 0
18.75 2.1650 0
18.27 2.37 0
17.76 2.48 0
17.23 2.48 0
16.72 2.37 0
16.25 2.16 0
15.82 1.85 0
15.47 1.46 0
15.21 1.016 0
15.05 0.5197 0
15 0 0
15.05 -0.51 0
15.21 -1.01 0
15.47 -1.46 0
15.82 -1.85 0
16.25 -2.16 0
16.72 -2.37 0
17.23 -2.48 0
17.76 -2.48 0
18.27 -2.37 0
18.75 -2.16 0
19.17 -1.85 0
19.52 -1.46 0
19.78 -1.01 0
19.94 -0.51 0
19 0 0
18.96 0.311 0
18.87 0.610 0
18.71 0.881 0
18.50 1.114 0
18.25 1.299 0
17.96 1.426 0
17.65 1.491 0
17.34 1.491 0
17.03 1.426 0
16.75 1.299 0
16.49 1.114 0
16.28 0.881 0
16.12 0.610 0
16.03 0.311 0
16 0 0
16.03 -0.311 0
16.12 -0.610 0
16.28 -0.881 0
16.49 -1.114 0
16.75 -1.299 0
17.03 -1.426 0
17.34 -1.491 0
17.65 -1.491 0
17.96 -1.426 0
18.25 -1.299 0
18.50 -1.114 0
18.71 -0.881 0
18.87 -0.610 0
18.96 -0.311 0

POLYGONS 30 150
4 0 30 31 1
4 1 31 32 2
4 2 32 33 3
4 3 33 34 4
4 4 34 35 5
4 5 35 36 6
4 6 36 37 7
4 7 37 38 8
4 8 38 39 9
4 9 39 40 10
4 10 40 41 11
4 11 41 42 12
4 12 42 43 13
4 13 43 44 14
4 14 44 45 15
4 15 45 46 16
4 16 46 47 17
4 17 47 48 18
4 18 48 49 19
4 19 49 50 20
4 20 50 51 21
[...]

PR #1100, opened 2019-06-07
fix(Rendering): an issue with using RGBA ScalarsToColors

martinken:
typo causing a crash


martinken:
fixes https://github.com/Kitware/vtk-js/issues/982

Issue #1104, opened 2019-06-12
Texture upload performances


mcmigliore:
There is currently a workaround in `vtkOpenGLTexture` for a Chrome 69 bug.  
Is this still needed? It slows down the function.

@manthey @laurennlam 

https://github.com/Kitware/vtk-js/blob/08091fbf41d939023f8d9d31cda82730c1b44c16/Sources/Rendering/OpenGL/Texture/index.js#L913

manthey:
This appears to only have been a problem with Chrome 69.  Assuming we don't care about Chrome 69, it can be removed.

mcmigliore:
I'll remove it, thanks!

PR #1105, opened 2019-06-12
fix(perf): Remove Chrome 69 fix

mcmigliore:
This workaround was causing performance issues.  
Removing it since the bug was fixed in Chrome 70.


mcmigliore:
Fix https://github.com/Kitware/vtk-js/issues/1104

mcmigliore:
@laurennlam @finetjul

jourdain:
Maybe you just want to guard it with `~navigator.userAgent.indexOf('Chrome/69')` ?

mcmigliore:
> Maybe you just want to guard it with `~navigator.userAgent.indexOf('Chrome/69')` ?

Thanks for the suggestion, it is indeed better.

martinken:
+1

PR #1106, opened 2019-06-12
fix(OpenGLVolume): Properly initialize keyMatrixTime with mtime=0

llangholz89:
Initialize model.keyMatrixTime with mtime=0 so that actor matrices are calculated on first pass


martinken:
+1

PR #1109, opened 2019-06-12
fix(VolumeMapper): Add Jittering options

swederik:
When working on https://github.com/Kitware/vtk-js/pull/1093 we noticed that turning off jittering helped avoid the stochastic noise we were seeing when using the volume mapper with the MPR-Slice Interactor Style (https://kitware.github.io/vtk-js/examples/InteractorStyleMPRSlice.html). This PR adds the 'jittering' option, which defaults to true, in order to allow this to be disabled when desired.

@martinken, feel free to trash this PR if you find another answer for the speckling. Note that this does not fix the MIP/MinIP/AverageIP case, it only fixes it for composite rendering.

This can be tested with `npm run example -- InteractorStyleMPRSlice`

### DIFF
--- a/Sources/Interaction/Style/InteractorStyleMPRSlice/example/index.js
+++ b/Sources/Interaction/Style/InteractorStyleMPRSlice/example/index.js
@@ -29,6 +29,11 @@ const actor = vtkVolume.newInstance();
 const mapper = vtkVolumeMapper.newInstance();
 actor.setMapper(mapper);
 
+// When we perform multiplanar reconstruction in this manner,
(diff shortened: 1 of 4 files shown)


martinken:
Thanks for catching the thin plane issue. It is a bug related to jitter and I'll whip up a fix for it.

pieper:
Would still be good to expose the jitter on/off option since (1) it exists in VTK and (2) because we shouldn't be jittering when our goal is to do MPR (we just want a trilinear interpolation of the front clipping plane).

PR #1113, opened 2019-06-15
ci(npm): Use npm ci instead of npm install

thewtex:
This is the preferred command for deployment that uses
package-lock.json.

### DIFF
--- a/Utilities/ci/azure-pipelines.yml
+++ b/Utilities/ci/azure-pipelines.yml
@@ -24,7 +24,7 @@ jobs:
     displayName: 'Install Node.js'
 
   - script: |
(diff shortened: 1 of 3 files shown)


thewtex:
Resemble.js 3.1.0 released this fix:

https://github.com/rsmbl/Resemble.js/pull/168

so we can update it in vtk.js -- it also updates the canvas dependency, which fixes installation issues.

jourdain:
@thewtex can you pull from master and update the lock file in order to fix the conflict?

jourdain:
did via side commits

thewtex:
Thanks, @jourdain

PR #1119, opened 2019-06-19
feat(ViewProxy): Focus camera to a given point

### LINKED ISSUE #1114: Centering camera on a given point
I would like to center the camera on a given point. I see the ViewProxy.resetCamera computes a new center from the bounds. 

Could it be possible to change this center without modifying the bounds? Are you interested by such feature?

BotellaA:
fix #1114

### DIFF
--- a/Sources/Proxy/Core/ViewProxy/index.js
+++ b/Sources/Proxy/Core/ViewProxy/index.js
@@ -516,6 +516,14 @@ function vtkViewProxy(publicAPI, model) {
   };
 
   // --------------------------------------------------------------------------
+
+  publicAPI.focusTo = macro.chain(
+    model.camera.setFocalPoint,
+    model.interactorStyle2D.setCenterOfRotation,
+    model.interactorStyle3D.setCenterOfRotation
+  );
+
+  // --------------------------------------------------------------------------
   // Initialization from state or input
   // --------------------------------------------------------------------------
 


Sources/Proxy/Core/ViewProxy/index.js
```diff
@@ -515,6 +515,14 @@ function vtkViewProxy(publicAPI, model) {
     model.renderWindow.render();
   };
 
+  // --------------------------------------------------------------------------
+
```
thewtex: Please describe the method and give an example of its inputs.

floryst:
@jourdain do you think this is useful enough to add in? You should be able to get the same effect with the following code (note: I did not test it):
```
view.setCameraFocalPoint(...focalPoint);
view.resetCamera();
```

The only extra things the above two lines do is it also resets the camera and the camera clipping range. This PR condenses two lines down to one line, which can be useful if it's a common enough operation.

BotellaA:
@floryst `view.resetCamera();` does not keep the previous `focalPoint` but recomputes it from the center of the bounds. My goal is to fix this center from a picked point and not from the center.

floryst:
Ok, then this looks good to me. As for the usefulness, I can see having a "focusTo" top-level method be easier to use than remembering to set it on the interactor styles.

BotellaA:
Do you intent to merge this? Is there something still missing?

floryst:
I'll merge in a bit once I get another fix-up PR ready. That way we cut a single release with a couple of merged PRs. Apologies for any delays!

PR #1121, opened 2019-06-24
fix(ITKHelper): Support multi-component pixel types

thewtex:
Support reading multi-component images from itk.js.

All pixel data is added to the vtkImageData's PointData "Scalars" to
facilitate multi-component volume rendering.

If the itk.js PixelType is Vector or CovariantVectors and the dimension
is 3 and there are 3 vector components, the DataArray is also set as the
vtkImageData's PointData "Vectors". VTK does not distinguish vectors and
covariant vectors. ITK images can be N-dimensional and the number of
vector components do not always match the image dimension.

If the itk.js PixelType is SymmetricSecondRankTensor or
DiffusionTensor3D and the dimension is 3 and there are 6 tensor
components, the DataArray is also set as the vtkImageData's PointData
"Tensors".


martinken:
+1

Issue #1120, opened 2019-06-20
Point representation always render as black


hevo66:
When rendering using point repsentation, the color does is not respected. The points are always rendered as black.

This can be reproduced in the GeoemtryViewer online example:

1. Open the diskout.vtp
2. Change representation to 'Points'
3. All points are black 
4. Change the solid color to any of the colors
5. Points are still Black

This issue seems to be introduced in v8.5.9.

danmindru:
I can reproduce that as well on the latest Chrome and Firefox.

martinken:
Should be an easy fix, I'll take a look this week

danmindru:
Great, thanks! Just let me know if I can help somehow.

hevo66:
Thank you for fixing this so quick :)

martinken:
Very welcome :-)

PR #1124, opened 2019-06-24
fix(Rendering): Fix lighting on points

martinken:
Fix issue https://github.com/Kitware/vtk-js/issues/1120


aronhelser:
prettier is complaining - `npm run reformat` :)

PR #1126, opened 2019-06-25
fix(AbstractMapper3D): Fix usage and implementation of getCenter

### LINKED ISSUE #1123: Picker: get wrong center in intersectWithLine
I am trying to use the `vtkPicker` and get the picked actors. In the `intersectWithLine` function, I have an issue with the `center` which is alway an empty array: https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/Picker/index.js#L44

It seems to work by changing this line to
```
    const center = mapper.getCenter();
```
But there is maybe a bug somewhere, maybe in `getCenter`...?
Why to you splice the `model.center` instead of always returning it?

floryst:
Fixes #1123 @BotellaA

### DIFF
--- a/Sources/Rendering/Core/AbstractMapper3D/index.js
+++ b/Sources/Rendering/Core/AbstractMapper3D/index.js
@@ -16,13 +16,10 @@ function vtkAbstractMapper3D(publicAPI, model) {
     }
   };
 
-  publicAPI.getCenter = (...center) => {
+  publicAPI.getCenter = () => {
     publicAPI.getBounds();
     for (let i = 0; i < 3; i++) {
       model.center[i] = (model.bounds[2 * i + 1] + model.bounds[2 * i]) / 2.0;
-    }
-    if (Array.isArray(center[0])) {
-      center[0] = model.center.splice(0);
     }
     return model.center;
   };
--- a/Sources/Rendering/Core/Picker/index.js
+++ b/Sources/Rendering/Core/Picker/index.js
@@ -40,8 +40,7 @@ function vtkPicker(publicAPI, model) {
       return Number.MAX_VALUE;
     }
 
-    const center = [];
-    mapper.getCenter(center);
+    const center = mapper.getCenter();
 
     const ray = [];
     for (let i = 0; i < 3; i++) {


Sources/Rendering/Core/AbstractMapper3D/index.js
```diff
@@ -16,14 +16,11 @@ function vtkAbstractMapper3D(publicAPI, model) {
...
-  publicAPI.getCenter = (...center) => {
+  publicAPI.getCenter = () => {
     publicAPI.getBounds();
     for (let i = 0; i < 3; i++) {
       model.center[i] = (model.bounds[2 * i + 1] + model.bounds[2 * i]) / 2.0;
     }
-    if (Array.isArray(center[0])) {
-      center[0] = model.center.splice(0);
-    }
     return model.center;
```
finetjul: To be consistent with getArray in macro.js, shouldn't we return a copy of the center ([].concat(model.center)). I'm afraid it could be error prone to return a reference and some developer ends up modifying it.

Sources/Rendering/Core/AbstractMapper3D/index.js
```diff
@@ -16,14 +16,11 @@ function vtkAbstractMapper3D(publicAPI, model) {
     }
   };
 
-  publicAPI.getCenter = (...center) => {
+  publicAPI.getCenter = () => {
```
finetjul: This changes the API. 
It probably shouldn't have just be a patch version release. Let's hope that no-one used the getCenter(center) API externally to VTK.JS :-)

Sources/Rendering/Core/AbstractMapper3D/index.js
floryst: Yes you are right. :P I will make that change.

Sources/Rendering/Core/AbstractMapper3D/index.js
floryst: Indeed. However, anyone using `getCenter(center)` would have incorrect code anyways, since `getCenter(center)` doesn't actually do anything. (`center` doesn't actually change.)

Issue #1118, opened 2019-06-19
ImageCropFilter.setCroppingPlanes(...planes) BUG


tanxu1120:
Examples -> Filters/General -> ImageCropFilter

Dragging K range results in an error.

![15609155588763](https://user-images.githubusercontent.com/37609756/59735146-e6628500-9286-11e9-9984-19b4a8e67693.png)

floryst:
I cannot seem to reproduce this even when closely matching your screenshot. (Tested on Firefox and Chrome.) Can you provide reproduction steps?

tanxu1120:
step 1 : Open https://kitware.github.io/vtk-js/ in the Chrome browser.
![1](https://user-images.githubusercontent.com/37609756/59815359-e2983680-934a-11e9-8efc-72c7c638e295.png)
step 2 : Click on Examples.
![2](https://user-images.githubusercontent.com/37609756/59815503-53d7e980-934b-11e9-8d78-4d68320b3dc3.png)
step 3 : Click on ImageCropFilter.
![3](https://user-images.githubusercontent.com/37609756/59816148-9b5f7500-934d-11e9-85df-835fc3d8ebce.png)
step 4 : Rotate image to coronal position and zoom in.
![4](https://user-images.githubusercontent.com/37609756/59816325-407a4d80-934e-11e9-8d2b-ad318234629a.png)
step 5 : Drag K range, the image shows a problem.
![5](https://user-images.githubusercontent.com/37609756/59816432-a9fa5c00-934e-11e9-9408-bbc61e8cdced.png)

floryst:
Hm, okay I cannot seem to reproduce the error. What browser are you using and what version? Does this behavior occur for other axes?

tanxu1120:
GPU : NVIDIA GeForce GT 730M
vtk.js version : 8.8.0
Browser : Google Chrome 75.0.3770.100
OS : win7 64bit service pack 1

No problems with other axes.

tanxu1120:
I tried it with a Mac, it works.

aronhelser:
Repro on Win10, chrome, nvidia graphics. The volume seems to compress along the K axis, and leave double images. Crops properly along I, J axes.

floryst:
@martinken

PR #1131, opened 2019-06-27
fix(Rendering): fix bugs in volume Rendering

martinken:
Fix an issue where a texture is reused multiple times
with cropped data.

Fix an issue with the crop filter where the inputs
extent is not considered when computing offsets.

Fix a rendering issue where the mapping of the volume
to pixels was not quite correct resulting in deformations
when rendering cropped images.


martinken:
This should fix issue https://github.com/Kitware/vtk-js/issues/1118

martinken:
@floryst see the changes to the corp filter (not the cause of the reported issue but I think they are needed)

PR #1132, opened 2019-06-27
perf(vtkMath): Export vtkMath as separate objects

### LINKED ISSUE #1038: "export default" convention in vtkMath prevents tree-shaking and causes bloat
The convention of having [a single default object exported by `vtkMath](https://github.com/Kitware/vtk-js/blob/772d31b47f39c04deb24eb76dfcb82f99364f954/Sources/Common/Core/Math/index.js#L2049) prevents other modules from directly importing only the parts of `vtkMath` which are needed, and allowing the rest to be dropped via tree-shaking.

Instead of exporting a default object, `vtkMath` (and perhaps other parts of `vtk.js` which follow the same anti-pattern) should provide named exports of all its individual utility functions.

brianhelba:
Fixes #1038


brianhelba:
See https://github.com/webpack/webpack/tree/master/examples/harmony-unused for more info on how this works.

PR #1139, opened 2019-07-03
fix(ShapeWidget): Cancel animation when losing focus in ShapeWidget.

RemiCecchinato:
@luciemac @finetjul


finetjul:
I am not familiar enough with the widgets to give my opinion on this PR. I think only @floryst  can.

floryst:
LGTM

Issue #1135, opened 2019-06-28
Unimplemented but documented Camera functions


mix3d:
The [Camera Documentation](https://kitware.github.io/vtk-js/api/Rendering_Core_Camera.html) states that the following functions are available:

`applyTransform(transform)`
>Apply a transform to the camera. The camera position, focal-point, and view-up are re-calculated using the transform’s matrix to multiply the old points by the new transform.

But this is not implemented at all.

Additionally, 
`getOrientation()`
>Get the orientation of the camera (x, y, z orientation angles from the transformation matrix).
and 
`getOrientationWXYZ()`
>Get the wxyz angle+axis representing the current orientation. The angle is in degrees and the axis is a unit vector.
are both implemented as `() => {};`

Several other functions are only half-documented, (explicitly stating a `getFunction` when there are both getters and setters) and some are even duplicated, (eg: `roll()` and `roll(angle)`, yet does not state if the angle is in degrees or radians)

Camera functions are a pretty important part of setting up a viewer. Either state that certain functions are unimplemented or don't include them in documentation until finished.

What can I do to help?

jourdain:
If you feel like it, and that would be of great help, you can update the [api.md](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/Camera/api.md) or improve the code in the [camera](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/Camera/index.js) by implementing the missing functions. For that you can look at the [VTK/C++](https://github.com/Kitware/VTK/blob/master/Rendering/Core/vtkCamera.cxx) counter part if need be.

mix3d:
I'm halfway through a docs update. Even in the C++ counterpart, I can't see where Scissor is actually or referenced, outside of its getters/setters.

If they aren't used _anywhere_, should we drop them from the spec, to reduce API clutter?

jourdain:
Yes, simpler is better

mix3d:
Running into more things, hopefully you can help:

Some functions that are implemented, but not called anywhere and look like they would be a _Bad Idea<sup>TM</sup>_:

**`setViewMatrix(mat4)`**
[...]

PR #1141, opened 2019-07-08
refactor(Core/Camera): Refactor Camera Docs, implement applyTransform()

mix3d:
Refactor Camera code for clarity, moving unimplemented functions together, implemented `applyTransform()` function.

Fix docs to remove bad references, update function defs to be more clear.

Removed "Scissor" references that were not implemented nor referenced anywhere, even in VTK C++ code.

Related to https://github.com/Kitware/vtk-js/issues/1135

### DIFF
--- a/Sources/Rendering/Core/Camera/index.js
+++ b/Sources/Rendering/Core/Camera/index.js
@@ -145,7 +145,6 @@ function vtkCamera(publicAPI, model) {
     computeViewPlaneNormal();
   };
 
-
   //----------------------------------------------------------------------------
   // Move the position of the camera along the view plane normal. Moving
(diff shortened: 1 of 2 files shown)


Sources/Rendering/Core/Camera/index.js
```diff
@@ -39,6 +39,14 @@ function vtkCamera(publicAPI, model) {
   const newPosition = vec3.create();
   const newFocalPoint = vec3.create();
 
+  // Internal Functions that don't need to be public
+  function computeViewPlaneNormal() {
```
mix3d: Moved to an internal function, off of publicAPI, since any publicAPI function that changes the directionofProjection should call this internally, a consumer of VTK.js should never need to call this.

PR #1144, opened 2019-07-15
feat(TriangleFilter): Add vtkTriangleFilter and vtkPolygon.

RemiCecchinato:


RemiCecchinato:
@finetjul @luciemac

Sources/Common/Core/PriorityQueue/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+function vtkPriorityQueue(publicAPI, model) {
+  // Set our classname
+  model.classHierarchy.push('vtkPriorityQueue');
+
+  publicAPI.push = (priority, element) => {
+    // naive algo
+    model.elements.push({ priority, element });
+    model.elements.sort(
+      ({ priorityLeft }, { priorityRight }) => priorityLeft - priorityRight
```
finetjul: you need to be careful with the sort algo. It is unstable with old chrome. Please do something in case of equality to enforce an order.

Sources/Common/Core/PriorityQueue/index.js
```diff
@@ -0,0 +1,61 @@
...
+    );
+  };
+
+  publicAPI.pop = () => {
+    if (model.elements.length > 0) {
+      const { element } = model.elements[0];
+      model.elements.shift();
```
finetjul:
```
const { element } = model.elements.shift();
```
?

Sources/Common/DataModel/Polygon/index.js
```diff
@@ -0,0 +1,248 @@
...
+    model.normal = [0, 0, 0];
+    const anchor = [...model.firstPoint.point];
+
+    let point = model.firstPoint;
+    for (let i = 0; i < model.pointCount; i++) {
+      vtkMath.subtract(point.point, anchor, v1);
+      vtkMath.subtract(point.next.point, anchor, v2);
+
+      const n = [0, 0, 0];
+      vtkMath.cross(v1, v2, n);
```
finetjul: what happens if v1 is [0,0,0] ?

Sources/Common/Core/PriorityQueue/index.js
RemiCecchinato: We don't want to return the priority, so we only return the element.

Sources/Common/DataModel/Polygon/index.js
RemiCecchinato: If v1 is [0, 0, 0] then two of the three points in the triangle are the same. Since this triangle is empty it's normal shouldn't change the polygon's normal.

Sources/Common/Core/PriorityQueue/index.js
RemiCecchinato: Ok.

Sources/Common/Core/PriorityQueue/index.js
finetjul: I know, but with my suggestion, you can combine 2 lines (21 &22) in just one line.

Alternatively:
```return model.elements.shift().element;```

File: Sources/Common/DataModel/Polygon/index.js [hunk 4]
```diff
@@ -0,0 +1,248 @@
...
+    let vtx = point;
+    for (let ptDId = 0; ptDId < model.pointCount; ptDId++) {
+      const previousVtx = vtx.previous;
+      val = vtkPlane.evaluate(model.normal, previous.point, vtx.point);
+      let sign = 0;
+      if (val > EPSILON) {
+        sign = 1;
+      } else if (val < EPSILON) {
+        sign = -1;
+      }
```
finetjul: how about 
```const sign = val > EPSILON? 1 : val < EPSILON ? -1: 0;```

File: Sources/Common/DataModel/Polygon/index.js [hunk 5]
```diff
@@ -0,0 +1,248 @@
...
+    let val = vtkPlane.evaluate(model.normal, previous.point, next.next.point);
+    let currentSign = 0;
+    if (val > EPSILON) {
+      currentSign = 1;
+    } else if (val < EPSILON) {
+      currentSign = -1;
+    }
+    let oneNegative = currentSign < 0 ? 1 : 0;
+
+    let vtx = point;
```
finetjul: rename vtx into vertex

Sources/Common/DataModel/Polygon/index.js
```diff
@@ -0,0 +1,248 @@
+import macro from 'vtk.js/Sources/macro';
+import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
+import vtkLine from 'vtk.js/Sources/Common/DataModel/Line';
+import vtkPlane from 'vtk.js/Sources/Common/DataModel/Plane';
+import vtkPriorityQueue from 'vtk.js/Sources/Common/Core/PriorityQueue';
+import { IntersectionState } from 'vtk.js/Sources/Common/DataModel/Line/Constants';
+
```
finetjul: if there is document in C++, can you copy it in a separate api.md file ?

Sources/Filters/General/TriangleFilter/index.js
```diff
@@ -0,0 +1,114 @@
...
+    const input = inData[0];
+    const polys = input.getPolys();
+    const points = input.getPoints().getData();
+    const cells = polys.getData();
+    let newCells = [];
+    let newPoints = [];
+
+    if (cells && cells.length > 0) {
+      let cellOffset = 0;
+      while (cellOffset < cells.length - 1) {
```
finetjul: use for loop instead ?

finetjul:
👍 +1

RemiCecchinato:
@finetjul can you merge it please ? I need it to create new PR.

Issue #612, opened 2018-02-26
row-major vs column-major conventions


agirault:
## Row-major vs column-major
There have been confusions regarding the data layout conventions for matrix in vtk.js:
https://en.wikipedia.org/wiki/Row-_and_column-major_order

## Existing conventions
- itk & itk.js are **row-major**
- vtk c++ is **row-major** (example with [vtkMatrix4x4](https://www.vtk.org/doc/nightly/html/classvtkMatrix4x4.html#details))
- vtk.js [Math](https://github.com/Kitware/vtk-js/blob/master/Sources/Common/Core/Math/index.js) is **row-major**
- vtk.js uses the **column-major** convention through **gl-matrix** for storing some variables and doing some operations  https://github.com/Kitware/vtk-js/search?l=JavaScript&q=gl-matrix

## glmatrix: "This may lead to some confusions... sorry about that"
As seen on http://glmatrix.net : 

> <img width="516" alt="screen shot 2018-02-26 at 11 20 32 am" src="https://user-images.githubusercontent.com/4910855/36681610-19ed1070-1ae7-11e8-833e-7e9352173c4f.png">

## Issues already found
This leads to issues like noted here:
https://github.com/Kitware/vtk-js/blob/d610c5f81080e974a8244e24d02098c8a227e6ba/Sources/Common/DataModel/ITKHelper/index.js#L40-L45
https://github.com/Kitware/vtk-js/blob/d610c5f81080e974a8244e24d02098c8a227e6ba/Sources/Rendering/Core/ImageMapper/index.js#L174-L187

[vtkMatrixBuilder](https://github.com/Kitware/vtk-js/pull/611/files/c0fc9a04a296976e344c86e16c05c506972175da#r170641029) ?

agirault:
@martinken @thewtex @jourdain @floryst

thewtex:
CC: @jcfr

jcfr:
Thanks @agirault  for summarizing the discussion and finding here. Documenting the convention will really help streamline the maintenance and help foster contributions to the toolkit :smile: 

I like the idea of using math.js and limiting the use (if still needed) of glMatrix to rendering library.

HaochengLIU:
```
vtk c++ is row-major (example with vtkMatrix4x4)
```
[...]

PR #1145, opened 2019-07-17
docs(MatrixBuilder): Create documentation for and refactor MatrixBuilder class

mix3d:
BREAKING CHANGE: Remove getVTKMatrix function as it had a side-effect, was poorly named, and was not referenced anywhere else. If the user wants to transpose the matrix, they can do it after calling `getMatrix()`

### DIFF
--- a/Sources/Common/Core/MatrixBuilder/api.md
+++ b/Sources/Common/Core/MatrixBuilder/api.md
@@ -1,6 +1,6 @@
 ## Introduction
 
-vtkMatrixBuilder class provides a system to create a mat4 transformation matrix. All functions return the MatrixBuilder Object instance, allowing transformations to be chained.
+The `vtkMatrixBuilder` class provides a system to create a mat4 transformation matrix. All functions return the MatrixBuilder Object instance, allowing transformations to be chained.
 
 Example:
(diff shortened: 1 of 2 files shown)


jourdain:
@agirault do you use the getVTKMatrix() in your projects?

mix3d:
> @agirault do you use the getVTKMatrix() in your projects?

If so, perhaps an alternative could be `.transpose()` as a function, replacing `getVTKMatrix()` with `vtkMatrixBuilder.buildFromDegree().translate().transpose().getMatrix()`?

jourdain:
The main issue is the matrix orientation convention between glMatrix and VTK/C++. 
As long as we stick with one for data manipulation, we are good, but when reading data, we run into issue. And I know that was brought up by @agirault. We just need to agree in term of API, but only @agirault and maybe @thewtex really know the corner use cases.

mix3d:
my main concern is the side effect. If you call .getVTKMatrix() twice, you'll end up with two different results.

agirault:
Not sure about that specific method, but that's probably related to this issue: https://github.com/Kitware/vtk-js/issues/612

PR #1148, opened 2019-07-18
feat(Core/Renderer): Implement removeAllActors(), removeAllVolumes()

mix3d:
Copies the intent of removeAllLights, which was implemented but not documented.


mix3d:
@jourdain any updates?

PR #1151, opened 2019-07-19
fix(InteractorStyleMPRSlice): Keep clipping range consistent

swederik:
fix(InteractorStyleMPRSlice): Keep clipping range consistent between mouse movement and initial setup

This is a tiny change that I have already discussed once with @floryst. Essentially we use 0.1 in one place:

https://github.com/Kitware/vtk-js/blob/c9fa0ceb23390af7b9059056b18024ffa96fa9c5/Sources/Interaction/Style/InteractorStyleMPRSlice/index.js#L295-L300

and 1 in the other for the width of the clipping range. It seems more sensible to make them both 0.1.

I'm only sending this PR because I ran into different behaviour in the mapper during mouse movement / initial setup.

Right now changing this won't really have much impact, but in an upcoming PR we have a special path for cases where the ray path distance is less than sampling distance, and in that case this value being smaller will improve the performance.


floryst:
LGTM. I can't merge at the moment bc of travel.

Issue #1146, opened 2019-07-18
Jitter-related artifact in thin-slice volume renderings


swederik:
Hi all,

There is a noticable artifact when using the thin-camera-clipping-range MPR approach. It looks like a bit of fuzziness. I suppose the artifact is also there in generic volume rendering but it's less obvious. If you disable jitter you can resolve it, but that doesn't seem like the final solution.

I'd like to take a stab at solving this problem so I figured it made sense to create a new issue to track it.

You can see the artifact here: https://kitware.github.io/vtk-js/examples/InteractorStyleMPRSlice.html

![Screenshot 2019-07-18 14 10 47](https://user-images.githubusercontent.com/607793/61456366-24c29100-a966-11e9-89ff-0e69c496a480.png)

Here is @martinken's description:

> And finally I *think* the issue with the thin slice is that the jitter residual is only handled on the end, but there is also a jitter residual at the start. Normally that makes little difference as the edges of the volume are transparent, but in this case, with such a thin slab, the missing jitter value from the beginning I think makes a difference. The fix will be to do a start residual, loop, then end residual. Right now we just have loop and end residual.

_Originally posted by @martinken in https://github.com/Kitware/vtk-js/pull/1093#issuecomment-501375681_

Thanks!

Relevant code:

https://github.com/Kitware/vtk-js/blob/c9fa0ceb23390af7b9059056b18024ffa96fa9c5/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl#L758

https://github.com/Kitware/vtk-js/blob/c9fa0ceb23390af7b9059056b18024ffa96fa9c5/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl#L553

martinken:
I think the fix is just to handle the start jitter as well. Basically do the same gist and the end residual but for start residual. Should be a fairly easy fix

swederik:
Oh you weren't kidding, that was a lot easier than I expected. I'll send the PR shortly!

mix3d:
will that Jitter fix also fix the case where you get weird artifacts when at the max volume range?

![image](https://user-images.githubusercontent.com/1474137/61478951-a656f680-a947-11e9-8358-0776449474bc.png)

swederik:
Weird, I don't see that artifact even before the changes. Which browser? I do see flickering at max volume range here though: https://kitware.github.io/vtk-js/examples/InteractorStyleMPRSlice.html

[...]

PR #1152, opened 2019-07-19
fix(vtkVolumeFS): Address sampling issues during thin-slice volume rendering

swederik:
Addresses #1146 

### Changes:

- Checks inside `computeIndexSpaceValues` if the length of the ray is smaller than the sample distance. If so, sets numSteps to 1.0 so we can skip some unnecessary computations
- Adds early exits for each blend mode in case numSteps is 1.0. In this case, we can skip the whole ray traversal loop and just take the first value.
- Stores the initial position along the ray so it can be passed to applyBlend as startResidualIS.

The idea is that the first sample along the ray would be from the initial position in the volume, and then there would be the usual sampling loop up to 1-step away from the end, and finally the residual step to sample the end position. That is why we are computing `initialPosIS = posIS - startResidual`.

I'm leaving this as a draft for now because I am still seeing the jitter speckling pattern on normal composite blend volume rendering (e.g. `npm run example -- VolumeMapper` and thick-slab MIP/MinIP, but not AverageIP) , though it is not as apparent now.

I figure that the speckling is caused by the texture being sampled outside of the texture dimensions, which is why it shows up as black spots.

### Testing
The easiest way to test this is to use `npm run example -- InteractorStyleMPRSlice`. If you want to see how it changes in thicker slabs, edit the +1 and +0.1 values here: https://github.com/Kitware/vtk-js/blob/master/Sources/Interaction/Style/InteractorStyleMPRSlice/index.js#L117

Set them both to something like 20 or 30.

and add mapper.setBlendModeToMaximumIntensity() in the example here: https://github.com/Kitware/vtk-js/blob/master/Sources/Interaction/Style/InteractorStyleMPRSlice/example/index.js#L41

### DIFF
--- a/Sources/Rendering/OpenGL/VolumeMapper/index.js
+++ b/Sources/Rendering/OpenGL/VolumeMapper/index.js
@@ -816,7 +816,6 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
           fbSize[0] !== Math.floor(size[0] * 0.7) ||
           fbSize[1] !== Math.floor(size[1] * 0.7)
         ) {
-          console.log('resizing');
           model.framebuffer.create(
             Math.floor(size[0] * 0.7),
             Math.floor(size[1] * 0.7)
--- a/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
+++ b/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
@@ -513,13 +513,14 @@ void getRayPointIntersectionBounds(
 //=======================================================================
 // Apply the specified blend mode operation along the ray's path.
 //
-void applyBlend(vec3 posIS, vec3 stepIS, vec3 tdims, float numSteps)
+void applyBlend(vec3 posIS, vec3 stepIS, vec3 tdims, float numSteps, vec3 startResidualIS)
 {
   vec3 tstep = 1.0/tdims;
 
   // integer number of steps to take and residual step size
   int count = int(numSteps - 0.05); // end slightly inside
   float residual = numSteps - float(count);
+  vec3 initialPosIS = posIS - startResidualIS;
 
   // local vars for the loop
   vec4 color = vec4(0.0, 0.0, 0.0, 0.0);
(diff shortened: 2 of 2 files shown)


martinken:
I have some changes to this but I'm not sure how to add them to this topic. I can post diffs if you want or maybe there is a way to add me to commit to this?

swederik:
Awesome! I've been pulled into other things but plan to get back to this
very soon. I'm happy with diffs, comments, or commits. I added you as a
collaborator on my fork so I think you should be able to push directly to
the branch now.

martinken:
@swederik Can you take a look?

swederik:
Your changes definitely make it cleaner!

I made some small changes to AverageIP blend mode. It was previously dividing the sum by `i`, which would mean that it would divide by all sampling steps even those which had values which were not inside the value range. Now it uses `stepsTravelled` so that it is only averaging values within the range.

Some questions:
- Why is the alpha at each step in composite blending now ` tColor.a = 1.0 - pow(1.0 - tColor.a, raySteps - stepsTraveled);` ? It seems like it's working fine, i just don't understand where that formula came from. Is it something I can find in here (https://en.wikipedia.org/wiki/Alpha_compositing or here https://en.wikipedia.org/wiki/Blend_modes ?)
- Why multiply `tColor.a *= raySteps;` in the thin-slice composite blend scenario? It looks like this line makes the image too transparent.

One issue:
- I think I am seeing some ring-shaped aliasing artifacts on the thin-slice composite case. Should we add jitter to posIS (or to the middle of posIS and endIS)?

Screenshot 2019-07-26 10 47 04

martinken:
This here https://en.wikipedia.org/wiki/Opacity_(optics) gives the basis for it. Note how the accumulated opacity is a power function with the exponent being the distance traveled. The old code I wrote incorrectly used a linear weighting to account for the smaller jitter start/end steps. The correct formula is now used.  

A simple test case can show the impact. Say a unit step gives 0.5 opacity, then two half steps would each give 0.25 opacity using a linear formula. Blending those values though is multiplicative resulting in  1.0 - 0.75*0.75 = 0.4375 which is clearly not the same as 0.5. Using the new code the half step opacity becomes 0.2929 which when blended gives 1.0 - .707*.707 = 0.5 whcih is correct.

martinken:
"    Why multiply tColor.a *= raySteps; in the thin-slice composite blend scenario? It looks like this line makes the image too transparent."

Because it is a bug :-) It should also use the pow formula ala
```
tColor.a = 1.0 - pow(1.0 - tColor.a, raySteps);
```

martinken:
Can you do a quick fix?

martinken:
Hmm, In terms of using jitter for this slice I'm not sure what folks would want.  You could try updating the code to do the following and see if you like it?  I suspect it will end up showing jitter noise instead of rings but it is an easy change to test.

```
 if (raySteps <= 1.0)
  {
     posIS += (jitter*(endIS - posIS));
   }
```

swederik:
Thanks for the explanation! I pushed the change to the alpha value and it's looking good now.

I've been playing with adding jitter for the thin slice and it doesn't seem to be making much impact. Either way this artifact is much less bothersome than the current artifact, so I don't mind merging this as is and coming back to the aliasing issue later if we need to.

mix3d:
Noticing some strange artifacting at small slabThicknesses with Average blend mode:

image

image

volumeMapper's sample distance is 1. If the "smaller than sample distance" stuff was the culprit, I'd be surprised the 1.9 was still creating such noise.

finetjul:
The [VolumeMapperBlendModes example](https://kitware.github.io/vtk-js/examples/VolumeMapperBlendModes.html) is broken. There is:
 - an artefact with the first and last slice
 - lighting seems odd when looking at the volume from the side

PR #1153, opened 2019-07-22
fix(CylinderSource): Translate points to center before rotate

floryst:
Fixes issue where points were not being centered around origin before applying rotation.


finetjul:
:+1

PR #1169, opened 2019-08-01
Polygon painting

RemiCecchinato:
@finetjul

### DIFF
--- a/Sources/Widgets/Widgets3D/SplineWidget/index.js
+++ b/Sources/Widgets/Widgets3D/SplineWidget/index.js
@@ -62,6 +62,7 @@ const DEFAULT_VALUES = {
     key: 'Shift',
     status: 'down',
   },
+  defaultCursor: 'pointer',
 };
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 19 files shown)


finetjul:
Warning, 1 commit has "SplineWidget" as feature, and the other "vtkSplineWidget", rename the 2nd one to "feat(SplineWidget)"

Sources/Widgets/Widgets3D/SplineWidget/index.js
```diff
@@ -62,6 +62,7 @@ const DEFAULT_VALUES = {
     key: 'Shift',
     status: 'down',
   },
+  defaultCursor: 'pointer',
```
finetjul: how about having that feature for ALL widgets ?

RemiCecchinato: Not all widgets change the cursor and some widgets might want some more complex systems than just defaultCursor/grabbingCursor (eg. IpmlicitPlaneWidget uses 4 different cursors).

finetjul: @floryst what do you think ?

floryst: Exposing an API to modify the default/dragging/etc. cursors seems like a reasonable idea. However, if this API is added to the widget superclass, then there has to be some way of accommodating various cursor states (e.g. if a widget has special cursors for dragging edges vs corners, etc.) Right now I think it'll be easier for users who want to change cursor state to subclass the widget.

Also, do you want to add a `macro.setGet` in extend() for `defaultCursor`?

RemiCecchinato: Thank you @floryst .

PR #1184, opened 2019-08-20
feat(vtkPlane): Add intersectWithPlane function

mix3d:
Similar to, but not the same as the c++ vtk plane logic, where the c++ version expects a
bound/finite plane and an infinite plane. While the function title might assume being passed another vtkPlane object, I did not build in such function overloading, so the function call expects the normal and origin as separate vectors.

Also updates the documentation for visual clarity and grammar.

With both Plane intersection and Line intersection, I can then build a "3 plane intersection" function.

### DIFF
--- a/Sources/Common/DataModel/Plane/index.js
+++ b/Sources/Common/DataModel/Plane/index.js
@@ -111,6 +111,79 @@ function intersectWithLine(p1, p2, origin, normal) {
   return outObj;
 }
 
+function intersectWithPlane(
+  plane1Origin,
+  plane1Normal,
+  plane2Origin,
+  plane2Normal
+) {
+  const outObj = {
+    intersection: false,
+    l0: [],
+    l1: [],
+    error: null,
+  };
+
+  const cross = [];
+  vtkMath.cross(plane1Normal, plane2Normal, cross);
+  const absCross = cross.map((n) => Math.abs(n));
+
+  // test if the two planes are parallel
+  if (absCross[0] + absCross[1] + absCross[2] < PLANE_TOLERANCE) {
+    // test if disjoint or coincide
+    const v = [];
+    vtkMath.subtract(plane1Origin, plane2Origin, v);
+    if (vtkMath.dot(plane1Normal, v) === 0) {
+      outObj.error = 'coincide';
+    } else {
+      outObj.error = 'disjoint';
+    }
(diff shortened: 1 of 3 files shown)


Sources/Common/DataModel/Plane/index.js
```diff
@@ -111,6 +111,79 @@ function intersectWithLine(p1, p2, origin, normal) {
...
+  vtkMath.cross(plane1Normal, plane2Normal, cross);
+  const absCross = cross.map((n) => Math.abs(n));
+
+  // test if the two planes are parallel
+  if (absCross[0] + absCross[1] + absCross[2] < PLANE_TOLERANCE) {
+    // test if disjoint or coincide
+    const v = [];
+    vtkMath.subtract(plane1Origin, plane2Origin, v);
+    if (vtkMath.dot(plane1Normal, v) === 0) {
+      outObj.error = 'coincide';
```
mix3d: I am more than happy to change the return values here, if they could be clearer

floryst:
Looks good to me. One suggestion for the error return values would be exported constants like ERR_COINCIDE = 'coincide', so others can safely test against the error field without magic strings.

mix3d:
> Looks good to me. One suggestion for the error return values would be exported constants like ERR_COINCIDE = 'coincide', so others can safely test against the error field without magic strings.

can do.

floryst:
:tada: Oh, and last thing: commits should be formatted as per [semantic-release](https://github.com/semantic-release/semantic-release) commit messages.

mix3d:
All of them, even if something as simple as that?

floryst:
Yup, it's for changelog reporting. semantic-release generates our changelogs automatically from properly formatted commits, so commits that aren't properly formatted will not trigger the correct version bump (if the PR doesn't have any correctly formatted commits), and will not show up in the generated changelog.

mix3d:
I see. I just assumed that if the first / main commit had the right formatting, small things like "I fixed a typo" didn't need to be included in the changelog, so the semantic-release stuff wouldn't need to pick it up, especially for a feature that hasn't been merged in yet. That's just my thinking though. 

Did my force-push work out for you?

floryst:
I'll be able to merge this in a short bit, along with the svg widget tweaks I have.

PR #1186, opened 2019-08-20
Misc SVG widget changes

floryst:
- change where SVG layer is inserted (immediately after canvas)
- don't buffer points when calling worldPointsToPixelSpace. This was from an old idea that was scrapped, but the code wasn't changed to reflect that we don't need to buffer anymore.


mix3d:
looking forward to this getting merged

PR #1190, opened 2019-08-26
fix(vtkPlane): Fix overloaded intersection in `intersectWithLine()` r…

### LINKED ISSUE #1189: vtkPlane's intersectWithLine() function return is mathematically confusing
the function is given a line defined by two points, but does not specify that the intersection HAS to happen between those two points for the result to return `intersect:true`.

A line defined by two points extends BEYOND those two points, and can still intersect, but if the intersection is outside of the defined points, you get a return with `intersect:false` but still get the coordinates of the intersection.

I think the function & corresponding return is misleading.

mix3d:
…esult.

See github issue #1189: result.intersect value is used both to say "Line and Plane are parallel
(false)" and "Plane intersects line between the line's provided points (true / false)". So the
latter intent is extracted into a new variable `result.betweenPoints`

BREAKING CHANGE: return result is changed if the line intersects the plane but outside of the
provided points. Previously would return intersect:false, but now intersect:true and
betweenPoints:false

fix #1189

### DIFF
--- a/Sources/Common/DataModel/Plane/api.md
+++ b/Sources/Common/DataModel/Plane/api.md
@@ -5,91 +5,110 @@ projecting points onto a plane, evaluating the plane equation, and
 returning plane normal.
 
 ### normal (set/get)
+
 Plane normal. Plane is defined by point and normal. Default is [0.0, 0.0, 1.0].
 
 ### origin (set/get)
+
 Point through which plane passes. Plane is defined by point and normal.
 Default is [0.0, 0.0, 0.0].
 
 ## Methods
 
 ### evaluateFunction(x)
+
 Evaluate plane equation for point x.
 Accepts both an array point representation and individual xyz arguments.
+
 ```js
 plane.evaluateFunction([0.0, 0.0, 0.0]);
 plane.evaluateFunction(0.0, 0.0, 0.0);
 ```
 
 ### evaluateGradient(xyz)
+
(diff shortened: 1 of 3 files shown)


finetjul:
@jourdain I have another "breaking change" MR about to be mergeable. I'm trying to get it ready asap

finetjul:
@jourdain What do you mean by "bundle at the same time"? Do you mean merge both MR at the same time (within a few seconds) or create a unique MR that contains both MR?

For information, in addition to my new comments to the issue #1189  I also left comments on this MR but apparently in the wrong commit: See https://github.com/mix3d/vtk-js/commit/77c4ae7a4e797141602b23471ed44316a3ec6a35#comments 

Therefore I am not feeling confident to merge this current MR as is.

finetjul:
It is not exactly the same behavior as in VTK C++.

In VTK C++, it returns `intersection=false` when line intersects plan when t<0 or t>1. Here, in VTK.JS, it will return `intersection=true` when t<0 or t>1.

So here is a summary of differences for `intersection`:

line is parallel to plane: 
   VTK.JS: false
   VTK C++: false

line intersects between p0 and p1: 
   VTK.JS: true
   VTK C++: true

line intersects outside p0 and p1: 
   VTK.JS: true
   VTK C++: false

finetjul:
Fair enough :-)

Issue #1207, opened 2019-09-13
Get strokeBuffer upon calling vtkPaintFilter's publicAPI.endStroke()


JamesAPetts:
Hey all,

When `publicAPI.endStroke()` is called on the paint widget, is there an easy way to retrieve the `strokeBuffer` from an external application? I'm trying to marry `cornerstoneTools` and `vtk-js`, and `cornerstoneTools` needs to know which 2D slices of the labelmap volume were painted on. This would be really easy to do if we had the `strokeLabelMap`.

It appears the `strokeBuffer` is used to calculate the diff and store history, then falls out of scope when `endStroke()` terminates, would anyone be opposed to returning the `strokeBuffer` from the `endStroke()` method? Is there a neater `vtk-js`-esque way to do this?

Thanks!

floryst:
You're right that there is currently no way to get the stroke buffer. Having it does seem useful, so right now I'm either thinking we return the `strokeBuffer` from `publicAPI.endStroke()`, or we have a property like `recentStrokeBuffer` which only stores the most recent stroke buffer.

JamesAPetts:
Hey Forrest,

Either would be functionally useful. Since the state is tracked internally, this will mostly be used for interfacing. I think returning the `strokeBuffer` is sufficient, as it can then be garbage collected if unused?

Also `recentStrokeBuffer` would have to be recalculated on an undo/redo.

I'm not near a computer right now, but can submit a PR later.

Cheers.

floryst:
Agreed, I would lean towards returning a `strokeBuffer`, since operations would be more concerned with handling strokes after a stroke occurred.

PR #1211, opened 2019-09-17
feat(Sources/Filters/General/PaintFilter/index.js): Return strokeBuffer promise from paint filter.

JamesAPetts:
@floryst , this PR just returns a promise from `endStroke` that resolves to the `strokeBuffer`.

re #1207


floryst:
Just some housekeeping: you can squash the eslint rule commit into your main commit. No need to record eslint style commits.

JamesAPetts:
Sure thing, ready when you are :).

PR #1212, opened 2019-09-19
WIP: feat(vtkDracoReader): Add Draco reader

boucaud:
This pull requests adds vtkDracoReader, a reader for (triangle) meshes that were compressed using Google's Draco library.

The reader decodes the draco file, and provides the decoded mesh as vtkPolyData. Supports normals and texture coordinates.

This PR has a dependency to the draco3d package. I'm not certain what the best course of action is to add this dependency.

Draco: https://github.com/google/draco
Note: I lack experience with both JS and vtk.js, so any feedback is welcome.

### DIFF
--- a/Sources/IO/Geometry/DracoReader/example/index.js
+++ b/Sources/IO/Geometry/DracoReader/example/index.js
@@ -1,0 +1,60 @@
+import 'vtk.js/Sources/favicon';
+
+import vtkActor from 'vtk.js/Sources/Rendering/Core/Actor';
+import vtkFullScreenRenderWindow from 'vtk.js/Sources/Rendering/Misc/FullScreenRenderWindow';
+import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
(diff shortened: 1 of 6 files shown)


jourdain:
Thanks @boucaud for contributing a Draco reader. 

I'll try it locally since our automated system does not seems to like your contribution. Also, we may need to adjust how the draco library is imported or loaded to make it somewhat optional inside vtk.js as it is not small.

By any chance do you have some sample files that I could use to try it out?

A quick glimpse at the code let me believe that your code is reasonable and should be easily integrated once we figure out how to handle the Draco dependency.

Thanks again for your contribution

boucaud:
Here is an archive containing some Draco encoded models, along with their originals for comparison. Please ask me if you need more specific data.
[draco samples.zip](https://github.com/Kitware/vtk-js/files/3631994/draco.samples.zip)

Another issue with the draco3d lib is that it also contains code for the encoder, which is not needed at all for this.

PR #1215, opened 2019-09-20
Better size handling

jourdain:
@agirault that's for you...

### DIFF
--- a/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
@@ -119,7 +119,7 @@ function vtkMouseRangeManipulator(publicAPI, model) {
     // Get proper pixel range used by viewport in rw size space
     const size = glRenderWindow.getViewportSize(renderer);
     // rescale size to match mouse event position
-    model.size = size.map((v) => v * ratio);
+    model.containerSize = size.map((v) => v * ratio);
   };
 
   //-------------------------------------------------------------------------
@@ -132,12 +132,14 @@ function vtkMouseRangeManipulator(publicAPI, model) {
     }
 
     if (model.horizontalListener) {
-      const dxNorm = (position.x - model.previousPosition.x) / model.size[0];
+      const dxNorm =
+        (position.x - model.previousPosition.x) / model.containerSize[0];
       const dx = scaleDeltaToRange(model.horizontalListener, dxNorm);
       processDelta(model.horizontalListener, dx);
     }
     if (model.verticalListener) {
-      const dyNorm = (position.y - model.previousPosition.y) / model.size[1];
+      const dyNorm =
+        (position.y - model.previousPosition.y) / model.containerSize[1];
       const dy = scaleDeltaToRange(model.verticalListener, dyNorm);
       processDelta(model.verticalListener, dy);
     }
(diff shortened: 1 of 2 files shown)


Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -170,6 +173,16 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
   };
 
+  publicAPI.getContainer = () => model.el;
+
+  publicAPI.getContainerSize = () => {
+    if (!model.containerSize && model.el) {
+      const { width, height } = model.el.getBoundingClientRect();
+      model.containerSize = [width, height];
+    }
+    return model.containerSize || [300, 300];
```
agirault: is the `[300, 300]` defined anywhere else?

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -112,6 +112,14 @@ function vtkMouseRangeManipulator(publicAPI, model) {
   //-------------------------------------------------------------------------
   publicAPI.onButtonDown = (interactor, renderer, position) => {
     model.previousPosition = position;
+    const glRenderWindow = interactor.getView();
+    // Ratio is the dom size vs renderwindow size
+    const ratio =
+      glRenderWindow.getContainerSize()[0] / glRenderWindow.getSize()[0];
```
agirault: Just curious how the container and the render window could have a different ratio? Is the mouse event position scaled by that ratio also? Can't we use the container size right away?

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -112,6 +112,14 @@ function vtkMouseRangeManipulator(publicAPI, model) {
...
+    // Get proper pixel range used by viewport in rw size space
+    const size = glRenderWindow.getViewportSize(renderer);
+    // rescale size to match mouse event position
+    model.size = size.map((v) => v * ratio);
```
agirault: `model.containerSize` to be more explicit? `size` itself won't mean much in for a `MouseRangeManipulator`... but it's not exposed so maybe that's fine.

Sources/Rendering/OpenGL/RenderWindow/index.js
agirault: Actually, why do we return `[300,300]` when `model.el === undefined` and not just return `undefined`?

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
jourdain: The mouse will be screen-based, but yes the container and the window can have some weird ratio. So no you can not use the container right away.

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
jourdain: Sure what ever

Sources/Rendering/OpenGL/RenderWindow/index.js
jourdain: Because we need to have valid numbers no matter what.

agirault: Why? `getContainer` can return something `undefined`. If there is no valid container, should we have valid numbers when requesting a size?

jourdain: Using the actual canvas size when no container is provided is even better.

agirault: Isn't that even more dangerous? We think we get a valid container size, but it's actually a canvas size which is not from the same space.

jourdain: No, in fact it is the perfect way to handle it since you have no container size constraint and will naturally have a 1 to 1 mapping. To me, that makes perfect sense. 
Remember we are dealing with a corner case.

agirault: okey!

PR #1216, opened 2019-09-23
Draco reader

jourdain:


jourdain:
Replacing https://github.com/Kitware/vtk-js/pull/1212 but need to find a way to dynamically fetch Draco decoder when used as it adds 800KB to the package otherwise.

boucaud:
I was thinking of a solution like this one:
```
    let dracoDecoder;
    try {
      // eslint-disable-next-line global-require, import/no-unresolved
      dracoDecoder = require('draco3d/draco_decoder_nodejs');
    } catch (e) {
      vtkErrorMacro(
        'Could not import draco. Please install the draco3d package.'
      );
    }
    if (typeof dracoDecoder !== 'undefined') {
      // Decode the file
    }
```

It does work, but it does not seem ideal.

boucaud:
Thanks @jourdain !

PR #1224, opened 2019-10-10
feat(Camera): add Thickness functions

mix3d:
Add unimplemented functions, both matching vtk c++ function and a new thickness function that seems more useful. Also update docs for clarity and style.

#Hacktober


jourdain:
LGTM @martinken, I let you merge if you are good with it.

mix3d:
One note is that I refactored-out the model's thickness parameter, instead calculating on the fly from the clipping range. Less chance of things getting out of sync, and we don't have to save the thickness value separately when setting the clipping range

PR #1228, opened 2019-10-16
feat(PaintFilter): improve and factorize PaintFilter's labelMap modification methods

boucaud:
While deriving a class from vtkPaintFilter to add new operations, users can not access the history variable, which stops them from being able to use undo/redo functionality in their new operations.

This exposes the filter's history to child classes.

### DIFF
--- a/Sources/Filters/General/PaintFilter/index.js
+++ b/Sources/Filters/General/PaintFilter/index.js
@@ -30,6 +30,19 @@ function vtkPaintFilter(publicAPI, model) {
     history.labels = [];
   }
 
+  function pushToHistory(snapshot, label) {
+    // Clear any "redo" info
+    const spliceIndex = history.index + 1;
+    const spliceLength = history.snapshots.length - history.index;
+    history.snapshots.splice(spliceIndex, spliceLength);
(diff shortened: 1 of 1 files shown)


boucaud:
@finetjul

finetjul:
LGTM
@floryst  WDYT?

floryst:
LGTM, since it can be useful. I assume this involves new operations that operate on the history and do more than just basic undo/redo? I ask since subclasses that are only interested in basic undo/redo still have access to the undo/redo methods.

boucaud:
> I assume this involves new operations that operate on the history and do more than just basic undo/redo?

Indeed. To be more specific, these functions don't use the filter's worker/workerPromise (no startStroke/endStroke). They set a new labelMap as a result, and compute the diff between new and old labelMap to append to the history.

boucaud:
Reverted the previous changes, factorized mask operation, history  handling, added a new method to directly change the labelMap.

@finetjul @floryst WDYT ?

finetjul:
LGTM

floryst:
LGTM

Issue #843, opened 2018-08-21
camera.setParallelProjection(true) issue


tanxu1120:
I just need to set camera.setParallelProjection(true), the volume will not be displayed.
Please help me find out what the problem is.
Thanks!!!

jourdain:
Did you try to reset the camera when in parallel projection?

tanxu1120:
@jourdain Yes, I did that.However, it still cannot be displayed.

mix3d:
following up on this, we're trying to use parallel projection and it's still not working.

I've tried the above methods, played with clipping points and parallelScale, but nothing seems to work.

in one case, I managed to get some things to render, but moving the camera (in a method needed for our MPR approach, moving the focal plane and camera distance together through the volume, with a small clipping plane) results in VERY strange cropping.

Looked through the code, and there doesn't seem to be any unit tests for the camera at all, so I'm not totally surprised.

How can we help debug this?

mix3d:
example of strange cropping issue:
![perspective progress](https://user-images.githubusercontent.com/1474137/67121542-9813b300-f1a0-11e9-8428-bf8d1ae8395b.gif)

The rendered volume slice stays the same dimensions/scale, but moving the camera closer/farther into the volume crops unusually. If I change the camera angle to one not defaulted to a cartesian coord, the clipping becomes even stranger.

mix3d:
Update: I made a small codepen trying to reproduce the problem. Seems to work in this very small case, so I'm going to keep adding to it until it breaks. 

One thing I've notices is that if you don't `resetCamera()`and the camera is significantly moved from the default view when switching to parallelProjection, the cone does not render (or renders off screen)

The above examples are using volumes that are `256 x 256 x ~100-500` in world dimensions, which may also be playing a part in the weirdness if most examples are in much smaller scale units.

mix3d:
My "hack" solution for this is to use a camera viewAngle of 1 degree. This approximates a parallel view well enough, but doesn't solve the core problem.

mix3d:
@jourdain @martinken any thoughts / is this on your radar? Perhaps a bug label at least?

martinken:
I'll try takjng a quick look. Should be a very small fix once I locate the issue.

JamesAPetts:
Hey @martinken ,

I've found the issue, I'll get a PR up soon!

martinken:
Awesome @JamesAPetts ! Tag me on the PR if you want and I'll do a review.

[...]

PR #1230, opened 2019-10-22
fix: 🐛 Fix parallel projection rendering using the VolumeMapper

JamesAPetts:
Fix the setCameraShaderParameters function of the VolumeMapper.

Adds tests and examples also.

Closes: #843


JamesAPetts:
@martinken  here is the fix for #843. The projection back onto the vc plane was incorrect for parallel projections for the VolumeMapper.

As with all these things it was a painstakingly difficult task to find this one line fix: https://github.com/Kitware/vtk-js/blob/fc914d3d75d5b6d67a26797cfa2a96ece2abea93/Sources/Rendering/OpenGL/VolumeMapper/index.js#L454

The rest of the PR is test cases, as its quite important this keeps working going forwards :).

P.S. @swederik  and I _think_ this is the correct solution, but feel free to push back/correct us if you think that `t` should be something else.

JamesAPetts:
Example in `react-vtkjs-viewport` addressing some of the issues in #843 :

https://media.giphy.com/media/KztKRqcEraorVHt1gE/giphy.gif

martinken:
Yes! I think you found the culprit which is the hard part. Can you try something? What about directly using pos for parallel projection. If my mathing is working I think that is the fix.  It expands the bounds out a tad farther I think. Something like this

```
    for (let i = 0; i < 8; ++i) {
      vec3.set(
        pos,
        bounds[i % 2],
        bounds[2 + (Math.floor(i / 2) % 2)],
        bounds[4 + Math.floor(i / 4)]
      );
      vec3.transformMat4(pos, pos, model.modelToView);
      if (!cam.getParallelProjection()) {
        vec3.normalize(dir, pos);

        // now find the projection of this point onto a
        // nearZ distance plane. Since the camera is at 0,0,0
        // in VC the ray is just t*pos and
        // t is -nearZ/dir.z
        // intersection becomes pos.x/pos.z
        const t = -crange[0] / pos[2];
        vec3.scale(pos, dir, t);
      }
      // now convert to DC
      vec3.transformMat4(pos, pos, keyMats.vcdc);

      dcxmin = Math.min(pos[0], dcxmin);
      dcxmax = Math.max(pos[0], dcxmax);
      dcymin = Math.min(pos[1], dcymin);
      dcymax = Math.max(pos[1], dcymax);
    }
```

JamesAPetts:
Let me give it a shot, thanks!

JamesAPetts:
That makes a lot more sense, and is a lot easier to read. Works great, let me just update the image for the karma tests :).

JamesAPetts:
@martinken Updated the method to your suggestion and also the subtly different test image 👍 .

martinken:
Thanks @JamesAPetts

JamesAPetts:
And thanks for the speedy review @martinken !

Issue #1217, opened 2019-09-25
Visualization precision issues with datasets that have 'large coordinates'


danmindru:
Hi there,

When trying to visualize a vtp with value ranges from 6200475.180468361 to 6200562.503912979, I've noticed that the visualization does not accurately represent the data.

For example, the vtp (see attached zip) is rendered in the vtk.js GeometryViewer example as:
![vtkjs-geometry-viewer](https://user-images.githubusercontent.com/1515742/65590564-4a35d180-df8b-11e9-840f-57d1ac9b4d1a.png)

Paraview renders it similarly: 
![paraview](https://user-images.githubusercontent.com/1515742/65590605-5d48a180-df8b-11e9-9b18-ec1ec29ac58f.png)

However, by visualizing the same data in i.e. QGIS, one can see that the bottom border is not accurately represented in either vtk.js or Paraview:
![qgis](https://user-images.githubusercontent.com/1515742/65590706-85380500-df8b-11e9-84dd-78ee3188ed2c.png)

[Used Data.zip](https://github.com/Kitware/vtk-js/files/3651799/Data.zip)

Are there any ways to increase the accuracy of the visualization that come to mind?
Thanks for the help,
Dan

-------------------------

About the data:
The data represents geographical coordinates in a given projection. The precision is necessary because the data is sometimes shown on top of a map. Any transform to this coordinate with either misrepresent the geographical data or effectively reproject the data.

jourdain:
Using ParaView 5.6 or 5.7, I can not reproduce the precision issue while I agree that on vtk.js the precision issue arises and you can see the geometry jumping around.

@martinken can we do similar things on vtk.js to properly handle double precision coordinates like it is done in VTK or ParaView without manually shifting the coordinates?

martinken:
Yes we can

martinken:
To be clear that is a "we could code the same technique", right now it isn't implemented though

martinken:
Latest PV should handle those coordinates fine (as you are seeing Seb)

danmindru:
Thanks for getting back so quickly. It sounds promising!
To clarify, would you prefer that this is pushed / funded by a customer?
(that's fine, just want to make sure that is the case 😊)

danmindru:
Hi @boucaud @finetjul,
Upon updating to v11.9.0 seems to improve the original issue, however I'm seeing new scaling issues when visualizing i.e. 2+ actors that have shifted coordinates.

`Example #1`:
[...]

PR #1231, opened 2019-10-22
feat: Add coordinate shift/scale to PolyDataMapper

boucaud:
This implements coordinate shifting/scaling in vtk-js, similarly to the way it is currently implemented in VTK.

This fixes issues with large coordinates:

Before:
image

After: 
image

re: #1217

Attached: [test data](https://github.com/Kitware/vtk-js/files/3756779/TestMeshNotFollowingGeometryFix-mesh.vtp.tar.gz)  from #1217

### DIFF
--- a/Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
+++ b/Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
@@ -1,9 +1,41 @@
+import { mat4, quat, vec3 } from 'gl-matrix';
+
 import macro from 'vtk.js/Sources/macro';
 import vtkBufferObject from 'vtk.js/Sources/Rendering/OpenGL/BufferObject';
 import { ObjectType } from 'vtk.js/Sources/Rendering/OpenGL/BufferObject/Constants';
 import { Representation } from 'vtk.js/Sources/Rendering/Core/Property/Constants';
 
 const { vtkDebugMacro, vtkErrorMacro } = macro;
+
+// ----------------------------------------------------------------------------
+// Static functions
+// ----------------------------------------------------------------------------
+
+function computeInverseShiftAndScaleMatrix(coordShift, coordScale) {
+  const inverseScale = vec3.create();
+  vec3.inverse(inverseScale, coordScale);
+
+  const matrix = mat4.create();
+  mat4.fromRotationTranslationScale(
+    matrix,
+    quat.create(),
+    coordShift,
+    inverseScale
+  );
+
+  return matrix;
+}
+
+function shouldApplyCoordShiftAndScale(coordShift, coordScale) {
+  if (coordShift === null || coordScale === null) {
+    return false;
+  }
+
+  return !(
+    vec3.exactEquals(coordShift, [0, 0, 0]) &&
+    vec3.exactEquals(coordScale, [1, 1, 1])
+  );
+}
 
 // ----------------------------------------------------------------------------
 // vtkOpenGLCellArrayBufferObject methods
@@ -200,13 +232,60 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
     let vboidx = 0;
     let ucidx = 0;
 
+    // Find out if shift scale should be used
+    // Compute squares of diagonal size and distance from the origin
+    let diagSq = 0.0;
+    let distSq = 0.0;
+    for (let i = 0; i < 3; ++i) {
+      const range = options.points.getRange(i);
+
+      const delta = range[1] - range[0];
+      diagSq += delta * delta;
+
+      const distShift = 0.5 * (range[1] + range[0]);
+      distSq += distShift * distShift;
+    }
+
+    const useShiftAndScale =
+      diagSq > 0 &&
+      (Math.abs(distSq) / diagSq > 1.0e6 || // If data is far from the origin relative to its size
+      Math.abs(Math.log10(diagSq)) > 3.0 || // If the size is huge when not far from the origin
+        (diagSq === 0 && distSq > 1.0e6)); // If data is a point, but far from the origin
+
+    if (useShiftAndScale) {
+      // Compute shift and scale vectors
+      const coordShift = vec3.create();
+      const coordScale = vec3.create();
+      for (let i = 0; i < 3; ++i) {
+        const range = options.points.getRange(i);
+        const delta = range[1] - range[0];
+
+        coordShift[i] = 0.5 * (range[1] + range[0]);
+        coordScale[i] = delta > 0 ? 1.0 / delta : 1.0;
+      }
+      publicAPI.setCoordShiftAndScale(coordShift, coordScale);
+    } else if (model.coordShiftAndScaleEnabled === true) {
+      // Make sure to reset
+      model.setCoordShiftAndScale(null, null);
+    }
+
     addAPoint = function addAPointFunc(i) {
       // Vertices
       pointIdx = i * 3;
 
-      packedVBO[vboidx++] = pointData[pointIdx++];
-      packedVBO[vboidx++] = pointData[pointIdx++];
-      packedVBO[vboidx++] = pointData[pointIdx++];
+      if (!model.coordShiftAndScaleEnabled) {
+        packedVBO[vboidx++] = pointData[pointIdx++];
+        packedVBO[vboidx++] = pointData[pointIdx++];
+        packedVBO[vboidx++] = pointData[pointIdx++];
+      } else {
+        // Apply shift and scale
+        packedVBO[vboidx++] =
+          (pointData[pointIdx++] - model.coordShift[0]) * model.coordScale[0];
+        packedVBO[vboidx++] =
+          (pointData[pointIdx++] - model.coordShift[1]) * model.coordScale[1];
+        packedVBO[vboidx++] =
+          (pointData[pointIdx++] - model.coordShift[2]) * model.coordScale[2];
+      }
 
       if (normalData !== null) {
         if (options.haveCellNormals) {
@@ -261,16 +340,52 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
     return cellCount;
   };
 
-  publicAPI.setCoordShiftAndScaleMethod = (shiftScaleMethod) => {
-    vtkErrorMacro('coordinate shift and scale not yet implemented');
-  };
+  publicAPI.setCoordShiftAndScale = (coordShift, coordScale) => {
+    if (
+      coordShift !== null &&
+      (coordShift.constructor !== Float32Array || coordShift.length !== 3)
+    ) {
+      vtkErrorMacro('Wrong type for coordShift, expected vec3 or null');
+      return;
+    }
 
-  publicAPI.setCoordShift = (shiftArray) => {
-    vtkErrorMacro('coordinate shift and scale not yet implemented');
-  };
+    if (
+      coordScale !== null &&
+      (coordScale.constructor !== Float32Array || coordScale.length !== 3)
+    ) {
+      vtkErrorMacro('Wrong type for coordScale, expected vec3 or null');
+      return;
+    }
 
-  publicAPI.setCoordScale = (scaleArray) => {
-    vtkErrorMacro('coordinate shift and scale not yet implemented');
+    if (
+      model.coordShift === null ||
+      coordShift === null ||
+      !vec3.equals(coordShift, model.coordShift)
+    ) {
+      model.coordShift = coordShift;
+    }
+
+    if (
+      model.coordScale === null ||
+      coordScale === null ||
+      !vec3.equals(coordScale, model.coordScale)
+    ) {
+      model.coordScale = coordScale;
+    }
+
+    model.coordShiftAndScaleEnabled = shouldApplyCoordShiftAndScale(
+      model.coordShift,
+      model.coordScale
+    );
+
+    if (model.coordShiftAndScaleEnabled) {
+      model.inverseShiftAndScaleMatrix = computeInverseShiftAndScaleMatrix(
+        model.coordShift,
+        model.coordScale
+      );
+    } else {
+      model.inverseShiftAndScaleMatrix = null;
+    }
   };
 }
 
@@ -290,6 +405,10 @@ const DEFAULT_VALUES = {
   colorComponents: 0,
   tcoordBO: null,
   customData: [],
+  coordShift: null,
+  coordScale: null,
+  coordShiftAndScaleEnabled: false,
+  inverseShiftAndScaleMatrix: null,
 };
 
 // ----------------------------------------------------------------------------
@@ -312,6 +431,13 @@ export function extend(publicAPI, model, initialValues = {}) {
     'colorOffset',
     'colorComponents',
     'customData',
+  ]);
+
+  macro.get(publicAPI, model, [
+    'coordShift',
+    'coordScale',
+    'coordShiftAndScaleEnabled',
+    'inverseShiftAndScaleMatrix',
   ]);
 
   // Object specific methods
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -358,6 +358,13 @@ function vtkOpenGLImageMapper(publicAPI, model) {
     const keyMats = model.openGLCamera.getKeyMatrices(ren);
     mat4.multiply(model.imagemat, keyMats.wcdc, model.imagemat);
 
+    if (cellBO.getCABO().getCoordShiftAndScaleEnabled()) {
+      const inverseShiftScaleMat = cellBO
+        .getCABO()
+        .getInverseShiftAndScaleMatrix();
+      mat4.multiply(model.imagemat, model.imagemat, inverseShiftScaleMat);
+    }
+
     program.setUniformMatrix('MCDCMatrix', model.imagemat);
   };
 
--- a/Sources/Rendering/OpenGL/PolyDataMapper/index.js
+++ b/Sources/Rendering/OpenGL/PolyDataMapper/index.js
@@ -8,7 +8,6 @@ import vtkOpenGLTexture from 'vtk.js/Sources/Rendering/OpenGL/Texture';
 import vtkProperty from 'vtk.js/Sources/Rendering/Core/Property';
 import vtkShaderProgram from 'vtk.js/Sources/Rendering/OpenGL/ShaderProgram';
 import vtkViewNode from 'vtk.js/Sources/Rendering/SceneGraph/ViewNode';
-
 import vtkPolyDataVS from 'vtk.js/Sources/Rendering/OpenGL/glsl/vtkPolyDataVS.glsl';
 import vtkPolyDataFS from 'vtk.js/Sources/Rendering/OpenGL/glsl/vtkPolyDataFS.glsl';
 
@@ -29,7 +28,6 @@ const { Representation, Shading } = vtkProperty;
 const { ScalarMode } = vtkMapper;
 const { Filter, Wrap } = vtkOpenGLTexture;
 const { vtkErrorMacro } = macro;
-
 const StartEvent = { type: 'StartEvent' };
 const EndEvent = { type: 'EndEvent' };
 
@@ -1490,6 +1488,15 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
     }
   };
 
+  function safeMat4Multiply(matrixArray) {
+    return matrixArray.reduce((res, matrix, index) => {
+      if (index === 0) {
+        return matrix ? mat4.copy(res, matrix) : mat4.identity(res);
+      }
+      return matrix ? mat4.multiply(res, res, matrix) : res;
+    }, model.tmpMat4);
+  }
+
   publicAPI.setCameraShaderParameters = (cellBO, ren, actor) => {
     const program = cellBO.getProgram();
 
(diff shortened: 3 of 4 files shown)


boucaud:
@finetjul

Sources/Rendering/OpenGL/CellArrayBufferObject/Constants.js
```diff
@@ -0,0 +1,10 @@
+export const CoordShiftAndScaleMethods = {
+  Disable: 0,
+  Auto: 1,
+  AlwaysAuto: 2,
+  Manual: 3,
+};
+
```
martinken: Can we just always have it on. Makes the code simpler. In C++ we have a path where the float point array can be directly uploaded to the VBO so auto has some advantage there as if there is no scaling we can take that fast path. But in vtkjs we always are copying the points into a new array.

boucaud: I just pushed changes to alwayshave it on.

boucaud:
@finetjul @martinken please review.

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -200,13 +232,71 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
...
+    ) {
+      useShiftAndScale = true;
+    } else if (model.coordShiftAndScaleEnabled === true) {
+      // Make sure to reset
+      model.coordShiftAndScaleEnabled = false;
+      model.coordScale = null;
+      model.coordShift = null;
+    }
+
+    if (useShiftAndScale === true) {
```
finetjul: if (useShiftAndScale)

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -200,13 +232,71 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
     let vboidx = 0;
     let ucidx = 0;
 
+    // Find out if shift scale should be used
+    let useShiftAndScale = false;
```
finetjul: move declaration after `for` loop

finetjul:
```
const useShiftAndScale = diagSq > 0 &&
      (Math.abs(distSq) / diagSq > 1.0e6 || // If data is far from the origin relative to its size
      Math.abs(Math.log10(diagSq)) > 3.0 || // If the size is huge when not far from the origin
        (diagSq === 0 && distSq > 1.0e6)) // If data is a point, but far from the origin
```

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -200,13 +232,71 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
...
+      for (let i = 0; i < 3; ++i) {
+        const range = options.points.getRange(i);
+        const delta = range[1] - range[0];
+
+        shift[i] = 0.5 * (range[1] + range[0]);
+        scale[i] = delta > 0 ? 1.0 / delta : 1.0;
+      }
+
+      publicAPI.setCoordShift(shift);
+      publicAPI.setCoordScale(scale);
```
finetjul: it would seem to be more efficient to set both at once, create a `setCoordShiftAndScale` ?

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -200,13 +232,71 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
...
+      (Math.abs(distSq) / diagSq > 1.0e6 || // If data is far from the origin relative to its size
+      Math.abs(Math.log10(diagSq)) > 3.0 || // If the size is huge when not far from the origin
+        (diagSq === 0 && distSq > 1.0e6)) // If data is a point, but far from the origin
+    ) {
+      useShiftAndScale = true;
+    } else if (model.coordShiftAndScaleEnabled === true) {
+      // Make sure to reset
+      model.coordShiftAndScaleEnabled = false;
+      model.coordScale = null;
+      model.coordShift = null;
```
finetjul: no need to reset model.inverseShiftAndScaleMatrix  ?

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -261,16 +351,44 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
     return cellCount;
   };
 
-  publicAPI.setCoordShiftAndScaleMethod = (shiftScaleMethod) => {
-    vtkErrorMacro('coordinate shift and scale not yet implemented');
-  };
-
   publicAPI.setCoordShift = (shiftArray) => {
-    vtkErrorMacro('coordinate shift and scale not yet implemented');
+    if (shiftArray.constructor !== Float32Array || shiftArray.length !== 3) {
```
finetjul: would be good to support a null shiftArray ?

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -358,7 +358,15 @@ function vtkOpenGLImageMapper(publicAPI, model) {
     const keyMats = model.openGLCamera.getKeyMatrices(ren);
     mat4.multiply(model.imagemat, keyMats.wcdc, model.imagemat);
 
-    program.setUniformMatrix('MCDCMatrix', model.imagemat);
+    if (cellBO.getCABO().getCoordShiftAndScaleEnabled() === true) {
```
finetjul: if (cellBO.getCABO().getCoordShiftAndScaleEnabled())

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -358,7 +358,15 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
+      const inverseShiftScaleMat = cellBO
+        .getCABO()
+        .getInverseShiftAndScaleMatrix();
+      mat4.multiply(model.tmpMat4, model.imagemat, inverseShiftScaleMat);
+      program.setUniformMatrix('MCDCMatrix', model.tmpMat4);
+    } else {
+      program.setUniformMatrix('MCDCMatrix', model.imagemat);
+    }
```
finetjul: no need for tmpMat4:
```
if (cellBO.getCABO().getCoordShiftAndScaleEnabled()) {
   const inverseShiftScaleMat = cellBO
        .getCABO()
        .getInverseShiftAndScaleMatrix();
      mat4.multiply(model.imagemat, model.imagemat, inverseShiftScaleMat);
}
program.setUniformMatrix('MCDCMatrix', model.imagemat);
```

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1501,38 +1499,79 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
     const camm = model.openGLCamera.getKeyMatrixTime().getMTime();
     const progm = program.getLastCameraMTime();
 
-    if (progm !== camm) {
-      if (actor.getIsIdentity()) {
-        program.setUniformMatrix('MCDCMatrix', keyMats.wcdc);
-        if (program.isUniformUsed('MCVCMatrix')) {
-          program.setUniformMatrix('MCVCMatrix', keyMats.wcvc);
+    if (cellBO.getCABO().getCoordShiftAndScaleEnabled() === true) {
```
finetjul: f (cellBO.getCABO().getCoordShiftAndScaleEnabled())

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1501,38 +1499,79 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
...
+        }
+        mat4.identity(model.tmpMat4);
+        mat4.multiply(model.tmpMat4, keyMats.wcdc, actMats.mcwc);
+        program.setUniformMatrix('MCDCMatrix', model.tmpMat4);
+        if (program.isUniformUsed('MCVCMatrix')) {
+          mat4.multiply(model.tmpMat4, keyMats.wcvc, actMats.mcwc);
+          program.setUniformMatrix('MCVCMatrix', model.tmpMat4);
+        }
+        // reset the cam mtime as actor modified the shader values
+        program.setLastCameraMTime(0);
```
finetjul: please factorize the code, that would be too hard to maintain otherwise...

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -11,29 +11,29 @@ const { vtkDebugMacro, vtkErrorMacro } = macro;
 // Static functions
 // ----------------------------------------------------------------------------
 
-function computeInverseShiftAndScaleMatrix(shiftArray, scaleArray) {
+function computeInverseShiftAndScaleMatrix(arrayShift, arrayScale) {
```
finetjul: now that I understood the whole thing, it is really a "shift" array of size 3 and not really the "shifts" of the (point data) array.  Sorry about that.

Maybe "shifts" and "scales"  would be better anyway :-)

boucaud:
@finetjul does this look ok ?

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -200,13 +231,70 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
...
+      model.coordShiftAndScaleEnabled = false;
+      model.coordScale = null;
+      model.coordShift = null;
+      model.inverseShiftAndScaleMatrix = null;
+    }
+
+    if (useShiftAndScale) {
+      // Compute shift and scale vectors
+      const shift = vec3.create();
+      const scale = vec3.create();
```
finetjul: shifts, scales

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -261,16 +349,37 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
...
 
-  publicAPI.setCoordScale = (scaleArray) => {
-    vtkErrorMacro('coordinate shift and scale not yet implemented');
+    if (model.coordScale === null || !vec3.equals(scales, model.coordScale)) {
+      model.coordScale = scales;
+    }
+
+    model.coordShiftAndScaleEnabled = computeIsCoordShiftAndScaleEnabled(
+      shifts,
+      model.coordScale
```
finetjul: `model.coordShifts, model.coordScale`

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -1,10 +1,41 @@
...
+    matrix,
+    quat.create(),
+    shifts,
+    inverseScale
+  );
+
+  return matrix;
+}
+
+function computeIsCoordShiftAndScaleEnabled(shifts, scales) {
```
finetjul: computeIsCoordShiftAndScaleEnabled -> isCoordShiftAndScaleEnabled

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -200,13 +231,70 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
...
+      for (let i = 0; i < 3; ++i) {
+        const range = options.points.getRange(i);
+        const delta = range[1] - range[0];
+
+        shift[i] = 0.5 * (range[1] + range[0]);
+        scale[i] = delta > 0 ? 1.0 / delta : 1.0;
+      }
+
+      publicAPI.setCoordShiftAndScale(shift, scale);
+      model.inverseShiftAndScaleMatrix = computeInverseShiftAndScaleMatrix(
```
finetjul: shouldn't computeInverseShiftAndScaleMatrix be called by setCoordShiftAndScale() ?

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -200,13 +231,70 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
...
+      (Math.abs(distSq) / diagSq > 1.0e6 || // If data is far from the origin relative to its size
+      Math.abs(Math.log10(diagSq)) > 3.0 || // If the size is huge when not far from the origin
+        (diagSq === 0 && distSq > 1.0e6)); // If data is a point, but far from the origin
+
+    if (!useShiftAndScale && model.coordShiftAndScaleEnabled === true) {
+      // Make sure to reset
+      model.coordShiftAndScaleEnabled = false;
+      model.coordScale = null;
+      model.coordShift = null;
+      model.inverseShiftAndScaleMatrix = null;
```
finetjul: Instead I would call `setCoordShiftAndScale(null, null)` that would set all variables automatically

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
finetjul: or `shouldApplyCoordShiftAndScale`

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -261,16 +349,37 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
     return cellCount;
   };
 
-  publicAPI.setCoordShiftAndScaleMethod = (shiftScaleMethod) => {
-    vtkErrorMacro('coordinate shift and scale not yet implemented');
-  };
+  publicAPI.setCoordShiftAndScale = (shifts, scales) => {
+    if (
+      shifts === null ||
```
finetjul: I would allow shifts/scales to be reset to null

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -261,16 +349,37 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
     return cellCount;
   };
 
-  publicAPI.setCoordShiftAndScaleMethod = (shiftScaleMethod) => {
-    vtkErrorMacro('coordinate shift and scale not yet implemented');
-  };
+  publicAPI.setCoordShiftAndScale = (shifts, scales) => {
```
finetjul: You should probably have the same name between `model.coordShift` and local `shifts`. Pick the one you prefer

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1490,6 +1488,17 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
     }
   };
 
+  function multiplyNonNullMatrices(keyMat, actMat, inverseShiftScaleMat) {
```
finetjul: This is a counter intuitive function name because you actually actMat and inverseShiftScaleMat can be null.

Rename to `safeMat4Multiply()` ?

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1501,38 +1510,50 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
...
-      if (program.isUniformUsed('MCVCMatrix')) {
-        mat4.multiply(model.tmpMat4, keyMats.wcvc, actMats.mcwc);
-        program.setUniformMatrix('MCVCMatrix', model.tmpMat4);
+      if (progm !== camm) {
+        if (program.isUniformUsed('cameraParallel')) {
+          program.setUniformi('cameraParallel', cam.getParallelProjection());
+          program.setLastCameraMTime(camm);
+        }
+      } else {
+        program.setLastCameraMTime(0);
```
finetjul: You removed the following comment, is it normal ? 
```
// reset the cam mtime as actor modified the shader values
```

Sources/Rendering/OpenGL/PolyDataMapper/index.js
finetjul: Maybe this function could be more generic e.g. safeMat4Multiply(m1, m2, m3) or even better, take an undefined number of matrices as input and multiply them all

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1490,6 +1488,17 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
     }
   };
 
+  function multiplyNonNullMatrices(keyMat, actMat, inverseShiftScaleMat) {
+    const tmpMat = mat4.clone(keyMat);
```
finetjul: You might want to use model.tmpMat4 to avoid instantiation of a new matrix each time

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1501,38 +1510,50 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
...
-      mat4.identity(model.tmpMat4);
-      mat4.multiply(model.tmpMat4, keyMats.wcdc, actMats.mcwc);
-      program.setUniformMatrix('MCDCMatrix', model.tmpMat4);
-      if (program.isUniformUsed('MCVCMatrix')) {
-        mat4.multiply(model.tmpMat4, keyMats.wcvc, actMats.mcwc);
-        program.setUniformMatrix('MCVCMatrix', model.tmpMat4);
+      if (progm !== camm) {
+        if (program.isUniformUsed('cameraParallel')) {
+          program.setUniformi('cameraParallel', cam.getParallelProjection());
+          program.setLastCameraMTime(camm);
```
finetjul: should be moved outside the isUniformUsed() `if` block

finetjul: Please make sure you did not changed the existing logical path

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1501,38 +1510,50 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
...
-        program.setUniformi('cameraParallel', cam.getParallelProjection());
+    const shiftScaleEnabled = cellBO.getCABO().getCoordShiftAndScaleEnabled();
+    const inverseShiftScaleMatrix = shiftScaleEnabled
+      ? cellBO.getCABO().getInverseShiftAndScaleMatrix()
+      : null;
+
+    const actorIsIdentity = actor.getIsIdentity();
+    const actMats = actorIsIdentity ? null : model.openGLActor.getKeyMatrices();
+
+    if (progm !== camm || !actorIsIdentity) {
```
finetjul: Do I understand correctly that if camera has not been changed and if actor matrice was not identity but just became identity, then no uniform are recomputed ? Is it really what we want ?

BotellaA:
Very interesting PR! :+1: 
I have a question: What appends if you have several objects in the camera? Is the scaling consistent for all the objects? For example, a large surface an a smaller one, both very close but very far from the origin.

Sources/Rendering/OpenGL/PolyDataMapper/index.js
boucaud: Before this change, uniforms were computed in two cases:
* If the camera has been changed and the actor matrix is identity
* If the actor matrix is not identity

This change keeps the same logic as before (updates uniforms if the camera has been changed or the actor matrix is not identity).

In your example, the uniforms would indeed not be recomputed, as would be the case in the previous version. Maybe a fix would be to also update uniforms if the actor's mtime is different to the program's.

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
boucaud: Picked model.coordShift

Sources/Rendering/OpenGL/PolyDataMapper/index.js
boucaud: After discussion, this should be a separate issue.

boucaud:
> Very interesting PR! 
> I have a question: What appends if you have several objects in the camera? Is the scaling consistent for all the objects? For example, a large surface and a smaller one, both very close but very far from the origin.

You would get different shift/scale values for each object, thus different camera matrices for each one.

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -200,13 +232,62 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
...
+      (Math.abs(distSq) / diagSq > 1.0e6 || // If data is far from the origin relative to its size
+      Math.abs(Math.log10(diagSq)) > 3.0 || // If the size is huge when not far from the origin
+        (diagSq === 0 && distSq > 1.0e6)); // If data is a point, but far from the origin
+
+    if (!useShiftAndScale && model.coordShiftAndScaleEnabled === true) {
+      // Make sure to reset
+      model.setCoordShiftAndScale(null, null);
+    }
+
+    if (useShiftAndScale) {
```
finetjul: `else if`

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -358,7 +358,15 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
-    program.setUniformMatrix('MCDCMatrix', model.imagemat);
+    if (cellBO.getCABO().getCoordShiftAndScaleEnabled()) {
+      const inverseShiftScaleMat = cellBO
+        .getCABO()
+        .getInverseShiftAndScaleMatrix();
+      mat4.multiply(model.imagemat, model.imagemat, inverseShiftScaleMat);
+      program.setUniformMatrix('MCDCMatrix', model.imagemat);
+    } else {
+      program.setUniformMatrix('MCDCMatrix', model.imagemat);
+    }
```
finetjul: move setUniformMatrix outside `if` block ?

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1490,6 +1488,17 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
...
 
+  function safeMat4Multiply(matrixArray) {
+    mat4.copy(model.tmpMat4, matrixArray[0]);
+    for (let i = 1; i < matrixArray.length; i++) {
+      const mat = matrixArray[i];
+      if (mat != null) {
+        mat4.multiply(model.tmpMat4, model.tmpMat4, mat);
+      }
+    }
+    return model.tmpMat4;
```
finetjul: return matrices.reduce((res, matrix, index) => {
    if (index === 0) {
      return matrix ? mat4.copy(res, matrix) : mat4.create();
     }
     return matrix ? mat4.multiply(res, res, matrix)) : res;
  },
  model.tmpMat4);

finetjul:
LGTM

PR #1234, opened 2019-10-24
Fix VolumeMapper FS for parallel projection.

JamesAPetts:
By casting the rays in the direction of the camera we fix the issues in the fragment shader, ~~but it goes wacky if I rotate it at the moment:~~ <--Fixed

I've got a testbed up in `npm run example -- VolumeMapperParallelProjection`, which uses a MIP with a wwwc covering the whole volume, and a cube sitting neatly on top of it for comaprison to perspective projection:

(The lag in the gif is due to my recording software)

Gif of progress


JamesAPetts:
Fixed now, there may be some optimisations shady savvy guys could make. I know if statements are bad in the shader. I followed the same pattern as the vertex shaders that check a `cameraParallel` uniform to control the flow, but if there is a better way, feel free! :)

JamesAPetts:
I now need to update the tests.

JamesAPetts:
The tests are passing locally, lets see if the CI gods agree.

JamesAPetts:
Tests are passing, but I'm not sure why the Linux build is failing, the CI logs don't give too much information.

jourdain:
The CI error could be ignored, everything is fine.

@JamesAPetts can you fix your commitizen as we don't use the one with the emoji. Ideally for this commit the message should have been `fix(VolumeMapper): Properly handle parallel projection`.

JamesAPetts:
Rebased 👍 .

jourdain:
@martinken can we merge?

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -815,10 +815,22 @@ void computeIndexSpaceValues(out vec3 pos, out vec3 endPos, out float sampleDist
...
-  // camera is at 0,0,0 so rayDir for perspective is just the vc coord
-  vec3 rayDirVC = normalize(vertexVCVSOutput);
+
+  vec3 rayDirVC;
+
+  if (cameraParallel == 1)
+  {
+    // Camera is parallel, so the rayDir is just the direction of the camera.
+    rayDirVC = cameraDirection;
+  } else {
```
martinken: I believe in parallel projection the rayDirVC is always vec3(0.0, 0.0, 1.0) can  you try that?

JamesAPetts: Oh interesting. Its actually always (0,0,-1) in view space. Which is what I was actually calculating haha. I'll update it.

JamesAPetts: Addressed.

martinken:
+1 technically we could remove the isUniformUsed as it is hardcoded into the shader right now. But that code only gets executing once per volume per render so it isn't a big deal.

PR #1236, opened 2019-10-24
fix(Manipulators): Ensure manipulators aren't null

agirault:
The array of manipulators can have `null` values since[ removing a manipulator will not remove its key from the list, but only set its value to `null`](https://github.com/Kitware/vtk-js/blob/a96658a/Sources/Interaction/Style/InteractorStyleManipulator/index.js#L181-L200).

Another fix instead would be to change the behavior of those `removeGesture` functions to actually remove the key/value and not just set the value to `null`, I can do that if you'd prefer @jourdain.

### DIFF
--- a/Sources/Interaction/Style/InteractorStyleManipulator/index.js
+++ b/Sources/Interaction/Style/InteractorStyleManipulator/index.js
@@ -179,74 +179,50 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
   };
 
   //-------------------------------------------------------------------------
-  publicAPI.removeMouseManipulator = (index) => {
-    if (model.mouseManipulators.length > index) {
-      model.mouseManipulators[index] = null;
+  const removeManipulator = (manipulator, list) => {
+    const index = list.indexOf(manipulator);
+    if (index === -1) {
+      return false;
     }
+    list.splice(index, 1);
+    publicAPI.modified();
+    return true;
   };
 
   //-------------------------------------------------------------------------
-  publicAPI.removeVRManipulator = (index) => {
-    if (model.vrManipulators.length > index) {
-      model.vrManipulators[index] = null;
+  publicAPI.removeMouseManipulator = (manipulator) =>
+    removeManipulator(manipulator, model.mouseManipulators);
+
+  //-------------------------------------------------------------------------
+  publicAPI.removeVRManipulator = (manipulator) =>
+    removeManipulator(manipulator, model.vrManipulators);
+
+  //-------------------------------------------------------------------------
+  publicAPI.removeGestureManipulator = (manipulator) =>
+    removeManipulator(manipulator, model.gestureManipulators);
+
+  //-------------------------------------------------------------------------
+  const addManipulator = (manipulator, list) => {
+    const index = list.indexOf(manipulator);
+    if (index !== -1) {
+      return false;
     }
+    list.push(manipulator);
+    publicAPI.modified();
+    return true;
   };
 
   //-------------------------------------------------------------------------
-  publicAPI.removeGestureManipulator = (index) => {
-    if (model.gestureManipulators.length > index) {
-      model.gestureManipulators[index] = null;
-    }
-  };
+  publicAPI.addMouseManipulator = (manipulator) =>
+    addManipulator(manipulator, model.mouseManipulators);
 
   //-------------------------------------------------------------------------
-  publicAPI.getMouseManipulator = (index) => {
-    let manipulator = null;
-    if (model.mouseManipulators.length > index) {
-      manipulator = model.mouseManipulators[index];
-    }
-    return manipulator;
-  };
+  publicAPI.addVRManipulator = (manipulator) =>
+    addManipulator(manipulator, model.vrManipulators);
 
   //-------------------------------------------------------------------------
-  publicAPI.getVRManipulator = (index) => {
-    let manipulator = null;
-    if (model.vrManipulators.length > index) {
-      manipulator = model.vrManipulators[index];
-    }
-    return manipulator;
-  };
-
-  //-------------------------------------------------------------------------
-  publicAPI.getGestureManipulator = (index) => {
-    let manipulator = null;
-    if (model.gestureManipulators.length > index) {
-      manipulator = model.gestureManipulators[index];
-    }
-    return manipulator;
-  };
-
-  //-------------------------------------------------------------------------
-  publicAPI.addMouseManipulator = (manipulator) => {
-    const index = model.mouseManipulators.length;
-    model.mouseManipulators.push(manipulator);
-    return index;
-  };
-
-  //-------------------------------------------------------------------------
-  publicAPI.addVRManipulator = (manipulator) => {
-    const index = model.vrManipulators.length;
-    model.vrManipulators.push(manipulator);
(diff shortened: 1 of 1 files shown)


jourdain:
Please go ahead and address the add/remove in a better way so you don't rely on an index that need to remain constant.

agirault:
@jourdain on top of this branch (to also keep the checks on the manipulators' validity), or instead of this branch?

jourdain:
The same branch is fine.

agirault:
Sounds good. That's technically going to be a breaking change... but should I just mark it as fix?

jourdain:
Mark it as a breaking change. That would be fine.

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -181,21 +181,21 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
   //-------------------------------------------------------------------------
   publicAPI.removeMouseManipulator = (index) => {
```
jourdain: You should expect the arg to be the manipulator not the index since, all your capture index will be messed up by any remove.

agirault: Why? That's a very normal API to be able to remove an element at an index that causes the content to shift:
* c++: [erase](http://www.cplusplus.com/reference/vector/vector/erase/)
* python: [pop](https://docs.python.org/3/tutorial/datastructures.html)
* js: [splice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)
* swift: [remove](https://developer.apple.com/documentation/swift/array/1641390-remove)

I agree we could have an API to use the manipulator as an input, but shouldn't that be new methods? So we can keep those methods a bit more consistent in behavior and input?

jourdain: And how do you get the index of the manipulator that you want to remove? The index that was given to you at the registration is what became invalid after the first remove.

jourdain: pop / erase and so on make sens because you control the list, here you don't hence the issue.

jourdain: The user should not have to know that the internal representation is a list of some kind.

The thing that the user cares about is to be able to add/remove his interactor style. You should not add any extra burden to the dev to actually know how the internal of that class is made of. If so that leads to bad design.

before the design was:

```
id = manip.addStyle(style);
manip.removeStyle(id);
```

the design now should be

```
manip.addStyle(style);
manip.removeStyle(style);
```

jourdain: And we can maintain the previous api valid by returning `style` when calling `addStyle`, that way it is not a breaking change.

agirault: So many comments 🙈

> And how do you get the index of the manipulator that you want to remove? The index that was given to you at the registration is what became invalid after the first remove.

Yeah indeed, that makes sense.

agirault: What about `getManipulator(id)` ?

jourdain: No

jourdain: I want all the following to work

```
id = manip.addStyle(style);
manip.removeStyle(id);
```

```
manip.addStyle(style);
manip.removeStyle(style);
```

jourdain:
```
add(obj) {
idx = list.indexOf(obj);
if (idx === -1) {
list.push(obj);
modified()
return obj;
}
}

remove(obj) {
idx = list.indexOf(obj);
if (idx === -1) {
list.splie(idx, 1);
modified();
return true;
}
}

```

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -179,74 +179,68 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
   };
 
   //-------------------------------------------------------------------------
-  publicAPI.removeMouseManipulator = (index) => {
-    if (model.mouseManipulators.length > index) {
-      model.mouseManipulators[index] = null;
+  const removeManipulator = (manipulator, list) => {
+    const index = list.indexOf(manipulator);
+    if (index !== -1) {
```
jourdain: === not !==

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -179,74 +179,68 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
...
 
   //-------------------------------------------------------------------------
-  publicAPI.getVRManipulator = (index) => {
-    let manipulator = null;
-    if (model.vrManipulators.length > index) {
-      manipulator = model.vrManipulators[index];
-    }
-    return manipulator;
+  const getManipulator = (index, list) => {
+    return index > list.length ? null : list[index];
```
jourdain: if index is out of bound, it would be fine. You should just do 

return list[index];

assuming we want to keep those getters?

Sources/Interaction/Style/InteractorStyleManipulator/index.js
agirault: Ah yeah, sorry.

Sources/Interaction/Style/InteractorStyleManipulator/index.js
agirault: We could have getters to return the whole arrays instead?

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -179,74 +179,68 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
...
+
+  //-------------------------------------------------------------------------
+  publicAPI.getGestureManipulator = (index) =>
+    getManipulator(index, model.gestureManipulators);
+
+  //-------------------------------------------------------------------------
+  const addManipulator = (manipulator, list) => {
+    const index = list.indexOf(manipulator);
+    if (index !== -1) {
+      return false
```
floryst: semicolon needed :)

floryst:
Do we want to keep the `getManipulator` API? Since it's an index-based API, we might not need it anymore since your changes to `addManipulator` makes it not return the index (which wouldn't be terribly useful anyways).

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -179,74 +179,50 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
...
   //-------------------------------------------------------------------------
-  publicAPI.removeMouseManipulator = (index) => {
-    if (model.mouseManipulators.length > index) {
-      model.mouseManipulators[index] = null;
+  const removeManipulator = (manipulator, list) => {
+    const index = list.indexOf(manipulator);
+    if (index === -1) {
+      return false;
     }
+    list.slice(index, 1);
```
floryst: I think this should be `list.splice` instead of `list.slice`.

PR #1240, opened 2019-10-28
fix(AbstractWidgetFactory): Incorrect unsubscribe logic

floryst:


Sources/Widgets/Core/AbstractWidgetFactory/index.js
```diff
@@ -153,13 +153,13 @@ function vtkAbstractWidgetFactory(publicAPI, model) {
   // Event Widget API
   // --------------------------------------------------------------------------
   let unsubscribe = NoOp;
-  publicAPI.delete = macro.chain(publicAPI.delete, unsubscribe);
+  publicAPI.delete = macro.chain(publicAPI.delete, () => unsubscribe());
```
jourdain: Do you really need to wrap it into another function?

jourdain: Oh, I see... Sorry, you are right!

floryst: Yes, because macro.chain will get the reference to `NoOp` here rather than the correct unsubscribe function.

floryst:
FYI I found another subscription bug in other code, so I'll tack it on to this PR before merging.

jourdain:
Cool thx

Sources/Widgets/Core/AbstractWidgetFactory/index.js
floryst: Haha github didn't show me your second message until after I posted.

PR #1241, opened 2019-10-29
Tweak hardware selector picking

floryst:
Adds `captureOn` option. When it is `CaptureOn.MOUSE_RELEASE`, capture will take place at end of scene animation, and will capture entire buffer. When it is `CaptureOn.MOUSE_MOVE`, capture will take place when mouse moves, and will capture just the pixel under the mouse cursor. Default value is MOUSE_RELEASE.

Addresses https://github.com/Kitware/paraview-glance/issues/270

### DIFF
--- a/Sources/Proxy/Core/ViewProxy/index.js
+++ b/Sources/Proxy/Core/ViewProxy/index.js
@@ -191,8 +191,8 @@ function vtkViewProxy(publicAPI, model) {
         return;
       }
       const devicePixelRatio = window.devicePixelRatio || 1;
-      const width = Math.max(10, devicePixelRatio * Math.floor(dims.width));
-      const height = Math.max(10, devicePixelRatio * Math.floor(dims.height));
+      const width = Math.max(10, Math.floor(devicePixelRatio * dims.width));
+      const height = Math.max(10, Math.floor(devicePixelRatio * dims.height));
       model.openglRenderWindow.setSize(width, height);
       publicAPI.invokeResize({ width, height });
       publicAPI.renderLater();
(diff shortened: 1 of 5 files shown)


floryst:
Also I need to do some manual testing of widgets to make sure I didn't break anything with this new option.

floryst:
My final force-push was just to clean up commit messages to reflect new API name.

finetjul:
Thanks @floryst for your fix. I've got few questions:
 - does/should it fix the discrepancy in capture performance between machines ? On some machines it was taking ~20m, on some others it was more aroun ~400ms
 - What should make us decide between MOUSE_RELEASE or MOUSE_MOVE ? Should it be widget dependent ?
 - picking was re-enabled automatically even if we explicitely told the widget manager to disable it. Will it fix it ?

Thanks.

floryst:
1) Switching to MOUSE_MOVE might help mitigate capture performance. From what I know, capture performance is dependent on effective webgl buffer size, so screens with high DPI will likely experience higher capture times.
2) I think it'll be application dependent. If you are primarily working with mostly static scenes with low frequency of interaction, then MOUSE_RELEASE might be good since you can capture once and then use that buffer for fast picking. So far, I think MOUSE_MOVE will be better for most other applications, but I think MOUSE_MOVE will suffer interactivity loss if your scene takes a noticeable amount of time to render.
3) Hum, this shouldn't be happening, and definitely seems like a bug. Does `disablePicking()` not work? I can take a look.

This will be good to add into the docs (which I forgot to do), so thanks for asking!

PR #1245, opened 2019-10-31
fix(LineWidget): Properly handle processEvents parameter

luciemac:


luciemac:
@finetjul can you review ?

PR #1251, opened 2019-11-07
fix(vtkOpenGLPolyDataMapper): Always update MCVC/MCDC uniform matrices

boucaud:
Always update MCVC/MCDC uniform matrices to make sure that inverseShiftAndScale matrix values are correct.

Currrently, view matrices in vtk.js are only updated in case of camera or actor changes.
With the introduction of coordinate shifting/scaling to the mapper, it is now necessary to properly update view matrices before each draw call, because inverse shift/scale matrices can be different for each CellArrayBufferObject and affect the value of the MCVC/MCDC matrices. Not doing so could result in incorrect rendering when modifying the mapper's input geometry, or rendering multiple objects.

For reference, this is similar to the way things are done in VTK's vtkOpenGLPolyDataMapper.

This fixes current issues with the LineWidget and HandleWidget examples.

### DIFF
--- a/Sources/Rendering/OpenGL/PolyDataMapper/index.js
+++ b/Sources/Rendering/OpenGL/PolyDataMapper/index.js
@@ -1488,13 +1488,14 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
     }
   };
 
-  function safeMat4Multiply(matrixArray) {
+  function safeMatrixMultiply(matrixArray, matrixType, tmpMat) {
+    matrixType.identity(tmpMat);
     return matrixArray.reduce((res, matrix, index) => {
       if (index === 0) {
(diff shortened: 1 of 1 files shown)


finetjul:
That's fine with me, but I don't know enough about it.
I let @martinken review :-)

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1516,46 +1516,44 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
     const actorIsIdentity = actor.getIsIdentity();
     const actMats = actorIsIdentity ? null : model.openGLActor.getKeyMatrices();
```
finetjul: I feel that if `actorIsIdentity` is false, setting `actMats` to `{mcwc: null, normalMatrix: null}`  would simplify the code

PR #1253, opened 2019-11-19
fix(HardwareSelector): Existing framebuffer not being bound to context

floryst:
This fixes an issue where the color picking buffer is antialiased when it shouldn't be, since beginSelection() wasn't binding the existing framebuffer.


Sources/Rendering/OpenGL/HardwareSelector/index.js
```diff
@@ -44,6 +45,7 @@ function vtkOpenGLHardwareSelector(publicAPI, model) {
         model.framebuffer.create(size[0], size[1]);
         model.framebuffer.populateFramebuffer();
       }
+      model.framebuffer.bind();
```
finetjul: shouldn't you call bind only in `else()` because `populateFramebuffer()` is called line 45/46 and your comment line 38 says `populateFramebuffer()` already calls `bind()` ?

floryst: Good point, I'll put it in the else block. I assum `bind` is idempotent, but best to not make assumptions here. :)

martinken:
+1 LGTM

PR #1259, opened 2019-11-25
feat(vtkImageData): add getIndexFromWorld function

mix3d:
* Write vtkImageData API Docs.
* Refactor vtkImageData functions to put similar functions in proximity, and return arrays in the worldToIndex and indexToWorld function, to match the behavior of the *bounds functions.
* Provide a new function to get the scalar data index from a given x,y,z world coordinate. This effectively matches the c++ function of `vtkImageData.getScalarComponentAsDouble()`

We needed the ability to select a point in screen space, convert to world space, then find the corresponding voxel index to get its value. Our c++ app used the above getScalar function, but I had to write the functionality locally. By having these functions in VTK, I can reduce my own code.

I debated also including a full `getScalarComponent` function that returned the actual voxel data, but wanted to start with this.

Docs are my best understanding of vtkImageData after spending a few days deepdiving and comparing C++ / JS code, while building this feature.

I welcome any suggestions, naming or otherwise.

### DIFF
--- a/Sources/Common/DataModel/ImageData/index.js
+++ b/Sources/Common/DataModel/ImageData/index.js
@@ -471,10 +471,11 @@ function vtkImageData(publicAPI, model) {
    */
   publicAPI.computeOffsetIndex = ([i, j, k]) => {
     const extent = publicAPI.getExtent();
-    const increments = publicAPI.computeIncrements(
-      extent,
-      publicAPI.getNumberOfComponents()
-    );
+    const numberOfComponents = publicAPI
+      .getPointData()
+      .getScalars()
+      .getNumberOfComponents();
+    const increments = publicAPI.computeIncrements(extent, numberOfComponents);
     // Use the array increments to find the pixel index
     // similar to c++ vtkImageData::GetArrayPointer
     // Math.floor to catch "practically 0" e^-15 scenarios.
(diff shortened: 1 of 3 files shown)


mix3d:
There was one part that I wasn't sure how to deal with, was the `length` parameter of the `computeIncrements()` function, which in C++ uses `scalars->GetNumberOfComponents()` to define the length, but I don't know how multi-value scalars are handled in the JS implementation.

Sources/Common/DataModel/ImageData/index.js
```diff
@@ -444,6 +449,60 @@ function vtkImageData(publicAPI, model) {
       sigma,
     };
   };
+
+  // TODO: use the unimplemented `vtkDataSetAttributes` for scalar length, that is currently also a TODO (GetNumberOfComponents).
+  // Scalar data could be tuples for color information?
+  publicAPI.computeIncrements = (extent, length = 1) => {
```
finetjul: rename length to "numberOfComponents" ?

floryst:
Maybe "flatIndexFromWorld" and "computeFlatIndex" would be a less confusing name? "index" as used in "indexToWorld" and similar functions assume a coordinate in index space rather than an offset in the underlying array.

For handling multiple-component values, you should use `getNumberOfComponents`. Storage-wise, components are stored contiguously in the underlying array.

Travis is complaining about some error in `indexFromWorld` as well.

mix3d:
> Maybe "flatIndexFromWorld" and "computeFlatIndex" would be a less confusing name? "index" as used in "indexToWorld" and similar functions assume a coordinate in index space rather than an offset in the underlying array.
What about `computeOffsetIndex` or `computeDataIndex`? I agree the naming isn't great, (definitions collision),  but it's not great in C++ either, nor is there easy naming replacements (No need to `getScalarPointer`) :/

> For handling multiple-component values, you should use `getNumberOfComponents`. Storage-wise, components are stored contiguously in the underlying array.
Is this available at the `ImageData` class level, or do I have to do `getPointData().getScalars()` first?


> Travis is complaining about some error in `indexFromWorld` as well.
Found and Fixed.

mix3d:
for multi-component scalar data, do we need to also then include the logic / options to get the Nth component? (get the B value of RGB tuples, etc)

Similar to the `comp` value in https://github.com/Kitware/VTK/blob/master/Common/DataModel/vtkImageData.cxx#L1306

floryst:
I'm good with `offsetIndex` as the name.

You have to use `getPointData().getScalars().getNumberOfComponents()` to get the number of components.

AFAIK those look to be wrappers around `getPointData().getScalars().[set/get]Component(index, comp, [value])`. Feel free to add those if you like to.

mix3d:
Updated to use `getNumberOfComponents` and added `getScalarValueFromWorld()`
I'm not really sure what the usecase for `setScalarValueFromWorld()` would be, but it sure wouldn't be hard, nor would it be hard for someone to just use the index and call the set on the scalar values themselves.

PR #1262, opened 2019-12-02
Colormap cleanup and additions

thewtex:
The additions are for multi-component volume rendering.


jourdain:
How do you know they were duplicates?

thewtex:
> How do you know they were duplicates?

The values were also the same -- just came across this when searching through the file :-)

PR #1264, opened 2019-12-02
Border fill

JamesAPetts:
This PR adds the option to render labelmaps with an outline of a specified thickness.

The logic is placed inside the VolumeMapperFS shader as it uses a lot of the same logic. If `useLabelOutline` is set true on the volume mapper the outline logic will be used.

The example is under `npm run example -- VolumeOutline`:
Example (Note the gif is compressed and the lines are smooth in actual usage):
example

I'm making this PR to get some initial feedback, but I understand the shader logic might need to be refactored somewhere else.

P.S. Quite new to writing shader code, if I've done anything stupid computationally I'd really appreciate constructive feedback! :)

### DIFF
--- a/Sources/Rendering/OpenGL/VolumeMapper/index.js
+++ b/Sources/Rendering/OpenGL/VolumeMapper/index.js
@@ -99,6 +99,15 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
       ).result;
     }
 
+    const vtkImageLabelOutline = actor.getProperty().getUseLabelOutline();
+    if (vtkImageLabelOutline === true) {
+      FSSource = vtkShaderProgram.substitute(
+        FSSource,
+        '//VTK::ImageLabelOutlineOn',
+        '#define vtkImageLabelOutlineOn'
+      ).result;
+    }
+
     const numComp = model.scalarTexture.getComponents();
     FSSource = vtkShaderProgram.substitute(
       FSSource,
@@ -130,6 +139,7 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
       (ext[3] - ext[2]) * spc[1],
       (ext[5] - ext[4]) * spc[2]
     );
+
     const maxSamples =
       vec3.length(vsize) / model.renderable.getSampleDistance();
 
@@ -513,7 +523,9 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
         Please either change the
         volumeMapper sampleDistance or its maximum number of samples.`);
     }
+
     const vctoijk = vec3.create();
+
     vec3.set(vctoijk, 1.0, 1.0, 1.0);
     vec3.divide(vctoijk, vctoijk, vsize);
     program.setUniform3f('vVCToIJK', vctoijk[0], vctoijk[1], vctoijk[2]);
@@ -568,6 +580,22 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
       // specify the planes in view coordinates
       program.setUniform3f(`vPlaneNormal${i}`, normal[0], normal[1], normal[2]);
       program.setUniformf(`vPlaneDistance${i}`, dist);
+
+      if (actor.getProperty().getUseLabelOutline()) {
+        const image = model.currentInput;
+        const worldToIndex = image.getWorldToIndex();
+
+        program.setUniformMatrix('vWCtoIDX', worldToIndex);
+
+        // Get the display coordinate to world coordinate transformation matrix.
+        mat4.invert(model.displayToWorld, keyMats.wcdc);
+        program.setUniformMatrix('DCWCMatrix', model.displayToWorld);
+
+        const size = publicAPI.getRenderTargetSize();
+
+        program.setUniformf('vpWidth', size[0]);
+        program.setUniformf('vpHeight', size[1]);
+      }
     }
 
     mat4.invert(model.displayToView, keyMats.vcdc);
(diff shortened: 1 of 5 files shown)


jourdain:
Could you rework the commit messages to follow commitizen so we could get a nice change-log?

JamesAPetts:
> Could you rework the commit messages to follow commitizen so we could get a nice change-log?

Sure thing.

JamesAPetts:
Rebased on the updates to master I missed (funny github didn't pick that up), and squashed to a single commitizen commit.

thewtex:
@JamesAPetts awesome contribution! :clap: :taco: :unicorn: 

Any plans to add support to `vtkImageMapper`?

JamesAPetts:
Hey @thewtex , not at the current time, as my client only requires outline rendering for the `vtkVolumeMapper`, as they are using the thin oblique MPR via small clipping range approach in OHIF.

I'd be very happy to describe the algorithm in more detail should someone else wish to take up the effort, however.

JamesAPetts:
I added another couple of commits to solve issues that arose when using the outline renderer in practice with multiple viewports.

I'll squash everything to a nice commit again after we itterate and review, as we are likely to commit a few more things as we itterate.

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -568,6 +580,30 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
+      if (actor.getProperty().getUseLabelOutline()) {
+        const image = model.currentInput;
+        const worldToIndex = image.getWorldToIndex();
+
+        program.setUniformMatrix('vWCtoIDX', worldToIndex);
+
+        // Compute the view coordinate to world coordinate transformation matrix.
+        model.viewToWorld = mat4.create();
+        mat4.invert(model.viewToWorld, keyMats.wcvc);
+
```
martinken: model.viewToWorld seems to be created every frame and not used maybe?

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -568,6 +580,30 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
+        const size = publicAPI.getRenderTargetSize();
+        const aspect = size[0] / size[1];
+        const viewToWorld = ren
+          .getActiveCamera()
+          .getCompositeProjectionMatrix(aspect, -1.0, 1.0);
+
```
martinken: viewToWorld here I think is really world to device. Which is stored already in keyMats.wcdc

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -568,6 +580,30 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
+        mat4.invert(viewToWorld, viewToWorld);
+        mat4.transpose(viewToWorld, viewToWorld);
+
+        program.setUniformMatrix('VCWCMatrix', viewToWorld);
```
martinken: Should be named DCWCMatrix I think as it is from device to world unless I am missing something

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -276,6 +288,19 @@ vec4 computeNormal(vec3 pos, float scalar, vec3 tstep)
...
 }
 
+vec3 fragCoordToIndexSpace(vec4 fragCoord) {
+  vec4 ndcPos = vec4(
+    (fragCoord.x / vpWidth - 0.5) * 2.0,
+    (fragCoord.y / vpHeight - 0.5) * 2.0,
+    (fragCoord.z - 0.5) * 2.0,
+    1.0);
+
+  vec4 worldCoord = VCWCMatrix * ndcPos;
```
martinken: Yeh, DCWCMatrix here

Sources/Rendering/OpenGL/VolumeMapper/index.js
JamesAPetts: You're correct, I thought I caught of my dead playing around code. Nice catch, I'll remove it.

Sources/Rendering/OpenGL/VolumeMapper/index.js
JamesAPetts: The naming conventions throughout vtkjs were the biggest source of confusion for me. I completely agree with you, but was trying to stay consistent with other parts of the codebase, exactly the same code is named `viewToWorld` here:

https://github.com/Kitware/vtk-js/blob/a5b4567c8a5b8eec4546eacf9c20045caeab2022/Sources/Rendering/Core/Renderer/index.js#L241

I've checked, and inverting keyMats.wcdc gives me the same result, thanks!

JamesAPetts:
Thanks for the review @martinken , points have been address and rebased.

martinken:
Assuming the tests all pass this looks fine to me. +1 

My only thought is that we are creating a mat4 every frame in the if statement. In the past when profiling JS apps I've found the garbage collector can be a real performance hit so I try to avoid creating objects on the heap during the render loop.  So it might be better to have that dcwc mat4 be created on class initialization and then just used in the if statement. But not a big deal as it only happens when the border is turned on.

martinken:
And yes the terminology in RenderWindow for worldToView is different and it drives me nuts. My fault and it needs to be cleaned up.

JamesAPetts:
> My only thought is that we are creating a mat4 every frame in the if statement.
Sure, I've changed it to happen on class construction, makes sense.

Also all karma tests are passing 👍 .

PR #1265, opened 2019-12-02
feat(ViewProxy): Add generic animatable moveCamera() function

psavery:
Add an animatable moveCamera() function to the 3D view proxy that allows the camera to change to any focal point, position, and view point. This function was based upon the updateOrientation() function above it.

### DIFF
--- a/Sources/Proxy/Core/ViewProxy/index.js
+++ b/Sources/Proxy/Core/ViewProxy/index.js
@@ -326,7 +326,6 @@ function vtkViewProxy(publicAPI, model) {
     } else {
       const skipWarning =
         requester === publicAPI ||
-        `${requester}`.indexOf('ViewProxy.updateOrientation.') === 0 ||
         `${requester}`.indexOf('ViewProxy.moveCamera.') === 0;
       model.renderWindow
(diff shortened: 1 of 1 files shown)


finetjul:
It would be great if you could factorize the animationStack with `updateOrientation()`.

Additionally, (and that might be outside of your scope) I wonder how those camera movements are interpolated without using quaternions. 
If that interests you, you have the following C++ examples: https://vtk.org/doc/nightly/html/classvtkCameraInterpolator.html
https://github.com/finetjul/bender/blob/master/Libs/VTK/Common/vtkDualQuaternion.h

jourdain:
I'm happy to merge it in its current state, but I agree with @finetjul. It would be great if we could refactor the code for the updateOrientation(). If you have time please do it. 

Having proper camera interpolation will indeed be great. This is just a simple linear interpolation.

psavery:
I refactored `updateOrientation()` to use `moveCamera()`.

I think improving the camera interpolation sounds good also - but it may end up being done in a separate PR in the future...

finetjul:
LGTM, thanks !

PR #1266, opened 2019-12-03
fix(VolumeRepresentationProxy): Edge gradient for multi-component vol…

thewtex:
…umes

### DIFF
--- a/Sources/Proxy/Representations/VolumeRepresentationProxy/index.js
+++ b/Sources/Proxy/Representations/VolumeRepresentationProxy/index.js
@@ -101,36 +101,38 @@ function updateDomains(dataset, dataArray, model, updateProp) {
 // ----------------------------------------------------------------------------
 
 function updateConfiguration(dataset, dataArray, { mapper, property }) {
-  const dataRange = dataArray.getRange();
-
   // Configuration
   // actor.getProperty().setInterpolationTypeToFastLinear();
   property.setInterpolationTypeToLinear();
 
-  // For better looking volume rendering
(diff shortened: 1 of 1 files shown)


Sources/Proxy/Representations/VolumeRepresentationProxy/index.js
```diff
@@ -101,36 +101,39 @@ function updateDomains(dataset, dataArray, model, updateProp) {
...
-  property.setGradientOpacityMaximumOpacity(0, 1.0);
+  const numberOfComponents = dataArray.getNumberOfComponents();
+  for (let component = 0; component < numberOfComponents; component++) {
+    // For better looking volume rendering
+    // - distance in world coordinates a scalar opacity of 1.0
+    property.setScalarOpacityUnitDistance(
+      component,
+      vtkBoundingBox.getDiagonalLength(dataset.getBounds()) /
+        Math.max(...dataset.getDimensions())
+    );
```
jourdain: Since that is const across components, should you have it computed outside of the loop?

jourdain:
Seems reasonable otherwise.

floryst:
LGTM. Can merge if you want to address Seb's one remark, but I'm okay with it since it's not a big deal performance-wise.

jourdain:
Agree not a big deal...

Sources/Proxy/Representations/VolumeRepresentationProxy/index.js
thewtex: Good catch -- done!

thewtex:
Thanks for the reviews!

PR #1272, opened 2019-12-06
feat(Manipulators): Add manipulators for a first person mode

psavery:
This adds two camera manipulators:

1. A keyboard manipulator that uses standard (but configurable) WASD-style movement.
2. A mouse manipulator that can be pointer locked (default) and moves the focal point based on mouse movements.

I'm currently starting the pointer lock and hooking into the events [directly](https://github.com/Kitware/vtk-js/compare/master...psavery:first-person-manipulators?expand=1#diff-1ce8257b17c30c3eaa97dd5f7c7cca7bR43-R52) because there is currently not a way to do so via the `RenderWindowInteractor`. Additionally, mouse movements are being hooked into directly as well because the style doesn't forward the mouse movement events [after the mouse is depressed](https://github.com/Kitware/vtk-js/blob/2202738cc033ad5d22d64687f72dc8fa43d9ab4e/Sources/Interaction/Style/InteractorStyleManipulator/index.js#L390-L404) (also, the mouse movements will need to be forwarded to the function as well. Currently, only the position of mouse is forwarded, and this does not change under pointer lock). These are things that can probably change.

Two identical examples were added (each of which use both of the above manipulators). You can run one of them via `npm run example -- KeyboardCameraManipulator`.

I'm open to suggestions as to how we may be able to improve performance.


psavery:
I'm thinking of a few things we could do to get the `RenderWindowInteractor`->`InteractorStyleManipulator`->`Manipulator` communication series to work with the pointer lock mode, if we wanted.

1. Add `event.movementX` and `event.movementY` to the callData [here](https://github.com/Kitware/vtk-js/blob/2202738cc033ad5d22d64687f72dc8fa43d9ab4e/Sources/Rendering/Core/RenderWindowInteractor/index.js#L419) that is forwarded to `InteractorStyleManipulator.handleMouseMove`. These are needed for pointer lock mode, since the mouse position does not change, but it does produce movement information.
2. Provide some kind of way for the manipulator to indicate that it wants to enter pointer lock mode, and have the `InteractorStyleManipulator` check that (maybe in `onButtonDown` [here](https://github.com/Kitware/vtk-js/blob/2202738cc033ad5d22d64687f72dc8fa43d9ab4e/Sources/Interaction/Style/InteractorStyleManipulator/index.js#L307-L338)) and then either enter pointer lock mode or signal to `RenderWindowInteractor` that it should enter pointer lock mode.
3. Don't end mouse interactions with the manipulator when the mouse is released (as is done [here](https://github.com/Kitware/vtk-js/blob/2202738cc033ad5d22d64687f72dc8fa43d9ab4e/Sources/Interaction/Style/InteractorStyleManipulator/index.js#L390-L404)) if we are in pointer lock mode.
4. Forward the movement information provided from number 1 to the manipulator in `handleMouseMove` [here](https://github.com/Kitware/vtk-js/blob/2202738cc033ad5d22d64687f72dc8fa43d9ab4e/Sources/Interaction/Style/InteractorStyleManipulator/index.js#L472-L481).
5. Have the `RenderWindowInteractor` listen for the `pointerlockchange` event, and forward to the `InteractorStyleManipulator` when the pointer lock mode becomes enabled, and then when it becomes disabled. When pointer lock mode becomes disabled, the `InteractorStyleManipulator` can then end mouse interactions as is done [here](https://github.com/Kitware/vtk-js/blob/2202738cc033ad5d22d64687f72dc8fa43d9ab4e/Sources/Interaction/Style/InteractorStyleManipulator/index.js#L390-L404).

Just some thoughts...

martinken:
+1  Ready for merge @psavery  ?

floryst:
I think we should switch away from using `setInterval` and use `interactor.onAnimation` instead. Using setInterval can cause visual tearing, and a quick perf test gives me dropped frames as opposed to using onAnimation (which uses requestAnimationFrame).

For the pointerlock API, maybe we can have the RenderWindowInteractor expose methods for starting and ending pointer lock, since manipulators receive the interactor anyways. We can then expose an "onPointerLockChange" event on the interactor, which corresponds to the "pointerlockchange" event on the document object. Just some thoughts.

floryst:
Oh, some other thought that would be cool after playing with this: support roll via rotating the camera around the direction of projection. For WASD controls, it could be Q and E to roll left and right, respectively.

martinken:
Yes, you can implement a handleAnimation method in your new class instead of using setInterval

It should get called automatically

martinken:
In it you can check the state and then do what you want.

martinken:
Specifically model.state

martinken:
Just add any states you need to 
```
onst stateNames = {
  Rotate: States.IS_ROTATE,
  Pan: States.IS_PAN,
  Spin: States.IS_SPIN,
  Dolly: States.IS_DOLLY,
  CameraPose: States.IS_CAMERA_POSE,
  WindowLevel: States.IS_WINDOW_LEVEL,
  Slice: States.IS_SLICE,
};
```

and then the start end methods will get created automatically

floryst:
LGTM

psavery:
I think this is ready to go. It might be good to add the ability to roll via Q and E in the future sometime, but probably in a separate PR...

martinken:
This isn't done quite the right way in terms of how we setup styles, but the functionality is there and we can make the changes later.

PR #1273, opened 2019-12-07
fix(ImageData): Support setDirection with Float32Array

thewtex:
Enables the use case:

  image2.setDirection(image1.getDirection())


thewtex:
> Can it be a Float64Array as well?

Sure, I'll add support for that as well.

PR #1274, opened 2019-12-08
feat(LookupTable): Support setTable to directly set colors

### LINKED ISSUE #1252: vtkColorTransferFunction
Hi,

I would like to know if there is a way to use a discretizable lookup table for mapping colors ? I would like to fix a number of color and then map a scalar value to any of these colors.

We have the vtkDiscretizableColorTransferFunction from C++ where we can specify the number of colors. Is there away to do this with vtk.js ?

Thanks

thewtex:
In order to support directly setting lookup table colors, support
`setTable`, which corresponds to the VTK C++ method:

  https://github.com/Kitware/VTK/blob/898d75800e5e33b6192b3f92b577715627a04718/Common/Core/vtkLookupTable.cxx#L707-L732

Given a vtkDataArray with the lookup table colors, set the internal
table based on these colors. The passed array should have unsigned char
type and 4 components per tuple for the RGBA colors.

The internal table is then augmented with special colors. The internally
table is kept as a `Array` versus a `vtkDataArray` as in the C++
implementation. This could possibly be a target for future refactoring.

ScalarToColors `setAnnotation` is fixed -- this was never covered and
had conflicting internal assumptions that its arguments were
vtkDataArray's or (Array or TypedArray's). Now, Array or TypedArray are
supported.

Closes #1252


Sources/Common/Core/ScalarsToColors/index.js
```diff
@@ -137,7 +133,7 @@ function vtkScalarsToColors(publicAPI, model) {
   //----------------------------------------------------------------------------
   publicAPI.resetAnnotations = () => {
     model.annotationArray = [];
-    model.annotatedValueMap = [];
+    model.annotatedValueMap = new Map();
```
jourdain: model should only have variables that should be serializable. If you don't respect it, you will break the pattern `vtk(inst.getState())` unless you have special handling inside an overridden version of getState().
If you need a Map, can it be private?

model.* is really for the protected vars.

Sources/Common/Core/ScalarsToColors/index.js
```diff
@@ -539,7 +536,7 @@ export function extend(publicAPI, model, initialValues = {}) {
 
   model.mappingRange = [0, 255];
   model.annotationArray = [];
-  model.annotatedValueMap = [];
+  model.annotatedValueMap = new Map();
```
jourdain: Same here... Should the var be private or can you simply use {}.

Sources/Common/Core/ScalarsToColors/index.js
thewtex: Ok, I will back out this change, then.

An Object instead of an Array was considered, but modern linters require the equivalent of a Map `has` to be called as `Object.prototype.hasOwnProperty.call(myObj, prop)` -- pretty annoying.

Would it be possible at some point to serialize a Map as an Object?

PR #1284, opened 2019-12-11
doc(develop_widget): Add title

### LINKED ISSUE #1279: Menu missing traduction for widget doc
https://kitware.github.io/vtk-js/docs/develop_widget.html
https://github.com/Kitware/vtk-js/blob/master/Documentation/tpl/__en__

We have an entry for `widgets` but not `widget`. ;-)

floryst:
Fixes #1279


jourdain:
That's not the problem...

PR #1294, opened 2019-12-16
Add texture and source LOD downloaders

psavery:
The TextureLODsDownloader is designed to download a series of texture images from low to high
resolution, and update the supplied vtkTexture as each download finishes. It currently downloads the
texture images sequentially rather than in parallel. There is also an option to provide a callback
for when each step finishes.

This feature was added to the HttpSceneLoader.

Signed-off-by: Patrick Avery <patrick.avery@kitware.com>


psavery:
When the updated textures are rendered for the first time, there is a pause/freeze in user interaction. This is especially noticeable when bigger texture images (>50 MB) are set, which can freeze the interaction on my laptop for like 3 - 10 seconds.

It'd be nice if this freezing could be avoided somehow. But I am not sure if it is possible.

Sources/Rendering/Misc/TextureLODsDownloader/index.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    let baseUrl = model.baseUrl;
+    if (baseUrl && !baseUrl.endsWith('/')) {
+      baseUrl += '/';
+    }
+
+    // Create the download stack
+    internal.downloadStack = [];
+    model.files.forEach((file) => internal.downloadStack.push(baseUrl + file));
```
jourdain: You probably want to use a template string. I would have thought the linter would warn you on it.

jourdain:
I'm wondering if the speed culprit is that [section](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/Texture/index.js#L893-L923).

If that's the case, you might be offload the data conversion to a [web worker](https://developers.google.com/web/updates/2018/08/offscreen-canvas).

To time a code section, you can use `console.time(label);` and `console.timeEnd(label);`.

psavery:
> I'm wondering if the speed culprit is that [section](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/Texture/index.js#L893-L923).
> 
> If that's the case, you might be able to offload the data conversion to a [web worker](https://developers.google.com/web/updates/2018/08/offscreen-canvas).
> 
> To time a code section, you can use `console.time(label);` and `console.timeEnd(label);`.

I gave it a try. It took 0.166ms for that section, so it looks like it might be something else...

jourdain:
Find the location and then let see if we can improve it.

martinken:
Odds are the delay is in opengl. There are two causes 

1) allocating a bug chunk of memory for the texture

2) uploading the texture data, which involved making a copy of the client side data and then sending it up to the GPU

For a large texture both of these operations can take time. Item (2) can be handled with some careful OpenGL code that uploads the texture in small pieces in a separate thread that is mutexed from the rendering thread to avoid multhreaded issues. It is tricky but the code to do it is here https://gitlab.kitware.com/ken-martin/mineview/blob/master/mvView.cxx#L1781  For item (1) the only solution I know of is to preallocate texture memory when the application starts, or at some other time that is better than when it is happening now. Maybe at the time of loading a file. But you will still have a delay. You are just changing when the delay will happen. If you allocate a large texture pool on application startup then you pay the price once, but you have to add a lot of logic to manage the texture memory pool etc. 

TLDR, if your timing results land on issue (1) or (2) then I think it is out of scope of this topic. If you find you are spending a bunch of time on something else then go for it.

Sources/Rendering/Misc/TextureLODsDownloader/index.js
psavery: I changed this to use a template string.

psavery:
Here are some of my timing results for when a ~70 MB texture gets set:

[vtkOpenGLTexture.render()](https://github.com/Kitware/vtk-js/blob/8db949fe930fef6dfe3b4c427efc0326caa4c679/Sources/Rendering/OpenGL/Texture/index.js#L19): 4428.747 ms
[vtkOpenGLTexture.create2DFromImage()](https://github.com/Kitware/vtk-js/blob/8db949fe930fef6dfe3b4c427efc0326caa4c679/Sources/Rendering/OpenGL/Texture/index.js#L867): 4428.271 ms
[model.context.texImage2D()](https://github.com/Kitware/vtk-js/blob/8db949fe930fef6dfe3b4c427efc0326caa4c679/Sources/Rendering/OpenGL/Texture/index.js#L925): 4427.764 ms

You can see that the vast majority of the time appears to be in that call to `texImage2D()`.

psavery:
I think we could potentially set the texture data in chunks using `texSubImage2D` (like @martinken is doing in [here](https://gitlab.kitware.com/ken-martin/mineview/blob/master/mvView.cxx#L1825), from the example above). But I might focus on other things first.

jourdain:
LGTM

Sources/IO/Core/DataAccessHelper/JSZipDataAccessHelper.js
```diff
@@ -193,6 +193,28 @@ function create(createOptions) {
...
+            return Promise.resolve(array);
+          });
+        }
+        return Promise.reject(new Error('Invalid compression'));
+      }
+
+      return zipRoot
+        .file(path)
+        .async('uint8array')
+        .then((data) => Promise.resolve(data));
```
jourdain: you have an option for arraybuffer, that might be better than uint8array.

psavery: Okay! I think I went with `uint8array` because the gzip data is converted to that ([here](https://github.com/Kitware/vtk-js/blob/6e6515f947e9ee8d6c4e1d383310293f93d07d4a/Sources/IO/Core/DataAccessHelper/JSZipDataAccessHelper.js#L157) and [here](https://github.com/Kitware/vtk-js/blob/6e6515f947e9ee8d6c4e1d383310293f93d07d4a/Sources/IO/Core/DataAccessHelper/JSZipDataAccessHelper.js#L182)). But I just gave `arraybuffer` a try and it seemed to work fine also. I'll switch over to that one.

psavery: Fixed. It seems to work fine.

psavery:
I tried out using Chrome's performance tool to take another look at the freeze that occurs for large textures (> 50 MB). It was very helpful! It looks like decoding the image is taking up a large part of the time:

performance

It appears that by default, the image does not get decoded right away. And then, when we try to render it, it blocks the GUI synchronously until it is decoded. However, we are able to decode asynchronously ([see here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decode), it looks like in "Usage Notes" a major use case is loading very large images like we are doing), and then use it only after it has been decoded.

However, for [this large image](https://github.com/psavery/psavery.github.io/blob/master/images/2018GaloreCreek/texture_1-16384x12603.jpg) (~70 MB), webgl in chrome seems to be able to load it successfully using the synchronous (blocking) decoding, but, at least on my laptop, the asynchronous decoding fails with a cryptic exception: `DOMException: The source image cannot be decoded.`

Firefox can decode that file both synchronously and asynchronously. However, when it runs `texImage2D` in webgl for either case, webgl crashes.

Smaller images seem to work fine for both sync and async in both firefox and chrome. Maybe we should put some kind of limit on the image sizes.

Sources/Rendering/Misc/TextureLODsDownloader/index.js
```diff
@@ -0,0 +1,184 @@
+import macro from 'vtk.js/Sources/macro';
+
+const { vtkErrorMacro } = macro;
+
+// ----------------------------------------------------------------------------
+
+const getRemoteFileSize = (url) => {
```
psavery: This function obtains the size of a remote file by checking the `Content-Length` header after making a `HEAD` request. For some reason, though, firefox is not returning a `Content-Length` header for cross origin requests. It seems to work fine for non-cross origin requests, though. But I am not sure why it doesn't work, nor what I could do to fix it, and I also do not know of another way to get the remote file size. Chrome seems to work fine in all cases I tried.

Sources/Rendering/Misc/TextureLODsDownloader/index.js
```diff
@@ -0,0 +1,185 @@
...
+    xhr.open('HEAD', url, true);
+
+    xhr.onreadystatechange = () => {
+      if (xhr.readyState === 4) {
+        if (xhr.status === 200) {
+          const size = xhr.getResponseHeader('Content-Length');
+          resolve(size);
+        } else {
+          console.log('Failed to get remote file size of', url);
+          reject(xhr.status, xhr.statusText);
```
jourdain: resolve and reject expect only 1 argument. `reject(xhr)`.

Sources/Rendering/Misc/TextureLODsDownloader/index.js
jourdain: Did you try without expecting readyState === 4 and status == 200 ?
What do you see in the network tab with firefox?

Sources/Rendering/Misc/TextureLODsDownloader/index.js
psavery: Fixed. Thanks Seb!

Sources/Rendering/Misc/TextureLODsDownloader/index.js
psavery: Yes, I tried that, but unfortunately, it does not help.

Strangely, in the network tab with firefox there is a `content-length` for the xhr CORS HEAD requests, which makes me think it must be present somewhere. But it does not show up in javascript, and I do not know why. If I print out `xhr.getAllResponseHeaders()` in firefox, I get something like the following:

```
cache-control: max-age=600
content-type: image/jpeg
expires: Tue, 31 Dec 2019 21:03:44 GMT
last-modified: Tue, 31 Dec 2019 02:34:55 GMT
```

You can see the `content-length` is not there. However, if the request has the same origin, the `content-length` IS there. And this is what `xhr.getAllResponseHeaders()` looks like in chrome, for the CORS HEAD request:

```
cache-control: max-age=600
content-length: 73223341
content-type: image/jpeg
expires: Tue, 31 Dec 2019 21:03:50 GMT
last-modified: Tue, 31 Dec 2019 02:34:55 GMT
```

Almost exactly the same, but the `content-length` is present.

I really do not know why it isn't working in firefox. But it makes it so that this function will not work for html files that we email around.

floryst: Firefox currently doesn't support accessing Content-Length headers from CORS responses, even though it's part of the spec: https://fetch.spec.whatwg.org/#cors-safelisted-response-header-name

I'll see if I can get a fix into Firefox, since I know where the fix should be.

psavery: @floryst Sounds good! Unfortunately, since I don't know of any work-arounds for this, we might just have to leave it as is for now, and wait until the issue is fixed. Until that time, standalone html files will not work with firefox with this function.

psavery:
This project is ending today, but I think this branch is in pretty good condition. If there are no more requested changes, I think it is ready for merging.

PR #1299, opened 2019-12-17
feat(WidgetManager): Split render method out from enablePicking

floryst:
This will allow for visually rendering widgets without having picking
enabled. In addition, this creates a clearly named method for rendering
widgets when some representation property has changed, e.g. SVG circle
color.

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -192,6 +192,26 @@ function vtkWidgetManager(publicAPI, model) {
     w.updateRepresentationForRender(model.renderingType);
   }
 
+  function renderPickingBuffer() {
+    model.renderingType = RenderingTypes.PICKING_BUFFER;
+    model.widgets.forEach(updateWidgetForRender);
+  }
+
+  function renderFrontBuffer() {
+    model.renderingType = RenderingTypes.FRONT_BUFFER;
+    model.widgets.forEach(updateWidgetForRender);
+  }
+
+  function captureBuffers(x1, y1, x2, y2) {
+    renderPickingBuffer();
+
+    model.selector.setArea(x1, y1, x2, y2);
+    model.selector.releasePixBuffers();
+
+    model.previousSelectedData = null;
+    return model.selector.captureBuffers();
+  }
+
   publicAPI.enablePicking = () => {
     model.pickingEnabled = true;
     model.pickingAvailable = true;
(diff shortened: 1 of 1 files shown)


floryst:
Don't merge yet. I need to test it.

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -194,18 +194,20 @@ function vtkWidgetManager(publicAPI, model) {
 
   publicAPI.enablePicking = () => {
     model.pickingEnabled = true;
+    model.pickingAvailable = true;
+    publicAPI.renderWidgets();
```
finetjul: how about not rendering widgets if pickingEnabled and pickingAvailable were already true ?

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -194,18 +194,20 @@ function vtkWidgetManager(publicAPI, model) {
...
+  };
 
-    model.renderingType = RenderingTypes.PICKING_BUFFER;
-    model.widgets.forEach(updateWidgetForRender);
+  publicAPI.renderWidgets = () => {
+    if (model.pickingEnabled && model.captureOn === CaptureOn.MOUSE_RELEASE) {
```
finetjul: Maybe `updateSelectionFromXY()` should also check `model.pickingEnabled`

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -194,18 +194,20 @@ function vtkWidgetManager(publicAPI, model) {
...
+      model.renderingType = RenderingTypes.PICKING_BUFFER;
+      model.widgets.forEach(updateWidgetForRender);
 
-    if (model.captureOn === CaptureOn.MOUSE_RELEASE) {
       const [w, h] = model.openGLRenderWindow.getSize();
       model.selector.setArea(0, 0, w, h);
       model.selector.releasePixBuffers();
       model.pickingAvailable = model.selector.captureBuffers();
       model.previousSelectedData = null;
```
finetjul: maybe those 7 last lines should be factorized with `updateSelectionFromXY()` into a unique picking function

Sources/Widgets/Core/WidgetManager/index.js
floryst: This was to preserve original functionality, where calling `enablePicking` would render the picking and front buffers regardless of `pickingEnabled`. Since some code already uses this, I decided to keep it this way instead of making it a breaking change.

Sources/Widgets/Core/WidgetManager/index.js
floryst: Good point, I'll go ahead and do that.

Sources/Widgets/Core/WidgetManager/index.js
floryst: `updateSelectionFromXY` has an implicit pre-condition that picking will be enabled, but it doesn't hurt to add it to be more explicit.

finetjul:
LGTM

PR #1304, opened 2019-12-20
feat(KeyboardCameraManipulator): Add ability to reset movement speed

psavery:
resetMovementSpeed() generates a movement speed based upon the longest length of the rendering
bounding box. It is intended to just be a reasonable guess as to what a good movement speed would
be. The user is able to modify the movement speed via setMovementSpeed(). If the movement speed is
not set in the options when the KeyboardCameraManipulator is constructed, upon the first movement,
resetMovementSpeed() will be used to generate a reasonable speed.

### DIFF
--- a/Sources/Interaction/Manipulators/KeyboardCameraManipulator/index.js
+++ b/Sources/Interaction/Manipulators/KeyboardCameraManipulator/index.js
@@ -1,4 +1,5 @@
 import macro from 'vtk.js/Sources/macro';
+import vtkBoundingBox from 'vtk.js/Sources/Common/DataModel/BoundingBox';
 import vtkCompositeKeyboardManipulator from 'vtk.js/Sources/Interaction/Manipulators/CompositeKeyboardManipulator';
 import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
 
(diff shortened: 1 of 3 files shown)


jourdain:
lgtm except that usually I prefer to inline the param of `newInstance({ ...});`

psavery:
@jourdain I fixed it.

jourdain:
LGTM, thx

Sources/Interaction/Manipulators/KeyboardCameraManipulator/index.js
```diff
@@ -30,12 +30,35 @@ function vtkKeyboardCameraManipulator(publicAPI, model) {
...
+    const bounds = renderer.computeVisiblePropBounds();
+    const lengths = [
+      Math.abs(bounds[1] - bounds[0]),
+      Math.abs(bounds[3] - bounds[2]),
+      Math.abs(bounds[5] - bounds[4]),
+    ];
+
+    // Just a number that seems to work okay for our examples...
+    const divisor = 500;
+    model.movementSpeed = Math.max(...lengths) / divisor;
```
finetjul: You might want to use BoundingBox::getMaxLength() instead.

Sources/Interaction/Manipulators/KeyboardCameraManipulator/index.js
```diff
@@ -225,7 +248,7 @@ function vtkKeyboardCameraManipulator(publicAPI, model) {
 // ----------------------------------------------------------------------------
 
 const DEFAULT_VALUES = {
-  movementSpeed: 0.02,
+  movementSpeed: null,
```
finetjul: maybe you could add a comment to explain the meaning of "null" (or 0) and what the unit is. (e.g. length travelled by the camera at each frame each 1/60 second)

Sources/Interaction/Manipulators/KeyboardCameraManipulator/index.js
psavery: Done.

Sources/Interaction/Manipulators/KeyboardCameraManipulator/index.js
psavery: Done. And actually, I also changed it to be an explicit check for `null` (as in, if the `movementSpeed === null`, then the `movementSpeed` gets reset when movement starts).

psavery:
I think this is ready for merging unless there are any more requested changes.

PR #1308, opened 2019-12-30
Expose base64 conversion

tjcorona:


tjcorona:
@jourdain Would you mind reviewing this merge request?

jourdain:
LGTM but you need to change the commit message otherwise it won't create a release.

`fix(Base64): Expose helper methods to convert base64 to byteArray`

jourdain:
actually expose it as `vtkBase64`

`import vtkBase64 from './Base64';`

tjcorona:
@jourdain I'm not sure why the  CI is failing. What am I doing wrong?

Sources/Common/Core/index.js
```diff
@@ -13,6 +14,7 @@ import vtkURLExtract from './URLExtract';
 import vtkVariantArray from './VariantArray';
 
 export default {
+  Base64,
```
jourdain: need to be `vtkBase64` now.

tjcorona: thanks!

PR #1310, opened 2020-01-03
fix(vtkRepresentationProxies): Propagate initialValues to constructor…

llangholz89:
… of abstract parent class


llangholz89:
This prevents, among other things, the issue of:
```
rep = vtkGeometryRepresentationProxy.newInstance({ rescaleOnColorBy: false });
console.log(rep.getRescaleOnColorBy())  // true
```

floryst:
LGTM

PR #1316, opened 2020-01-13
feat(Picking): Support returning coordinates of

martinken:
If selector.captureZValues is set to true in the
HardwareSelector then a selection will include both
displayPosition and worldPosition values of the selection.

Where multiple pixels are selected the closest pixel is
used. That is, the pixel closest to the front clipping plane.


martinken:
@jourdain PTAL as my JS is a bit rusty

jourdain:
I'll check tomorrow. Thanks!

PR #1317, opened 2020-01-15
doc(Citation): Add Zenodo DOI badge

thewtex:
Provides a citation for referencing a specific version of the software. The DOI and citation
content is automatically updated on release by Zenodo.


thewtex:
@aylward @wschroed @bradking

PR #1321, opened 2020-01-22
fix(HttpDataSetReader): Only register newly loaded arrays to dataset

llangholz89:
Previously,
```
reader.enableArray('pointData', 'arrayA');
reader.loadData();
```
would not effect the reader's output dataset, because `arrayA` had already been registered through `pointData.addArray`.

This change ensures array registration only occurs when the array data has been fetched by `reader.loadData()`.

It seems to work for my use case, but if I'm missing something please let me know!


jourdain:
I don't remember the details but it seems reasonable.

PR #1322, opened 2020-01-29
Scene loader with jpg texture

jourdain:


jourdain:
FYI @psavery

jourdain:
@floryst it might be good to update Glance to support scene with textures along with the fix that @psavery did so we can provide an updated standalone version on ParaView bundle.

We might not be ready for the ParaView RC2 which is going to be cut today at 1pm.

PR #1324, opened 2020-01-30
docs(widgets): Add initial mixin docs

floryst:
Adds some docs for mixins. Specifically meant to clarify the colorMixin.


thewtex:
Thanks for adding this @floryst !

PR #1335, opened 2020-02-19
fix(DracoReader): Provide correct argument

zac92:


jourdain:
Can you rename your commit to follow semantic-release?
Something like `fix(DracoReader): Provide correct argument`

jourdain:
Thank you for your contribution.

PR #1336, opened 2020-02-19
feat(Rendering): add wide line support

martinken:
Add basic wide line support that relies on the lineWidth
function in webgl. Note that lineWidth in webgl is only
required to support a width of 1.0 (how dumb is that) so
this implementation may not work on your system. Support
for other lineWidths is vendor dependent.

### DIFF
--- a/Sources/Rendering/OpenGL/PolyDataMapper/index.js
+++ b/Sources/Rendering/OpenGL/PolyDataMapper/index.js
@@ -1661,6 +1661,8 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
       actor.getProperty().getEdgeVisibility() &&
       representation === Representation.SURFACE;
 
+    gl.lineWidth(actor.getProperty().getLineWidth());
+
     // for every primitive type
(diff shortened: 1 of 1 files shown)


Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1680,6 +1682,8 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
         model.primitiveIDOffset += cabo.getElementCount() / stride;
       }
     }
+    // reset the line width
+    gl.lineWidth(1);
```
danmindru: hi @martinken.
Unless I misunderstanding something, wouldn't this undo L:1664 and set the line width to `1` anyway?

martinken: The lines are drawn between the two calls. We reset it to 1 so that it doesn't impact other mappers etc.

danmindru: got it. Makes sense, thanks for the explanation.

PR #1340, opened 2020-02-20
Fix OpenglRenderWindow::SetContainer

BotellaA:
Do not attach the background image on an undefined container


jourdain:
Can you rewrite the commit like so `fix(OpenGLRenderWindow): Only attach bg image when valid container available`

PR #1341, opened 2020-02-21
fix(WSLinkClient): delay notification when idle

BotellaA:


jourdain:
Your change makes the notification happen right away when idle and later when busy. The comment suggest we rather have it the other way.

PR #1343, opened 2020-02-22
docs: Fix typos identified using codespell

jcfr:


jcfr:
@floryst @jourdain  Let me know if you think `fix: Fix typo in ivar identified using codespell` should be categorized as a breaking change or not.

Sources/Rendering/Core/Camera/index.js
```diff
@@ -380,7 +380,7 @@ function vtkCamera(publicAPI, model) {
     let t = thickness;
     if (t < 1e-20) {
       t = 1e-20;
-      vtkDebugMacro('Thickness is set to minimum.');
+      vtkDebugMacro('Thi./Sources/Rendering/Core/Camera/index.jsckness is set to minimum.');
```
finetjul: looks odd to me

jcfr: Good catch

jcfr: I don't have access to my dev computer (at the airport now), is it something you could help address ? If not, I will take care of it when I return on March 2nd.

jourdain:
This could even be a `chore(spell): Update doc and iVar` as no API change is truly happening.

jourdain:
Was fixed in https://github.com/Kitware/vtk-js/pull/1344

PR #1346, opened 2020-02-25
clean up coordinate system names

martinken:
view is now consistent between shaders and vtkCoordinate
projection is a new vtkCoordinate value that matches
what shaders used to call device coordinates.

This also makes the camer methods make more sense as
GetViewMatrix actually returns WCVC matrix and
GetProjectionMatrix returns VCPC


martinken:
@jourdain PTAL

jourdain:
seems reasonable except the commit text. ;-)

martinken:
oops fixed

jourdain:
Once thing I noticed though...

Is that ok that to matrix as only Float32?
For better precision, you should:
- replace ` mat3.create()` by  `new Float64Array(9)`
- replace ` mat4.create()` by  `new Float64Array(16)`

PR #1348, opened 2020-02-26
fix(vtkOpenGLRenderWindow): Fix wrong call to projectionToNormalizedDisplay

boucaud:
Fixes a small bug where a call to projectionToNormalizedDisplay was done as a member of the renderWindow (method doesn't exist) instead of the renderer.


boucaud:
@jourdain

martinken:
+1

PR #1360, opened 2020-03-12
feat(vtkColorTransferFunction): Add discretize option

boucaud:
@finetjul

### DIFF
--- a/Sources/Rendering/Core/ColorTransferFunction/index.js
+++ b/Sources/Rendering/Core/ColorTransferFunction/index.js
@@ -494,6 +494,28 @@ function vtkColorTransferFunction(publicAPI, model) {
         x = 0.5 * (xStart + xEnd);
       }
 
+      // Linearly map x from mappingRange to [0, numberOfValues-1],
+      // discretize (round down to the closest integer),
+      // then map back to mappingRange
+      if (model.discretize) {
+        const range = model.mappingRange;
+        if (x >= range[0] && x <= range[1]) {
+          const numberOfValues = model.numberOfValues;
+          const deltaRange = range[1] - range[0];
+          if (numberOfValues <= 1) {
+            x = range[0] + deltaRange / 2.0;
+          } else {
+            // normalize x
+            const xn = (x - range[0]) / deltaRange;
+            // discretize
+            const discretizeIndex = vtkMath.floor(numberOfValues * xn);
+            // get discretized x
+            x =
+              range[0] + (discretizeIndex / (numberOfValues - 1)) * deltaRange;
+          }
+        }
+      }
+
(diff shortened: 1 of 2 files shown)


Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -494,6 +494,24 @@ function vtkColorTransferFunction(publicAPI, model) {
         x = 0.5 * (xStart + xEnd);
       }
 
+      if (model.discretize) {
+        const range = model.mappingRange;
+        if (x > range[0] && x < range[1]) {
+          const numberOfValues = model.numberOfValues;
+          const deltaRange = range[1] - range[0];
+          if (numberOfValues <= 1) {
+            x = range[0] + deltaRange / 2.0;
```
finetjul: if numberOfValues == 1 then you should have the same color for x = range[0], x = range[1] and x = (range[0]+range[1]) /2)
here you will have 3 colors (because of your `if` condition line 499)

Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -494,6 +494,25 @@ function vtkColorTransferFunction(publicAPI, model) {
         x = 0.5 * (xStart + xEnd);
       }
 
+      if (model.discretize) {
```
finetjul: Please add a comment to explain the idea. E.g. 
Linearly map x to [0, numberOfValues-1], round down to the closest integer (i.e. discretize) and map back to mappingRange.

finetjul:
LGTM

finetjul:
@jourdain @martinken what do you think ? Is it the right approach to inject discretization directly in vtkColorTransferFunction? 

It is clearly different frmo VTK-C++. Here the idea is to have a unique class for color transfer function where VTK-C++ has 2 classes: vtkColorTransferFunction and vtkDiscretizableColorTransferFunction. 

vtkDiscretizableColorTransferFunction seems confusing to us partly because it internally uses a vtkLookupTable. But maybe we ignored some subtilities...

jourdain:
@martinken has a better grasp on the color handling within VTK and its origin. I let him comment out.
Hopefully we can take an easier and less confusing path.

martinken:
The VTK C++ arrangement is confusing to me so having something clearer would be great. Thinking out loud here we have
ScalarsToColors -> should stay really generic, more of an api definition
 - lookuptable - uses hue/sat ranges to implement
 - ctf - uses piecewise functions to implement

So this MR modifies the output of ctf, I would not add another class. Keep it here like you have done IMO.

finetjul:
thanks @martinken

PR #1378, opened 2020-04-01
fix(ProxyManager): Consistent args to proxy registration callback for…

llangholz89:
… register and unregister


llangholz89:
I could not think of a good reason why the proxy was not included in the registration callback when unregistering. It has not been deleted yet and, at least in my case, could potentially be used for clean up purposes.  If there's something I'm missing please let me know and feel free to disregard the PR!

PR #1384, opened 2020-04-03
fix(ImageProperty): Return true/false on set methods

llangholz89:
I like the new multi-component options for slice rendering, but we have an extension of `vtkImageProperty` and saw some issues because
```
if (superSetRGBTransferFunction(func)) {
  ...
}
```
no longer works.

I went ahead and updated the `vtkVolumeProperty` as well for consistency.

### DIFF
--- a/Sources/Rendering/Core/ImageProperty/index.js
+++ b/Sources/Rendering/Core/ImageProperty/index.js
@@ -111,11 +111,13 @@ function vtkImageProperty(publicAPI, model) {
     return model.componentData[index].componentWeight;
   };
 
-  publicAPI.setInterpolationTypeToNearest = () =>
-    publicAPI.setInterpolationType(InterpolationType.NEAREST);
+  publicAPI.setInterpolationTypeToNearest = () => {
+    return publicAPI.setInterpolationType(InterpolationType.NEAREST);
+  };
 
-  publicAPI.setInterpolationTypeToLinear = () =>
-    publicAPI.setInterpolationType(InterpolationType.LINEAR);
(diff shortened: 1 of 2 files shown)


jourdain:
LGTM

jourdain:
@scottwittenburg it is important to keep the return behavior of setters.

scottwittenburg:
> @scottwittenburg it is important to keep the return behavior of setters.

At first it wasn't clear to me how that PR changed the return behavior of any methods.  But I guess the issue is that the auto-generate get/set methods had a return value?

jourdain:
Yes

jourdain:
Do we know why the branch is failing?

scottwittenburg:
> Do we know why the branch is failing?

Looks like some style checks failing, e.g.:

```
/home/vsts/work/1/s/Sources/Rendering/Core/VolumeProperty/index.js
      183:7  error  Arrow function expected no return value  consistent-return
      185:5  error  Arrow function expected no return value  consistent-return
```

llangholz89:
Interesting. Is that implying that 
```
publicAPI.setInterpolationTypeToNearest = () =>
    publicAPI.setInterpolationType(InterpolationType.NEAREST);
```
needs to be written as
```
publicAPI.setInterpolationTypeToNearest = () => {
    return publicAPI.setInterpolationType(InterpolationType.NEAREST);
};
```
?

jourdain:
That's weird. Let's make the explicit return path as it would be more obvious for the next dev.

jourdain:
Already merged

PR #1386, opened 2020-04-05
fix(ImageProperty): Return true/false on set methods

jourdain:


Sources/Rendering/Core/ImageProperty/index.js
```diff
@@ -86,14 +90,16 @@ function vtkImageProperty(publicAPI, model) {
   publicAPI.setComponentWeight = (index, value) => {
     if (index < 0 || index >= VTK_MAX_VRCOMP) {
       vtkErrorMacro('Invalid index');
-      return;
+      return false;
```
jourdain, line 93: that was the issue FYI @laurennlam

Sources/Rendering/Core/VolumeProperty/index.js
```diff
@@ -159,14 +173,16 @@ function vtkVolumeProperty(publicAPI, model) {
   publicAPI.setComponentWeight = (index, value) => {
     if (index < 0 || index >= VTK_MAX_VRCOMP) {
       vtkErrorMacro('Invalid index');
-      return;
+      return false;
```
jourdain, line 176: that was the issue FYI @llangholz89

Sources/Rendering/Core/ImageProperty/index.js
llangholz89: Aha! Mystery solved, thanks!

Issue #1377, opened 2020-04-01
Coincident topology and vtkImageMapper


boucaud:
Hello,
We are having issues with coincident topology and the image mapper, and we would like to be able to use setRelativeCoincidentTopologyPolygonOffsetParameters with vtkImageMapper to fix those issues.

I tried to implement this in a simple way, and got it to work by essentially doing this:

1. Derive vtkImageMapper from vtkMapper instead of vtkAbstractMapper
2. Add shader replacements/set uniforms in vtkOpenGLImageMapper

My question is about this first step. I'm not sure whether it is to be avoided, and whether it has side effects that I didn't notice. An alternative would be to move the coincident topology methods up to vtkAbstractMapper, or to create an intermediary class.

What do you think ? Does this approach seem ok, or do you prefer one of the alternatives ? Did I miss something ?
Thanks,
Adrien

finetjul:
@martinken @jourdain what do you think ?

PR #1387, opened 2020-04-06
feat(vtkImageMapper): support coincident topology

boucaud:
re #1377, this is the proposed implementation.

@finetjul fyi

### DIFF
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -17,6 +17,7 @@ import { InterpolationType } from 'vtk.js/Sources/Rendering/Core/ImageProperty/C
 
 import vtkPolyDataVS from 'vtk.js/Sources/Rendering/OpenGL/glsl/vtkPolyDataVS.glsl';
 import vtkPolyDataFS from 'vtk.js/Sources/Rendering/OpenGL/glsl/vtkPolyDataFS.glsl';
+import vtkReplacementShaderMapper from 'vtk.js/Sources/Rendering/OpenGL/ReplacementShaderMapper';
 
 const { vtkErrorMacro } = macro;
 
@@ -74,6 +75,13 @@ function vtkOpenGLImageMapper(publicAPI, model) {
     if (prepass) {
       publicAPI.render();
     }
+  };
+
+  publicAPI.getCoincidentParameters = (ren, actor) => {
+    if (model.renderable.getResolveCoincidentTopology()) {
+      return model.renderable.getCoincidentTopologyPolygonOffsetParameters();
+    }
+    return null;
   };
 
   // Renders myself
@@ -300,6 +308,8 @@ function vtkOpenGLImageMapper(publicAPI, model) {
 
     shaders.Vertex = VSSource;
     shaders.Fragment = FSSource;
+
+    publicAPI.replaceShaderCoincidentOffset(shaders, ren, actor);
   };
 
   publicAPI.getNeedToRebuildShaders = (cellBO, ren, actor) => {
@@ -471,6 +481,16 @@ function vtkOpenGLImageMapper(publicAPI, model) {
       cellBO
         .getProgram()
         .setUniformi('depthRequest', model.renderDepth ? 1 : 0);
+    }
+
+    // handle coincident
+    if (cellBO.getProgram().isUniformUsed('coffset')) {
+      const cp = publicAPI.getCoincidentParameters(ren, actor);
+      cellBO.getProgram().setUniformf('coffset', cp.offset);
+      // cfactor isn't always used when coffset is.
+      if (cellBO.getProgram().isUniformUsed('cfactor')) {
+        cellBO.getProgram().setUniformf('cfactor', cp.factor);
+      }
     }
 
     const texColorUnit = model.colorTexture.getTextureUnit();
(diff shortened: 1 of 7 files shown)


Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -94,6 +101,75 @@ function vtkOpenGLImageMapper(publicAPI, model) {
     shaders.Geometry = '';
   };
 
+  publicAPI.replaceShaderCoincidentOffset = (shaders, ren, actor) => {
```
finetjul, line 104: This is duplicate with OpenGL/PolyDataMapper/index.js.

Maybe you can create an OpenGL/ReplacementShaderMapper/index.js that contains a `replaceShaderCoincidentOffset = (publicAPI, model, shaders, ren, actor)` function, that is binded here ?
```
import {replaceShaderCoincidentOffset} from 'ReplacementShaderMapper/index.js';
...
publicAPI.replaceShaderCoincidentOffset = replaceShaderCoincidentOffset.bind(publicAPI, model);
```

finetjul:
Please note that this commit may break backward compatibility due to moving ImageMapper superclass from vtkAbstractMapper to vtkMapper.
Please flag commit accordingly.

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -94,6 +104,12 @@ function vtkOpenGLImageMapper(publicAPI, model) {
     shaders.Geometry = '';
   };
 
+  publicAPI.replaceShaderCoincidentOffset = replaceShaderCoincidentOffset.bind(
+    null,
+    publicAPI,
+    model
+  );
```
finetjul, line 111: @jourdain is there something more elegant doable here ? would `bind(this)` work ?

jourdain: Just do like everything else (closure). 

`createReplaceShaderCoincidentOffset(publicAPI, model) => (...) => {}`

jourdain: notice the `create` which try to make things more explicit.

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -94,6 +104,11 @@ function vtkOpenGLImageMapper(publicAPI, model) {
     shaders.Geometry = '';
   };
 
+  publicAPI.replaceShaderCoincidentOffset = createReplaceShaderCoincidentOffset(
+    publicAPI,
+    model
+  );
```
jourdain, line 110: Sorry for keep changing my mind but I don't necessary have a good picture of what you are doing but what about doing something like assuming the binding of the functions to publicAPI should always be the same ones and so on... What do you think?

```
  // Inheritance
  vtkViewNode.extend(publicAPI, model, initialValues);
  vtkReplacementShaderMapper.implementReplaceShaderCoincidentOffset(publicAPI, model, initialValues);
  [...]
}
```

boucaud: Is this new change what you had in mind ?

jourdain: Basically when you start providing `(publicAPI, model)` to a function to generate something it really feels like our `extend` methods and therefore it could be used that way rather than rewiring methods outside of that call. But again, I don't have a good picture of what you are doing. 
So I don't know, I'm just making suggestion here on how that code could be refactored.

So far, your change is "good enough" and does not require anything else.

boucaud: Well we're basically trying to factorize a single method between Image and Polydata mappers to avoid duplicates.

As it's only one function, it would be a bit heavy handed to create a parent class, so indeed the current way of adding this function to `publicAPI` during `extend` seems like a good compromise.

jourdain: This is technically what you already did, not a new class hierarchy. But since you already provide `publicAPI` there is no reason that you can not do the following:

```
function implementReplaceShaderCoincidentOffset(publicAPI, model) {
   publicAPI.replaceShaderCoincidentOffset = (a,b,c) => {};
}
++
implementReplaceShaderCoincidentOffset(publicAPI, model);
```

Rather than doing

```
function createReplaceShaderCoincidentOffset(publicAPI, model) {
   return (a,b,c) => {};
}

++
publicAPI.replaceShaderCoincidentOffset = createReplaceShaderCoincidentOffset(
    publicAPI,
    model
  );
```

This is especially true if the methods name is always the same and therefore no typo can be made...

boucaud: I completely agree. I changed the implementation to apply this suggestion in the latest change.

finetjul: This is interesting, thanks @jourdain .

Now the question is, should the same mechanism be done for exposing "resolve coincident 
 topology" parameters to vtkImageMapper?

Current implementation changes the superclass of vtkImageMapper to be vtkMapper instead of vtkAbstractMapper but maybe that is not needed or worst, that is not wanted. The difference between vtkAbstractMapper, vtkAbstractMapper3D and vtkMapper is not clear to me.

jourdain: @martinken might be able to shade some light. 

Otherwise with vtk-js I tried to enable as much composition as possible to prevent writing too much code and reduce code mistake.
 
The macro.js is indeed providing the common method generator, but other class could start gathering several `implementation` rather than having only one like you have in your `ReplacementShaderMapper`.

finetjul:
Please update ImageMapper documentation (.md) accordingly.

jourdain:
LGTM

martinken:
LGTM +1

PR #1388, opened 2020-04-06
fix(WidgetManager): Guard against object deletion

### LINKED ISSUE #1382: Uncaught (in promise) TypeError: Cannot read property 'appendChild' of undefined
This happens in `WidgetManager`::`updateSvg()` promise, line `model.svgRoot.appendChild(node);`

This is because at the time the promise is ran, `WidgetManager` has already been destroyed. 

@floryst  Is the fix as simple as `return`ing if `model.deleted` is `true` ?

floryst:
WidgetManager may be deleted between updateSvg() and the promise
resolution callback.

Fixes #1382 @finetjul


finetjul:
Thanks @floryst

PR #1389, opened 2020-04-06
fix(ImageMapper): Provide reasonable default if no function

scottwittenburg:
If no piecewise function is specified for a given component, fill with the max value as a default.

### DIFF
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -690,18 +690,23 @@ function vtkOpenGLImageMapper(publicAPI, model) {
 
         for (let c = 0; c < numIComps; ++c) {
           pwfun = actorProperty.getPiecewiseFunction(c);
-          const pwfRange = pwfun.getRange();
-          pwfun.getTable(pwfRange[0], pwfRange[1], pwfWidth, tmpTable, 1);
-          // adjust for sample distance etc
-          if (iComps) {
-            for (let i = 0; i < pwfWidth; i++) {
-              pwfFloatTable[c * pwfWidth * 2 + i] = 255.0 * tmpTable[i];
-              pwfFloatTable[c * pwfWidth * 2 + i + pwfWidth] =
-                255.0 * tmpTable[i];
-            }
+          if (pwfun === null) {
+            // Piecewise constant max if no function supplied for this component
+            pwfFloatTable.fill(255.0);
           } else {
-            for (let i = 0; i < pwfWidth; i++) {
-              pwfFloatTable[c * pwfWidth * 2 + i] = 255.0 * tmpTable[i];
+            const pwfRange = pwfun.getRange();
+            pwfun.getTable(pwfRange[0], pwfRange[1], pwfWidth, tmpTable, 1);
+            // adjust for sample distance etc
+            if (iComps) {
+              for (let i = 0; i < pwfWidth; i++) {
+                pwfFloatTable[c * pwfWidth * 2 + i] = 255.0 * tmpTable[i];
+                pwfFloatTable[c * pwfWidth * 2 + i + pwfWidth] =
+                  255.0 * tmpTable[i];
+              }
+            } else {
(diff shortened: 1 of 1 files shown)


scottwittenburg:
@thewtex To me, this seems like a reasonable thing to do here, but another option is changing the viewer to always supply a pwf for each component.

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -690,18 +690,23 @@ function vtkOpenGLImageMapper(publicAPI, model) {
 
         for (let c = 0; c < numIComps; ++c) {
           pwfun = actorProperty.getPiecewiseFunction(c);
```
thewtex, line 692: Re: line 682: `pwfun` in `pwfunToString`: this is now not including all component piecewise functions; Should `pwfunToString` include these?

Note: the VolumeMapper is only using the property `MTime`:

https://github.com/Kitware/vtk-js/blob/bb172973bbf0fc5d6a734c3475f5cb9e12f91ed7/Sources/Rendering/OpenGL/VolumeMapper/index.js#L1099-L1100

Is this another issue, or do we not need to use the piecewise function `MTime` in this situation?

@martinken

scottwittenburg: @thewtex I think we should consider all component functions when computing that string, because the buffer is composed of rows of pixels from all the components, so if any of those functions change, we need to rebuild the whole buffer.  I've added another commit to this PR which does that for per-component color transfer functions and piecewise functions.

I think the volume mapper probably needs the same logic, what do you think?

jourdain: Does `vprop.getMTime()` return the Max of all mtimes including all `actorProperty.getPiecewiseFunction(n)` and itself mtime? 

If so, you are good.

thewtex: It looks like the `VolumeProperty` is using all piecewise function `MTime`'s in estimating its `MTime`:

https://github.com/Kitware/vtk-js/blob/f351cbc8a1790f9238f29b1706e6c28b7d901136/Sources/Rendering/Core/VolumeProperty/index.js#L19-L56

and we have a similar construct for the `ImageProperty`:

https://github.com/Kitware/vtk-js/blob/f351cbc8a1790f9238f29b1706e6c28b7d901136/Sources/Rendering/Core/ImageProperty/index.js#L17-L38

so just the image property `MTime` will be sufficient here.

scottwittenburg: Yes, I see now that the `MTime` of the component functions are used in coming up with the `MTime` of the image property.  However, I think we'll still need to include some extra things in the string, which are not considered when computing the `MTime` of the image property, but for which changes would require rebuilding the buffer.  For example, whether or not the image property uses independent components and the number of components.

jourdain: You know best what is required but taking the max of all your dependencies mtimes would be enough by design. Make sure you include all your dependencies.

scottwittenburg: Thanks @jourdain.  

The `independentComponents` field on the `ImageProperty` corresponds to auto-generated get/set methods.  But I'm looking at `macro.js`, and it seems the generated `set` methods *do* update the objects `MTime`, so in that case I wouldn't need to add this field explicitly to the string.

However, the `ImageProperty` doesn't know the number of components in the image, and if that changes with `independentComponents` turned on, then the color/tfun buffers will need to be rebuilt.

Just thinking out loud in some way, and recording that thought process so @thewtex will be aware.

thewtex: Then, if the piecewise function check includes 1.) the `ImageProperty` `MTime` and 2.) `independentComponents`?

I am wondering if extra string state generation is produces more of a performance hit than it is saving? :-)

PR #1397, opened 2020-04-10
More imagemapper fixes

### LINKED ISSUE #1396: ImageMapper rendering issues
Since v13.12.0, there are some broken behavior. Please find below 2 issues I found:
 - the [ImageOutlineFilter ](https://kitware.github.io/vtk-js/examples/ImageOutlineFilter.html) example renders all black.
 - Global plane opacity on vtkImageSlice property no longer works:
```
const actor = vtkImageSlice.newInstance();
...
actor.getProperty().setOpacity(0.5);
```

scottwittenburg:
This PR reduces the computation done to generate the lookup table and piecewise function strings used to determine whether those buffers should be rebuilt, while still causing them to rebuilt when necessary.  Also addresses an issue introduced by changing `setScalarOpacity` method to `setPiecewiseFunction` without providing the backward compatibility allowing the method to be called with a single argument.

Fixes #1396

### DIFF
--- a/Sources/Rendering/Core/ImageProperty/index.js
+++ b/Sources/Rendering/Core/ImageProperty/index.js
@@ -60,8 +60,14 @@ function vtkImageProperty(publicAPI, model) {
 
   // Set the piecewise function
   publicAPI.setPiecewiseFunction = (index, func) => {
-    if (model.componentData[index].piecewiseFunction !== func) {
-      model.componentData[index].piecewiseFunction = func;
+    let idx = index;
+    let transferFunc = func;
+    if (!Number.isInteger(index)) {
+      transferFunc = index;
+      idx = 0;
+    }
+    if (model.componentData[idx].piecewiseFunction !== transferFunc) {
+      model.componentData[idx].piecewiseFunction = transferFunc;
       publicAPI.modified();
       return true;
     }
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -731,20 +731,19 @@ function vtkOpenGLImageMapper(publicAPI, model) {
           pwfun = actorProperty.getPiecewiseFunction(c);
           if (pwfun === null) {
             // Piecewise constant max if no function supplied for this component
-            pwfFloatTable.fill(255.0);
+            pwfFloatTable.fill(1.0);
           } else {
             const pwfRange = pwfun.getRange();
             pwfun.getTable(pwfRange[0], pwfRange[1], pwfWidth, tmpTable, 1);
             // adjust for sample distance etc
             if (iComps) {
               for (let i = 0; i < pwfWidth; i++) {
-                pwfFloatTable[c * pwfWidth * 2 + i] = 255.0 * tmpTable[i];
-                pwfFloatTable[c * pwfWidth * 2 + i + pwfWidth] =
-                  255.0 * tmpTable[i];
+                pwfFloatTable[c * pwfWidth * 2 + i] = tmpTable[i];
+                pwfFloatTable[c * pwfWidth * 2 + i + pwfWidth] = tmpTable[i];
               }
             } else {
               for (let i = 0; i < pwfWidth; i++) {
-                pwfFloatTable[c * pwfWidth * 2 + i] = 255.0 * tmpTable[i];
+                pwfFloatTable[c * pwfWidth * 2 + i] = tmpTable[i];
               }
             }
           }
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -28,21 +28,10 @@ const { SlicingMode } = Constants;
 // ----------------------------------------------------------------------------
 
 function computeFnToString(property, fn, numberOfComponents) {
-  let pwfun = fn.apply(property);
+  const pwfun = fn.apply(property);
   if (pwfun) {
     const iComps = property.getIndependentComponents();
-    let pwfunToString = `${property.getMTime()}-${iComps}-0-${pwfun.getMTime()}`;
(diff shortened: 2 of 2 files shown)


Sources/Rendering/Core/ImageProperty/index.js
```diff
@@ -60,8 +60,14 @@ function vtkImageProperty(publicAPI, model) {
 
   // Set the piecewise function
   publicAPI.setPiecewiseFunction = (index, func) => {
```
jourdain, line 62: Why don't we use `(func, index=0)` if it is only new code that use the second argument?

thewtex: The reason this approach was taken previously was to maintain a consistent API for these functions throughout vtk.js / VTK -- the index comes first for functions like these.

scottwittenburg: Yes, that was my inclination as well.  But the volume property uses the (index, func) ordering for similar methods, and supposedly other places as well, according to @thewtex, so he preferred it this way.

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -742,20 +731,19 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
           } else {
             const pwfRange = pwfun.getRange();
             pwfun.getTable(pwfRange[0], pwfRange[1], pwfWidth, tmpTable, 1);
             // adjust for sample distance etc
             if (iComps) {
               for (let i = 0; i < pwfWidth; i++) {
-                pwfFloatTable[c * pwfWidth * 2 + i] = 255.0 * tmpTable[i];
-                pwfFloatTable[c * pwfWidth * 2 + i + pwfWidth] =
-                  255.0 * tmpTable[i];
+                pwfFloatTable[c * pwfWidth * 2 + i] = tmpTable[i];
```
scottwittenburg, line 741: @thewtex These do need to be in the range `[0.0, 1.0]`, rather than `[0.0, 255.0]` in order to be used as opacity values in the case of dependent components.   The reason it was working for independent components, even though they weren't in that range, is due to how we normalize them per-fragment anyway.

finetjul:
LGTM

Sources/Rendering/Core/ImageProperty/index.js
jourdain: Well we already decided to not follow VTK/C++ for `setInputData(ds, port=0)` as 95% of the case you only have 1 port. So 95% of the case, you can skip the port and have an easy to read code.

This use case is exactly the same... 

Since JS does not allow to easily override methods based on argument types, using C++ as reference is probably the best move.

Sources/Rendering/OpenGL/ImageMapper/index.js
thewtex: @scottwittenburg :+1: :1st_place_medal:

Sources/Rendering/Core/ImageProperty/index.js
finetjul: I did not know about such rule, but I guess it makes sense.
@jourdain , is there some documentation somewhere related to differences between VTK and VTK.JS ? Thanks.

jourdain: This is not really a rule but more something to take into account when you create a JavaScript library meant to be used by JS developers and not VTK/C++ developers. It is true that we want to map as much as possible to what has been done in VTK/C++ that worked well. But if we can fix design issue or provide improvements when possible, we should not prevent it just because *we don't follow the C++ way*.
We should enforce code simplicity and prevent user/developer error. VTK is complex enough.

I've been wanted to write more about the macro stuff and provide a real API doc and compatibility map, but unfortunately I never got a chance to do it. So I do not have anything more than [that](https://kitware.github.io/vtk-js/docs/develop_class.html) to offer.

@floryst did a great job explaining the new Widget architecture that is nowhere close to what VTK/C++ is doing but actually solve many issue that medical are having with VTK/C++.

thewtex: @jourdain yes, you are right -- it make sense to make API improvements when possible and that are appropriate to JavaScript. 

But, do you agree that there will be less errors when it is:

```
volumeProp.setRGBTransferFunction(index1, func1)
volumeProp.setScalarOpacity(index2, func2)
imageProp.setRGBTransferFunction(index3, func3)
imageProp.setPiecewiseFunction(index3, func4)
```

versus:

```
volumeProp.setRGBTransferFunction(index1, func1)
volumeProp.setScalarOpacity(index2, func2)
imageProp.setRGBTransferFunction(index3, func3)
imageProp.setPiecewiseFunction(func4, index4)
```

In the first, a new developer can confidently guess what the order is without referencing the API everytime they make a call or without knowing whether the history of whether it came from C++.

jourdain:
I prefer consistency and would put the index in 2nd if possible.
If not possible because those api have been defined like that for a while,
then so be it. Let’s keep (index, fn).

If we have many (fn, index) arg we could create a helper to extract them
regardless the order. I don’t want several implementation around.

What started to worry me was the bug introduced because of the add-on index
arg. And the “complex” added code to solve the problem.

thewtex:
The CI build test failures appear to have started with this merge:

https://dev.azure.com/Kitware/vtk-js/_build?definitionId=2&_a=summary

scottwittenburg:
Yeah, that view makes it pretty clear.  I don't get why all the checks passed before merging this PR, but then I guess failed on the merge to master afterwards.

scottwittenburg:
> The CI build test failures appear to have started with this merge:
> 
> https://dev.azure.com/Kitware/vtk-js/_build?definitionId=2&_a=summary

As I dig into this a bit more, right where those tests start failing is also precisely when they start to get run with Firefox 75 instead of Firefox 71.

PR #1399, opened 2020-04-11
feat(Prop3D): add setOrientation function

xavArtley:
Implement the function setOrientation to take into account an initial orientation of vtkProp3D objects

### DIFF
--- a/Sources/Rendering/Core/Prop3D/index.js
+++ b/Sources/Rendering/Core/Prop3D/index.js
@@ -87,7 +87,7 @@ function vtkProp3D(publicAPI, model) {
       y === model.orientation[1] &&
       z === model.orientation[2]
     ) {
-      return;
+      return false;
     }
     model.orientation = [x, y, z];
     mat4.identity(model.rotation);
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/Prop3D/index.js
```diff
@@ -81,6 +81,22 @@ function vtkProp3D(publicAPI, model) {
     publicAPI.modified();
   };
 
+  publicAPI.setOrientation = (x, y, z) => {
+    if (
+      x === model.orientation[0] &&
+      y === model.orientation[1] &&
+      z === model.orientation[2]
+    ) {
+      return;
```
jourdain, line 90: This needs to return false.

Sources/Rendering/Core/Prop3D/index.js
```diff
@@ -81,6 +81,22 @@ function vtkProp3D(publicAPI, model) {
...
+    }
+    model.orientation = [x, y, z];
+    mat4.identity(model.rotation);
+    publicAPI.rotateZ(z);
+    publicAPI.rotateX(x);
+    publicAPI.rotateY(y);
+    publicAPI.modified();
+  };
```
jourdain, line 98: need to return true.

PR #1400, opened 2020-04-13
feat: improve support for webGL2 features

boucaud:
This brings a few fixes and features to some of the WebGL utility classes, with a focus on improving
WebGL2 support:

* Allow attaching and detaching multiple color textures to framebuffers if WebGL 2 is enabled
* Allow attaching and detaching a depth texture to framebuffers if WebGL 2 is enabled
* Fix an error in vtkFrameBuffer.releaseGraphicsResources
* Use user-specified internal texture type if present
* Fix an error in ShaderProgram
* Use 32 bit floats for depth textures (when type is Float32Array and with WebGL 2)

### DIFF
--- a/Sources/Rendering/OpenGL/Framebuffer/index.js
+++ b/Sources/Rendering/OpenGL/Framebuffer/index.js
@@ -69,22 +69,87 @@ function vtkFramebuffer(publicAPI, model) {
     model.glFramebuffer.height = height;
   };
 
-  publicAPI.setColorBuffer = (texture, mode) => {
+  publicAPI.setColorBuffer = (texture, attachment = 0) => {
     const gl = model.context;
+
+    let glAttachment = gl.COLOR_ATTACHMENT0;
+    if (attachment > 0) {
+      if (model.openGLRenderWindow.getWebgl2()) {
+        glAttachment += attachment;
+      } else {
+        macro.vtkErrorMacro(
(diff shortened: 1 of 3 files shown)


Sources/Rendering/OpenGL/Framebuffer/index.js
```diff
@@ -69,22 +69,87 @@ function vtkFramebuffer(publicAPI, model) {
...
+    gl.framebufferTexture2D(
+      gl.FRAMEBUFFER,
+      glAttachment,
+      gl.TEXTURE_2D,
+      null,
+      0
+    );
+  };
+
+  publicAPI.setDepthBuffer = (texture, mode) => {
```
finetjul, line 120: mode -> attachment to be consistent with setColorBuffer ?

boucaud: I believe you can only attach a single depth buffer.

`mode` was unused here though, so I removed it.

boucaud:
Failing tests seem to also be failing on master (locally at least). It seems to be an issue with webGL contexts not being destroyed. I'm looking for the cause.

boucaud:
@finetjul as I mentionned earlier, tests fail even with no changes

boucaud:
@martinken can you review this ?

PR #1409, opened 2020-04-21
feat(Box): Add intersectPlane for vtkBox

lihouxing-sjtu:
Add a static function intersectPlane for vtkBox , the code is borrowed from C++ version

### DIFF
--- a/Sources/Common/DataModel/Box/index.js
+++ b/Sources/Common/DataModel/Box/index.js
@@ -1,5 +1,6 @@
 import macro from 'vtk.js/Sources/macro';
 import vtkBoundingBox from 'vtk.js/Sources/Common/DataModel/BoundingBox';
+import vtkPlane from 'vtk.js/Sources/Common/DataModel/Plane';
 
 // ----------------------------------------------------------------------------
 // Global methods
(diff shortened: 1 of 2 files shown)


jourdain:
Thanks for your contribution.

Sources/Common/DataModel/Box/index.js
```diff
@@ -82,12 +83,45 @@ function intersectBox(bounds, origin, dir, coord, tolerance) {
...
 
+// Plane intersection with box
+// The plane is infinite in extent and defined by an origin and normal.The function indicates
+// whether the plane intersects, not the particulars of intersection points and such
+// The function returns non-zero if the plane and box intersect; zero otherwise.
+function intersectPlane(bounds, origin, normal) {
+  const p = [];
+  let d = 0;
+  let sign = 1;
+  let firstOne = 1;
```
finetjul, line 94: here you could use `sign=undefined`  to know if it is the first time is has been set

Sources/Common/DataModel/Box/index.js
```diff
@@ -82,12 +83,45 @@ function intersectBox(bounds, origin, dir, coord, tolerance) {
...
+
+  // Evaluate the eight points. If there is a sign change, there is an intersection
+  for (let z = 4; z <= 5; ++z) {
+    p[2] = bounds[z];
+    for (let y = 2; y <= 3; ++y) {
+      p[1] = bounds[y];
+      for (let x = 0; x <= 1; ++x) {
+        p[0] = bounds[x];
+        d = vtkPlane.evaluate(normal, origin, p);
```
finetjul, line 103: `d` could have been declared only here

PR #1410, opened 2020-04-21
fix(vtkOpenGLRenderWindow): Only update canvas size when needed

boucaud:
Firefox 75 introduced a regression in vtk-js. The LabelWidget and HandleWidget examples for example show some strange flickering when moving the cursor over the widgets.

[This](https://hg.mozilla.org/integration/autoland/pushloghtml?fromchange=ae2a8d6e0c958b9d2d27c21b797100994bb488a2&tochange=73572c522da83cb8e8392c723da1f6a01c619250) is the change that introduced the regression.

What the change does is stop presenting the draw buffer when canvas size is modified but the size is the same as the current canvas size.

This affected us in vtk-js because changing cursor type (which widgets do a lot) also caused the canvas sizes to be modified. That's what this commit is preventing, and it seems like an effective workaround.

### DIFF
--- a/Sources/Rendering/OpenGL/RenderWindow/index.js
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.js
@@ -72,11 +72,19 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
   model.classHierarchy.push('vtkOpenGLRenderWindow');
 
   // Auto update style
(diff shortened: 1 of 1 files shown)


boucaud:
@finetjul fyi

jourdain:
LGTM

PR #1411, opened 2020-04-23
fix(LookupTableProxy): Support midpoint/sharpness in rgbpoints

scottwittenburg:
Currently the `vtkLookupTableProxy` ignores the midpoint and sharpness values we might provide when calling `setRGBPoints`.  This PR allows us to do the following:

```
rgbPoints = [];
rgbPoints.push([v, r, g, b, mid, sharp]);
...
lutProxy.setRGBPoints(rgbPoints);
```

### DIFF
--- a/Sources/Proxy/Core/LookupTableProxy/index.js
+++ b/Sources/Proxy/Core/LookupTableProxy/index.js
@@ -30,7 +30,7 @@ function vtkLookupTableProxy(publicAPI, model) {
     }
   };
 
-  // Takes an array of points [x, r, g, b]
+  // Takes an array of points [x, r, g, b, m=0.5, s=0.0]
   publicAPI.setRGBPoints = (rgbPoints) => {
     if (model.rgbPoints !== rgbPoints) {
       model.rgbPoints = (rgbPoints || Defaults.RGBPoints).slice();
@@ -38,7 +38,7 @@ function vtkLookupTableProxy(publicAPI, model) {
     }
   };
 
-  // Takes an array of points [x, h, s, v]
+  // Takes an array of points [x, h, s, v, m=0.5, s=0.0]
   publicAPI.setHSVPoints = (hsvPoints) => {
     if (model.hsvPoints !== hsvPoints) {
       model.hsvPoints = (hsvPoints || Defaults.HSVPoints).slice();
@@ -75,14 +75,14 @@ function vtkLookupTableProxy(publicAPI, model) {
       case Mode.RGBPoints:
         model.lookupTable.removeAllPoints();
         model.rgbPoints.forEach((point) =>
-          model.lookupTable.addRGBPoint(...point)
+          model.lookupTable.addRGBPointLong(...point)
         );
         break;
 
       case Mode.HSVPoints:
         model.lookupTable.removeAllPoints();
         model.hsvPoints.forEach((point) =>
-          model.lookupTable.addHSVPoint(...point)
+          model.lookupTable.addHSVPointLong(...point)
         );
         break;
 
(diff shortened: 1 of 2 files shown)


scottwittenburg:
@jourdain Mind taking a look for any potential gotchas here?  When you get a chance.  Thanks!

scottwittenburg:
@thewtex This is to help us make sure we don't do any color interpolation when doing volume rendering of the label maps.

scottwittenburg:
Weird.  A couple volume rendering tests seem to be failing the image match comparison.  I ran them using Firefox 75 on my macbook and saw a small image diff (0.07%).  Then I ran the same test with master and saw the same diff (0.07%).  Still investigating.

scottwittenburg:
I notice in the test output, an `` tag with a base64 data url is generated.  Is that the failing image?  Or the diff?  Is there some convenient way to view those?

scottwittenburg:
Now I've just noticed those same tests seem to have failed on two other recent PRs: #1409 and #1410.  So I suspect my change here didn't introduce them.  However, is it possible that it's a real issue with volume rendering on Firefox 75 on Linux?  I don't have any linux machines handy to test this at the moment.

thewtex:
> using Firefox 75 on my macbook

Note: it is possible to reproduce the GitHub CI test environment locally with Docker by running ./Utilities/ci/run-tests-in-docker.sh. You can also start up a VNC server to debug by passing the `-d` flag.

> a small image diff (0.07%)

Maybe this is fine and can be worked around by setting a tolerance on the baseline image comparison?

> I notice in the test output, an  tag with a base64 data url is generated. Is that the failing image? Or the diff? Is there some convenient way to view those?

More tips on this here: https://kitware.github.io/vtk-js/docs/develop_test.html

> The test execution creates a file Documentation/content/coverage/tests.md.
> Rename this file as tests.html and open it in the browser.

scottwittenburg:
> using Firefox 75 on my macbook
> 
> Note: it is possible to reproduce the GitHub CI test environment locally with Docker by running ./Utilities/ci/run-tests-in-docker.sh. You can also start up a VNC server to debug by passing the -d flag.
> 
> a small image diff (0.07%)
> 
> Maybe this is fine and can be worked around by setting a tolerance on the baseline image comparison?

Sorry if I wasn't clear, that small diff of 0.07% was on my machine.  In this PR and the others before it, the diff was like 60% or something when the tests ran on linux with Firefox 75.  

> 
> I notice in the test output, an  tag with a base64 data url is generated. Is that the failing image? Or the diff? Is there some convenient way to view those?
> 
> More tips on this here: https://kitware.github.io/vtk-js/docs/develop_test.html
> 
> The test execution creates a file Documentation/content/coverage/tests.md.
> Rename this file as tests.html and open it in the browser.

I think what you suggest above is for when you're developing and running the tests locally, correct?  When the test fails on CI, I haven't found good/easy ways to see the failing test image diff.  And in this case, I cannot reproduce the failure locally.

I noticed in the text output, there are a few `` tags containing base64 image data.  By opening that file in raw mode, copy-pasting the three image tags, and then putting them in an html file, I can see the new test image, the baseline, and the diff.  But that's a pain, so I was wondering if there was some easy way to see the image results from tests run on CI.

For example, one of the failing volume test results:

Screen Shot 2020-04-28 at 6 42 32 PM

scottwittenburg:
I plan to merge this today unless there are objections.  This PR didn't introduce any new failures, and I feel the volume rendering regressions should be fixed in a separate PR.

thewtex:
> I think what you suggest above is for when you're developing and running the tests locally, correct? 

It is for developing and running the tests locally in the exact same environment that runs in CI thanks to Docker. That's the idea, at least :-).

PR #1422, opened 2020-04-29
Orientation support for ShapeWidget

agirault:
The widget was constrained to face the X, Y, or Z axis using the `setSlicingMode` API, which was limiting when wanting to create the widget on an arbitrary plane (example: IJK slicing of a dataset with a non-identity direction matrix).

The widgets now face the camera like other 2D widgets (SplineWidget, DistanceWidget) when being placed. It then holds on to its initial plane when being manipulated, even if the camera axis changes.

### DIFF
--- a/Sources/Rendering/Core/Glyph3DMapper/Constants.js
+++ b/Sources/Rendering/Core/Glyph3DMapper/Constants.js
@@ -1,6 +1,7 @@
 export const OrientationModes = {
   DIRECTION: 0,
   ROTATION: 1,
+  MATRIX: 2,
 };
 
 export const ScaleModes = {
--- a/Sources/Rendering/Core/Glyph3DMapper/index.js
+++ b/Sources/Rendering/Core/Glyph3DMapper/index.js
@@ -34,6 +34,8 @@ function vtkGlyph3DMapper(publicAPI, model) {
     publicAPI.setOrientationMode(OrientationModes.DIRECTION);
   publicAPI.setOrientationModeToRotation = () =>
     publicAPI.setOrientationMode(OrientationModes.ROTATION);
+  publicAPI.setOrientationModeToMatrix = () =>
+    publicAPI.setOrientationMode(OrientationModes.MATRIX);
   publicAPI.getOrientationArrayData = () => {
     const idata = publicAPI.getInputData(0);
     if (!idata || !idata.getPointData()) {
@@ -143,6 +145,17 @@ function vtkGlyph3DMapper(publicAPI, model) {
           const orientation = [];
           oArray.getTuple(i, orientation);
           switch (model.orientationMode) {
+            case OrientationModes.MATRIX: {
+              // prettier-ignore
+              const rotMat4 = [
+                ...orientation.slice(0, 3), 0,
+                ...orientation.slice(3, 6), 0,
+                ...orientation.slice(6, 9), 0,
+                0, 0, 0, 1,
+              ];
+              mat4.multiply(z, z, rotMat4);
+              break;
+            }
             case OrientationModes.ROTATION:
               mat4.rotateZ(z, z, orientation[2]);
               mat4.rotateX(z, z, orientation[0]);
--- a/Sources/Widgets/Representations/RectangleContextRepresentation/index.js
+++ b/Sources/Widgets/Representations/RectangleContextRepresentation/index.js
@@ -59,9 +59,8 @@ function vtkRectangleContextRepresentation(publicAPI, model) {
     const dataset = vtkPolyData.newInstance();
 
     if (state.getVisible()) {
-      const bounds = state.getBounds();
-      const point1 = [bounds[0], bounds[2], bounds[4]]; // so wrong, can't guess points from bounds
-      const point2 = [bounds[1], bounds[3], bounds[5]]; // so wrong, can't guess points from bounds
+      const point1 = state.getOrigin();
+      const point2 = state.getCorner();
       const diagonal = [0, 0, 0];
       vec3.subtract(diagonal, point2, point1);
       const up = state.getUp();
--- a/Sources/Widgets/Representations/RectangleContextRepresentation/index.js
+++ b/Sources/Widgets/Representations/RectangleContextRepresentation/index.js
@@ -4,7 +4,7 @@ import vtkHandleRepresentation from 'vtk.js/Sources/Widgets/Representations/Hand
 import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
 import vtkPolyData from 'vtk.js/Sources/Common/DataModel/PolyData';
 
-import { SlicingMode } from 'vtk.js/Sources/Rendering/Core/ImageMapper/Constants';
+import { vec3 } from 'gl-matrix';
 
 // ----------------------------------------------------------------------------
 // vtkRectangleContextRepresentation methods
@@ -60,52 +60,31 @@ function vtkRectangleContextRepresentation(publicAPI, model) {
 
     if (state.getVisible()) {
       const bounds = state.getBounds();
-      const point1 = [bounds[0], bounds[2], bounds[4]];
-      const point2 = [bounds[1], bounds[3], bounds[5]];
+      const point1 = [bounds[0], bounds[2], bounds[4]]; // so wrong, can't guess points from bounds
+      const point2 = [bounds[1], bounds[3], bounds[5]]; // so wrong, can't guess points from bounds
+      const diagonal = [0, 0, 0];
+      vec3.subtract(diagonal, point2, point1);
+      const up = state.getUp();
+      const upComponent = vec3.dot(diagonal, up);
 
       const points = new Float32Array(4 * 3);
-
       points[0] = point1[0];
       points[1] = point1[1];
       points[2] = point1[2];
-
+      points[3] = point1[0] + upComponent * up[0];
+      points[4] = point1[1] + upComponent * up[1];
+      points[5] = point1[2] + upComponent * up[2];
       points[6] = point2[0];
       points[7] = point2[1];
       points[8] = point2[2];
-
-      const slicingMode = state.getDirection().indexOf(1);
-
-      if (slicingMode === SlicingMode.I) {
-        points[3] = point1[0];
-        points[4] = point1[1];
-        points[5] = point2[2];
-
-        points[9] = point2[0];
-        points[10] = point2[1];
-        points[11] = point1[2];
-      } else if (slicingMode === SlicingMode.J) {
-        points[3] = point1[0];
-        points[4] = point1[1];
-        points[5] = point2[2];
-
-        points[9] = point2[0];
-        points[10] = point2[1];
-        points[11] = point1[2];
-      } else if (slicingMode === SlicingMode.K) {
-        points[3] = point1[0];
-        points[4] = point2[1];
-        points[5] = point1[2];
-
-        points[9] = point2[0];
-        points[10] = point1[1];
-        points[11] = point2[2];
-      }
-
-      // Triangles
-      const polys = new Uint32Array([4, 0, 1, 2, 3]);
+      points[9] = point2[0] - upComponent * up[0];
+      points[10] = point2[1] - upComponent * up[1];
+      points[11] = point2[2] - upComponent * up[2];
 
       dataset.getPoints().setData(points, 3);
+
       if (model.drawFace) {
+        const polys = new Uint32Array([4, 0, 1, 2, 3]);
         dataset.getPolys().setData(polys, 1);
       }
       if (model.drawBorder) {
--- a/Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
@@ -1,6 +1,5 @@
 import shapeBehavior from 'vtk.js/Sources/Widgets/Widgets3D/ShapeWidget/behavior';
 import { vec3 } from 'gl-matrix';
-import vtkBoundingBox from 'vtk.js/Sources/Common/DataModel/BoundingBox';
 
 export default function widgetBehavior(publicAPI, model) {
   // We inherit shapeBehavior
@@ -9,38 +8,28 @@ export default function widgetBehavior(publicAPI, model) {
 
   model.classHierarchy.push('vtkEllipseWidgetProp');
 
-  publicAPI.setBounds = (bounds) => {
-    if (superClass.setBounds) {
-      superClass.setBounds(bounds);
+  publicAPI.setCorners = (point1, point2) => {
+    if (superClass.setCorners) {
+      superClass.setCorners(point1, point2);
     }
 
-    const center = vtkBoundingBox.getCenter(bounds);
-    const scale3 = vtkBoundingBox.computeScale3(bounds);
-
-    model.shapeHandle.setOrigin(center);
-    model.shapeHandle.setScale3(scale3);
-  };
-
-  publicAPI.setBoundsFromRadius = (center, pointOnCircle) => {
-    const radius = vec3.distance(center, pointOnCircle);
-
-    publicAPI.setBounds([
-      center[0] - radius,
-      center[0] + radius,
-      center[1] - radius,
-      center[1] + radius,
-      center[2] - radius,
-      center[2] + radius,
-    ]);
-  };
-
-  publicAPI.setBoundsFromDiameter = (point1, point2) => {
     const center = [
       0.5 * (point1[0] + point2[0]),
       0.5 * (point1[1] + point2[1]),
       0.5 * (point1[2] + point2[2]),
     ];
 
-    publicAPI.setBoundsFromRadius(center, point1);
+    const diagonal = [0, 0, 0];
+    vec3.subtract(diagonal, point2, center);
+
+    const right = model.shapeHandle.getRight();
+    const up = model.shapeHandle.getUp();
+    const dir = model.shapeHandle.getDirection();
+    const rightComponent = vec3.dot(diagonal, right);
+    const upComponent = vec3.dot(diagonal, up);
+    const dirComponent = vec3.dot(diagonal, dir);
+
+    model.shapeHandle.setOrigin(center);
+    model.shapeHandle.setScale3([rightComponent, upComponent, dirComponent]);
   };
 }
--- a/Sources/Widgets/Widgets3D/EllipseWidget/state.js
+++ b/Sources/Widgets/Widgets3D/EllipseWidget/state.js
@@ -25,11 +25,10 @@ export default function generateState() {
     })
     .addStateFromMixin({
       labels: ['ellipseHandle'],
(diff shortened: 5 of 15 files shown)


agirault:
@finetjul can you tell me what you think of this? There technically is a breaking change, by removing `setSlicingMode` in `ShapeWidgetBehavior`... though I think it's worth it, let me know your thoughts.

Also, I'm having some issues in terms of interaction with the `SplineWidget`: I can't seem to be able to stop moving its handles to "complete" it. Pressing "Enter" does trigger the `onEndInteractionEvent` though, so I'm thinking that's a bug? If you agree, I might try to look into fixing that (could do in that PR), and welcome any pointer.

finetjul:
I think it is fine and will make it easier on developers (no need to call setSlicingMode)
However, you could have kept, by making it optional, the feature to enforce on widgets a plane different from the camera. I do not have a clear usecase in mind, but it seems like something useful.

I am not surprised that "Enter" does not work for you. Remi's use case was to use the SplineWidget for painting, so on `endInteractionEvent`, he was deleting the widget. I am afraid I won't be of much help here, it's @RemiCecchinato who knows it best.

agirault:
> However, you could have kept, by making it optional, the feature to enforce on widgets a plane different from the camera. I do not have a clear usecase in mind, but it seems like something useful.

I thought of that as well initially, though:
1. as you said, we currently don't have a clear use case in mind, so I'd rather not spend time on it
2. no other widgets offers that
3. it's still possible to place a widget on an arbitrary plane, different from the camera plane, by setting the widget state (only can't do that in "placing" interactive mode)

So if that's alright with you I'll pass on that unless a use case comes up before we want to merge this.

> I am not surprised that "Enter" does not work for you. Remi's use case was to use the SplineWidget for painting, so on endInteractionEvent, he was deleting the widget. 

Sounds good. As we discussed, I'll work on customizing the spline widget with behavior modifiers as parameters, similar to the shape widgets.

agirault:
Rebased on top of master, and improved the spline widget (see notes from the commit message):
```
In index.js:
- Add `resetAfterEnter` boolean to maintain or reset the widget after
placing its control points and pressing "Enter" (similar to ShapeWidget)

In behavior.js:
- Make private methods actually private (updateHandlesSize, addPoint,
getHoveredHandle)
- Use `macro.setGet` to add public APIs around the model properties, and
remove manual setters (sethandleSizeInPixels and setFreehandMinDistance)
- Override setHandleSizeInPixels and setResolution to update the model
representations
- Remove updateResolution which hardcoded a non-standard behavior when
pressing "Shift". Instead, the setResolution API can be used by the
application with a custom keyDown listener.
- Add missing event callbacks in grab/loseFocus

In the example:
- Improve the control Panel to interact with most spline parameters

Notes:
- color, scale and visibility of the handles should probably be a property
of the viewWidgets and not of the state.
- there are confusions between the model properties on the widget
(widget factory) and the behavior (view widget). The former can be set to
define the default values when instantiating a view widget, but the later
should be used to edit the properties of a view widget instance.
```

finetjul:
In SplineWidget example, when changing the resolution with the slider and moving the mouse on the render view,  the clickable area (when you mouse over the shape) is not updated: the previous resolution is still used

agirault:
> when changing the resolution with the slider and moving the mouse on the render view, the clickable area (when you mouse over the shape) is not updated: the previous resolution is still used

Would you have an idea why? I set the resolution on the representation (`model.representations[1]`) the same way it was done before my changes in `updateResolution`. My guess is that it didn't work before either, though it could not be tested since the spline did not stay. Still, I agree that would be nice to address.

> after closing the shape (Enter), it is abnormally slow after releasing a handle

What specifically is slow? Which browser? Not sure I experience the same thing.

> when placing a spline widget, a click (no drag) on first (or last) placed point or last point should close the shape.

Good feature idea (though I would think only for the first placed point). I'll look to see if I can easily implement that. However, my current (commercial) funding won't support that so it might need to wait.

finetjul:
> > when placing a spline widget, a click (no drag) on first (or last) placed point or last point should close the shape.
>
> ~Good feature idea (though I would think only for the first placed point). I'll look to see if I can easily implement that.~ Edit: done! I started messing up with the cursors to try to change them up, but that was becoming finicky so I just left the current behavior.

That feature was already existing (see [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Widgets3D/SplineWidget/behavior.js#L231-L247 ))

>> after closing the shape (Enter), it is abnormally slow after releasing a handle
>
> What specifically is slow? Which browser? Not sure I experience the same thing.
On Chrome, a simple click (no drag) on a closed spline handle takes more than 600ms for mouseup event:
```
[Violation] 'mouseup' handler took 389ms index.js:526
[Violation] 'mouseup' handler took 672ms index.js:526
[Violation] 'mouseup' handler took 672ms index.js:526
[Violation] 'mouseup' handler took 669ms index.js:526
[Violation] 'mouseup' handler took 647ms
```

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
```diff
@@ -5,31 +5,10 @@ export default function widgetBehavior(publicAPI, model) {
...
-
-  publicAPI.sethandleSizeInPixels = (handleSizeInPixels) => {
-    model.handleSizeInPixels = handleSizeInPixels;
-  };
-
+  // Private methods
   // --------------------------------------------------------------------------
 
-  publicAPI.updateHandlesSize = () => {
+  const updateHandlesSize = () => {
```
finetjul, line 11: does not seem to work (anymore?)

agirault: Can you expand on your issue? It seems to work here:

ezgif-4-f5d7ac9c53f0

It does not update with camera changes, though that's not a regression since it didn't beforehand either. I know @floryst is looking into bringing some scalable handle representation from Glance back to vtk.js, so we could wait on that and leverage them.

agirault:
> That feature was already existing (see here)

Thanks, I leveraged that instead

> On Chrome, a simple click (no drag) on a closed spline handle takes more than 600ms for mouseup event:

It seems to come from this `cancelAnimation` [here](https://github.com/Kitware/vtk-js/blob/e32e688/Sources/Widgets/Widgets3D/SplineWidget/behavior.js#L224), which we never encountered before since the example would not lose focus. @floryst do you know why it would be so slow?

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
finetjul: My bad, I thought that `updateHandesSize` was for the latter case. Nevermind then.

agirault:
Note: the change I did to the `ShapeWidget` isn't sufficient: it works to place the points properly, and it works for the ellipse/circle widget to create the contour based on the bounds, but not yet on the rectangle widget.

finetjul:
> @finetjul I'm wondering if that concept of setting `bounds` for those shape widget makes sense, or if we should just set two points/corners for all the shape widgets? 

If you want something generic, I think that 2 points (corners) is not enough if you want to support non axis-aligned shapes. Maybe OBB (Oriented Bounding Box) is the way to go...

finetjul:
Out of date picking may be fixed with 
```widgetManager.setCaptureOn(CaptureOn.MOUSE_MOVE);```

**Comment by agirault** [comment]
Thanks a lot @finetjul. I moved all the discussions around the spline widget to the #1443 PR.

File: Sources/Widgets/Representations/RectangleContextRepresentation/index.js [hunk 2]
```diff
@@ -59,53 +59,35 @@ function vtkRectangleContextRepresentation(publicAPI, model) {
     const dataset = vtkPolyData.newInstance();
 
     if (state.getVisible()) {
-      const bounds = state.getBounds();
-      const point1 = [bounds[0], bounds[2], bounds[4]];
-      const point2 = [bounds[1], bounds[3], bounds[5]];
+      const normal = state.getDirection();
+      const vertical = state.getUp();
+      const horizontal = state.getRight();
+      vec3.cross(horizontal, vertical, normal);
```
finetjul, line 65: I find it confusing that those vectors variable names are different (up/vertical, right/horizontal, normal/orient/direction)

Sources/Widgets/Representations/RectangleContextRepresentation/index.js
```diff
@@ -59,53 +59,35 @@ function vtkRectangleContextRepresentation(publicAPI, model) {
     const dataset = vtkPolyData.newInstance();
 
     if (state.getVisible()) {
-      const bounds = state.getBounds();
-      const point1 = [bounds[0], bounds[2], bounds[4]];
-      const point2 = [bounds[1], bounds[3], bounds[5]];
+      const normal = state.getDirection();
+      const vertical = state.getUp();
+      const horizontal = state.getRight();
```
finetjul, line 64: it is confusing to get a variable to recompute it just after.

Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
```diff
@@ -9,38 +8,26 @@ export default function widgetBehavior(publicAPI, model) {
...
       0.5 * (point1[2] + point2[2]),
     ];
 
-    publicAPI.setBoundsFromRadius(center, point1);
+    const d = [0, 0, 0];
+    vec3.subtract(d, point2, center);
+
+    const horizontal = model.shapeHandle.getRight();
+    const vertical = model.shapeHandle.getUp();
+    const sh = vec3.dot(d, horizontal);
```
finetjul, line 27: rename `sh` into `horizontalScale` (same for `sv`)

Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
```diff
@@ -9,38 +8,26 @@ export default function widgetBehavior(publicAPI, model) {
...
-  publicAPI.setBoundsFromDiameter = (point1, point2) => {
     const center = [
       0.5 * (point1[0] + point2[0]),
       0.5 * (point1[1] + point2[1]),
       0.5 * (point1[2] + point2[2]),
     ];
 
-    publicAPI.setBoundsFromRadius(center, point1);
+    const d = [0, 0, 0];
+    vec3.subtract(d, point2, center);
```
finetjul, line 23: rename `d` to `diameter`.

However that looks more like a `radius`, but an incorrectly computed radius . (ellipse/circle is **contained** within the rectangle defined by corner points)

Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
```diff
@@ -9,38 +8,26 @@ export default function widgetBehavior(publicAPI, model) {
...
+
+    const horizontal = model.shapeHandle.getRight();
+    const vertical = model.shapeHandle.getUp();
+    const sh = vec3.dot(d, horizontal);
+    const sv = vec3.dot(d, vertical);
+
+    model.shapeHandle.setOrigin(center);
+    model.shapeHandle.setScale3([sh, sv, 1]);
```
finetjul, line 31: why not support z scale ? (in case corner points are not within the same plane)

Sources/Widgets/Widgets3D/RectangleWidget/behavior.js
```diff
@@ -7,11 +7,12 @@ export default function widgetBehavior(publicAPI, model) {
 
   model.classHierarchy.push('vtkRectangleWidgetProp');
 
-  publicAPI.setBounds = (bounds) => {
-    if (superClass.setBounds) {
-      superClass.setBounds(bounds);
+  publicAPI.setCorners = (point1, point2) => {
+    console.warn('setCorners');
```
finetjul, line 11: to remove

Sources/Widgets/Widgets3D/EllipseWidget/state.js
```diff
@@ -25,11 +25,10 @@ export default function generateState() {
     })
     .addStateFromMixin({
       labels: ['ellipseHandle'],
-      mixins: ['origin', 'color', 'scale3', 'visible', 'direction'],
+      mixins: ['origin', 'color', 'scale3', 'visible', 'orientation'],
```
finetjul, line 28: isn't `corner` mixin missing ?

Sources/Widgets/Representations/RectangleContextRepresentation/index.js
agirault: I didn't like up and right because they could be down or left for that matter, so I used vertical and horizontal... I'll go back to up and right though since that shows the relative coordinate system

Sources/Widgets/Representations/RectangleContextRepresentation/index.js
agirault: That's a leftover from before using Right and Up, we don't even need the direction/normal anymore... will remove, thanks.

Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
agirault: will rename to rightComponent and upComponent

Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
agirault: will rename to direction

Sources/Widgets/Widgets3D/EllipseWidget/state.js
agirault: Might be why it doesn't work :) Will look more, thanks

Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
agirault: As far as whether it's a radius, diameter, corner to corner, or center to corner, that's properly handled already based on the `BehaviorCategory` [here](https://github.com/Kitware/vtk-js/blob/5f7c6f66c51d44c84d7741006328d4a3de2bd7ca/Sources/Widgets/Widgets3D/ShapeWidget/behavior.js#L289-L316) which can call [those functions](https://github.com/Kitware/vtk-js/blob/5f7c6f66c51d44c84d7741006328d4a3de2bd7ca/Sources/Widgets/Widgets3D/ShapeWidget/behavior.js#L230-L255) to properly extract the true corners that define the rectangle which contains the ellipse/circle.

Some parts of that PR still don't work (scaling, orienting the `CircleContextRepresentation` properly), but that part should.

Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
agirault: I'm having issues finishing that part up: we have two corners and they should be on the same plane as the right/up plane that's parallel to the camera plane. The scaling here `[sh, sv, 1]` is actually in that plane coordinate system, but I am not sure if `scale3` should be in that local coordinate system, or in the world XYZ coordinate system...?

Whichever it is/should be, those current values coupled with how the orientation is applied to it in the `CircleContextRepresentation` and in the `Glyph3DMapper` just doesn't quite work yet... an attempt was made in the last `WIP` commit of this PR to fix the orientation in those classes, but my brain was fried last Friday evening and I could not make that work.
cc: @floryst who witnessed my brain-melting 🧠🔥

agirault: The scale is in local space, not world space, so we can actually put any value in that third component, it will always be ignored because the `circleContextRepresentation` is a 2D representation.

We've investigated more with @floryst and came up with what can be seen in [this commit](https://github.com/Kitware/vtk-js/pull/1422/commits/664e1eac81ccf4b5922054022c001ed937cfa621):
- removed some code from the `circleContextRepresentation` (no need to reorient the scale)
- added `MATRIX` to `OrientationModes` that is implemented in `Glyph3DMapper` and used by `CircleContextRepresentation`

Debug only:
- modified the `Glyph3DMapper` example to observe the scaling behavior and test/compare `OrientationModes.MATRIX` and `OrientationModes.ROTATION` -> they work 🎉
- modified the `ShapeWidget` example to add an orientation widget to try to understand why the `CircleContextRepresentation` is not shaping properly -> still haven't figured it out 🤕

@floryst plans to look at it again at some point to figure that out.

agirault:
@laurennlam @boucaud: would one of you be able to give this a look?

Sources/Rendering/Core/Glyph3DMapper/Constants.js
```diff
@@ -1,6 +1,7 @@
 export const OrientationModes = {
   DIRECTION: 0,
   ROTATION: 1,
+  MATRIX: 2,
```
agirault, line 4: @martinken we added a new way to input orientation to the `3DGlyphMapper` (3x3 matrix). Could you give this commit a look when you get a chance? It's then used by the `CircleContextRepresentation` in [this commit](https://github.com/Kitware/vtk-js/pull/1422/commits/c50bc37733e63ad71c5f35a4cd5c5dc5de634fa7). Thank you!

Sources/Rendering/Core/Glyph3DMapper/index.js
```diff
@@ -143,6 +145,14 @@ function vtkGlyph3DMapper(publicAPI, model) {
           const orientation = [];
           oArray.getTuple(i, orientation);
           switch (model.orientationMode) {
+            case OrientationModes.MATRIX: {
+              const rotQuat = quat.create();
+              quat.fromMat3(rotQuat, orientation);
+              const rotMat4 = mat4.create();
+              mat4.fromQuat(rotMat4, rotQuat);
```
floryst, line 152: We can just copy the mat3 to the mat4 directly without going through quat, which avoids going through unnecessary calculations.

floryst:
Overall, LGTM

Sources/Rendering/Core/Glyph3DMapper/index.js
agirault: addressed

Sources/Widgets/Widgets3D/PaintWidget/index.js
```diff
@@ -41,24 +44,34 @@ function widgetBehavior(publicAPI, model) {
       model.activeState &&
       model.activeState.getActive()
     ) {
+      const normal = model.camera.getDirectionOfProjection();
+      const up = model.camera.getViewUp();
+      const right = vec3.create();
```
jourdain, line 49: Each time I see a `{glmatrix}.create()`, I worry that we will lose precision and we could get bad interaction with models that are far from the [0,0,0] origin. It might not be an issue here, but in general, *never* use the `{glMatrix}.create()` but instead provide the array itself `[0,0,0]` or `new Float64Array(3)` as by default glMatrix will do `new Float32Array(size) + identity(for matrix)`.

So ideally do a search and replace for those `.create()` calls.

agirault: sounds good, will do.

agirault: replaced by `[]`

PR #1424, opened 2020-04-30
feat(SynchronizableRenderWindow): More generic updater to fetch arrays

xavArtley:
To continue the discussion https://github.com/holoviz/panel/pull/1248#issuecomment-621724971 I open this PR.
It merge imageDataUpdater and PolyDataUpdater into the genericUpdater

In the same time I added an updater for PieceWiseFunctions  and extended alias with new classes to make it work with image slices and volumes


jourdain:
Thanks @xavArtley, but you tried to do too much and your changes are may break things. I wanted to do the dataset handling via another step. Once the export of dataset rely on `state.arrays`, we could start deprecating the old updater. But for backward compatibility I wanted to migrate it progressively.

I think there is a lot of good parts, but a smaller increment without touching the dataset handling at first should be better. The genericUpdated should not do any `isA` or custom handling GEOM....

jourdain:
Rework done [here](https://github.com/Kitware/vtk-js/pull/1427).

jourdain:
https://github.com/Kitware/vtk-js/pull/1427

PR #1427, opened 2020-05-01
Sync ren win improvements

jourdain:


jourdain:
FYI @xavArtley
Once we improve the serializers (Python/C++), we could simplify the datasetUpdater by removing the custom array handling. But it is not a big deal leaving it  that way for a long time.

xavArtley:
Hello @jourdain 
I tried the new version of vtkjs with the work I developped i panel.
However I can't make it works anymore with volumes.
```
VM3524 vtk.js:282 Uncaught (in promise) TypeError: Cannot read property 'getNumberOfComponents' of null
    at Object.e.buildBufferObjects (VM3524 vtk.js:282)
    at Object.e.updateBufferObjects (VM3524 vtk.js:282)
    at Object.e.renderPieceStart (VM3524 vtk.js:282)
    at Object.e.renderPiece (VM3524 vtk.js:282)
```
I had this error before and I could get rid off by waiting all compressed arrays to be decoded. However now it does not work anymore

xavArtley:
The errors are due to the fact that I add widgets which call the render function when the scene has not finished loading
I think a way should be to be able to set a custom CallbackWhenReady function when the synchronize function il called

xavArtley:
@jourdain 
I made some change here : https://github.com/Kitware/vtk-js/compare/master...xavArtley:xav/sync_ren_expose_ready
Tell me if it makes sense, I'll submit a PR in this case

jourdain:
I'm working on that PR, but do you need to prevent the default render? Or you are just looking for adding other handling when the synchronization is done?

jourdain:
Can you test and review it? Thanks...
https://github.com/Kitware/vtk-js/pull/1428

xavArtley:
I made a quick test and it seems perfect

Le lun. 4 mai 2020 à 11:47, Sebastien Jourdain <notifications@github.com> a
écrit :

jourdain:
Also something that you can made use of https://github.com/Kitware/vtk-js/pull/1429

PR #1430, opened 2020-05-04
fix(Widgets): Apply coincident topology parameters

boucaud:
Applies coincident topology parameters to mappers created by widget representations.

Default values are -1.0, -1.0, but users can change those (see Sources/Widgets/Core/WidgetManager/example/index.js)

### DIFF
--- a/Sources/Widgets/Core/AbstractWidgetFactory/index.js
+++ b/Sources/Widgets/Core/AbstractWidgetFactory/index.js
@@ -45,10 +45,13 @@ function vtkAbstractWidgetFactory(publicAPI, model) {
 
       // Create representations for that view
       /* eslint-disable no-shadow */
+      const widgetInitialValues = initialValues; // Avoid shadowing
       widgetModel.representations = publicAPI
         .getRepresentationsForViewType(viewType)
         .map(({ builder, labels, initialValues }) =>
-          builder.newInstance(Object.assign({ labels }, initialValues))
+          builder.newInstance(
+            Object.assign({ labels }, initialValues, widgetInitialValues)
+          )
         );
       /* eslint-enable no-shadow */
 
@@ -60,10 +63,9 @@ function vtkAbstractWidgetFactory(publicAPI, model) {
       });
 
       model.behavior(widgetPublicAPI, widgetModel);
-
       // Forward representation methods
-      if (model.methodsToLink) {
-        model.methodsToLink.forEach((methodName) => {
+      ['coincidentTopologyParameters', ...(model.methodsToLink || [])].forEach(
+        (methodName) => {
           const set = `set${macro.capitalize(methodName)}`;
           const get = `get${macro.capitalize(methodName)}`;
           const methods = {
@@ -91,8 +93,8 @@ function vtkAbstractWidgetFactory(publicAPI, model) {
               widgetPublicAPI[name] = macro.chain(...calls);
             }
           });
-        });
-      }
+        }
+      );
 
       // Custom delete to detach from parent
       widgetPublicAPI.delete = macro.chain(() => {
--- a/Sources/Widgets/Core/WidgetManager/example/index.js
+++ b/Sources/Widgets/Core/WidgetManager/example/index.js
@@ -31,6 +31,20 @@ const WIDGET_BUILDERS = {
         label: 'Polyline',
       })
     );
+    instance.setCoincidentTopologyParameters({
+      Point: {
+        factor: -1.0,
+        offset: -1.0,
+      },
+      Line: {
+        factor: -1.5,
+        offset: -1.5,
+      },
+      Polygon: {
+        factor: -2.0,
+        offset: -2.0,
+      },
+    });
     instance.setActiveScaleFactor(0.9);
     instance.setGlyphResolution(60);
     return instance;
@@ -39,7 +53,24 @@ const WIDGET_BUILDERS = {
     const instance = widgetManager.addWidget(
       vtkPolyLineWidget.newInstance({
         label: 'Closed Polyline',
-      })
+      }),
+      null,
+      {
+        coincidentTopologyParameters: {
+          Point: {
+            factor: -1.0,
+            offset: -1.0,
+          },
+          Line: {
+            factor: -1.5,
+            offset: -1.5,
+          },
+          Polygon: {
+            factor: -2.0,
+            offset: -2.0,
+          },
+        },
+      }
     );
     instance.setActiveScaleFactor(1.1);
     instance.setGlyphResolution(30);
--- a/Sources/Widgets/Representations/SphereHandleRepresentation/index.js
+++ b/Sources/Widgets/Representations/SphereHandleRepresentation/index.js
@@ -47,7 +47,7 @@ function vtkSphereHandleRepresentation(publicAPI, model) {
   // model.displayActor.getProperty().setOpacity(0); // don't show in 3D
   model.displayActor.setMapper(model.displayMapper);
   model.displayMapper.setInputConnection(publicAPI.getOutputPort());
-  model.actors.push(model.displayActor);
+  publicAPI.addActor(model.displayActor);
   model.alwaysVisibleActors = [model.displayActor];
 
   model.mapper = vtkGlyph3DMapper.newInstance({
@@ -65,7 +65,7 @@ function vtkSphereHandleRepresentation(publicAPI, model) {
   model.mapper.setInputConnection(model.glyph.getOutputPort(), 1);
   model.actor.setMapper(model.mapper);
 
-  model.actors.push(model.actor);
+  publicAPI.addActor(model.actor);
 
   // --------------------------------------------------------------------------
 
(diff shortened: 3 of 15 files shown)


boucaud:
@finetjul

Sources/Widgets/Core/AbstractWidgetFactory/index.js
```diff
@@ -45,11 +45,15 @@ function vtkAbstractWidgetFactory(publicAPI, model) {
...
       // Create representations for that view
       /* eslint-disable no-shadow */
+      const paramValues = initialValues; // Avoid shadowing
       widgetModel.representations = publicAPI
         .getRepresentationsForViewType(viewType)
-        .map(({ builder, labels, initialValues }) =>
-          builder.newInstance(Object.assign({ labels }, initialValues))
-        );
+        .map(({ builder, labels, initialValues }) => {
+          console.log(Object.assign({ labels }, initialValues, paramValues));
```
finetjul, line 52: to remove

Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -161,6 +161,38 @@ function vtkWidgetRepresentation(publicAPI, model) {
...
+  publicAPI.initializeMapper = (mapper) => {
+    applyCoincidentTopologyParametersToMapper(
+      mapper,
+      model.coincidentTopologyPolygonOffsetParameters
+    );
+    model.initializedMappers.push(mapper);
+    return mapper;
+  };
+
+  publicAPI.setCoincidentTopologyPolygonOffsetParameters = (parameters) => {
```
finetjul, line 184: How about `setCoincidentTopologyParameters` (because it is for PolygonOffset, LineOffset and PointOffset parameters)?

Further on, parameters could then be { PolygonOffset: {factor, offset}, LineOffset: {factor, offset})...)

Sources/Widgets/Representations/SphereHandleRepresentation/index.js
```diff
@@ -49,12 +49,13 @@ function vtkSphereHandleRepresentation(publicAPI, model) {
   model.displayMapper.setInputConnection(publicAPI.getOutputPort());
   model.actors.push(model.displayActor);
   model.alwaysVisibleActors = [model.displayActor];
-
-  model.mapper = vtkGlyph3DMapper.newInstance({
-    scaleArray: 'scale',
-    colorByArrayName: 'color',
-    scalarMode: ScalarMode.USE_POINT_FIELD_DATA,
-  });
+  model.mapper = publicAPI.initializeMapper(
```
finetjul, line 52: `this initializeMapper` is not very userfriendly.

How about using `getActors()` or have a similar mechanism for mappers ?

I thought mappers could be initialized in `extend()`

boucaud: I think a similar mechanism for mappers might be redundant with `getActors`, as they are mostly initialized at the same place, but I'm not sure of exceptions ?

If we want to be able to use `getActors()`, then we need to initialize mappers and actors in `extend()`, and append them to the actor list in `extend()` anyways. This is because we want to automatically apply the topology parameters at the end of `WidgetRepresentation`'s `extend()`.

If that seems ok then let's do it like this.

boucaud: That implies making a rule that all mappers always have to be initialized in extend, and all actors have to be added to the list. I'm not sure that's ok, and if it could be enforced @finetjul ?

Alternatively, we could implement a method to replace `model.actors.push(actor)` with something custom that also initializes the mapper, but that looks a lot like `initializeMapper`

finetjul: yes, if what you have in mind is something like that:
```
extend() {
   ...
   getActors().forEach(actor -> initializeMapper(actor.getMapper()))
}
```

finetjul: You're right, it can also be done in a `addActor()` convenient method.

I feel it is better to initialize as late a possible in order to have as much context as possible.

boucaud: Yes, that's what I was thinking.

Do you think there's a way to enforce that all mappers/actors should be created in extend ?

finetjul: I believe it is implicitly already the case (@floryst ?). Maybe that should simply be documented.

boucaud: Right now they are mostly initialized in the (child) representation's constructor, but that's too late for us if we want to go  the `getActors()` route.

Sources/Widgets/Core/AbstractWidgetFactory/index.js
```diff
@@ -45,10 +45,13 @@ function vtkAbstractWidgetFactory(publicAPI, model) {
 
       // Create representations for that view
       /* eslint-disable no-shadow */
+      const paramValues = initialValues; // Avoid shadowing
```
finetjul, line 48: rename `paramValues` into `widgetInitialValues` ?

Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -161,6 +161,42 @@ function vtkWidgetRepresentation(publicAPI, model) {
...
+    mapper.setRelativeCoincidentTopologyPolygonOffsetParameters(factor, offset);
+    mapper.setRelativeCoincidentTopologyLineOffsetParameters(factor, offset);
+    mapper.setRelativeCoincidentTopologyPointOffsetParameters(factor, offset);
+  }
+
+  // Add warning to model.actors.push
+  model.actors.push = (...args) => {
+    console.warn(
+      'You should use publicAPI.addActor() to initialize the actor properly'
+    );
```
finetjul, line 178: @jourdain is it a best-practice in VTK-JS ?

Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -161,6 +161,42 @@ function vtkWidgetRepresentation(publicAPI, model) {
     }
   };
 
+  function applyCoincidentTopologyParametersToMapper(
+    mapper,
+    { factor, offset }
+  ) {
+    mapper.setResolveCoincidentTopologyToPolygonOffset();
```
finetjul, line 168: check that mapper exists and has those methods available

Sources/Widgets/Core/WidgetManager/example/index.js
```diff
@@ -31,6 +31,10 @@ const WIDGET_BUILDERS = {
         label: 'Polyline',
       })
     );
+    instance.setCoincidentTopologyPolygonOffsetParameters({
+      factor: -1.0,
```
finetjul, line 35: it does not make sense to set the same value as the default.

Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -173,17 +215,27 @@ const DEFAULT_VALUES = {
...
   Object.assign(model, DEFAULT_VALUES, initialValues);
+  // Initialize coincident topology parameters
+  CATEGORIES.forEach((category) => {
+    if (!model.coincidentTopologyParameters[category]) {
+      model.coincidentTopologyParameters[category] = {
+        factor: -1,
+        offset: -1,
+      };
+    }
+  });
```
finetjul, line 233: I am not sure it is more readable than setting those into DEFAULT_VALUE

Sources/Widgets/Representations/WidgetRepresentation/index.js
agirault: @floryst

Sources/Widgets/Core/WidgetManager/example/index.js
```diff
@@ -39,7 +53,24 @@ const WIDGET_BUILDERS = {
...
+          Line: {
+            factor: -1.5,
+            offset: -1.5,
+          },
+          Polygon: {
+            factor: -2.0,
+            offset: -2.0,
+          },
+        },
+      }
```
agirault, line 73: Since this is the same on line 34 and here, could you define this as a static object and pass it to both `instance.setCoincidentTopologyParameters` and `widgetManager.addWidget`?

Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -173,6 +215,20 @@ const DEFAULT_VALUES = {
...
+    },
+    Line: {
+      factor: -1.0,
+      offset: -1.0,
+    },
+    Polygon: {
+      factor: -1.0,
+      offset: -1.0,
+    },
+  },
```
agirault, line 231: Why are those default values different here than from the widget manager example? Is one more recommended?

Sources/Widgets/Core/WidgetManager/example/index.js
agirault: I just realized that was in the example and no the actual class, so doesn't matter as such.

Sources/Widgets/Representations/WidgetRepresentation/index.js
boucaud: The example values are different for an examples' sake, just so we don't set values to their defaults.

The default values should usually be fine.

Sources/Widgets/Representations/WidgetRepresentation/index.js
floryst: I would use `vtkWarningMacro` here.

floryst: You could also invoke `publicAPI.addActor` here as well, unless you want to leave `model.actors.push` as a means of adding an actor without applying the topology parameters.

PR #1431, opened 2020-05-05
fix(PaintFilter): fix polygon paint filter painting the wrong slice

boucaud:
The polygon paint filter would sometimes paint to the slice above the current slice.


boucaud:
@finetjul this fixes our issue.

finetjul:
LGTM

PR #1433, opened 2020-05-06
Display-scaled widget representations

floryst:
This sphere handle will scale to keep the same screen-space size. It
requires widgets to invoke `attachScaleListener` to their behaviors for
this to work.

This is more or less copied from what I played with in Glance. Right now the API requires a whole new representation (the ScaledSphereHandleRepresentation), since initially I thought to make the distinction between widget state scale1 and the camera-dependent scaling in this PR. This could technically be achieved by using the regular SphereHandleRepresentation, and having the camera listener call `setScale1(scale)`, but then you wouldn't be able to control the relative size of the sphere handle via the widget state. That's why this is currently in a separate representation.

Another idea is to just merge the `attachScaleListener` into SphereHandleRepresentation, add a `handleScale` property to that class, and throw out this extra ScaledSphereHandleRepresentation class. It'll still work as before, but if you use `attachScaleListener`, then your spheres will start scaling according to the camera. Thoughts?

@agirault @finetjul

### DIFF
--- a/Sources/Widgets/Representations/SphereHandleRepresentation/index.js
+++ b/Sources/Widgets/Representations/SphereHandleRepresentation/index.js
@@ -132,6 +132,10 @@ function vtkSphereHandleRepresentation(publicAPI, model) {
         (!state.isVisible || state.isVisible() ? 1 : 0) *
         (state.getScale1 ? state.getScale1() : model.defaultScale);
 
+      if (publicAPI.getScaleByDisplay()) {
+        typedArray.scale[i] *= publicAPI.getDisplayWorldHeightAt(coord);
+      }
+
       typedArray.color[i] =
         model.useActiveColor && isActive ? model.activeColor : state.getColor();
     }
--- a/Sources/Widgets/Representations/WidgetRepresentation/index.js
+++ b/Sources/Widgets/Representations/WidgetRepresentation/index.js
@@ -1,5 +1,6 @@
 import macro from 'vtk.js/Sources/macro';
 import vtkProp from 'vtk.js/Sources/Rendering/Core/Prop';
+import { subtract, dot } from 'vtk.js/Sources/Common/Core/Math';
 
 import { Behavior } from 'vtk.js/Sources/Widgets/Representations/WidgetRepresentation/Constants';
 import { RenderingTypes } from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
@@ -203,6 +204,21 @@ function vtkWidgetRepresentation(publicAPI, model) {
     });
   };
 
+  publicAPI.getDisplayWorldHeightAt = (worldCoord) => {
+    const {
+      dispHeightFactor,
+      cameraPosition,
+      cameraDir,
+      isParallel,
+    } = model.displayScaleParams;
+    if (isParallel) {
+      return dispHeightFactor;
+    }
+    const worldCoordToCamera = [...worldCoord];
+    subtract(worldCoordToCamera, cameraPosition, worldCoordToCamera);
+    return dot(worldCoordToCamera, cameraDir) * dispHeightFactor;
+  };
+
   // Make sure setting the labels at build time works with string/array...
   publicAPI.setLabels(model.labels);
 }
@@ -229,6 +245,13 @@ const DEFAULT_VALUES = {
       offset: -1.0,
     },
   },
+  scaleByDisplay: false,
+  displayScaleParams: {
+    dispHeightFactor: 1,
+    cameraPosition: [0, 0, 0],
+    cameraDir: [1, 0, 0],
+    isParallel: false,
+  },
 };
 
 // ----------------------------------------------------------------------------
@@ -240,6 +263,8 @@ export function extend(publicAPI, model, initialValues = {}) {
   vtkProp.extend(publicAPI, model, initialValues);
   macro.algo(publicAPI, model, 1, 1);
   macro.get(publicAPI, model, ['labels', 'coincidentTopologyParameters']);
+  macro.set(publicAPI, model, ['displayScaleParams']);
+  macro.setGet(publicAPI, model, ['scaleByDisplay']);
 
   // Object specific methods
   vtkWidgetRepresentation(publicAPI, model);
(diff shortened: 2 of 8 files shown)


Sources/Widgets/Representations/ScaledSphereHandleRepresentation/index.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+  });
+}
+
+function attachScaleListener(
+  publicAPI,
+  model,
+  scaleHandler = defaultScaleHandler
+) {
+  let camsub = null;
```
finetjul, line 17: rename to `cameraSubscription` ?

Sources/Widgets/Representations/ScaledSphereHandleRepresentation/index.js
```diff
@@ -0,0 +1,91 @@
...
+  const setScaleFromCamera = (camera) => {
+    const scale = camera.getParallelProjection()
+      ? camera.getParallelScale() / 0.85
+      : camera.getDistance() / 3.37;
```
finetjul, line 21: explain what those magic numbers mean

finetjul: That works only if handle is in the same plane than focalPoint

finetjul:
I agree, it should probably not be into a new representation. To me that feature should be in WidgetRepresentation (maybe with a property such as "fix[ed]Size" or "scaleInDisplayCoordinates").

To the least, CircleContextRepresentation should probably have that feature.

floryst:
I'm revisiting the scaled representations and I have re-written the API for this. I'm still not relying on pixels for sizing (esp since I would imagine high DPI screens wouldn't like that), so instead I provide an API to scale based on some fraction of the world height of the viewport.

Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -203,6 +204,21 @@ function vtkWidgetRepresentation(publicAPI, model) {
     });
   };
 
+  publicAPI.getDisplayWorldHeightAt = (point3d) => {
```
finetjul, line 207: rename `point3d` into `worldCoord`

Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -203,6 +204,21 @@ function vtkWidgetRepresentation(publicAPI, model) {
...
+    const {
+      isParallel,
+      cameraPosition,
+      cameraDir,
+      dispHeightFactor,
```
finetjul, line 212: `get` parameters in same order than when `set` ?

Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -203,6 +204,21 @@ function vtkWidgetRepresentation(publicAPI, model) {
...
+    } = model.displayScaleParams;
+    if (isParallel) {
+      return dispHeightFactor;
+    }
+    const p = [...point3d];
```
finetjul, line 217: rename `p` to `worldCoordToCamera`

Sources/Widgets/Widgets3D/PolyLineWidget/state.js
```diff
@@ -8,7 +8,7 @@ export default function generateState() {
       mixins: ['origin', 'color', 'scale1', 'visible'],
       name: 'moveHandle',
       initialValues: {
-        scale1: 0.1,
+        scale1: 0.05,
```
finetjul, line 11: it would be good to add some explanation about this magic number

Sources/Widgets/Representations/SphereHandleRepresentation/index.js
```diff
@@ -132,6 +132,10 @@ function vtkSphereHandleRepresentation(publicAPI, model) {
         (!state.isVisible || state.isVisible() ? 1 : 0) *
         (state.getScale1 ? state.getScale1() : model.defaultScale);
 
+      if (publicAPI.getScaleByDisplay()) {
+        typedArray.scale[i] *= publicAPI.getDisplayWorldHeightAt(coord);
```
finetjul, line 136: I wonder if using `center` is not more reliable than `origin`

Sources/Widgets/Widgets3D/PolyLineWidget/state.js
floryst: When interpreted as a display scale for SphereHandleRepresentation, it's a fraction of the vertical height of the viewport. I'll add a comment about it.

Sources/Widgets/Representations/SphereHandleRepresentation/index.js
floryst: What do you mean by `center`?

finetjul: center of the `bounds` vs `origin` of the `bounds`.

floryst: `origin` is assumed to be the center of the sphere handle, which should match with the center of the sphere bounds.

finetjul: Fair enough

Sources/Widgets/Widgets3D/PolyLineWidget/index.js
```diff
@@ -38,9 +38,21 @@ function vtkPolyLineWidget(publicAPI, model) {
...
+              scaleByDisplay: true,
+            },
+          },
+          {
+            builder: vtkSphereHandleRepresentation,
+            labels: ['moveHandle'],
+            initialValues: {
+              scaleByDisplay: true,
+            },
+          },
```
agirault, line 54: @floryst could this have been shorten to this?
```js
          {
            builder: vtkSphereHandleRepresentation,
            labels: ['handles', 'moveHandle'],
            initialValues: {
              scaleByDisplay: true,
            },
          },
```

floryst: Yeah, good point. I'll refactor that at some other time.

PR #1435, opened 2020-05-07
Fix azure builds

floryst:
Installing the canvas package is having some issues, most notably node-pre-gyp isn't pulling prebuilts for it. So, we support fallback build-from-source. But even that doesn't work the first time you run `npm ci`, so run it twice for it to work.

This (hopefully) fixes the build issue where `karma` isn't found on LinuxNode12. That was due to the install being prematurely terminated because of a failed install of canvas.

### DIFF
--- a/Utilities/ci/azure-pipelines.yml
+++ b/Utilities/ci/azure-pipelines.yml
@@ -79,6 +79,7 @@ jobs:
       brew install pkg-config pixman cairo pango fribidi
       npm i -g npm@6.4.1
       npm ci
+      npm i --ignore-scripts # if canvas fails to load, then not all binaries will install
       npm run build:release
     displayName: 'Build'
 
--- a/Utilities/ci/azure-pipelines.yml
+++ b/Utilities/ci/azure-pipelines.yml
@@ -26,6 +26,7 @@ jobs:
     displayName: 'Install Node.js'
 
   - script: |
+      sudo apt install build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
       npm ci
       npm run build:release
     displayName: 'Build'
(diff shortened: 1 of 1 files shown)


floryst:
Huh, guess the first `npm ci` worked. I'll likely remove that second `npm ci` invocation then.

Utilities/ci/azure-pipelines.yml
```diff
@@ -64,6 +67,29 @@ jobs:
...
+
+  - script: |
+      brew install pkg-config pixman cairo pango
+      npm i -g npm@6.4.1
+      npm ci
+      npm run build:release
+    displayName: 'Build'
+
+- job: 'BuildTestWindows'
+  displayName: 'Build and test'
```
thewtex, line 89: -> *Build and test* -> *Build* (we are only running the tests on Linux because that is where we have a WebGL environment).

Utilities/ci/azure-pipelines.yml
```diff
@@ -47,8 +50,8 @@ jobs:
       testRunner: JUnit
       testResultsFiles: 'Utilities/TestResults/TESTS-*.xml'
 
-- job: 'Build'
-  displayName: 'Build'
+- job: 'BuildTestMacOS'
+  displayName: 'Build and test'
```
thewtex, line 54: -> *Build and test* -> *Build* (we are only running the tests on Linux because that is where we have a WebGL environment).

Utilities/ci/azure-pipelines.yml
```diff
@@ -64,6 +67,29 @@ jobs:
         poolName: 'Hosted macOS'
         imageName: 'macos-10.13'
```
thewtex, line 68: We may also want to update the VM's at some point.

floryst:
I've updated the VMs:
- ubuntu 16.04 -> 18.04
- windows server 2016 -> 2019
- macos 10.13 -> 10.15

PR #1440, opened 2020-05-11
doc(TS): Generated initial typescript definitions

jourdain:
This was just generated via the typescript compiler by running `npx tsc`. 

I have several questions:
1) Is that even useful or should we just write them by hand one by one? My guess is that it may help vtk.js dev to fill the gaps by fixing types and documentation in a progressive manner.
2) Can someone fix the root index.d.ts file? It is a bit funky and I'm not sure how to write it as ts so nested .d.ts can actually be resolved. Also, how to publish `vtk.d.ts` and `vtk-lite.d.ts` so CDN user can leverage those definitions?
3) What is the recommendation in term of API/doc generation? (typedoc)? When I started vtk.js I was hoping to rely on .d.ts file to provide both IDE and online documentation once the technology matured. I'm hoping that's the case. Please share any input or suggestion.


matthova:
> This was just generated via the typescript compiler by running `npx tsc`.
> 
> I have several questions:
> 
>     1. Is that even useful or should we just write them by hand one by one? My guess is that it may help vtk.js dev to fill the gaps by fixing types and documentation in a progressive manner.
> 
>     2. Can someone fix the root index.d.ts file? It is a bit funky and I'm not sure how to write it as ts so nested .d.ts can actually be resolved. Also, how to publish `vtk.d.ts` and `vtk-lite.d.ts` so CDN user can leverage those definitions?
> 
>     3. What is the recommendation in term of API/doc generation? (typedoc)? When I started vtk.js I was hoping to rely on .d.ts file to provide both IDE and online documentation once the technology matured. I'm hoping that's the case. Please share any input or suggestion.

1. imo, as a consumer of the library in a ts-app these type stubs are great placeholders until we can fill the gaps.  

2.a. Hmm, global variables are tough. Will think about this one a bit.  

2.b. It looks like RXJS had a similar question where [this](https://github.com/ReactiveX/rxjs/issues/1481#issuecomment-273854834) was one proposed solution  

3. js-doc with .d.ts declarations sounds like a good way to go. It looks like [three.js](https://github.com/mrdoob/three.js/blob/dev/src/extras/core/Curve.d.ts#L16) has taken the same route.

floryst:
Thanks @matthova for your comments. I'm going to update this branch with newly generated definitions, and take a look at your link to the rxjs proposed solution. Ideally I can revive this PR and get it merged.

jourdain:
@daker please submit a PR if you have something better like you are showing. Even if it is WIP, it could help us see the path you are taking and it will give us the opportunity to talk about it. Thanks! This looks great...

jourdain:
It looks like @daker is making good progress on that topic. Closing that specific PR. Feel free to create an Issue to keep track of the remaining pieces or what has been done so far on that topic.

Thanks all for your invaluable help!

PR #1441, opened 2020-05-11
fix(SliceRepresentationProxy): Use spacing in slice values

floryst:
When an image's spacing is not 1, then we may either skip slices or have too many slice values map to the same slice. This fixes that issue.

### DIFF
--- a/Sources/Proxy/Representations/SliceRepresentationProxy/index.js
+++ b/Sources/Proxy/Representations/SliceRepresentationProxy/index.js
@@ -245,24 +245,6 @@ function vtkSliceRepresentationProxy(publicAPI, model) {
     };
   };
 
-  // Used for UI
-  publicAPI.getSliceValues = (slicingMode = model.slicingMode) => {
-    const ds = publicAPI.getInputDataSet();
-    if (!ds) {
-      return [];
-    }
-    const values = [];
-    const bds = ds.getBounds();
-    const axisIndex = 'XYZ'.indexOf(slicingMode);
-    const endValue = bds[axisIndex * 2 + 1];
-    let currentValue = bds[axisIndex * 2];
-    while (currentValue <= endValue) {
-      values.push(currentValue);
-      currentValue++;
-    }
-    return values;
-  };
-
   const parentSetColorBy = publicAPI.setColorBy;
   publicAPI.setColorBy = (arrayName, arrayLocation, componentIndex = -1) => {
     if (arrayName === null) {
--- a/Sources/Proxy/Core/View2DProxy/index.js
+++ b/Sources/Proxy/Core/View2DProxy/index.js
@@ -201,13 +201,13 @@ function vtkView2DProxy(publicAPI, model) {
         update();
         nbListeners++;
       }
-      if (representation.getSlice && representation.getSliceValues) {
+      const domain = representation.getPropertyDomainByName('slice');
+      if (representation.getSlice && domain) {
         const update = () => setSlice(representation.getSlice());
-        const values = representation.getSliceValues();
         model.rangeManipulator.setScrollListener(
-          values[0],
-          values[values.length - 1],
-          values[1] - values[0] || 1,
+          domain.min,
+          domain.max,
+          domain.step,
           representation.getSlice,
           setSlice
         );
--- a/Sources/macro.js
+++ b/Sources/macro.js
@@ -1260,7 +1260,8 @@ export function proxy(publicAPI, model) {
   publicAPI.getPropertyByName = (name) =>
     getProperties().find((p) => p.name === name);
 
-  publicAPI.getPropertyDomainByName = (name) => propertyMap[name].domain;
+  publicAPI.getPropertyDomainByName = (name) =>
+    (propertyMap[name] || {}).domain;
 
   // ui section
   publicAPI.getProxySection = () => ({


Sources/Proxy/Representations/SliceRepresentationProxy/index.js
```diff
@@ -255,10 +255,11 @@ function vtkSliceRepresentationProxy(publicAPI, model) {
     const bds = ds.getBounds();
     const axisIndex = 'XYZ'.indexOf(slicingMode);
     const endValue = bds[axisIndex * 2 + 1];
+    const spacing = ds.getSpacing()[axisIndex];
```
agirault, line 258: That's not right: if you did IJK slicing, you could reuse the spacing values from the image data. When you are in XYZ, you can not just take that value if it's not aligned (worse: it could be along another direction).

I'll provide more insight and suggestions soon.

floryst: Ah, good point. I was only testing with IJK slicing.

agirault: `getClosestAxis` should help us here.

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
```diff
@@ -255,10 +255,12 @@ function vtkSliceRepresentationProxy(publicAPI, model) {
     const bds = ds.getBounds();
     const axisIndex = 'XYZ'.indexOf(slicingMode);
     const endValue = bds[axisIndex * 2 + 1];
+    const ijk = model.mapper.getClosestIJKAxis().ijkMode;
+    const spacing = ds.getSpacing()[ijk];
     let currentValue = bds[axisIndex * 2];
     while (currentValue <= endValue) {
       values.push(currentValue);
-      currentValue++;
+      currentValue += spacing;
```
agirault, line 263: Can the proxy be in IJK mode? If it's in IJK mode, the values should iterate by 1. Not sure that `SliceRepresentationProxy` supports it, or only XYZ.

floryst: Oh you're right, SliceRep does support IJK, but View2DProxy does not. I've been looking at this from the perspective of View2DProxy, but SliceRep does support IJK.

floryst: There's actually a lot of code in SliceRep that assumes XYZ and not IJK, so changing things would be more involved...

agirault: Yeah, I actually have two projects that use this in IJK mode when looking back.

What is the goal of `getSliceValues` when we have `updateDomains` already defining all the necessary values? See [here](https://github.com/Kitware/vtk-js/commit/7182b5a16c778ba92e788d61814d56113969a3ee) from the commit two years ago.

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
```diff
@@ -255,10 +255,13 @@ function vtkSliceRepresentationProxy(publicAPI, model) {
     const bds = ds.getBounds();
     const axisIndex = 'XYZ'.indexOf(slicingMode);
     const endValue = bds[axisIndex * 2 + 1];
+    const ijkAxis = model.mapper.getClosestIJKAxis().ijkMode;
+    const isXYZ = 'XYZ'.indexOf(model.mapper.getSlicingMode()) !== -1;
```
agirault, line 259: isn't
```js
'XYZ'.indexOf(model.mapper.getSlicingMode())
```
the same as
```js
const axisIndex = 'XYZ'.indexOf(slicingMode);
```
above?

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
floryst: I don't actually know why we have getSlicingValues(). I assumed it was to handle non-uniform spacing, but the logic wasn't even there for that. I'll look into using the results of updateDomains in Glance, rather than mess with getSlicingValues().

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
agirault: Also seems like all those values could be computed generically from looking at the content of `getState` which is computed by `updateDomains`?

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
agirault: Thanks @floryst. Looks like the result of `updateDomains` is cached in `state`.

jourdain: Looking at your commit @agirault , you renamed `getSliceIndexValues` by `getSliceValues`. So you replaced an `index` based method to a generic one when the domain was supposed to take care of the generic path. So now, indeed, the `getSliceValues` is useless and should probably be removed.

jourdain: My bet is that `getSliceIndexValues()` was pre-dating the domain part.

floryst:
I've removed `getSlicingValues`, since I don't think anybody was using it. View2DProxy now utilizes the slice domain computed in SliceRepresentationProxy in order to step through each slice.

Sources/Proxy/Representations/SliceRepresentationProxy/index.js
agirault: Thanks @jourdain.

I guess `getSliceValues` (and the previous `getSliceIndexValues`) could be useful if you don't only want the min, max, and step, but want all the values?

It's ironic that I linked to the commit where I made that change and did not properly support IJK at the time. Apologies for my 2y old mistake 🙏I was probably only using `updateDomains` and did not need all the slice values.

PR #1443, opened 2020-05-12
Spline widget fixes and improvements

### LINKED ISSUE #1423: SplineWidget issue if the projection axis isn't Z
Apply the following diff to the [`SplineWidget`](https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Widgets3D/SplineWidget/example/index.js) example:
```diff
diff --git a/Sources/Widgets/Widgets3D/SplineWidget/example/index.js b/Sources/Widgets/Widgets3D/SplineWidget/example/index.js
index ac1e5e62cf..3bfcacede9 100644
--- a/Sources/Widgets/Widgets3D/SplineWidget/example/index.js
+++ b/Sources/Widgets/Widgets3D/SplineWidget/example/index.js
@@ -27,6 +27,7 @@ const widget = vtkSplineWidget.newInstance();
 widgetManager.addWidget(widget);
 
 renderer.resetCamera();
+renderer.getActiveCamera().setDirectionOfProjection(0, 1, 0); // Anything but (0, 0, 1) or (0, 0, -1)
 widgetManager.enablePicking();
 
 // -----------------------------------------------------------
```

Leads to:
```error
macro.js:506 Uncaught RangeError: Invalid number of values for array setter (origin)
    at Object.publicAPI.<computed> [as setOrigin] (macro.js:506)
    at Object.publicAPI.handleMouseMove (behavior.js:297)
    at Object.<anonymous> (index.js:59)
    at _loop2 (macro.js:827)
    at Object.invoke [as invokeMouseMove] (macro.js:816)
    at Object.publicAPI.<computed> [as mouseMoveEvent] (index.js:754)
    at HTMLDivElement.publicAPI.handleMouseMove (index.js:463)
```

Analysis:
- `worldCoords` is passed as an empty array in `SplineWidget.handleMouseMove` [here](
https://github.com/Kitware/vtk-js/blob/e32e688/Sources/Widgets/Widgets3D/SplineWidget/behavior.js#L297)
- that value was queried slightly above from the `vtkPlaneManipulator` [here
](https://github.com/Kitware/vtk-js/blob/e32e688/Sources/Widgets/Widgets3D/SplineWidget/behavior.js#L297)
- which was queried in `intersectDisplayWithPlane` of `vtkPlaneManipulator` [here](https://github.com/Kitware/vtk-js/blob/e32e6885815181c23e2691ce6809c8f5d50549a1/Sources/Widgets/Manipulators/PlaneManipulator/index.js#L15)
- with `vtkPlane.intersectWithLine` failing to compute an intersection and returning [here](https://github.com/Kitware/vtk-js/blob/e32e6885815181c23e2691ce6809c8f5d50549a1/Sources/Common/DataModel/Plane/index.js#L100)
  - Example of values passed to `intersectWithLine` when the camera direction is set to [0, 1, 0]:
    - `p1 = (3) [0, 0, -0.43150606751441956]`
    - `p2 = (3) [0, 0, -0.6140561699867249]`
    - `origin = (3) [0, 0, 0]`
    - `normal = (3) [0, 1, 0]`
  - Example of values passed to `intersectWithLine` when the camera has the default direction (Z):
    - `p1 = (3) [-2.249490737915039, 1.2567908763885498, 1.1569218635559082]`
    - `p2 = (3) [-3.2011451721191406, 1.788480520248413, -1.1847667694091797]`
    - `origin = (3) [0, 0, 0]`
    - `normal = (3) [0, 0, 1]`
- It seems that `p1` and `p2` create a line parallel to the plane. They are computed as `mear` and `far` [here](https://github.com/Kitware/vtk-js/blob/e32e6885815181c23e2691ce6809c8f5d50549a1/Sources/Widgets/Manipulators/PlaneManipulator/index.js#L12-L13)

agirault:
_Extracted from #1422_

- The normal and origin of the plane manipulator should be set before searching for an intersection. Fixes #1423
- Improve the API
  - Make private methods actually private (updateHandlesSize, addPoint,
getHoveredHandle)
  - Use `macro.setGet` to add public APIs around the model properties, and
remove manual setters (sethandleSizeInPixels and setFreehandMinDistance)
  - Override setHandleSizeInPixels and setResolution to update the model
representations
  - Remove updateResolution which hardcoded a non-standard behavior when
pressing "Shift". Instead, the setResolution API can be used by the
application with a custom keyDown listener.
  - Add missing event callbacks in grab/loseFocus
- Improve the example:
  - Improve the control Panel to interact with most spline parameters

Screen Shot 2020-04-30 at 3 25 34 PM

Notes:
- color, scale and visibility of the handles should probably be a property
of the viewWidgets and not of the state.
- there are confusions between the model properties on the widget
(widget factory) and the behavior (view widget). The former can be set to
define the default values when instantiating a view widget, but the later
should be used to edit the properties of a view widget instance.

### DIFF
--- a/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
@@ -250,14 +250,6 @@ export default function widgetBehavior(publicAPI, model) {
       model.widgetState.deactivate();
     }
 
-    if (
-      (model.hasFocus && !model.activeState) ||
-      (model.activeState && !model.activeState.getActive())
-    ) {
-      model.widgetManager.enablePicking();
-      model.interactor.render();
-    }
-
     model.freeHand = false;
     model.isDragging = false;
     model.draggedPoint = false;
@@ -378,7 +370,6 @@ export default function widgetBehavior(publicAPI, model) {
     model.moveHandle.setVisible(false);
     model.activeState = null;
     model.interactor.render();
-    model.widgetManager.enablePicking();
 
     model.hasFocus = false;
   };
--- a/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
@@ -205,6 +205,7 @@ export default function widgetBehavior(publicAPI, model) {
       model.isDragging = true;
       model.openGLRenderWindow.setCursor('grabbing');
       model.interactor.requestAnimation(publicAPI);
+      publicAPI.invokeStartInteractionEvent();
     }
 
     return macro.EVENT_ABORT;
@@ -220,6 +221,7 @@ export default function widgetBehavior(publicAPI, model) {
         model.openGLRenderWindow.setCursor(model.defaultCursor);
         model.widgetState.deactivate();
         model.interactor.cancelAnimation(publicAPI);
+        publicAPI.invokeEndInteractionEvent();
       } else {
         model.moveHandle.setOrigin(...model.activeState.getOrigin());
         model.activeState.deactivate();
--- a/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
@@ -272,11 +272,14 @@ export default function widgetBehavior(publicAPI, model) {
     if (
       !model.activeState ||
       !model.activeState.getActive() ||
-      !model.pickable
+      !model.pickable ||
+      !model.manipulator
     ) {
       return macro.VOID;
     }
 
+    model.manipulator.setOrigin(model.activeState.getOrigin());
+    model.manipulator.setNormal(model.camera.getDirectionOfProjection());
     const worldCoords = model.manipulator.handleEvent(
       callData,
       model.openGLRenderWindow
@@ -290,9 +293,6 @@ export default function widgetBehavior(publicAPI, model) {
       model.moveHandle.setVisible(true);
       model.openGLRenderWindow.setCursor(model.defaultCursor);
     }
-
-    model.manipulator.setOrigin(worldCoords);
-    model.manipulator.setNormal(model.camera.getDirectionOfProjection());
 
     if (model.lastHandle) {
       model.lastHandle.setVisible(true);
--- a/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
@@ -5,28 +5,10 @@ export default function widgetBehavior(publicAPI, model) {
   model.classHierarchy.push('vtkSplineWidgetProp');
 
   // --------------------------------------------------------------------------
-  // Display 2D
+  // Private methods
   // --------------------------------------------------------------------------
 
-  publicAPI.setDisplayCallback = (callback) =>
-    model.representations[0].setDisplayCallback(callback);
-
-  // --------------------------------------------------------------------------
-  // Public methods
-  // --------------------------------------------------------------------------
-
-  publicAPI.getPoints = () =>
-    model.representations[1].getOutputData().getPoints().getData();
-
-  // --------------------------------------------------------------------------
-
-  publicAPI.sethandleSizeInPixels = (handleSizeInPixels) => {
-    model.handleSizeInPixels = handleSizeInPixels;
-  };
-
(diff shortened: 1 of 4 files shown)


agirault:
Discussion from #1422:

@agirault:
> Also, I'm having some issues in terms of interaction with the `SplineWidget`: I can't seem to be able to stop moving its handles to "complete" it. Pressing "Enter" does trigger the `onEndInteractionEvent` though, but the last handle still follow the mouse, so I'm thinking that's a bug? If you agree, I might try to look into fixing that (could do in that PR), and welcome any pointer. My hope is to be able to create a spline, then move previously created handles without having a handle stuck to the mouse.

@finetjul:
>  I am not surprised that "Enter" does not work for you. Remi's use case was to use the SplineWidget for painting, so on `endInteractionEvent`, he was deleting the widget. I am afraid I won't be of much help here, it's @RemiCecchinato who knows it best.

@agirault:
> Sounds good. As we discussed, I'll work on customizing the spline widget with behavior modifiers as parameters, similar to the shape widgets.

@finetjul:
> In SplineWidget example:
> - when changing the resolution with the slider and moving the mouse on the render view,  the clickable area (when you mouse over the shape) is not updated: the previous resolution is still used
> - after closing the shape (Enter), it is abnormally slow after releasing a handle [...] On Chrome, a simple click (no drag) on a closed spline handle takes more than 600ms for mouseup event: `[Violation] 'mouseup' handler took 672ms index.js:526`
> - when placing spline widget, a click (no drag) on first (or last) placed point or last point should close the shape. [...] That feature was already existing (see [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Widgets3D/SplineWidget/behavior.js#L231-L247 ))

@agirault:
> - [resolution] Still don't know why this occurs. I set the resolution on the representation (`model.representations[1]`) the same way it was done before my changes in `updateResolution`. One observation: as you as you click once in the renderer, then the pointer will get back to change properly above the spline area representation based on the new resolution.
> - [speed] It seems to come from this `cancelAnimation` [here](https://github.com/Kitware/vtk-js/blob/e32e688/Sources/Widgets/Widgets3D/SplineWidget/behavior.js#L224), which we never encountered before since the example would not lose focus. @floryst do you know why it would be so slow?
> - [close] Thanks @finetjul, I leveraged that instead.

@finetjul:
>Out of date picking may be fixed with `widgetManager.setCaptureOn(CaptureOn.MOUSE_MOVE);`

agirault:
@finetjul: The speed issue is handled by removing `widgetManager.enablePicking()` (already called by `interactor.cancelAnimation()`), hence cutting the time by half, as well as switching the widget manager `captureOn` to `MOUSE_MOVE`, done in #1452. This also handles properly updating the picking which fixes the issue after changing the resolution.

agirault:
@finetjul want to check this PR? Same as on the previous PR, with fix for the resolution and the speed issue handled by the last commit and #1452

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
```diff
@@ -317,24 +312,24 @@ export default function widgetBehavior(publicAPI, model) {
...
-      if (key === 'Escape') {
-        publicAPI.reset();
         publicAPI.loseFocus();
-      } else if (key === 'Delete' || key === 'Backspace') {
-        if (model.lastHandle) {
-          model.widgetState.removeHandle(model.lastHandle);
+      }
+    } else if (key === 'Escape') {
+      publicAPI.reset();
+      publicAPI.loseFocus();
```
agirault, line 326: @finetjul @floryst what do you think about using escape to only complete (loseFocus, which calls `invokeEndInteractionEvent`) without resetting/hiding the representation, and leave it to the application to reset it if it wants by listening to that callback.

Then we would have it lose focus on `Escape`, and maybe `Enter` as well (though `Enter` isn't used in other widgets AFAIK)? If we wanted to configure something on this widget, I'd suggest that would be a list of keys that would end the interaction (defaulting to `['Escape']`).

agirault: @floryst : I talked with @finetjul and he'd like to keep the current behavior as a default, at least for now, until we maybe add some kinf of "Action Manager" to configure behavior (and not just representation). I'm fine with that, and keeping this as is.

floryst: I'm also okay with this then for now. It would be nice to have this customizable from an application perspective in the future.

PR #1445, opened 2020-05-13
Reslice cursor new widget architecture

laurennlam:
Todo:

- [x] Improve speed

- [ ] More tolerant picking --> improvement for next MR

- [x] Update cursor on different handle

- [ ] Update handle representation color when mouse is on it --> improvement for next MR

- [ ] Replace sphere by circle --> improvement for next MR

@finetjul

### DIFF
--- a/Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
+++ b/Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
@@ -1,0 +1,233 @@
+import macro from 'vtk.js/Sources/macro';
+
+import vtkActor from 'vtk.js/Sources/Rendering/Core/Actor';
+import vtkLineSource from 'vtk.js/Sources/Filters/Sources/LineSource';
+import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
+import vtkPolyData from 'vtk.js/Sources/Common/DataModel/PolyData';
+import vtkSphereSource from 'vtk.js/Sources/Filters/Sources/SphereSource';
+import vtkWidgetRepresentation from 'vtk.js/Sources/Widgets/Representations/WidgetRepresentation';
+
+import { RenderingTypes } from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
+
+// ----------------------------------------------------------------------------
+// vtkResliceCursorContextRepresentation methods
+// ----------------------------------------------------------------------------
+
+function vtkResliceCursorContextRepresentation(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkResliceCursorContextRepresentation');
+
+  // --------------------------------------------------------------------------
+  // Generic rendering pipeline
+  // --------------------------------------------------------------------------
+
+  model.mapper = vtkMapper.newInstance();
+  model.actor = vtkActor.newInstance();
+  model.mapper.setInputConnection(publicAPI.getOutputPort());
+  model.actor.setMapper(model.mapper);
+  publicAPI.addActor(model.actor);
+
+  model.pipelines = {};
+  model.pipelines.center = {
+    source: vtkSphereSource.newInstance(),
+    mapper: vtkMapper.newInstance(),
+    actor: vtkActor.newInstance(),
+  };
+  model.pipelines.axes = [];
+  // Create axis 1
+  const axis1 = {};
+  axis1.line = {
+    source: vtkLineSource.newInstance(),
+    mapper: vtkMapper.newInstance(),
+    actor: vtkActor.newInstance({ pickable: true }),
+  };
+  axis1.rotation1 = {
+    source: vtkSphereSource.newInstance(),
+    mapper: vtkMapper.newInstance(),
+    actor: vtkActor.newInstance({ pickable: true }),
+  };
+  axis1.rotation2 = {
+    source: vtkSphereSource.newInstance(),
+    mapper: vtkMapper.newInstance(),
+    actor: vtkActor.newInstance({ pickable: true }),
+  };
+  // Create axis 2
+  const axis2 = {};
+  axis2.line = {
+    source: vtkLineSource.newInstance(),
+    mapper: vtkMapper.newInstance(),
+    actor: vtkActor.newInstance({ pickable: true }),
+  };
+  axis2.rotation1 = {
+    source: vtkSphereSource.newInstance(),
+    mapper: vtkMapper.newInstance(),
+    actor: vtkActor.newInstance({ pickable: true }),
+  };
+  axis2.rotation2 = {
+    source: vtkSphereSource.newInstance(),
+    mapper: vtkMapper.newInstance(),
+    actor: vtkActor.newInstance({ pickable: true }),
+  };
+
+  model.pipelines.axes.push(axis1);
+  model.pipelines.axes.push(axis2);
+
+  // Improve actors rendering
+  model.pipelines.center.actor.getProperty().setAmbient(1, 1, 1);
+  model.pipelines.center.actor.getProperty().setDiffuse(0, 0, 0);
+
+  vtkWidgetRepresentation.connectPipeline(model.pipelines.center);
+  publicAPI.addActor(model.pipelines.center.actor);
+
+  model.pipelines.axes.forEach((axis) => {
+    Object.values(axis).forEach((lineOrRotationHandle) => {
+      vtkWidgetRepresentation.connectPipeline(lineOrRotationHandle);
+      const actor = lineOrRotationHandle.actor;
+      actor.getProperty().setAmbient(1, 1, 1);
+      actor.getProperty().setDiffuse(0, 0, 0);
+      publicAPI.addActor(actor);
+    });
+  });
+
+  publicAPI.setSphereRadius = (radius) => {
+    model.pipelines.center.source.setRadius(radius);
+    model.pipelines.axes[0].rotation1.source.setRadius(radius);
+    model.pipelines.axes[0].rotation2.source.setRadius(radius);
+    model.pipelines.axes[1].rotation1.source.setRadius(radius);
+    model.pipelines.axes[1].rotation2.source.setRadius(radius);
+  };
+
+  publicAPI.setSphereRadius(0.05);
+
+  function updateRender(state, axis) {
+    const color = state.getColor();
+    axis.line.actor.getProperty().setColor(color);
+    axis.rotation1.actor.getProperty().setColor(color);
+    axis.rotation2.actor.getProperty().setColor(color);
+
+    axis.line.source.setPoint1(state.getPoint1());
+    axis.line.source.setPoint2(state.getPoint2());
+    axis.rotation1.source.setCenter(state.getRotationPoint1());
+    axis.rotation2.source.setCenter(state.getRotationPoint2());
+  }
+
+  publicAPI.requestData = (inData, outData) => {
+    const state = inData[0];
+
+    const origin = state.getCenter();
+    model.pipelines.center.source.setCenter(origin);
+
+    const getAxis1 = `get${model.axis1Name}`;
+    const getAxis2 = `get${model.axis2Name}`;
+    const axis1State = state[getAxis1]();
+    const axis2State = state[getAxis2]();
+
+    updateRender(axis1State, model.pipelines.axes[0]);
+    updateRender(axis2State, model.pipelines.axes[1]);
+
+    publicAPI.setSphereRadius(state.getSphereRadius());
+
+    outData[0] = vtkPolyData.newInstance();
+  };
+
+  publicAPI.updateActorVisibility = (
+    renderingType,
+    wVisible,
+    ctxVisible,
+    hVisible
+  ) => {
+    if (renderingType === RenderingTypes.PICKING_BUFFER) {
+      publicAPI.getActors().forEach((actor) => actor.setVisibility(wVisible));
+    } else {
+      publicAPI
+        .getActors()
+        .forEach((actor) => actor.setVisibility(wVisible && hVisible));
+    }
+  };
+
+  publicAPI.getSelectedState = (prop, compositeID) => {
+    const state = model.inputData[0];
+    state.setActiveViewName(model.viewName);
+
+    const getAxis1 = `get${model.axis1Name}`;
+    const getAxis2 = `get${model.axis2Name}`;
+    const axis1State = state[getAxis1]();
+    const axis2State = state[getAxis2]();
+
+    let activeLineState = null;
+    let activeRotationPointName = '';
+    let methodName = '';
+
+    switch (prop) {
+      case model.pipelines.axes[0].line.actor:
+        activeLineState = axis1State;
+        methodName = 'translateAxis';
+        break;
+      case model.pipelines.axes[1].line.actor:
+        activeLineState = axis2State;
+        methodName = 'translateAxis';
+        break;
+      case model.pipelines.axes[0].rotation1.actor:
+        activeLineState = axis1State;
+        activeRotationPointName = 'RotationPoint1';
+        methodName = 'rotateLine';
+        break;
+      case model.pipelines.axes[0].rotation2.actor:
+        activeLineState = axis1State;
+        activeRotationPointName = 'RotationPoint2';
+        methodName = 'rotateLine';
+        break;
+      case model.pipelines.axes[1].rotation1.actor:
+        activeLineState = axis2State;
+        activeRotationPointName = 'RotationPoint1';
+        methodName = 'rotateLine';
+        break;
+      case model.pipelines.axes[1].rotation2.actor:
+        activeLineState = axis2State;
+        activeRotationPointName = 'RotationPoint2';
+        methodName = 'rotateLine';
+        break;
+      default:
+        methodName = 'translateCenter';
+        break;
+    }
+
+    state.setActiveLineState(activeLineState);
+    state.setActiveRotationPointName(activeRotationPointName);
+    state.setUpdateMethodName(methodName);
+
+    return state;
+  };
+}
+
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {
+  axis1Name: '',
+  axis2Name: '',
+  viewName: '',
+};
+
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  vtkWidgetRepresentation.extend(publicAPI, model, initialValues);
+
+  // Object specific methods
+  vtkResliceCursorContextRepresentation(publicAPI, model);
+}
+
+// ----------------------------------------------------------------------------
+
+export const newInstance = macro.newInstance(
+  extend,
+  'vtkResliceCursorContextRepresentation'
+);
+
+// ----------------------------------------------------------------------------
+
+export default { newInstance, extend };
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
@@ -1,0 +1,190 @@
+import { mat4 } from 'gl-matrix';
+
+import macro from 'vtk.js/Sources/macro';
+import vtkLine from 'vtk.js/Sources/Common/DataModel/Line';
+import vtkPlaneManipulator from 'vtk.js/Sources/Widgets/Manipulators/PlaneManipulator';
+import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
+
+import {
+  getAssociatedLinesName,
+  updateState,
+} from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers';
+
+export default function widgetBehavior(publicAPI, model) {
+  let isDragging = null;
+
+  publicAPI.updateCursor = () => {
+    switch (model.activeState.getUpdateMethodName()) {
+      case 'translateCenter':
+        model.openGLRenderWindow.setCursor('move');
+        break;
+      case 'rotateLine':
+        model.openGLRenderWindow.setCursor('alias');
+        break;
+      case 'translateAxis':
+        model.openGLRenderWindow.setCursor('pointer');
+        break;
+      default:
+        model.openGLRenderWindow.setCursor('default');
+        break;
+    }
+  };
+
+  publicAPI.handleLeftButtonPress = (callData) => {
+    if (!model.activeState || !model.activeState.getActive()) {
+      return macro.VOID;
+    }
+    isDragging = true;
+    const viewName = model.widgetState.getActiveViewName();
+    const currentPlaneNormal = model.widgetState[`get${viewName}PlaneNormal`]();
+    model.planeManipulator.setOrigin(model.widgetState.getCenter());
+    model.planeManipulator.setNormal(currentPlaneNormal);
+
+    model.interactor.requestAnimation(publicAPI);
+    publicAPI.invokeStartInteractionEvent();
+
+    return macro.EVENT_ABORT;
+  };
+
+  publicAPI.handleMouseMove = (callData) => {
+    if (isDragging && model.pickable) {
+      return publicAPI.handleEvent(callData);
+    }
+    return macro.VOID;
+  };
+
+  publicAPI.handleLeftButtonRelease = () => {
+    if (isDragging) {
+      publicAPI.invokeEndInteractionEvent();
+      model.interactor.cancelAnimation(publicAPI);
+    }
+    isDragging = false;
+    model.widgetState.deactivate();
+  };
+
+  publicAPI.handleEvent = (callData) => {
+    if (model.activeState.getActive()) {
+      publicAPI[model.activeState.getUpdateMethodName()](callData);
+      publicAPI.invokeInteractionEvent();
+      return macro.EVENT_ABORT;
+    }
+    return macro.VOID;
+  };
+
+  publicAPI.translateAxis = (calldata) => {
+    const stateLine = model.widgetState.getActiveLineState();
+    const worldCoords = model.planeManipulator.handleEvent(
+      calldata,
+      model.openGLRenderWindow
+    );
+
+    const point1 = stateLine.getPoint1();
+    const point2 = stateLine.getPoint2();
+
+    // Translate the current line along the other line
+    const otherLineName = getAssociatedLinesName(stateLine.getName());
+    const otherLine = model.widgetState[`get${otherLineName}`]();
+    const otherLinePoint1 = otherLine.getPoint1();
+    const otherLinePoint2 = otherLine.getPoint2();
+    const otherLineVector = [];
+    vtkMath.subtract(otherLinePoint2, otherLinePoint1, otherLineVector);
+    vtkMath.normalize(otherLineVector);
+    const axisTranslation = otherLineVector;
+
+    const currentLineVector = [];
+    vtkMath.subtract(point2, point1, currentLineVector);
+    vtkMath.normalize(currentLineVector);
+
+    const dot = vtkMath.dot(currentLineVector, otherLineVector);
+    // lines are colinear, translate along perpendicular axis from current line
+    if (dot === 1 || dot === -1) {
+      vtkMath.cross(
+        currentLineVector,
+        model.planeManipulator.getNormal(),
+        axisTranslation
+      );
+    }
+
+    const closestPoint = [];
+    vtkLine.distanceToLine(worldCoords, point1, point2, closestPoint);
+
+    const translationVector = [];
+    vtkMath.subtract(worldCoords, closestPoint, translationVector);
+    const translationDistance = vtkMath.dot(translationVector, axisTranslation);
+
+    const center = model.widgetState.getCenter();
+    const newOrigin = [
+      center[0] + translationDistance * axisTranslation[0],
+      center[1] + translationDistance * axisTranslation[1],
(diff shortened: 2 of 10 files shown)


Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+  function updatePlanePoints(line1, line2, origin, planeSource) {
+    const point1 = getClosestPoint(origin, [
+      line1.getPoint1(),
+      line1.getPoint2(),
+    ]);
+    const point2 = getClosestPoint(origin, [
+      line2.getPoint1(),
+      line2.getPoint2(),
+    ]);
+    planeSource.setPoint1(point2);
```
finetjul, line 46: planeSource.setPoint1(point2); -> planeSource.setPoint1(point1); ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+    planeSource.setPoint2(point1);
+  }
+
+  function computeReslicePlaneOrigin(viewType) {
```
finetjul, line 50: let's try to keep the use of `viewType` as limited as possible. Use `[plane]normal`/orientation` instead.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+        bounds[2] + offset[1],
+        bounds[5] - offset[2]
+      );
+    }
+    return planeSource;
+  }
+
+  function updateCamera(renderer, normal) {
+    // When the reslice plane is changed, update the camera to look at the
+    // normal to the reslice plane always.
```
finetjul, line 131: remove `always`

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+
+    const focalPoint = renderer.getActiveCamera().getFocalPoint();
+    const position = renderer.getActiveCamera().getPosition();
+
+    const distance = Math.sqrt(
```
finetjul, line 136: renderer.getActiveCamera().getDistance();

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+      focalPoint[1] + distance * normal[1],
+      focalPoint[2] + distance * normal[2],
+    ];
+
+    // intersect with the plane to get updated focal point
+    const intersection = vtkPlane.intersectWithLine(
+      focalPoint,
+      estimatedCameraPosition,
+      model.widgetState.getCenter(),
+      normal
```
finetjul, line 151: odd that you have normal as parameter but not origin. You should be consistent

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+
+    renderer
+      .getActiveCamera()
+      .setPosition(
+        newCameraPosition[0],
+        newCameraPosition[1],
+        newCameraPosition[2]
+      );
+
+    // // Renderer may not have yet actor bounds
```
finetjul, line 173: remove double commet

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+      case ViewTypes.SLICE:
+      case ViewTypes.VOLUME:
+      default:
+        return [];
+    }
+  };
+
+  publicAPI.setImage = (image) => {
+    model.widgetState.setImage(image);
+    model.image = image;
```
finetjul, line 243: why store image if it is already in widgetState ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+      case ViewTypes.GEOMETRY:
+      case ViewTypes.SLICE:
+      case ViewTypes.VOLUME:
+      default:
+        return [];
+    }
+  };
+
+  publicAPI.setImage = (image) => {
+    model.widgetState.setImage(image);
```
finetjul, line 242: should this really go in widgetState ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+  // Methods
+  // --------------------------------------------------------------------------
+
+  publicAPI.resetCamera = (renderer, viewType) => {
+    let viewName = 'Axial';
+    if (viewType === ViewTypes.SAGITTAL) {
+      viewName = 'Sagittal';
+    } else if (viewType === ViewTypes.CORONAL) {
+      viewName = 'Coronal';
+    }
```
finetjul, line 259: create convenient function to convert from ViewTypes to viewName

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+
+    const bounds = model.image.getBounds();
+    const center = [
+      (bounds[0] + bounds[1]) / 2,
+      (bounds[2] + bounds[3]) / 2,
+      (bounds[4] + bounds[5]) / 2,
+    ];
```
finetjul, line 266: image.getCenter() ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+    const estimatedFocalPoint = center;
+    const estimatedCameraPosition = [
+      estimatedFocalPoint[0] + distance * normal[0],
+      estimatedFocalPoint[1] + distance * normal[1],
+      estimatedFocalPoint[2] + distance * normal[2],
+    ];
+
+    renderer
+      .getActiveCamera()
+      .setFocalPoint(
```
finetjul, line 287: setFocalPoint(estimatedFocalPoint)

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+        estimatedCameraPosition[0],
+        estimatedCameraPosition[1],
+        estimatedCameraPosition[2]
+      );
+
+    // Project focalPoint onto image plane and preserve distance
+    updateCamera(renderer, normal);
+
+    // Reset clipping range
+    updateCamera(renderer, normal);
```
finetjul, line 305: duplicated call

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+      0,
+    ]);
+    imageReslice.setOutputExtent([0, extentX - 1, 0, extentY - 1, 0, 0]);
+
+    return newResliceAxes;
+  };
+
+  publicAPI.getPlaneAndPlaneSourceFromViewType = (type) => {
+    const returnedPlaneSource = vtkPlaneSource.newInstance();
+    const returnedPlane = vtkPlane.newInstance();
```
finetjul, line 499: returnedPlane -> plane

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+      0.5 * outputSpacingY,
+      0,
+    ]);
+    imageReslice.setOutputExtent([0, extentX - 1, 0, extentY - 1, 0, 0]);
+
+    return newResliceAxes;
+  };
+
+  publicAPI.getPlaneAndPlaneSourceFromViewType = (type) => {
+    const returnedPlaneSource = vtkPlaneSource.newInstance();
```
finetjul, line 498: returnedPlaneSource -> planeSource

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,562 @@
...
+    const returnedPlane = vtkPlane.newInstance();
+    const widgetState = publicAPI.getWidgetState();
+    const origin = widgetState.getCenter();
+    returnedPlaneSource.setOrigin(origin);
```
finetjul, line 502: the origin of a planesource should be the lowerleft corner, not necessarily the intersection point of all planes.

Sources/Widgets/Widgets3D/ResliceCursorWidget/state.js
```diff
@@ -0,0 +1,118 @@
...
+    .addField({ name: 'activeHandle', initialValue: null })
+    .addField({ name: 'image', initialValue: null })
+    .addField({ name: 'activeViewName', initialValue: '' })
+    .addField({ name: 'sphereRadius', initialValue: 5 })
+    .addField({
+      name: 'updateMethodName',
+    })
+    .addField({ name: 'sagittalPlaneNormal', initialValue: [1, 0, 0] })
+    .addField({ name: 'coronalPlaneNormal', initialValue: [0, -1, 0] })
+    .addField({ name: 'axialPlaneNormal', initialValue: [0, 0, 1] })
```
finetjul, line 116: sometimes you call it axial, and sometimes Z, it would be less confusing if you pick only 1 naming convention

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -0,0 +1,157 @@
...
+    return macro.VOID;
+  };
+
+  publicAPI.handleLeftButtonRelease = () => {
+    if (isDragging) {
+      publicAPI.invokeEndInteractionEvent();
+      model.interactor.cancelAnimation(publicAPI);
+    }
+    isDragging = false;
+    model.widgetState.deactivate();
```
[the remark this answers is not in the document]
laurennlam: This piece of code comes from the ImplicitPlaneWidget. I don't know the best.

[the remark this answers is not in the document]
finetjul: I think it is a mistake in ImplicitPlaneWidget :-)

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -0,0 +1,157 @@
...
+
+    const vOriginPrevious = [];
+    vtkMath.subtract(previousWorldPosition, origin, vOriginPrevious);
+    vtkMath.normalize(vOriginPrevious);
+
+    const vOriginNext = [];
+    vtkMath.subtract(worldCoords, origin, vOriginNext);
+    vtkMath.normalize(vOriginNext);
+
+    const rotationAngle = vtkMath.angleBetweenVectors(
```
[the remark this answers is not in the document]
laurennlam: rotation angle will be zero, won't it ?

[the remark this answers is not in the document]
finetjul: indeed

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: Not sure to understand whyyou want to use the orientation. The goal of this method is to setup a basic plane with bounded points. Then, we apply the effective normal to the plane, which will update the point.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: Not sure to understand. Did you mean that I should replace `model.widgetState.getCenter()` by a constant named origin ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: I need the image's bounds when I update the state. Maybe only storing the bounds in the state is enough.

laurennlam: Well, I also need the center, so I'll keep the image only in the state.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
finetjul: I don't like `viewType` very much (a bit hack-ish), therefore I want to make sure it is used as few as possible.
If viewType was replaced by slicingMode (0, 1, 2 for i, j, k), you could then factorize setOrigin, setPoint1 and setPoint2. e.g.
```
const origin = [
  bounds[0] + offset[0],
  bounds[2] + offset[1],
 bounds[2] + offset[1]
];
origin[slicingMode] = center[slicingMode];
```
Now, if you have slicingMode working, you can simply support the more generic version:
```
const mainOrientation = normal.reduce((max, component) => Math.max(max, component);
slicingMode = normal.indexOf(mainOrientation);
```

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
finetjul: Me neither, I do not remember what I had in mind

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
finetjul: Yes, I think you must keep only bounds. 
Center can be computed with vtkBoundingBox (or vtkBox)

laurennlam: Is it relevant to recompute the center each time we update the state ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -0,0 +1,157 @@
...
+
+    const closestPoint = [];
+    vtkLine.distanceToLine(worldCoords, point1, point2, closestPoint);
+
+    const move = [];
+    vtkMath.subtract(worldCoords, closestPoint, move);
+    const dot = vtkMath.dot(move, currentLineVector);
+
+    const origin = model.widgetState.getCenter();
+    const newOrigin = [
```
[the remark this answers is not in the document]
laurennlam: If lines are colinear, we don't want to allow the axis translation, is it correct ?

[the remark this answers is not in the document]
laurennlam: But along which axis the line will be translated ? Not sure that it makes sense.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: I didn't know that it was recomputed. So let's store only the bounds.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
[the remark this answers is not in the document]
finetjul: perpendicular to the current line

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
finetjul: Me neither :-)

However, we tend to not store the same (or derived) information multiple times. It typically leads to synchronization problems/mistakes.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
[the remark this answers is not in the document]
laurennlam: If we do that, then we'll have to update the second axis too. Which is not the goal.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: I don't understand the issue. Why is it not a good idea to use it here ? The viewType is only use in this file.
Do you mean that if I don't use viewType, then in function updateReslicePlane, I'll have to convert viewType into an int (0, 1 or 2 ) ? Which I think, is a little bit more confusing if we want to follow the workflow. I'd like to have a readable code and I'm afraid that this solution complicate the readability of the code (conversion between number and explicit enum)

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: New information after rechecking. I also need the spacing of the image. What's the best ? Storing the whole image or storing only bounds and spacing ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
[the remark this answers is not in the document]
finetjul: Not translating the line is not what the user would expect.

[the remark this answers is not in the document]
laurennlam: So what's the best way ?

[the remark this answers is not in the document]
finetjul: Wouldn't the following work: 
Leave it as is for regular cases, if other line is colinear with current line, translate center perpendicularly to current line.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
finetjul: I do not like the new ViewTypes enum values. However I have nothing better to suggest. So leave it as is

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
```diff
@@ -0,0 +1,247 @@
...
+
+import vtkActor from 'vtk.js/Sources/Rendering/Core/Actor';
+import vtkLineSource from 'vtk.js/Sources/Filters/Sources/LineSource';
+import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
+import vtkPolyData from 'vtk.js/Sources/Common/DataModel/PolyData';
+import vtkSphereSource from 'vtk.js/Sources/Filters/Sources/SphereSource';
+import vtkWidgetRepresentation from 'vtk.js/Sources/Widgets/Representations/WidgetRepresentation';
+import WidgetManagerConst from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
+
+const { RenderingTypes } = WidgetManagerConst;
```
finetjul, line 11: combine both lines into:
```
import { RenderingTypes } from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
```

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
```diff
@@ -0,0 +1,247 @@
...
+
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  vtkWidgetRepresentation.extend(publicAPI, model, initialValues);
+
+  model.image = null;
+  model.center = [0, 0, 0];
```
finetjul, line 232: still needed ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -0,0 +1,158 @@
...
+import vtkLine from 'vtk.js/Sources/Common/DataModel/Line';
+import vtkPlaneManipulator from 'vtk.js/Sources/Widgets/Manipulators/PlaneManipulator';
+import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
+
+import {
+  getAssociatedLinesName,
+  updateState,
+} from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers';
+
+import { mat4 } from 'gl-matrix';
```
finetjul, line 11: I think external includes should be done before VTK-JS includes

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -0,0 +1,223 @@
...
+    case 'axisXinZ':
+      return 'AxisYinZ';
+    case 'axisYinX':
+      return 'AxisZinX';
+    case 'axisYinZ':
+      return 'AxisXinZ';
+    case 'axisZinX':
+      return 'AxisYinX';
+    case 'axisZinY':
+      return 'AxisXinY';
```
finetjul, line 91: the casing inconsistency is puzzling

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -0,0 +1,223 @@
...
+  lineState.setPoint1(p1);
+  lineState.setPoint2(p2);
+  lineState.setRotationPoint1(rotationP1);
+  lineState.setRotationPoint2(rotationP2);
+}
+
+// Update the reslice cursor state according to the three planes normals and the origin
+export function updateState(widgetState) {
+  // Compute axis
+  const XNormal = widgetState.getXPlaneNormal();
```
finetjul, line 142: do not start variables with uppercase

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -0,0 +1,223 @@
...
+    widgetState.getAxisZinX(),
+    center,
+    newYAxis,
+    pdLength,
+    YRotationLength
+  );
+}
+
+// Get the closest point between rotation point of the line to the worldCoord
+export function getClosestLinePoint(worldCoord, lineState) {
```
finetjul, line 211: not needed anymore

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,526 @@
...
+  function updatePlanePoints(line1, line2, origin, planeSource) {
+    const point1 = getClosestPoint(origin, [
+      line1.getPoint1(),
+      line1.getPoint2(),
+    ]);
+    const point2 = getClosestPoint(origin, [
+      line2.getPoint1(),
+      line2.getPoint2(),
+    ]);
+    planeSource.setPoint1(point1);
```
finetjul, line 47: Taking the closest line (i.e.g segment) point as plane source point1 does not seem like the right approach.

Or said differently, I am not sure what that means.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -0,0 +1,526 @@
...
+
+    // Renderer may not have yet actor bounds
+    const rendererBounds = renderer.computeVisiblePropBounds();
+    const bounds = model.widgetState.getImage().getBounds();
+    rendererBounds[0] = Math.min(bounds[0], rendererBounds[0]);
+    rendererBounds[1] = Math.max(bounds[1], rendererBounds[1]);
+    rendererBounds[2] = Math.min(bounds[2], rendererBounds[2]);
+    rendererBounds[3] = Math.max(bounds[3], rendererBounds[3]);
+    rendererBounds[4] = Math.min(bounds[4], rendererBounds[4]);
+    rendererBounds[5] = Math.max(bounds[5], rendererBounds[5]);
```
finetjul, line 179: use vtkBoundingBox or vtkBox for such work on bounds

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
laurennlam: It's still a good way to initialize attribute, isn't it ?

finetjul: There are other places where it can be initialized (e.g. in DEFAULT_VALUES).

But it seems that model.center and model.image are never used, are they?

laurennlam: My bad. I thought it was in the ResliceCursorWidget/index.js file. Yes, indeed we can remove it.

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
laurennlam: I updated it with uppercase. Because the generated getter need uppercase for the first letter.

finetjul: Ok, but you could have used the `capitalize()` function in macro.js

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: There is no method which doing that job in these classes. Is it relevant to adding a method which takes two bounds and return the maximum bounds ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: I'll keep the image for now.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
finetjul: `vtkBoundingBox` has a `addBounds` method that does what you want...

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
laurennlam: I didn't know this function...

jourdain: The getters use internally the `capitalize()` function to generate the proper `getFieldName()` method when you provide `macro.get(publicAPI, model, ['fieldName'])`.

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
```diff
@@ -0,0 +1,243 @@
...
+
+  model.pipelines.axes.push(axis1);
+  model.pipelines.axes.push(axis2);
+
+  // Improve actors rendering
+  model.pipelines.center.actor.getProperty().setAmbient(1, 1, 1);
+  model.pipelines.center.actor.getProperty().setDiffuse(0, 0, 0);
+
+  for (let i = 0; i < model.pipelines.axes.length; i++) {
+    Object.values(model.pipelines.axes[i]).forEach((axis) => {
```
finetjul, line 80:
```
for (let i = 0; i < model.pipelines.axes.length; i++) {
    Object.values(model.pipelines.axes[i]).
```
->
``` 
model.pipelines.axes.forEach((axis) => {
  Object.values(axis).forEach((lineOrRotationHandle) => {
...
}
```

jourdain: If you want speed, the first one will be faster.

laurennlam: So maybe we can leave it as it is now ?

finetjul: Yes, you can leave it as is now, it was just "Bonus".

I find `forEach()` more readable and less error prone than a `for` loop.

@jourdain, performance not being an issue here, is there a best practice in VTK-JS?

jourdain: I tend also to prefer `.map()`, `.forEach()` but when things are in tight loop, the `for(let i = 0; i < array.length; i++)` are faster especially if the lambda function don't already exist.

I would enforce readability unless we need speed. More of the rendering code should look very much like C++. For the rest, it could be more like JS.

I'm on the same boat as you @finetjul. I just wanted to share that small detail since you had "perf" in one of your MR title at some point.

laurennlam:
Does it look good to you ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -0,0 +1,194 @@
+import vtkBoundingBox from 'vtk.js/Sources/Common/DataModel/BoundingBox';
+import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
+
+import { ViewTypes } from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
+
+// Project point (inPoint) to the bounds of the image according to a plane
+// defined by two vectors (v1, v2)
+export function boundPoint(inPoint, v1, v2, image) {
```
finetjul, line 8: take bounds instead of image to be more generic

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: This method is used to recompute the right three points that define the plane. We only need the normal and one point on the plane.

finetjul: For the record, I think it might lead to bugs later on, but it is fine for now

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
finetjul: For the record, I think it might lead to problems later on, but we can leave it as is for time being

finetjul:
LGTM. Please rebase because I can't merge it as is.

PR #1450, opened 2020-05-15
feat(RectangleWidget): add option to keep the representation plane ort..

boucaud:
…hogonal to the manipulator's normal

The widget can have two points with different heights (usually when used with the imagemapper).
This can cause parts of the representation to be hidden when one of the two points is below the current slice. 
This change adds an option to use the highest of the two points to set the representation plane's height, to make sure that the representation stays on top of the ImageMapper.

### DIFF
--- a/Sources/Widgets/Core/StateBuilder/index.js
+++ b/Sources/Widgets/Core/StateBuilder/index.js
@@ -7,6 +7,7 @@ import color from 'vtk.js/Sources/Widgets/Core/StateBuilder/colorMixin';
 import direction from 'vtk.js/Sources/Widgets/Core/StateBuilder/directionMixin';
 import manipulator from 'vtk.js/Sources/Widgets/Core/StateBuilder/manipulatorMixin';
 import name from 'vtk.js/Sources/Widgets/Core/StateBuilder/nameMixin';
+import normal from 'vtk.js/Sources/Widgets/Core/StateBuilder/normalMixin';
 import orientation from 'vtk.js/Sources/Widgets/Core/StateBuilder/orientationMixin';
 import origin from 'vtk.js/Sources/Widgets/Core/StateBuilder/originMixin';
 import scale1 from 'vtk.js/Sources/Widgets/Core/StateBuilder/scale1Mixin';
@@ -25,6 +26,7 @@ const MIXINS = {
   direction,
   manipulator,
   name,
+  normal,
   orientation,
   origin,
   scale1,
--- a/Sources/Widgets/Representations/RectangleContextRepresentation/index.js
+++ b/Sources/Widgets/Representations/RectangleContextRepresentation/index.js
@@ -65,6 +65,18 @@ function vtkRectangleContextRepresentation(publicAPI, model) {
 
       const points = new Float32Array(4 * 3);
 
+      const slicingMode = state.getDirection().indexOf(1);
+
+      if (model.stayOnTop) {
+        // Use the highest (depending on the normal) component for plane height
+        // I, J, or K, so we only need the normal for direction (sign).
+        const min = Math.min(point1[slicingMode % 3], point2[slicingMode % 3]);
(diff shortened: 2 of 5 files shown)


boucaud:
@finetjul

Sources/Widgets/Core/StateBuilder/index.js
```diff
@@ -25,6 +26,7 @@ const MIXINS = {
   direction,
   manipulator,
   name,
+  normal,
```
finetjul, line 29: isn't it direction/orientation in charge of being the "normal" ? @agirault @floryst

Sources/Widgets/Representations/RectangleContextRepresentation/index.js
```diff
@@ -131,6 +141,7 @@ function vtkRectangleContextRepresentation(publicAPI, model) {
 const DEFAULT_VALUES = {
   drawBorder: false,
   drawFace: true,
+  stayOnTop: true,
```
finetjul, line 144: add documentation

finetjul:
LGTM

@agirault  WDYT ?

Sources/Widgets/Core/StateBuilder/index.js
boucaud: The problem I had with `direction` was that it was always positive, regardless of the normal's orientation.

finetjul: I think it is a mistake then. To me, a `direction` is:  `orientation` + `sign`. 

What do you think @floryst  ?

boucaud: It seems here that `direction` is currently expected to be two 0s and a 1. See line 68:

`const slicingMode = state.getDirection().indexOf(1);`

That does seem strange.

floryst: `direction` is typically interpreted as `normal`, especially when defining planes. `RectangleContextRepresentation` shouldn't be assuming that the direction is positive and axis-aligned. However, I see that the representation is assuming IJK oriented planes, so that'll explain the assumption that the direction is axis-aligned.

floryst: The widget can use `direction` however they so please, so some widgets might assume constraints on `direction`. Though it would be nice to have widgets that aren't constrained to be axis-aligned when they don't need to be.

finetjul: Thanks @floryst 

So I guess the easiest to do for now is keep `direction` instead of `normal`.  Just convert `state.getDirection().indexOf(1)` to `state.getDirection().indexOf(state.getDirection().reduce((max, comp) Math.max(max, Math.abs(comp)));`  or something like that...

agirault:
This affects changes that are made in #1422 which - even if not yet complete for the ellipse widget - supports arbitrary orientation and seems to work well for the rectangle widget. I'd rather we complete that other PR, then look if there is something missing that @boucaud's PR addresses. Happy to discuss our options through a call if needed.

Sources/Widgets/Core/StateBuilder/index.js
agirault: Some of it was already mentioned, but:
- `direction` should be the same as `normal`
- it isn't right now because it's limited to some slicing modes
- `orientation` is not just the `direction/normal`, but a full rotation (so also `up` and `right` vectors). We can use `orientation` instead of `direction` to properly define arbitrary orientations. That's required to keep the edges of the shapes parallel with the edges of the camera view when the camera plane isn't XY, YZ, or XZ.

#1422 should handle all of that fine for the rectangle widget.

finetjul:
Does #1717 make this PR obsolete ?

boucaud:
> Does #1717 make this PR obsolete ?

It seems that is indeed the case.

PR #1458, opened 2020-05-19
feat(stlwriter): sTLWriter: writes a PolyData object to an STL file (…

### LINKED ISSUE #1442: STL Writer
Hello everyone,

I would like to be able to generate STL files from a *vtkPolyData* object in vtk.js. I see that for now, it is only possible to write this object to the *.vtp* format. My questions are thus:

- Is this a feature that you are planning to implement soon?
- To implement it myself, should I just transcribe the *vtkSTLWriter* from VTK to javascript? Are there some particular considerations/ changes between VTK and vtk.js to take into account?
- Any hints/tips on how to implement this are welcome!

Thanks!

thomas-beznik:
…in binary or ascii format)

The STLWriter writes a vtkPolyData object to a dataview for the binary format and to a string in the
ascii format. The example is not yet working.

fix #1442


Sources/IO/Geometry/STLWriter/index.js
```diff
@@ -0,0 +1,216 @@
...
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  // Make this a VTK object
+  macro.obj(publicAPI, model);
+
+  // Also make it an algorithm with one input and one output
+  macro.algo(publicAPI, model, 1, 1);
```
finetjul, line 202: it is an algorithm with 0 output .

See ZipMultiDataSetWriter and XMLWriter

Sources/IO/Geometry/STLWriter/example/index.js
```diff
@@ -0,0 +1,68 @@
...
+  // const fileContents = writer.write(reader.getOutputData());
+
+  global.readerOutput = reader.getOutputData();
+  writer.setInputData(reader.getOutputData());
+  const fileContents = writer.getOutputData();
+  
+  // Try to read it back.
+  /*const textEncoder = new TextEncoder();
+  
+  writerReader.parseAsArrayBuffer(textEncoder.encode(fileContents.buffer.prototype.buffer));
```
finetjul, line 35: did you have issue using STLReader ?

Sources/IO/Geometry/STLWriter/index.js
```diff
@@ -0,0 +1,216 @@
+// Author: Thomas Beznik, with the help of Julien Finet (https://github.com/Kitware/vtk-js/issues/1442)
```
finetjul, line 1: @jourdain do we have a more standardized way to cite authors in VTK.js ?

Sources/IO/Geometry/STLWriter/index.js
```diff
@@ -0,0 +1,216 @@
...
+function vtkSTLWriter(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkSTLWriter');
+
+  function writeBinary(polyData, transform = null) {
+    const polys = polyData.getPolys().getData();
+    const points = polyData.getPoints().getData();
+    const strips = polyData.getStrips() ? polyData.getStrips().getData() : null;
+    const buffer = new ArrayBuffer(80 + 4 + (50 * polys.length) / 4); // buffer for the full file; size = header (80) + num cells (4) +  50 bytes per poly
+    const dataview = new DataView(buffer);
```
finetjul, line 39: dataView

Sources/IO/Geometry/STLWriter/index.js
```diff
@@ -0,0 +1,216 @@
...
+
+// ----------------------------------------------------------------------------
+// vtkSTLWriter methods
+// ----------------------------------------------------------------------------
+
+function vtkSTLWriter(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkSTLWriter');
+
+  function writeBinary(polyData, transform = null) {
```
finetjul, line 34: Structure of ascii and binary files is similar .

How about factorizing it with something like the following ?

```
const binaryWriter = () => {
  let offset = 0;
  let dataView = null;
  return {
    init: (polyData) => {
      const buffer = new ArrayBuffer(80 + 4 + (50 * polys.length) / 4);
      dataview = new DataView(buffer);
    },
    writeHeader: (polyData) => {},
    writeTriangle: (v1, v2, v3, dn) => {},
    writeFooter: (polyData) => {},
};
...
writeSTL(polyData, binaryWriter(), transform);
...
function writeSTL(polyData, writer, transform = null) {
  ...
  writer.init(polyData);
  writer.header(polyData);
  for (...) { writer.writeTriangle(...);}
  ...
}
```
Maybe @jourdain has a more elegant way of doing it...

Sources/IO/Geometry/STLWriter/index.js
jourdain: I thought that was elegant to provide the data as output. Especially for the web context.

We could expose some helper that provide the output via a different method name. But I like that approach.

Sources/IO/Geometry/STLWriter/index.js
jourdain: https://jsdoc.app/tags-author.html

Sources/IO/Geometry/STLWriter/index.js
finetjul: Gotcha, so the other writers should probably have the same change (does not have to be done in this PR)

Sources/IO/Geometry/STLWriter/index.js
jourdain: That sounds reasonable and seems that they could be used as static methods. If so, they should be exposed as such and used inside the writer class.

That way used could rely on either a writer instance of function to convert a polydata to stl.

```
const stlContent = vtkSTLWriter.convert(polydata, vtkSTLWriter.binaryEncoder);

// or

const writer = vtkSTLWriter.newInstance({ encoding: 'binary' });
writer.setInputdata(polydata);
const stlContent = writer.getOuptuData();
```

Sources/IO/Geometry/STLWriter/index.js
jourdain: Yes, I would think so to have some consistency. 

Then we may want to have helper methods to get the output like:

```
getOutputAsBlob();

// Only for binary format
getOutputAsArrayBuffer() {
   setEncodingAsBinary();
   return getOutputData();
}

// Only for/when ascii format
getOuptutAsString() { 
   setEncodingAsString();
   return getOutputData();
}
```

But like you said, that could be done via a refactoring pass on all the writer we have. 
No need to do it in that PR.

finetjul:
hi @thomas-beznik do you have by any chance the time to look at our minor reviews ?

thomas-beznik:
Hello @finetjul  and @jourdain !

Very sorry for the long delay! The end of the year brought many deadlines (and vacation), and I completely lost track of this PR... I have taken into account your comments and corrected the code. I have also corrected the example! 

There is just one thing left: I wasn't able to make a static method as suggested by @jourdain. I followed this [tutorial](https://kitware.github.io/vtk-js/docs/develop_class.html) but it doesn't seem to work (it doesn't find the `writeSTL` method). Any help there would be welcome!

Sorry again for the delay!
Best,
Thomas Beznik

finetjul:
LGTM

Great work @thomas-beznik  !

Thanks !

Sources/IO/Geometry/STLWriter/index.js
```diff
@@ -0,0 +1,213 @@
...
+  vtkSTLWriter(publicAPI, model);
+}
+
+// ----------------------------------------------------------------------------
+
+export const newInstance = macro.newInstance(extend, 'vtkSTLWriter');
+
+// ----------------------------------------------------------------------------
+
+export default { newInstance, extend };
```
jourdain, line 213: `export default { newInstance, extend, ...STATIC };`

jourdain:
I let you merge it @finetjul. Even without the STATIC export it looks great.

Sources/IO/Geometry/STLWriter/index.js
thomas-beznik: I am fixing this right now!

PR #1466, opened 2020-05-22
fix(VolumeRepresentationProxy): Adjust on a log scale

thewtex:
Gradient magnitudes for natural images are likely to have a power-law
distribution. To provide linear scale to a user adjusting gradient
opacity with a slider, make gradient opacity maximum and minimum steps
on a log scale.


thewtex:
CI seems not to be triggering but is passing locally.

PR #1481, opened 2020-06-08
Fix cardinal spline

### LINKED ISSUE #1479: Invalid Cardinal Spline
Tested by replacing the `vtkSpline3D` kind from `KOCHANEK_SPLINE` to `CARDINAL_SPLINE`:

Screen Shot 2020-06-08 at 3 22 11 PM

agirault:
Fixes #1479

Shown in the `SplineWidget` example:
Screen Shot 2020-06-08 at 5 09 58 PM
Screen Shot 2020-06-08 at 5 10 05 PM

### DIFF
--- a/Sources/Common/DataModel/Spline3D/index.js
+++ b/Sources/Common/DataModel/Spline3D/index.js
@@ -37,7 +37,7 @@ function vtkSpline3D(publicAPI, model) {
       if (model.intervals.length === 0) {
         intervals = new Float32Array(size);
         for (let i = 0; i < intervals.length; i++) {
-          intervals[i] = i / (size - 1);
+          intervals[i] = i;
         }
       } else {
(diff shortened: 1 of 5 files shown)


agirault:
I pushed a new commit that exposes more options to configure the Kochanek spline... but maybe only tension is really interesting for a widget. I could remove bias and continuity you all think it's not interesting for the widget.

I could also remove the spline kind, and just have a Kochanek spline with the tension exposed (since Kochanek with the tension of -1 is the same as the cardinal).

agirault:
I could change to "spline type" but that would be a breaking change. I'm personally fine with it. "What kind of spline?" sounds good to me.

agirault:
Rebased on master. I've got three approvals so I'll merge once the checks complete.

PR #1500, opened 2020-06-18
fix(RenderWindowInteractor): Reset lastFrameStart with tab switch

thewtex:
When tabs are not active browser will step calling
requestAnimationFrame callbacks. See:

  https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame

This causes the RenderWindowInteractor lastFrameStart to be inaccurate,
which results in an incorrect lastFrameTime. This can cause the
VolumeMapper, which adjusts its the sampling distance based on the
framerate, to assume too low of a frame rate and produce poor quality
renderings.

Use the Page Visibility API:

  https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

to detected when tabs are switched, and reset the RenderWindowInteractor
lastFrameStart.

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -932,14 +932,35 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     }
   };
 
+  publicAPI.handleVisibilityChange = () => {
+    model.lastFrameStart = Date.now();
+  };
+
   // Stop animating if the renderWindowInteractor is deleted.
   const superDelete = publicAPI.delete;
   publicAPI.delete = () => {
     while (animationRequesters.size) {
       publicAPI.cancelAnimation(animationRequesters.values().next().value);
     }
+    if (typeof document.hidden !== 'undefined') {
+      document.removeEventListener(
+        'visibilitychange',
+        publicAPI.handleVisibilityChange
+      );
+    }
     superDelete();
   };
+
+  // Use the Page Visibility API to detect when we switch away from or back to
+  // this tab, and reset the lastFrameStart. When tabs are not active, browsers
+  // will stop calling requestAnimationFrame callbacks.
+  if (typeof document.hidden !== 'undefined') {
+    document.addEventListener(
+      'visibilitychange',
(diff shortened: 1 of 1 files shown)


finetjul:
The fix looks good to me, thanks.

I had that problem too. To fix it I investigated why I still had an animation when the tab was not active. Are you sure it makes sense to have an animation at that time ?

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -940,6 +940,19 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     }
     superDelete();
   };
+
+  // Use the Page Visiblity API to detect when we switch away from or back to
```
agirault, line 944: `Visibility` typo

thewtex:
> The fix looks good to me, thanks.

@finetjul :+1: thanks for the review.

> I had that problem too. To fix it I investigated why I still had an animation when the tab was not active. Are you sure it makes sense to have an animation at that time ?

I have an animation that just continuously rotates the camera for the user. They can switch tabs at any tabs, when the browser will just say:

image

finetjul:
Haha, fair enough

Sources/Rendering/Core/RenderWindowInteractor/index.js
thewtex: :+1:  fixed

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -940,6 +940,19 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     }
     superDelete();
   };
+
+  // Use the Page Visibility API to detect when we switch away from or back to
+  // this tab, and reset the lastFrameStart. When tabs are not active, browsers
+  // will stop calling requestAnimationFrame callbacks.
+  if (typeof document.hidden !== 'undefined') {
+    document.addEventListener(
```
floryst, line 948: Probably a good idea to remove the event listener in `delete()`.

thewtex: :+1: good idea

PR #1513, opened 2020-06-24
fix(renderwindowinteractor): return first interactive renderer viewpo…

### LINKED ISSUE #1487: Overlay ViewPort Interaction
Hello, I have a scene with 2 viewports:
ezgif com-video-to-gif (7)

However I'd like to be able to interact only with the upper left viewport when the mouse is on it.
Is it possible?

If needed, the link for the synchronizable scene is here :
https://drive.google.com/file/d/1SgWOJBR6qYVlU6AujDxPE3ZyVMIYqicF/view?usp=sharing

xavArtley:
…rt found as pokedRenderer

Fix the problem when multiple viewports overlay one over the other. The actual behavior iterate over
all renderer and return the last one instead the first one

fix #1487

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -668,14 +668,8 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     if (!model.view) {
       return null;
     }
-    const rc = model.view
-      .getRenderable()
-      .getRenderersByReference()
-      .sort((a, b) => {
-        if (a.getLayer() < b.getLayer()) return -1;
-        if (a.getLayer() > b.getLayer()) return 1;
-        return 0;
-      });
+    const rc = model.view.getRenderable().getRenderersByReference();
+    rc.sort((a, b) => a.getLayer() - b.getLayer());
     let interactiveren = null;
     let viewportren = null;
     let currentRenderer = null;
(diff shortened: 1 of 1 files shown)


jourdain:
We need to keep the renderer that is on the top [layer](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/Renderer/index.js#L654). 
You can not assume anything in term of ordering. So you need to check that layer to only keep the top one.

jourdain:
Or sort it.

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -668,7 +668,14 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     if (!model.view) {
       return null;
     }
-    const rc = model.view.getRenderable().getRenderersByReference();
+    const rc = model.view
+      .getRenderable()
+      .getRenderersByReference()
+      .sort((a, b) => {
```
jourdain, line 674:
```
const rc = model.view.getRenderable().getRenderersByReference();
rc.sort((a, b) => a.getLayer() - b.getLayer());
```

@martinken is it ok to sort that array in place?

jourdain:
LGTM, thanks @xavArtley !
If it is working for you, I'll merge it.

xavArtley:
> 
> 
> LGTM, thanks @xavArtley !
> If it is working for you, I'll merge it.

it's ok for me thx.

PR #1515, opened 2020-06-29
fix(LabelRepresentation): Handle devicePixelRatio

### LINKED ISSUE #1509: labelWidget is not working!
labelWidget is not working properly at example.

floryst:
Device pixel ratio is not handled correctly when going from display space to CSS space and vice-versa.

Fixes #1509


finetjul:
LGTM, thanks !

PR #1520, opened 2020-07-02
Piecewise gaussian tweaks

thewtex:
After:

localhost_8080

Before:

localhost_8080 (1)


floryst:
LGTM. The 10->20 change looks like increasing the interaction region size. Maybe add a comment about that for future info?

thewtex:
Yes, I think *20* is a better magic number than *10*, it could arguably still go higher.

> LGTM. The 10->20 change looks like increasing the interaction region size. Maybe add a comment about that for future info?

Yes, I'll add a comment.

PR #1530, opened 2020-07-12
Add widget pixel scaling

floryst:
This PR adds the ability for widget representations to be pixel-scaled.

### DIFF
--- a/Documentation/content/docs/concepts_widgets.md
+++ b/Documentation/content/docs/concepts_widgets.md
@@ -109,18 +109,27 @@ rendered.
 
 ### Scaling representations to be fixed in display space
 
-In order to scale a representation such that it retains the same size in display space,
-it must use the `scaleByDisplay` flag and the `getDisplayWorldHeightAt(coord)` method.
-The flag is used to allow users of the representation to choose between display-scaled
-vs world-scaled sizing.
+In order to scale a representation such that it retains the same size in
+display space, qt must use the `scaleByDisplay` and `scaleInPixels` flag and
+the`getDisplayScaleAtCoord(coord)` method.  The flag is used to allow users of
+the representation to choose between display-scaled vs world-scaled sizing.
 
-`getDisplayWorldHeightAt(coord)` will return the world height at a particular point that
-corresponds to the viewport height. In other words, it is the height of the rectangle formed
-by the intersection of the plane (defined by the camera normal and coord) and the view
-frustrum. A typical usage of this method is to compute some fraction of this height and
-set that as the representation scale.
+`getDisplayScaleAtCoord(coord)` will return some scaling to be applied to a
+widget representation. What it returns depends on what flags are set. If
+`scaleByDisplay=true` and `scaleInPixels=false`, the method will return the
+viewport height in world coordinates at a particular point. In other words, it
+is the height of the rectangle formed by the intersection of the plane (defined
+by the camera normal and the provided coord) and the view frustrum. In this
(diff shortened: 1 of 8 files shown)


Documentation/content/docs/concepts_widgets.md
```diff
@@ -109,18 +109,27 @@ rendered.
...
-
-`getDisplayWorldHeightAt(coord)` will return the world height at a particular point that
-corresponds to the viewport height. In other words, it is the height of the rectangle formed
-by the intersection of the plane (defined by the camera normal and coord) and the view
-frustrum. A typical usage of this method is to compute some fraction of this height and
-set that as the representation scale.
-
-Look at the `SphereHandleRepresentation` as an example for how `getDisplayWorldHeightAt`
+In order to scale a representation such that it retains the same size in
+display space, qt must use the `scaleByDisplay` and `scaleInPixels` flag and
```
thewtex, line 113: Is this a qt reference? If this is just a reference, it would help if it is noted as such, or just remove it altogether for clarity.

Documentation/content/docs/develop_widget.md
```diff
@@ -145,6 +145,11 @@ This mixin adds an `origin` property.
 
 These mixins adds a single scale factor and a 3-component scale factor, respectively.
 
+If `scaleByDisplay` is set to true for a representation, then scale1 will be interpreted as
+a fraction of the display height. (Only for representations that support this; grep for
+"scaleByDisplay" to see which representations do.) If `scaleByDisplay` and `scaleInPixels`
+are true, then scale1 will be treated as a pixel height value.
```
thewtex, line 151: If I am following correctly, the scale is

1. In world coordinates (default)
2. In display coordinates (scaleByDisplay)
3. In pixel coordinates (scaleInPixels)

?

Would a set of constants / vtk.js enumerations work?

Documentation/content/docs/concepts_widgets.md
floryst: Oh that's a typo. Not sure how "qt" got in there haha.

Documentation/content/docs/develop_widget.md
floryst: The API is something worth discussing, and I like your bullet points since they provide clarity, so I'll add it in the docs.

The scale is as such:
1. Default: scale only describes world coordinate sizing
2. scaleByDisplay: the name is confusing. The scale is interpreted to be a fraction of the display height, so if the canvas is 200px and the scale is 0.05, then a sphere handle will take up 10px in height.
3. scaleInPixels: The scale is interpreted as the height in pixels.

Note this all depends on the height rather than the width, since the camera view angle (which is used for the scale calculations) determines height of the viewport.

I like the constants idea. I avoided it initially since it is an API breaking change, but after re-thinking it I'm going to go ahead with that.

thewtex: Cool, thanks @floryst

floryst: In addition, I'm thinking of removing the 2nd functionality. I would suppose that most people will just beeline to setting pixel height/size, since that's what everyone does for sizing controls anyways. It arguably is odd if widget controls grew in size if the viewport were to be resized along the height dimension.

floryst:
Update: I've re-written the API to only support pixel sizing. I couldn't come up with a good reason to keep the "scale as a fraction of the display height" functionality, so I replaced it.

finetjul:
LGTM

PR #1547, opened 2020-07-29
doc(WidgetManager): Fix invalid example html

floryst:
This was causing builds to fail. Here's a quick post-mortem:
- The issue only manifested when kw-doc was invoked with minify option, so terser was complaining about something.
- Parallel-webpack wasn't logging any webpack errors, hence the build error of "file not found" remained cryptic.
- terser-plugin minifies html, and this html was invalid.


floryst:
@jourdain I'm running a docs build just to make sure nothing else is complaining.

floryst:
Everything builds on my end.

PR #1551, opened 2020-08-06
fix(OpenGL/Texture): Allow changing internal formats

floryst:
ImageMapper has a case where it's possible to switch between
UNSIGNED_CHAR and FLOAT textures, depending on if there is a piecewise
function for opacity (see [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/ImageMapper/index.js#L744)).This commit relaxes a restriction where the
internal format can only be set once.

I'll add a test for this.

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -310,19 +310,8 @@ function vtkOpenGLTexture(publicAPI, model) {
 
   //----------------------------------------------------------------------------
   publicAPI.getInternalFormat = (vtktype, numComps) => {
-    if (!model.internalFormat) {
-      model.internalFormat = publicAPI.getDefaultInternalFormat(
-        vtktype,
-        numComps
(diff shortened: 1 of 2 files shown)


floryst:
This is the first step to supporting toggle-able color slices in Glance (https://github.com/Kitware/paraview-glance/issues/351).

martinken:
```
/**
   * Reset format, internal format, and type of the texture.

   * This method is useful when a texture is reused in a
   * context same as the previous render call. In such
   * cases, texture destruction does not happen and therefore
   * previous set values are used.
   */
  void ResetFormatAndType();
```

martinken:
You need to use a reset method (or something else) because some mappers Set the internal format as a method of overriding the default format calc. This change would make that impossible.

floryst:
Oh thanks Ken, that's good to know. I'll make a change to vtkOpenGLImageMapper that resets the format and type as needed.

PR #1552, opened 2020-08-07
fix(OpenGL/ImageMapper): Reset texture format for pwf

floryst:
Before, adding a piecewise function to a texture after the texture has been rendered would not do anything. Now, piecewise functions added/removed during the lifetime of the texture will be used appropriately.


martinken:
LGTM +1

PR #1558, opened 2020-08-17
fix(FullScreenRenderWindow): Use 100% height/width instead of 100vw/vh

floryst:
Instead of filling the viewport, fill the parent element. This fixes
issues with how Chrome on Android treats 100vh as including the address
bar height, resulting in the lower ~56px being cut off. This should have
minor (but positive) impacts on desktop browsers, as vw/vh may include
scrollbar width/height, causing the "fullscreen" aspect to actually
extend beyond the visible viewport.

### DIFF
--- a/Sources/Rendering/Misc/FullScreenRenderWindow/index.js
+++ b/Sources/Rendering/Misc/FullScreenRenderWindow/index.js
@@ -18,8 +18,8 @@ const STYLE_CONTAINER = {
   position: 'absolute',
   top: '0',
   left: '0',
-  width: '100vw',
-  height: '100vh',
+  width: '100%',
+  height: '100%',
   overflow: 'hidden',
 };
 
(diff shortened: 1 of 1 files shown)


jourdain:
I'm not sure you will get the full height on a desktop. If the issue is mostly on Android, we may need to fix it just for Android.

floryst:
The best way to ensure that 100% gets full height is to apply 100% height to html, body, and the container. I suppose that since people expect FullScreenRenderWindow to be full screen, we can take control of html/body CSS and do something like `document.documentElement.style.height = document.body.style.height = "100%";`.

I would rather not do something specific for Android. I think this also occurs on iOS Safari (I don't have a device to test), and really it would probably manifest in any mobile browser that hides the address bar but treats the viewport height as the entire height (including the address bar).

jourdain:
Also, FullScreenRenderWindow is also used with custom root container to indeed be 100% of that container while not being full screen. I know that is a stretch especially since that class was designed for smaller and simpler examples...

jourdain:
We could update the style of body by making sure the rootContainer is actually the body. If not, we let the user deal with it.

floryst:
Makes sense. If the user uses the default body, then we can set 100% on html+body. That would necessitate changing all of the examples that use a custom root container (e.g. VolumeViewer).

Another line of thinking is to always apply 100% to html and body heights, since if you are doing fullscreen then chances are your page will be wanting to have those properties anyways, and you aren't scrolling anywhere. For now, I will only apply 100% css if rootContainer === body.

PR #1560, opened 2020-08-18
fix(ImageReslice): Do not use TypedArray.from()

### LINKED ISSUE #1557: vtkImageReslice dose not support Safari(V13.1.2)
Hello,

I used vtkImageReslice to get image pixel for display in cornerstone canvas but I found an error of function vtkImageReslice.getOutputData().

this is an error "Unhandled Promise Rejection: TypeError: this dose not subclass a TypedArray constructor"

Tools: VueJS (v2.6.11), vtk.js (V14.15.2), Safari (V13.1.2), macOS Catalina (V10.15.6)

I tried to test change vtk.js version 8.3.9 to 14.15.2 in this project https://github.com/QSolutionsLLC/cornerstone-vtk-mpr. The result is this project can display images. So I change vtk.js in my project to 8.3.9. It can get output data from vtkImageReslice but I need to use some function of the latest version.

floryst:
Some versions of Webkit may complain about invalid TypedArray
constructors when using `.from`. Use the TypedArray constructor instead.

Fixes #1557

@finetjul If there is some subtlety for using `.from` here rather than the TypedArray constructor, please let me know.

### DIFF
--- a/Sources/Imaging/Core/ImageReslice/index.js
+++ b/Sources/Imaging/Core/ImageReslice/index.js
@@ -17,7 +17,7 @@ import {
 } from 'vtk.js/Sources/Imaging/Core/AbstractImageInterpolator/InterpolationInfo';
 import SlabMode from './Constants';
 
-const { capitalize, vtkErrorMacro, vtkDebugMacro } = macro;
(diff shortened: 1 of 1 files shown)


jourdain:
Thx! The code is easier to read as well now with that change.

PR #1563, opened 2020-08-26
fix(OpenGL/CellArrayBufferObject): in some scenarios, the value of ca…

gaoxinyu18:
In some scenarios, the value of caboCount is negative, causing array allocation to fail

### DIFF
--- a/Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
+++ b/Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
@@ -224,6 +224,11 @@ function vtkOpenGLCellArrayBufferObject(publicAPI, model) {
       index += array[index] + 1;
     }
 
+    if (caboCount <= 0) {
+      model.elementCount = 0;
+      return 0;
+    }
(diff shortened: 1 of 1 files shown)


jourdain:
Do we know when `caboCount` can be negative?

martinken:
The fix needs to be in the cell build/count functions.  Because they need to be consistent with each other

For example the lines to wireframe cell builder will emit zero points when given a zero point line but the cell count function will return -2. Checking the total resulting cell count being negative would not catch the case were a number of good lines are present mixed with a zero point line. In that case the total will still be positive but the count will be lower than it should be.

So I think the the following need fixing to be correct

builder.polysToWireframe (ignore one or zero point polys)
builder.stripsToWireframe (ignore for one or zero point strips)
counter.linesToWireframe (return zero for zero point lines)
counter.polysToWireframe (return zero for one or zero point polys)
counter.stripsToWireframe (return zero for one or zero point strips)
counter.stripsToSurface (return zero for two, one or zero point strips)

Issue #1381, opened 2020-04-03
GL framebuffer incomplete


xavArtley:
Hello,
it is not really an issue but more a question,

I noticed when loading the example InteractiveOrientationWidget:
https://kitware.github.io/vtk-js/examples/InteractiveOrientationWidget.html

There are several warnings trace in the js console:
![Capture d’écran de 2020-04-03 10-03-23](https://user-images.githubusercontent.com/18531147/78338084-6b0e5d80-7592-11ea-9d69-9c41a2899b91.png)

I was wondering if someone know where it comes from and if it's really harmeless

jourdain:
I'm thinking that we might be calling stuff that should be delayed until the GLContext is properly created. That's just a wild guess as everything is fine once the app is ready and running.

@martinken do you have any idea?

martinken:
No idea, but it looks like a bug. Our framebuffer should be complete etc so something is up.

xavArtley:
Don't know if it's help for this issue but the warning appear only when the mouse cursor enter into the renderwindow
Moreover the marker widget size increase as well
![invalid_framebuffer](https://user-images.githubusercontent.com/18531147/94006327-f5948480-fd9f-11ea-88b1-deabeab40648.gif)

jourdain:
@floryst

floryst:
I can no longer reproduce this issue in vtk.js v19. If the warnings still show up, please re-open and let us know.

PR #1590, opened 2020-09-25
fix(OrientationMarkerWidget): Update viewport on modified

floryst:
Previously, the viewport was not being re-computed when min/max sizes
were set.

Fixes the second issue mentioned in #1381

### DIFF
--- a/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
+++ b/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
@@ -13,12 +13,15 @@ function vtkOrientationMarkerWidget(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkOrientationMarkerWidget');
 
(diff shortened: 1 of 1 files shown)


Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
```diff
@@ -13,12 +13,15 @@ function vtkOrientationMarkerWidget(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkOrientationMarkerWidget');
 
+  const superClass = { ...publicAPI };
```
jourdain, line 16: I don't see you using it?

jourdain (approved):

Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
floryst: I'm using it in `publicAPI.delete()`

PR #1591, opened 2020-09-25
fix(OpenGL/Framebuffer): Set framebuffer when complete

### LINKED ISSUE #1381: GL framebuffer incomplete

floryst:
Previously, the framebuffer would get set even when the framebuffer was
incomplete, resulting in "incomplete framebuffer" warnings.

Fixes #1381 (from what I've tested)

@martinken are there any cases where this would break something? Also, if the framebuffer is incomplete, would it make sense to call `setActiveFramebuffer(null)`?

### DIFF
--- a/Sources/Rendering/OpenGL/Framebuffer/index.js
+++ b/Sources/Rendering/OpenGL/Framebuffer/index.js
@@ -60,7 +60,12 @@ function vtkFramebuffer(publicAPI, model) {
     if (model.colorTexture) {
       model.colorTexture.bind();
     }
-    model.openGLRenderWindow.setActiveFramebuffer(publicAPI);
+    if (
+      model.context.checkFramebufferStatus(model.context.FRAMEBUFFER) ===
+      model.context.FRAMEBUFFER_COMPLETE
+    ) {
+      model.openGLRenderWindow.setActiveFramebuffer(publicAPI);
+    }
   };
 
   publicAPI.create = (width, height) => {


martinken:
Attempting to bind an incomplete framebuffer is an error. We should never be doing that. My gut is that this may hide the problem but still something is messed up if we are trying to bind an incomplete framebuffer

floryst:
I've noticed that, in [`vtkOpenGLFramebuffer.bind()`](https://github.com/Kitware/vtk-js/blob/8f7a46374543eb30136cf2b8068ab4b6ab76c9bd/Sources/Rendering/OpenGL/Framebuffer/index.js#L55), if the colorTexture is null, then the framebuffer remains incomplete. Running the `InteractiveOrientationWidget` and moving the mouse over the scene verifies this, as I see `model.colorTexture` being false and the state of the framebuffer is still incomplete.

The backtrace is as follows:
1. [`vtkOpenGLFramebuffer.bind()`](https://github.com/Kitware/vtk-js/blob/8f7a46374543eb30136cf2b8068ab4b6ab76c9bd/Sources/Rendering/OpenGL/Framebuffer/index.js#L55)
2. [`vtkOpenGLFramebuffer.populateFramebuffer()`](https://github.com/Kitware/vtk-js/blob/8f7a46374543eb30136cf2b8068ab4b6ab76c9bd/Sources/Rendering/OpenGL/Framebuffer/index.js#L187)
3. [`vtkOpenGLHardwareSelector.beginSelection()`](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/HardwareSelector/index.js#L40)

PR #1595, opened 2020-10-02
fix(Math): Cross results incorrect when doing in-place

### LINKED ISSUE #1594: ImageCroppingWidget not working properly
Hello I was trying the https://kitware.github.io/vtk-js/examples/ImageCroppingWidget/index.html example
However it appears not to work

cropwidget

floryst:
If x or y is the same array as out, then the result will be wrong due to
inter-dependence between components.

Fixes #1594

### DIFF
--- a/Sources/Common/Core/Math/api.md
+++ b/Sources/Common/Core/Math/api.md
@@ -117,6 +117,7 @@ Returns out.
 
 Computes cross product of 3D vectors x and y.
 Returns out.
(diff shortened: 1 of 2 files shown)


finetjul:
LGTM, maybe you can also update api.md to say it if safe to use input `x` or `y` as `out`

floryst:
Sounds good. I'll update the API and merge.

PR #1599, opened 2020-10-09
Tube filter copy celldata

floryst:
- Copy cell data from tube filter
- initialize attribute copy flags on DataSetAttributes creation
- Remove a call to `initialize` in `copy<NAME>Off`, since `initialize` clears `model.arrays`, which is not desired behavior when turning copy off

### DIFF
--- a/Sources/Filters/General/TubeFilter/index.js
+++ b/Sources/Filters/General/TubeFilter/index.js
@@ -707,6 +707,24 @@ function vtkTubeFilter(publicAPI, model) {
       output.getPointData().addArray(newArray); // concat newArray to end
     }
 
+    // loop over cellData arrays and resize based on numNewCells
+    let numNewCells = inLines.getNumberOfCells() * model.numberOfSides;
+    if (model.capping) {
+      numNewCells += 2;
+    }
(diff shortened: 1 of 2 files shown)


Sources/Filters/General/TubeFilter/index.js
```diff
@@ -707,6 +707,21 @@ function vtkTubeFilter(publicAPI, model) {
       output.getPointData().addArray(newArray); // concat newArray to end
     }
 
+    // loop over cellData arrays and resize based on numNewCells
+    const numNewCells = inLines.getNumberOfCells() * model.numberOfSides + 2;
```
jourdain, line 711: Why +2 ?

jourdain:
Just glancing through. At high level that seems reasonable.

jourdain (approved):
LGTM at a high level

Sources/Filters/General/TubeFilter/index.js
floryst: The +2 is to account for capping. Though I can probably save that extra 2 cell slots if I check to see if capping is enabled.

PR #1601, opened 2020-10-11
Generate super-generic typescript typings

matthova:
This PR is a first-shot at generating generic typescript typings for all files in this package.  
The types were generated using [js2dts](https://github.com/whxaxes/js2dts) for all `.js` files in `Sources` and `Utilities` that didn't already have a matching `.d.ts` file.  
My hope is that a lot of the more granular type-generation can be automated, but in any case, this could be a step towards having some nice typings for the library.  

All of the (non-automated) work can be found in `package.json` and `generateDTSFiles.js`


floryst:
How does `js2dts` differ from using `tsc` to auto-generate d.ts files? I'm asking because I've never used js2dts.

Also, here is an earlier effort that needs to be updated: https://github.com/Kitware/vtk-js/pull/1440. Now that you've brought up typescript support again, I'll make sure to spend time this week to look at it.

matthova:
Whoops, I had missed the other PR. It looks like the built-in type generation in `tsc` should work just fine.
For the sake of conversation, here's a sameple of `js2dts` vs `tsc`

js2dts:
```.d.ts
declare function extractCellSizes(cellArray: any): any;
declare function getNumberOfCells(cellArray: any): any;
export interface T100 {
  extractCellSizes: typeof extractCellSizes;
  getNumberOfCells: typeof getNumberOfCells;
}
export const STATIC: T100;
export interface T101 {
  [key: string]: any;
}
declare function extend_1(publicAPI: any, model: any, initialValues?: T101): void;
export const extend: typeof extend_1;
export const newInstance: any;
export interface T102 {
  extractCellSizes: typeof extractCellSizes;
  getNumberOfCells: typeof getNumberOfCells;
  newInstance: any;
  extend: typeof extend_1;
}
declare const T103: T102;
export default T103;
```

tsc:
```.d.ts
export function extend(publicAPI: any, model: any, initialValues?: {}): void;
export namespace STATIC {
    export { extractCellSizes };
    export { getNumberOfCells };
}
export const newInstance: any;
declare var _default: any;
export default _default;
declare function extractCellSizes(cellArray: any): any;
declare function getNumberOfCells(cellArray: any): any; 
```

It looks like `js2dts` is generating a bit more boilerplate types but with little extra value.
If these auto-generated files are placeholders for future work where the files are properly typed, I think using `tsc` is the way to go 👍

PR #1606, opened 2020-10-19
improvement(reslicecursorwidget): add mouse scrolling to move slice

laurennlam:
@finetjul

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
@@ -1,6 +1,7 @@
 import { mat4 } from 'gl-matrix';
 
 import macro from 'vtk.js/Sources/macro';
+import vtkBoundingBox from 'vtk.js/Sources/Common/DataModel/BoundingBox';
 import vtkLine from 'vtk.js/Sources/Common/DataModel/Line';
 import vtkPlaneManipulator from 'vtk.js/Sources/Widgets/Manipulators/PlaneManipulator';
 import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
@@ -66,6 +67,67 @@ export default function widgetBehavior(publicAPI, model) {
     }
     isDragging = false;
     model.widgetState.deactivate();
+  };
+
+  publicAPI.handleStartMouseWheel = (callData) => {
+    publicAPI.invokeStartInteractionEvent();
+    // When interacting, plane actor and lines must be re-rendered on other views
+    publicAPI.getViewWidgets().forEach((viewWidget) => {
+      viewWidget.getInteractor().requestAnimation(publicAPI);
+    });
+  };
+
+  publicAPI.handleMouseWheel = (calldata) => {
+    const direction = calldata.spinY;
+    const currentCenter = model.widgetState.getCenter();
+    const image = model.widgetState.getImage();
+    const imageSpacing = image.getSpacing();
+    const imageBounds = image.getBounds();
+    const boundingBox = vtkBoundingBox.newInstance({ bounds: imageBounds });
+
+    // Get the normal use to translate the center along
+    const normal = calldata.pokedRenderer
+      .getRenderWindow()
+      .getRenderers()[0]
+      .getActiveCamera()
+      .getDirectionOfProjection();
+
+    // Direction of the projection if the inverse of what we want
+    const translationVector = vtkMath.multiplyScalar(normal, -1);
+
+    let newCenter = [
+      currentCenter[0] + direction * translationVector[0] * imageSpacing[0],
+      currentCenter[1] + direction * translationVector[1] * imageSpacing[1],
+      currentCenter[2] + direction * translationVector[2] * imageSpacing[2],
+    ];
+
+    if (!boundingBox.containsPoint(...newCenter)) {
+      // Clamp the center to the bounds of the image
+      const imageOrigin = image.getOrigin();
+      const localRoundedCenter = currentCenter.map(
+        (value, index) =>
+          Math.round(value - imageOrigin[index]) / imageSpacing[index]
+      );
+
+      // Get it back in world coordinates
+      newCenter = localRoundedCenter.map(
+        (value, index) => value * imageSpacing[index] + imageOrigin[index]
+      );
+    }
+
+    model.widgetState.setCenter(newCenter);
+    updateState(model.widgetState);
+
+    publicAPI.invokeInteractionEvent();
+
+    return macro.EVENT_ABORT;
+  };
+
+  publicAPI.handleEndMouseWheel = (calldata) => {
+    publicAPI.invokeEndInteractionEvent();
+    publicAPI.getViewWidgets().forEach((viewWidget) => {
+      viewWidget.getInteractor().cancelAnimation(publicAPI);
+    });
   };
 
   publicAPI.handleEvent = (callData) => {
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
@@ -12,9 +12,11 @@ import {
   updateState,
 } from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers';
 
+import { ScrollingMethods } from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/Constants';
+
 export default function widgetBehavior(publicAPI, model) {
   let isDragging = null;
-  let middlePressed = false;
+  let isScrolling = false;
 
   publicAPI.updateCursor = () => {
     switch (model.activeState.getUpdateMethodName()) {
@@ -34,16 +36,26 @@ export default function widgetBehavior(publicAPI, model) {
   };
 
   publicAPI.handleLeftButtonPress = (callData) => {
-    if (!model.activeState || !model.activeState.getActive()) {
+    if (model.activeState && model.activeState.getActive()) {
+      isDragging = true;
+      const viewName = model.widgetState.getActiveViewName();
+      const currentPlaneNormal = model.widgetState[
+        `get${viewName}PlaneNormal`
+      ]();
+      model.planeManipulator.setOrigin(model.widgetState.getCenter());
+      model.planeManipulator.setNormal(currentPlaneNormal);
+
+      publicAPI.startInteraction();
+    } else if (
+      model.widgetState.getScrollingMethod() ===
+      ScrollingMethods.LEFT_MOUSE_BUTTON
+    ) {
+      isScrolling = true;
+      model.previousPosition = callData.position;
+      publicAPI.startInteraction();
+    } else {
       return macro.VOID;
     }
-    isDragging = true;
-    const viewName = model.widgetState.getActiveViewName();
-    const currentPlaneNormal = model.widgetState[`get${viewName}PlaneNormal`]();
-    model.planeManipulator.setOrigin(model.widgetState.getCenter());
-    model.planeManipulator.setNormal(currentPlaneNormal);
-
-    publicAPI.startInteraction();
 
     return macro.EVENT_ABORT;
   };
@@ -52,8 +64,8 @@ export default function widgetBehavior(publicAPI, model) {
     if (isDragging && model.pickable) {
       return publicAPI.handleEvent(callData);
     }
-    if (middlePressed) {
-      if (model.previousPosition !== callData.position) {
+    if (isScrolling) {
+      if (model.previousPosition.y !== callData.position.y) {
         const step = model.previousPosition.y - callData.position.y;
         publicAPI.translateCenterOnCurrentDirection(
           step,
@@ -68,11 +80,33 @@ export default function widgetBehavior(publicAPI, model) {
   };
 
   publicAPI.handleLeftButtonRelease = () => {
-    if (isDragging) {
+    if (isDragging || isScrolling) {
       publicAPI.endInteraction();
     }
     isDragging = false;
+    isScrolling = false;
     model.widgetState.deactivate();
+  };
+
+  publicAPI.handleRightButtonPress = (calldata) => {
+    if (
+      model.widgetState.getScrollingMethod() ===
+      ScrollingMethods.RIGHT_MOUSE_BUTTON
+    ) {
+      model.previousPosition = calldata.position;
+      isScrolling = true;
+      publicAPI.startInteraction();
+    }
+  };
+
+  publicAPI.handleRightButtonRelease = (calldata) => {
+    if (
+      model.widgetState.getScrollingMethod() ===
+      ScrollingMethods.RIGHT_MOUSE_BUTTON
+    ) {
+      isScrolling = false;
+      publicAPI.endInteraction();
+    }
   };
 
   publicAPI.handleStartMouseWheel = (callData) => {
@@ -93,14 +127,24 @@ export default function widgetBehavior(publicAPI, model) {
   };
 
   publicAPI.handleMiddleButtonPress = (calldata) => {
-    middlePressed = true;
-    model.previousPosition = calldata.position;
-    publicAPI.startInteraction();
+    if (
+      model.widgetState.getScrollingMethod() ===
+      ScrollingMethods.MIDDLE_MOUSE_BUTTON
+    ) {
+      isScrolling = true;
+      model.previousPosition = calldata.position;
+      publicAPI.startInteraction();
+    }
   };
 
   publicAPI.handleMiddleButtonRelease = (calldata) => {
-    middlePressed = false;
-    publicAPI.endInteraction();
+    if (
+      model.widgetState.getScrollingMethod() ===
+      ScrollingMethods.MIDDLE_MOUSE_BUTTON
+    ) {
+      isScrolling = false;
+      publicAPI.endInteraction();
+    }
   };
 
   publicAPI.handleEvent = (callData) => {
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
@@ -14,6 +14,7 @@ import {
 
 export default function widgetBehavior(publicAPI, model) {
   let isDragging = null;
+  let middlePressed = false;
 
   publicAPI.updateCursor = () => {
     switch (model.activeState.getUpdateMethodName()) {
@@ -42,12 +43,7 @@ export default function widgetBehavior(publicAPI, model) {
     model.planeManipulator.setOrigin(model.widgetState.getCenter());
     model.planeManipulator.setNormal(currentPlaneNormal);
 
-    publicAPI.invokeStartInteractionEvent();
-
-    // When interacting, plane actor and lines must be re-rendered on other views
-    publicAPI.getViewWidgets().forEach((viewWidget) => {
-      viewWidget.getInteractor().requestAnimation(publicAPI);
-    });
+    publicAPI.startInteraction();
 
     return macro.EVENT_ABORT;
   };
@@ -56,54 +52,36 @@ export default function widgetBehavior(publicAPI, model) {
     if (isDragging && model.pickable) {
       return publicAPI.handleEvent(callData);
     }
+    if (middlePressed) {
+      if (model.previousPosition !== callData.position) {
+        const step = model.previousPosition.y - callData.position.y;
+        publicAPI.translateCenterOnCurrentDirection(
+          step,
+          callData.pokedRenderer
+        );
+        model.previousPosition = callData.position;
+
+        publicAPI.invokeInteractionEvent();
+      }
+    }
     return macro.VOID;
   };
 
   publicAPI.handleLeftButtonRelease = () => {
     if (isDragging) {
-      publicAPI.invokeEndInteractionEvent();
-      publicAPI.getViewWidgets().forEach((viewWidget) => {
-        viewWidget.getInteractor().cancelAnimation(publicAPI);
-      });
+      publicAPI.endInteraction();
     }
     isDragging = false;
     model.widgetState.deactivate();
   };
 
   publicAPI.handleStartMouseWheel = (callData) => {
-    publicAPI.invokeStartInteractionEvent();
-    // When interacting, plane actor and lines must be re-rendered on other views
-    publicAPI.getViewWidgets().forEach((viewWidget) => {
-      viewWidget.getInteractor().requestAnimation(publicAPI);
-    });
+    publicAPI.startInteraction();
   };
 
   publicAPI.handleMouseWheel = (calldata) => {
-    const direction = calldata.spinY;
-    const oldCenter = model.widgetState.getCenter();
-    const image = model.widgetState.getImage();
-    const imageSpacing = image.getSpacing();
-
-    // Get the normal use to translate the center along
-    const dirProj = calldata.pokedRenderer
-      .getRenderWindow()
-      .getRenderers()[0]
-      .getActiveCamera()
-      .getDirectionOfProjection();
-
-    // Direction of the projection is the inverse of what we want
-    const normal = vtkMath.multiplyScalar(dirProj, -1);
-
-    // Define the potentially new center
-    let newCenter = [
-      oldCenter[0] + direction * normal[0] * imageSpacing[0],
-      oldCenter[1] + direction * normal[1] * imageSpacing[1],
-      oldCenter[2] + direction * normal[2] * imageSpacing[2],
-    ];
-    newCenter = publicAPI.getBoundedCenter(newCenter);
-
-    model.widgetState.setCenter(newCenter);
-    updateState(model.widgetState);
+    const step = calldata.spinY;
+    publicAPI.translateCenterOnCurrentDirection(step, calldata.pokedRenderer);
 
     publicAPI.invokeInteractionEvent();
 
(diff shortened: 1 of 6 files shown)


Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -68,6 +69,42 @@ export default function widgetBehavior(publicAPI, model) {
...
+    const correctNormal = vtkMath.multiplyScalar(normal, -1);
+
+    const newCenter = [
+      currentCenter[0] + direction * correctNormal[0] * imageSpacing[0],
+      currentCenter[1] + direction * correctNormal[1] * imageSpacing[1],
+      currentCenter[2] + direction * correctNormal[2] * imageSpacing[2],
+    ];
+
+    // Don't allow to move out of the image
+    if (bounds.containsPoint(...newCenter)) {
```
finetjul, line 100: that may prevent you from going on the edge of the bounds :-/

laurennlam: if you think that it is not a requirement, I can remove it.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -68,6 +69,42 @@ export default function widgetBehavior(publicAPI, model) {
...
+
+    // Get the normal use to translate the center along
+    const normal = calldata.pokedRenderer
+      .getRenderWindow()
+      .getRenderers()[0]
+      .getActiveCamera()
+      .getDirectionOfProjection();
+
+    // Direction of the projection if the inverse of what we want
+    const correctNormal = vtkMath.multiplyScalar(normal, -1);
```
finetjul, line 91: correctNormal -> translationVector

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -148,14 +148,14 @@ for (let i = 0; i < 3; i++) {
 // Load image
 // ----------------------------------------------------------------------------
 
-function updateReslice(viewtype, reslice, actor, renderer) {
+function updateReslice(viewtype, reslice, actor, renderer, renderwindow) {
```
finetjul, line 151: you can get renderwindow from renderer, no ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -182,11 +182,19 @@ reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
         // No need to update plane nor refresh when interaction
         // is on current view. Plane can't be changed with interaction on current
         // view. Refreshs happen automatically with `animation`.
-        .filter((_, index) => index !== i)
+        // Note: Need to refresh also the current view because of adding the mouse wheel
+        // to change slicer
+        // .filter((_, index) => index !== i)
```
finetjul, line 187: this will dramatically reduce performances. 

I'm surprised the view does not refresh itself when scrolling.
Alternatively, pass a parameter to the InteractionEvent to know if you need to refresh self

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -148,14 +148,14 @@ for (let i = 0; i < 3; i++) {
...
-function updateReslice(viewtype, reslice, actor, renderer) {
+function updateReslice(viewtype, reslice, actor, renderer, renderwindow) {
   const modified = widget.updateReslicePlane(reslice, viewtype);
   if (modified) {
     // Get returned modified from setter to know if we have to render
     actor.setUserMatrix(reslice.getResliceAxes());
     widget.resetCamera(renderer, viewtype);
   }
-
+  renderwindow.render();
```
finetjul, line 158: This will dramatically decrease perfs. You used to not need it, what happened ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
finetjul: This is a requirement :-) Sorry I was not being clear.

I meant you should project new center onto the edge if it is outside (you can ask Paul for what he did on some other project)

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
laurennlam: Yes, didn't see it!

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
laurennlam: The display lines on the others views are not updated when I scroll on the third one, except if I move the mouse on it after scrolled.

laurennlam: My bad ! I forgot to add something in the behavior.js.

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
laurennlam: Thanks to the previous modifications (remove the render on renderwindow) then it make sens to keep it has it is now.
If you translate/rotate on view X, it will only update the moving views. If you scroll on view X, it will only update the view X thanks to `modified` attribute in `updateReslice` function

finetjul: If you scroll on the view X, you need to refresh the views Y and Z because the line X has changed

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -68,6 +68,43 @@ export default function widgetBehavior(publicAPI, model) {
...
+    ];
+
+    model.widgetState.setCenter(newCenter);
+    updateState(model.widgetState);
+
+    publicAPI.invokeInteractionEvent();
+
+    // When interacting, plane actor and lines must be re-rendered on other views
+    publicAPI.getViewWidgets().forEach((viewWidget) => {
+      viewWidget.getInteractor().requestAnimation(publicAPI);
```
finetjul, line 103: why don't you do that with the other events ?

finetjul: shouln't you also "cancelAnimation" also ?

laurennlam: I finally find out whats happening. When we start scrolling, we want to request animation. Each time we scroll, we animate. At the end of the scrolling, we cancel all animations.
It works better now. But, it doesn't solve the preivous issue where I have to update 3 views here:
`// Note: Need to refresh also the current view because of adding the mouse wheel
        // to change slicer
        // .filter((_, index) => index !== i)`

The scrolling directly affects the current imageReslice. So, we have to update the slice by calling `updateReslice` on it.
Doest it seem clearer now ? The `updateReslice` function update the reslice filter. But only the modified one will be "rerender". So, when I will scroll on view X, only the viewX will be rerender.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -68,6 +68,51 @@ export default function widgetBehavior(publicAPI, model) {
...
+    model.widgetState.setCenter(newCenter);
+    updateState(model.widgetState);
+
+    publicAPI.invokeInteractionEvent();
+
+    return macro.EVENT_ABORT;
+  };
+
+  publicAPI.handleEndMouseWheel = (calldata) => {
+    publicAPI.invokeEndInteractionEvent();
```
finetjul, line 110: Could't find invokeStartInteractionEvent() ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
finetjul: Do you mean that `updateReslicePlane` will return false for the view X when you do not "scroll" (but rotate line Y) ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: My bad.

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
laurennlam: The updateReslice function only modify the Reslice filters. Not the representation of the reslice cursor (spheres and lines).
The representation is updated thanks to the requestAnimation. That's what I understand of the request Animation

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: Yes! For example, if you rotate the green line on the X view, then it will only call 
`    actor.setUserMatrix(reslice.getResliceAxes());
    widget.resetCamera(renderer, viewtype);` of the Y view.
But, we are agreed that the `updateReslicePlane` method, will be called on each views, each time the widgetState is updated. No matter if you translate, rotate, scroll.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -68,6 +68,50 @@ export default function widgetBehavior(publicAPI, model) {
...
+    // Direction of the projection if the inverse of what we want
+    const translationVector = vtkMath.multiplyScalar(normal, -1);
+
+    const newCenter = [
+      currentCenter[0] + direction * translationVector[0] * imageSpacing[0],
+      currentCenter[1] + direction * translationVector[1] * imageSpacing[1],
+      currentCenter[2] + direction * translationVector[2] * imageSpacing[2],
+    ];
+
+    model.widgetState.setCenter(newCenter);
```
finetjul, line 100: isn't it going to go outside the bounds ?

laurennlam: I thaught you didn't want to check where the center is ?

finetjul: I said:  _you SHOULD PROJECT new center onto the edge if it is outside_

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: I didn't see the last comment. That's why I didn't understand your last comment..

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -68,6 +69,67 @@ export default function widgetBehavior(publicAPI, model) {
...
+    // Direction of the projection if the inverse of what we want
+    const translationVector = vtkMath.multiplyScalar(normal, -1);
+
+    let newCenter = [
+      currentCenter[0] + direction * translationVector[0] * imageSpacing[0],
+      currentCenter[1] + direction * translationVector[1] * imageSpacing[1],
+      currentCenter[2] + direction * translationVector[2] * imageSpacing[2],
+    ];
+
+    if (!boundingBox.containsPoint(...newCenter)) {
```
finetjul, line 104: does not work when translationVector is not an orthogonal axis.

image


You should use `intersectPlane`

finetjul: Move back newCenter by the shortest distance between all bbox corners.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -83,36 +85,93 @@ export default function widgetBehavior(publicAPI, model) {
...
+    // Find in which view we are
+    let otherViews = ['X', 'Y'];
+    let otherNormals = [XPlaneNormal, YPlaneNormal];
+    if (vtkMath.dot(normal, XPlaneNormal) > 0.9) {
+      otherViews = ['Y', 'Z'];
+      otherNormals = [YPlaneNormal, ZPlaneNormal];
+    } else if (vtkMath.dot(normal, YPlaneNormal) > 0.9) {
+      otherViews = ['X', 'Z'];
+      otherNormals = [XPlaneNormal, ZPlaneNormal];
+    }
```
finetjul, line 113: I'm surprised you do not already have utility functions for that. 
Please add it as utility function next to the "GetAssociatedLine..."

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -83,36 +85,93 @@ export default function widgetBehavior(publicAPI, model) {
...
+
+      const otherViewsPlaneOutOfBounds = [];
+      let otherPlanesAreOutsideBounds = false;
+      for (let i = 0; i < otherViews.length; i++) {
+        const outBounds =
+          !vtkBox.intersectPlane(imageBounds, newCenter, otherNormals[i]) &&
+          vtkBox.intersectPlane(imageBounds, currentCenter, otherNormals[i]);
+        otherViewsPlaneOutOfBounds.push(outBounds);
+        otherPlanesAreOutsideBounds = otherPlanesAreOutsideBounds && outBounds;
+      }
```
finetjul, line 138:
```
const otherViewsPlaneOutOfBounds = otherViews.map( ... );
const otherPlanesAreOutsideBounds = otherViewsPlaneOutOfBounds.every(...)
```

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -83,36 +85,93 @@ export default function widgetBehavior(publicAPI, model) {
...
+            newCenter = vtkPlane.intersectWithLine(
+              axis.getPoint1(),
+              axis.getPoint2(),
+              newCenter,
+              normal
+            ).x;
+            break;
+          }
+        }
+      }
```
finetjul, line 174: couldn't you use `vtkBox.intersectBox` ? (maybe its naming is not good and should be changed (by preserving backward compatibility)).

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -83,36 +85,93 @@ export default function widgetBehavior(publicAPI, model) {
...
-      // Get it back in world coordinates
-      newCenter = localRoundedCenter.map(
-        (value, index) => value * imageSpacing[index] + imageOrigin[index]
-      );
+    const newCenterInsideBounds = bounds.containsPoint(...newCenter);
+
+    // The center is out of the image
+    if (!newCenterInsideBounds) {
+      // Find which plane is out of the bounds
+      const isCurrentPlaneOutOfBounds =
```
finetjul, line 126: isn't it `isNewPlaneOutOfBounds` according to the following 2 lines ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -83,36 +85,93 @@ export default function widgetBehavior(publicAPI, model) {
...
+        // Find new reslice cursor center in local coordinates and round it to be on a valid slice
+        const imageOrigin = image.getOrigin();
+        const localRoundedCenter = currentCenter.map(
+          (value, index) =>
+            Math.round(value - imageOrigin[index]) / imageSpacing[index]
+        );
+
+        // Get it back in world coordinates
+        newCenter = localRoundedCenter.map(
+          (value, index) => value * imageSpacing[index] + imageOrigin[index]
```
finetjul, line 151: if image origin is 0,0,0, spacing (1,1,1) and dimensions( 10, 10, 10).

if currentCenter is (5,5,5) and newCenter is (15, 5, 5), then localRoundedCenter would be 5, 5, 5 and newCenter (5, 5, 5) but we want (10, 5, 5).

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: The variable tests if the current position is correct and the next one, out of the bounds. We can call it both.
Current only means here that it's the current view type (for exemple X if we are scrolling on X view)

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: I don't want to intersect with a box. I really want to get the intersection between a plane and a line.
Why would you use `vtkBox.intersectBox` ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: Indeed. That's a mistake. It's only use to get the correct id of a slice.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
finetjul: here "current" means too things: `currentCenter` vs `newCenter` and current as in currentView. 
Maybe you could rename `currentCenter` by `oldCenter` to avoid confusion.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
finetjul: If newCenter is outside the box, you "project" `newCenter` onto the box (the box being the bounds).
The projection is done by creating a line betwen `oldCenter` and `newCenter` and see where it intersect the bounds. `newCenter` is then moved onto the intersection point.

finetjul: Looks like ResliceCursorWidget/helpers.js `boundPoint` does what you want, no ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -81,6 +81,35 @@ export function getViewPlaneNameFromViewType(viewType) {
...
+  const viewY = {
+    name: 'Y',
+    normal: YPlaneNormal,
+  };
+  const viewZ = {
+    name: 'Z',
+    normal: ZPlaneNormal,
+  };
+  let otherViews = [viewX, viewY];
+  if (vtkMath.dot(currentNormal, XPlaneNormal) > 0.9) {
```
finetjul, line 105: this does not seem very robust (ViewX and ViewY can eventually share the same normal).
You should probably use `viewType` to find other views.

laurennlam: `viewType `? The main issue I have here is that I cannot know in which view I am. I only catch the mouse wheel on a renderer but I don't know which one of the three.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: Maybe but I'm not sure. As it works like this, I don't want to change because I'm afraid it would take me too many times to think on an other way to use this method.

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
finetjul: Isn't `model.widgetState.getActiveViewName()` returning the current viewType ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: As we are moving spacing by spacing with the mouse wheel, these code only clamped the value to the closest slice.
Following your example, if the currentCenter is (9.5, 5, 5), then after using the mouse wheel, you'll have (10.5, 5, 5). This code will clamp it to (10, 5, 5)

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
laurennlam: Unfortunately not. These state attribute is only set when you select a pickable object:
https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js#L184

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -68,6 +72,110 @@ export default function widgetBehavior(publicAPI, model) {
...
+            // Get the correct line that will be used to find the new center
+            // by computing the intersection between current plane and the line
+            const axisName = `Axis${outViewName}in${displayViewOutOfBounds}`;
+            const axis = model.widgetState[`get${axisName}`]();
+            newCenter = vtkPlane.intersectWithLine(
+              axis.getPoint1(),
+              axis.getPoint2(),
+              newCenter,
+              normal
+            ).x;
```
finetjul, line 157: isn't `axis` within the plane defined by oldCenter and normal ? 
If so, how can you get an intersection between  the line `axis` and oldCenter and normal ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
finetjul: I think for the sake of genericity, it should not be assumed that you are moving spacing by spacing.

finetjul: (as a side note, if currentCenter is (9.1, 5, 5) then after a mouse wheel of 1, you'll have (10.5, 5, 5). The code will clamp it to (9, 5, 5) and you will therefore never be able to see the slice 10)

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: No. When you enter into the else, it means that only one of the two other planes (not the current one) is out of the bounds.
The cas when it's the current view which is out of the bounds is manage in the if.

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
finetjul: In behavior.js , you can do `model.viewType`

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
finetjul: btw, if middle button is down and moved, then it should scroll (not just mouse wheel).

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
laurennlam: Indeed ! My bad. It's really such a better way to do that. Thanks!

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -48,6 +49,18 @@ export function boundPoint(inPoint, v1, v2, bounds) {
   return outPoint;
 }
 
+// Compute the intersection between p1 and p2 on bounds
+export function boundPoint2(p1, p2, bounds) {
```
finetjul, line 53: Rename `boundPoint` into `boundPointOnPlane`

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -81,6 +94,35 @@ export function getViewPlaneNameFromViewType(viewType) {
...
+    name: 'Z',
+    normal: ZPlaneNormal,
+  };
+  let otherViews = [viewX, viewY];
+  if (viewType === ViewTypes.SAGITTAL) {
+    otherViews = [viewY, viewZ];
+  } else if (viewType === ViewTypes.CORONAL) {
+    otherViews = [viewX, viewZ];
+  }
+  return otherViews;
```
finetjul, line 123: I would split that function into 2 functions (1 function to get the other view types and 1 function to get the normal of a viewType).

But it seems to no longer be needed...

laurennlam: Indeed.

finetjul:
It is only missing the middle mouse button interaction and we should be all set.

laurennlam:
Then I will working on the customizable scrolling.

laurennlam:
@finetjul , I finished the customizable scrolling. That can be reviewed.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -1,17 +1,22 @@
...
+  boundPointOnPlane,
   getAssociatedLinesName,
   updateState,
 } from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers';
 
+import { ScrollingMethods } from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/Constants';
+
 export default function widgetBehavior(publicAPI, model) {
   let isDragging = null;
+  let onScrolling = false;
```
finetjul, line 19: onScrolling -> isScrolling ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -54,20 +64,89 @@ export default function widgetBehavior(publicAPI, model) {
     if (isDragging && model.pickable) {
       return publicAPI.handleEvent(callData);
     }
+    if (onScrolling) {
+      if (model.previousPosition !== callData.position) {
```
finetjul, line 68: shouldn't you compare element per element ?

finetjul:
LGTM

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: Is it necessary ?

finetjul: well it depends what you want to compare... It is a semantical question.
```
a = {x: 1, y: 1}
b = {x: 1, y: 1}
b !== a // it is true 
c = a
c !== a // it is false
```

laurennlam: Then, I do not understand how it can works now.

finetjul: it works because for each new movement, callData.position is different (pointers are different).

laurennlam: Yes but that sounds weird. But, maybe in order to not being confused, I'll check the y values.

PR #1608, opened 2020-10-20
fix(renderwindowinteractor): ignore mouse events from back and forward buttons

jbeezley:
Currently, when you try to navigate with a mouse's side buttons while hovering over a vtk context, an error is thrown in the console.

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -295,7 +295,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
   };
 
   publicAPI.handleMouseDown = (event) => {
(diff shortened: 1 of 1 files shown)


jourdain:
@floryst what do you think?

floryst:
`event.which` is actually a deprecated web standard, so while I would suppose browsers will support it for a long time coming (on the basis of supporting legacy code), we should move off of it if possible. Can you see if you can re-write it using `event.key` or `event.code`? Here's some more info: https://stackoverflow.com/questions/49278648/alternative-for-events-deprecated-keyboardevent-which-property

jbeezley:
`button` appears to return compatible information.  I made the switch.

floryst:
LGTM. Once you rebase on master then we should be good to merge.

PR #1612, opened 2020-10-28
Fix reslice cursor

laurennlam:
- Fix rendering when volume origin is not [0, 0, 0]
- Setup te correct camera's view up on sagittal view

@finetjul

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
@@ -1,6 +1,5 @@
 import macro from 'vtk.js/Sources/macro';
 import vtkAbstractWidgetFactory from 'vtk.js/Sources/Widgets/Core/AbstractWidgetFactory';
-import vtkBoundingBox from 'vtk.js/Sources/Common/DataModel/BoundingBox';
 import vtkPlane from 'vtk.js/Sources/Common/DataModel/Plane';
 import vtkPlaneSource from 'vtk.js/Sources/Filters/Sources/PlaneSource';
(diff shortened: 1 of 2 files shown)


Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -267,6 +263,12 @@ function vtkResliceCursorWidget(publicAPI, model) {
     renderer.getActiveCamera().setFocalPoint(...estimatedFocalPoint);
     renderer.getActiveCamera().setPosition(...estimatedCameraPosition);
 
+    // Update view up in sagittal view
+    if (viewType === ViewTypes.SAGITTAL) {
+      const viewUp = [0, 0, 1];
+      renderer.getActiveCamera().setViewUp(viewUp);
```
finetjul, line 269: can't it be done in `computeReslicePlaneOrigin()` ?

laurennlam: We don't have access to camera in this method. Do you want to return an object ?

finetjul: I wonder if the same end result as setting a view up can be achieved by changing what "point1" and "point2" mean.

laurennlam: Okay! Just rechecked the point0/1/2 set in the planeSource.
There are already well defined:
0 = bottom left
1 = bottom right
2 = top left
So that we can compute the view up thanks to points 0 and 1

finetjul:
LGTM if that works with oblique views...

laurennlam:
Good point, I didn't test it with oblique and not a surprised, that won't work...

laurennlam:
@finetjul That should be good now.

finetjul:
LGTM

PR #1614, opened 2020-10-30
fix(reslicecursorwidget): fix reslice cursor context representation i…

finetjul:
…nitial opacity

updateActorVisibility() is called before requestData().
At initialization, model.opacity is undefined. Because it is undefined
actor opacity was unset and requestData() was not called. Therefore
opacity was not initialized.


finetjul:
@laurennlam

laurennlam:
LGTM Thanks!

PR #1615, opened 2020-10-30
fix(Widgets): Guard against empty worldCoords

floryst:
Manipulators may return an empty result, so a guard must be in place.

FYI @laurennlam @finetjul I know you've been doing widgets stuff, so if any of these changes conflict or apply to your PRs, feel free to copy the changes to your PRs as to avoid conflicts.


finetjul:
LGTM

It should not be an issue for us, we always have world coords.

PR #1616, opened 2020-10-30
Update deps

### LINKED ISSUE #1596: Incompatibility with latest worker-loader package
Hello, 
I have an issue creating a project with vuejs about the compatibility of `worker-loader` and `vtk.js` package.
Using the latest version of `vtk.js` and a version of `worker-loader` >= 3.0.0 I got the following error:
```
 error  in ./node_modules/vtk.js/Sources/Filters/General/PaintFilter/PaintFilter.worker.js

Module build failed (from ./node_modules/worker-loader/dist/cjs.js):
ValidationError: Invalid options object. Worker Loader has been initialized using an options object that does not match the API schema.
 - options has an unknown property 'fallback'. These properties are valid:
   object { worker?, publicPath?, filename?, chunkFilename?, inline?, esModule? }
 - options.inline should be one of these:
   "no-fallback" | "fallback"
    at validate (./vtk-test/node_modules/schema-utils/dist/validate.js:98:11)
    at Object.pitch (./vtk-test/node_modules/worker-loader/dist/index.js:57:28)

```

, decreasing the worker-loader version to 2.0.0 everything works.

Do you have any idea if there is some configuration I'm missing?

[Here](https://github.com/paologf/vtk-test) you could find the sample project.

Thanks

Paolo

floryst:
BREAKING CHANGE: This brings in webpack 5 and a few other major version changes to dependencies.

Fixes #1596 

@finetjul We can merge this alongside the LineWidget when that PR is done, since this is also a breaking change.


jourdain (approved):
Seems reasonable.

Did you try to run the following to make sure you did not miss anything?

```
npm run doc:www
npm run example -- SimpleCone
```

floryst:
I'm currently working through fixing `doc:www` and other npm commands.

floryst:
Since this transition is making the jump from webpack 4 to 5, we have to wait on webpack-dev-server to become compatible with webpack v5 before upgrading.

The published version of karma-webpack does not support webpack v5, nor is there any activity indicating it will (yet), so I'm using their `next` branch for the time being (which somehow works under webpack v5).

floryst:
I'm going to use `webpack serve` for now in lieu of webpack-dev-server. This does mean that running examples will not auto-open a browser.

PR #1619, opened 2020-11-08
Fix plane source

finetjul:


finetjul:
@laurennlam

Sources/Filters/Sources/PlaneSource/index.js
```diff
@@ -129,23 +129,22 @@ function vtkPlaneSource(publicAPI, model) {
...
+          vtkMath.cross(model.normal, n, rotationVector);
           theta = vtkMath.degreesFromRadians(Math.acos(dp));
         }
         // Create rotation matrix
         const transform = mat4.create();
         const negCenter = [];
-
         vec3.negate(negCenter, model.center);
 
-        mat4.translate(transform, transform, negCenter);
```
laurennlam, line 141: `vec3.negate` doesn't modify model.center. Only `negCenter`. As `negCenter` is not used, you can remove `vec3.negate(negCenter, model.center)`

laurennlam: My bad, didn't see it's used right after.

laurennlam:
LGTM, thanks for the fix!

PR #1622, opened 2020-11-09
Increase picking tolerance

finetjul:


Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -257,6 +257,7 @@ const DEFAULT_VALUES = {
     cameraPosition: [0, 0, 0],
     cameraDir: [1, 0, 0],
     isParallel: false,
+    rendererPixelDims: [1, 1],
```
finetjul, line 260: @floryst had to set it up to avoid crash. let me know if it is not a good default

finetjul:
@laurennlam

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
```diff
@@ -89,24 +91,64 @@ function vtkResliceCursorContextRepresentation(publicAPI, model) {
...
-    axis.line.source.setPoint1(state.getPoint1());
-    axis.line.source.setPoint2(state.getPoint2());
+    const vector = [0, 0, 0];
+    vtkMath.subtract(state.getPoint2(), state.getPoint1(), vector);
+    const center = [0, 0, 0];
+    vtkMath.multiplyAccumulate(state.getPoint1(), vector, 0.5, center);
+    axis.line.source.setCenter(center);
+    const length = vtkMath.normalize(vector);
+    axis.line.source.setDirection(vector);
+    axis.line.source.setHeight(length);
```
laurennlam, line 151: My understanding of this is that lines will be drawn only between two rotations points ?
But now, lines are longer than this distance.

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
```diff
@@ -184,6 +227,17 @@ function vtkResliceCursorContextRepresentation(publicAPI, model) {
...
       }
     });
+    let lineThickness = state.getLineThickness();
+    if (renderingType === RenderingTypes.PICKING_BUFFER) {
+      lineThickness = Math.max(3, lineThickness);
+    }
+    publicAPI.setLineThickness(lineThickness);
+
+    let radius = state.getSphereRadius();
+    if (renderingType === RenderingTypes.PICKING_BUFFER) {
```
laurennlam, line 237: Is there a reason to have two `if (renderingType === RenderingTypes.PICKING_BUFFER)` ?
Else, maybe we can factorize:
`
let lineThickness = state.getLineThickness();
let radius = state.getSphereRadius();
if (renderingType === RenderingTypes.PICKING_BUFFER) {
  lineThickness = Math.max(3, lineThickness);
  radius += 1;
}
publicAPI.setLineThickness(lineThickness);
publicAPI.setSphereRadius(radius);
`

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
finetjul: the cylinder will be as long as the previous lines.

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
finetjul: This is an interesting good point.

In this particular case, for some reason (and I admit I do not have a scientific proof to provide), I find it better (more readable) to keep linethickness separated from sphere radius.

But if you feel strong about it, I can factorize like you suggest.

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
laurennlam: ok sounds good then!

Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
laurennlam: I was just surprised, especially from you, to have e dupplication in your code :)
But it's just a style, you can keep it as you feel.

Sources/Widgets/Representations/WidgetRepresentation/index.js
floryst: Seems good to me.

PR #1623, opened 2020-11-10
fix(vtkforwardpass): check to see if a renderer should be drawn

JamesAPetts:
Hey,

Renderers currenlty have a `draw` property, which the api docs show should prevent a redraw if set to false, however this is never checked.

I've added the check to the forwardPass, confirmed locally it works when you have multiple `renderer`s on a single `renderWindow`.


martinken:
LGTM +1

JamesAPetts:
Good to merge then @jourdain? :)

PR #1625, opened 2020-11-11
Fix reslice cursor rotation

finetjul:


CONTRIBUTING.md
```diff
@@ -71,6 +71,14 @@ Please follow the coding style:
     $ npm test
     ```
 
+## Tests
+
+- To create a test, create a testFuncNameToTest.js in a "test" folder of the feature to test.
+- Add the test path into Sources/tests.js and temporarily comment the other tests
+- Edit karma.conf.js and uncomment commented lines
+- Run `npm test`
+- In the opened window, click the Debug button and place breakpoints in browser debugger.
```
finetjul, line 80: @jourdain is it the best way to debug tests ?

jourdain: There is a `npm run test:debug` but you need to comment out (temporary) all the tests that you don't care in the `Sources/test.js`

finetjul: thanks

laurennlam:
LGTM

PR #1633, opened 2020-11-18
feat(InteractorStyleUnicam) Add InteractorStyleUnicam

laurentmalka:
Add InteractorStyleUnicam, an interactor style that uses a single button to function


Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+        break;
+    }
+
+    model.previousPosition = position;
+  };
+
+  //--------------------------------------------------------------------------
+  publicAPI.handleLeftButtonRelease = (callData) => {
+    model.leftButtonPressed = false;
+
```
finetjul, line 144: Add ```renderer = callData.pokedRenderer; ```

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+        model.focusSphere.setPosition(...model.downPoint);
+
+        const camera = callData.pokedRenderer.getActiveCamera();
+        const cameraPosition = camera.getPosition();
+
+        const cameraToPointVec = [];
+        for (let index = 0; index < 3; index++) {
+          cameraToPointVec[index] =
+            model.downPoint[index] - cameraPosition[index];
+        }
```
finetjul, line 162: vtkMath.subtract

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+
+        camera.computeDistance();
```
finetjul, line 164: why do you need to call it. Isn't it already called internally when needed ?

Sources/Interaction/Style/InteractorStyleUnicam/example/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+document.querySelector('.useWorldUpVector').oninput = () => {
+  updateWorldUpVector();
+  renderWindow.render();
+};
+document.querySelector('.coordinates').oninput = () => {
+  updateWorldUpVector();
+  renderWindow.render();
+};
+
```
finetjul, line 101: please add a checkbox that toggles from perspective to orthographic camera projection

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+    return { x: nx, y: ny };
+  };
+
+  model.getRightVAndUpV = (downPoint, camera) => {
+    const cameraPosition = camera.getPosition();
+    const cameraToPointVec = [0, 0, 0];
+
+    for (let index = 0; index < 3; index++) {
+      cameraToPointVec[index] = downPoint[index] - cameraPosition[index];
+    }
```
finetjul, line 376: vtkMath.subtract

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+
+    const atV = camera.getViewPlaneNormal();
+    vtkMath.normalize(atV);
+    const l = vtkMath.dot(cameraToPointVec, atV);
+    const viewAngle = (camera.getViewAngle() * vtkMath.Pi()) / 180.0;
```
finetjul, line 381: vtkMath.radiansFromDegrees

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+    const rightV = [];
+    vtkMath.cross(upV, atV, rightV);
+    vtkMath.cross(atV, rightV, upV);
+    vtkMath.normalize(rightV);
+    vtkMath.normalize(upV);
+
+    for (let index = 0; index < 3; index++) {
+      rightV[index] *= scaleX;
+      upV[index] *= scaleY;
+    }
```
finetjul, line 397: vtkMath.multiplyScalar

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+  model.translateCamera = (camera, offset) => {
+    const cameraPosition = camera.getPosition();
+    const cameraFocalPoint = camera.getFocalPoint();
+    const newCameraPosition = [];
+    const newCameraFocalPoint = [];
+
+    for (let index = 0; index < 3; index++) {
+      newCameraPosition[index] = cameraPosition[index] + offset[index];
+      newCameraFocalPoint[index] = cameraFocalPoint[index] + offset[index];
+    }
```
finetjul, line 412: vtkMath.add

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+
+    camera.setPosition(...newCameraPosition);
+    camera.setFocalPoint(...newCameraFocalPoint);
```
finetjul, line 415: optimize that by creating a new publicAPI.translate(x, y, z) method in vtkCamera

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+    mat4.identity(transform);
+    mat4.rotate(transform, transform, angle, vec3.fromValues(ax, ay, az));
+    const newCameraViewUp = [];
+    vec3.transformMat4(newCameraViewUp, cameraViewUp, transform);
+
+    camera.setPosition(...newCameraPosition);
+    camera.setFocalPoint(...newCameraFocalPoint);
+    camera.setViewUp(...newCameraViewUp);
+
+    camera.computeDistance();
```
finetjul, line 447: why do you need it ?

Sources/Interaction/Style/InteractorStyleUnicam/example/index.js
```diff
@@ -0,0 +1,107 @@
...
+function updateWorldUpVec() {
+  let useWorldUpVec = document.querySelector('.useWorldUpVec').checked;
+
+  let coordinates = Array.from(document.querySelector('.coordinates').children);
+  coordinates.forEach((element) => {
+    element.disabled = !useWorldUpVec;
+  });
+
+  if (useWorldUpVec) {
+    coordinates = coordinates.map((item) => {
```
finetjul, line 75: it is not great to repurpose your array. it is better to create a new array
```
const coordinateElements = ...
if () {
  const coordinates = coordinatesElements.map
```

Sources/Interaction/Style/InteractorStyleUnicam/example/index.js
```diff
@@ -0,0 +1,107 @@
...
+document.querySelector('.useWorldUpVec').oninput = updateWorldUpVec;
+document.querySelector('.coordinates').oninput = updateWorldUpVec;
+
+document.querySelector('.useOrientationMarker').oninput = (ev) => {
+  orientationWidget.setEnabled(ev.target.checked);
+};
+
+document.querySelector('.useParallelCamera').oninput = (ev) => {
+  renderer.getActiveCamera().setParallelProjection(ev.target.checked);
+  updateWorldUpVec();
```
finetjul, line 102: I do not see why update world up vec when changing parallel projection

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+      const upVec = model.useWorldUpVec ? model.worldUpVec : camera.getViewUp();
+      vtkMath.normalize(upVec);
+
+      model.rotateCamera(camera, ...center, ...upVec, angle);
+
+      const dVec = [];
+      const cameraPosition = camera.getPosition();
+      for (let index = 0; index < 3; index++) {
+        dVec[index] = cameraPosition[index] - center[index];
+      }
```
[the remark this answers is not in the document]
finetjul: can't you use vtkMath.subtract ?

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,497 @@
...
+      let rDist = normalizedPosition.y - normalizedPreviousPosition.y;
+      vtkMath.normalize(dVec);
+
+      const atV = camera.getViewPlaneNormal();
+      const upV = camera.getViewUp();
+      const rightV = [];
+      vtkMath.cross(upV, atV, rightV);
+      vtkMath.normalize(rightV);
+
+      if (model.useWorldUpVec) {
```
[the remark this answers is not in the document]
finetjul: I meant to copy them every where in this file, not just here

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,486 @@
...
+      camera.zoom(1 - delta.y);
+    } else {
+      const movec = [];
+      vtkMath.subtract(model.downPoint, cameraPosition, movec);
+      const offset1 = [];
+      for (let index = 0; index < 3; index++) {
+        offset1[index] = movec[index] * delta.y * -4;
+      }
+
+      camera.translate(...offset1);
```
finetjul, line 309: I think in Parallel projection, camera must ALSO be translated

Sources/Interaction/Style/InteractorStyleUnicam/index.js
```diff
@@ -0,0 +1,486 @@
...
+    }
+
+    const { rightV } = getRightVAndUpV(model.downPoint, camera);
+
+    const offset2 = [];
+    for (let index = 0; index < 3; index++) {
+      offset2[index] = delta.x * rightV[index];
+    }
```
finetjul, line 317: use vtkMath.multiplyScalar

Sources/Rendering/Core/Camera/index.js
```diff
@@ -338,6 +338,16 @@ function vtkCamera(publicAPI, model) {
...
   };
 
+  publicAPI.translate = (x, y, z) => {
+    const offset = [x, y, z];
+
+    vtkMath.add(model.position, offset, newPosition);
+    vtkMath.add(model.focalPoint, offset, newFocalPoint);
+
+    publicAPI.setPosition(...newPosition);
+    publicAPI.setFocalPoint(...newFocalPoint);
```
finetjul, line 348: the idea of creating translate is to avoid calling modified and compute distance twice, but only once.

finetjul (changes requested):

Sources/Interaction/Style/InteractorStyleUnicam/example/index.js
```diff
@@ -0,0 +1,109 @@
...
+import vtkOrientationMarkerWidget from 'vtk.js/Sources/Interaction/Widgets/OrientationMarkerWidget';
+
+import controlPanel from './controlPanel.html';
+
+// ----------------------------------------------------------------------------
+// Renderer setup
+// ----------------------------------------------------------------------------
+
+const fullScreenRenderWindow = vtkFullScreenRenderWindow.newInstance({
+  background: [0, 0, 0],
```
finetjul, line 18: set it to not black to see flickering when picking with orientation marker

finetjul:
There is a Z-buffer picking issue when enabling Orientation marker widget (that renders in a separate renderer). Additionally, if background is originally not black, it flickers when picking (when Orientation marker widget is on).
@martinken any idea of what is going on ?

finetjul:
@laurentmalka : please add dolly on scroll wheel.

Also, add simple parameters to associate a specific action (dolly, pan, choose, rotate...) to a button (left, middle, right). 

Finally, add a rotation factor parameter to rotate faster.(you should be able to do a 180 degree rotation at once).

Sources/Interaction/Manipulators/index.js
```diff
@@ -6,6 +6,7 @@ import vtkCompositeVRManipulator from './CompositeVRManipulator';
 import vtkGestureCameraManipulator from './GestureCameraManipulator';
 import vtkKeyboardCameraManipulator from './KeyboardCameraManipulator';
 import vtkMouseCameraAxisRotateManipulator from './MouseCameraAxisRotateManipulator';
+import vtkMouseCameraUnicamManipulator from './MouseCameraUnicamManipulator';
 import vtkMouseCameraSliceManipulator from './MouseCameraSliceManipulator';
```
finetjul, line 10: alphabetical sorting

Sources/Interaction/Manipulators/index.js
```diff
@@ -26,6 +27,7 @@ export default {
   vtkGestureCameraManipulator,
   vtkKeyboardCameraManipulator,
   vtkMouseCameraAxisRotateManipulator,
+  vtkMouseCameraUnicamManipulator,
   vtkMouseCameraSliceManipulator,
```
finetjul, line 31: alphabetical sorting

boucaud:
It tested this out and ran into an error using the 'unicam' manipulator, when clicking on an object :

```
index.js?2597:330 Uncaught TypeError: Cannot read property '0' of null
    at Object.publicAPI.getPixelInformation (index.js?2597:330)
    at Object.publicAPI.generateSelection (index.js?2597:468)
    at Object.publicAPI.select (index.js?2597:76)
    at Object.publicAPI.onButtonDown (index.js?f9e1:381)
    at Object.publicAPI.onButtonDown (index.js?3786:347)
    at Object.publicAPI.handleLeftButtonPress (index.js?3786:267)
    at Object.eval (index.js?8e0e:59)
    at _loop2 (macro.js?b5a9:850)
    at Object.invoke [as invokeLeftButtonPress] (macro.js?b5a9:839)
    at Object.publicAPI.<computed> [as leftButtonPressEvent] (index.js?8f5f:778)
```

The error is in vtkHardwareSelector. Do I need to initialize something in the renderer ?

Commenting lines 381:384 in MouseCameraUnicamManipulator (forcing use of the vtkPicker) seems to fix the issue.

laurentmalka:
I can't reproduce this bug, did you run the example or custom code ? If it's the latter, could you share it so I can test it ?

> * Is there a way to make the sphere radius bigger ? I would like to be able to set the color as well, it is hard to see when the data is colored yellow.

This is a good point, I'll add a way to change these properties

> * Rotating requires two clicks right now, one to place the sphere, then another one to rotate around it. Is it possible to do it in a single click (i.e place the sphere on button down then rotate around it until button up), or do you think this has to be implemented in a derived manipulator ?

Another way to rotate the camera is to grab the borders of the screen (10% on each side) which will start a rotation around the previous focus sphere, or around the (0,0,0) if no focus sphere has been placed before.
The problem with a single click like you described is that it would conflict with the pan and dolly interactions, which already use a hold of the click, so it would have to be a different manipulator which does not offer the ability to pan, dolly or both.

boucaud:
> I can't reproduce this bug, did you run the example or custom code ? If it's the latter, could you share it so I can test it ?

I've investigated further, and it seems the issue is not linked to this MR. (it occurs when adding renderPasses to a renderer. The hardwareSelector seems to assume the first pass is a forwardPass, so that causes errors). I'll try to work around it.

> This is a good point, I'll add a way to change these properties

Thanks !

> Another way to rotate the camera is to grab the borders of the screen (10% on each side) which will start a rotation around the previous focus sphere, or around the (0,0,0) if no focus sphere has been placed before.
The problem with a single click like you described is that it would conflict with the pan and dolly interactions, which already use a hold of the click, so it would have to be a different manipulator which does not offer the ability to pan, dolly or both.

Indeed, what I need is to disable pan and dolly, so I'll probably derive a new manipulator for this one.

finetjul:
I like the idea of having a "rotation-only" option for the unicam manipulator (especially if you can associate other manipulators to other buttons).

boucaud:
> I like the idea of having a "rotation-only" option for the unicam manipulator (especially if you can associate other manipulators to other buttons).
> 
> @boucaud there used to be a pb with the [orientation widget](https://github.com/Kitware/vtk-js/pull/1642), could it be related ?

No, the issue is directly caused by adding a pre-pass in the renderer. The hardware selector takes the first renderPass in the renderer and calls setDepthRequested on it, which of course does nothing with our preprocessing pass

I don't think it'll be easy to make these custom passes we use work well with the hardware selector. I think having an option to disable hardware selection entirely in the manipulator could be very helpful. @laurentmalka do you think you could add a flag for this ?

Sources/Interaction/Style/InteractorStyleManipulator/Presets.js
```diff
@@ -14,6 +14,7 @@ const MANIPULTOR_TYPES = {
   gestureCamera: Manipulators.vtkGestureCameraManipulator,
   movement: Manipulators.vtkKeyboardCameraManipulator,
   freeLook: Manipulators.vtkMouseCameraTrackballFirstPersonManipulator,
+  unicam: Manipulators.vtkMouseCameraUnicamManipulator,
```
boucaud, line 17: Should the rotate manipulator be added here ?

Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
```diff
@@ -0,0 +1,346 @@
...
+    button: model.button,
+    shift: model.shift,
+    control: model.control,
+    alt: model.alt,
+    dragEnabled: model.dragEnabled,
+    scrollEnabled: model.scrollEnabled,
+  });
+
+  //----------------------------------------------------------------------------
+  const updateAndRender = (interactor) => {
```
finetjul, line 43: does not seem to update anymore. Is this function still needed ?

Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
```diff
@@ -0,0 +1,346 @@
...
+    model.rotateManipulator.onButtonDown(interactor, renderer, position);
+  };
+
+  //----------------------------------------------------------------------------
+  publicAPI.onMouseMove = (interactor, renderer, position) => {
+    if (!model.buttonPressed) {
+      return;
+    }
+
+    if (model.rotateManipulator.getState() === States.IS_ROTATE) {
```
finetjul, line 247: move it in the switch

Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
```diff
@@ -0,0 +1,346 @@
...
+      model.selector.setArea(position.x, position.y, position.x, position.y);
+      selections = model.selector.select();
+    }
+
+    if (selections && selections.length !== 0) {
+      model.downPoint = selections[0].getProperties().worldPosition;
+    } else {
+      model.picker.pick([position.x, position.y, position.z], renderer);
+      model.downPoint = model.picker.getPickPosition();
+    }
```
finetjul, line 236: isn't `rotateManipulator.onButtonDown()` already doing all this work already ?

Sources/Interaction/Manipulators/MouseCameraUnicamRotateManipulator/index.js
```diff
@@ -0,0 +1,368 @@
...
+
+  // Setup focus dot
+  const sphereSource = vtkSphereSource.newInstance();
+  sphereSource.setThetaResolution(6);
+  sphereSource.setPhiResolution(6);
+  const sphereMapper = vtkMapper.newInstance();
+  sphereMapper.setInputConnection(sphereSource.getOutputPort());
+
+  // XXX - Would like to make the focus sphere not be affected by
+  // XXX - the lights-- i.e., always be easily easily seen.  i'm not sure
```
finetjul, line 45: set only ambient color, no diffuse

Sources/Interaction/Style/InteractorStyleManipulator/Presets.js
finetjul: I agree

Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
```diff
@@ -0,0 +1,311 @@
...
+  model.classHierarchy.push('vtkMouseCameraUnicamManipulator');
+
+  // Setup HardwareSelector and Picker to pick points
+  model.selector = vtkOpenGLHardwareSelector.newInstance({
+    captureZValues: true,
+  });
+  model.selector.setFieldAssociation(
+    FieldAssociations.FIELD_ASSOCIATION_POINTS
+  );
+  model.picker = vtkPicker.newInstance();
```
finetjul, line 29: you don't need those variables anymore, do you ?

Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
```diff
@@ -0,0 +1,311 @@
...
+    button: model.button,
+    shift: model.shift,
+    control: model.control,
+    alt: model.alt,
+    dragEnabled: model.dragEnabled,
+    scrollEnabled: model.scrollEnabled,
+  });
+
+  //----------------------------------------------------------------------------
+  const normalize = (position, interactor) => {
```
finetjul, line 43: use RotateUnicam normalize instead ?

Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
```diff
@@ -0,0 +1,311 @@
...
+  // Public API methods
+  //----------------------------------------------------------------------------
+  publicAPI.onButtonDown = (interactor, renderer, position) => {
+    model.buttonPressed = true;
+    model.startPosition = position;
+    model.previousPosition = position;
+    model.time = Date.now() / 1000.0;
+    model.dist = 0;
+
+    model.rotateManipulator.onButtonDown(interactor, renderer, position);
```
finetjul, line 209: add comment that you delegate the picking to the rotateManipulator

Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    model.previousPosition = position;
+  };
+
+  //--------------------------------------------------------------------------
+  publicAPI.onButtonUp = (interactor) => {
+    model.buttonPressed = false;
+    if (model.state === States.IS_NONE) {
+      model.rotateManipulator.onButtonUp(interactor);
```
finetjul, line 245: it's too bad that rotateManipulator calls render() even if we were not in rotation mode

Sources/Interaction/Manipulators/MouseCameraUnicamRotateManipulator/index.js
```diff
@@ -0,0 +1,368 @@
...
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {
+  focusSphereRadiusFactor: 1,
+  useHardwareSelector: true,
+  useWorldUpVec: true,
+  // set WorldUpVector to be z-axis by default
+  worldUpVec: [0, 0, 1],
```
finetjul, line 332: I do not see the 2 modes:
 - down&drag to rotate
 - click then drag to rotate

laurentmalka: These two interactions are already available in this manipulator, but I did not define two different modes because they are not overlapping. I can split it in two modes if you think there is a point in doing so.

finetjul: nope, sounds good to me. thanks.

Sources/Interaction/Manipulators/MouseCameraUnicamRotateManipulator/index.js
```diff
@@ -0,0 +1,368 @@
...
+
+      model.selector.setArea(position.x, position.y, position.x, position.y);
+      selections = model.selector.select();
+    }
+
+    if (selections && selections.length !== 0) {
+      model.downPoint = selections[0].getProperties().worldPosition;
+    } else {
+      model.picker.pick([position.x, position.y, position.z], renderer);
+      model.downPoint = model.picker.getPickPosition();
```
boucaud, line 237: This method seems to pick points that are way out of the model. Can you try using picker.getPickedPositions() instead ? Usually sorting by distance to camera lets you get the best point.

boucaud (commented):
After testing the rotate manipulator, it seems the sphere only appears after I release the mouse button, and if the mouse was not move. Is it possible to show the sphere immediately after button down, and hide it on button up ?

boucaud (approved):
+1 for me (tested without hardwareSelector).

finetjul (approved):

PR #1639, opened 2020-11-22
The selection condition was added

gandis0713:
When adding a widget to the environment of 1 renderwindow and n renderers, selectedState does not work properly in mouse move event.

I think It compares the pokedRenderer that is passed to the mouse move event and the Renderer that WindgetManager has, and if they are different, it seems that there is no need to handle picking.

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -321,7 +321,11 @@ function vtkWidgetManager(publicAPI, model) {
     );
 
     subscriptions.push(
-      model.interactor.onMouseMove(({ position }) => {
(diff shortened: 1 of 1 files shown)


floryst:
Looks reasonable to me. I'll first want to test it out before merging.

floryst:
Hi, if you can rebase this on master then I can merge. Also, can you rewrite your commits to follow [semantic-release](https://github.com/semantic-release/semantic-release) conventions (excluding your merge commit)? This should be a fix commit. Thanks!

gandis0713:
Sorry for not following the contributing process.
I have rebase this branch on master. Also, rewritten commits.

floryst:
Thanks!

PR #1642, opened 2020-11-30
fix(OrientationMarkerWidget): Use resize observer

floryst:
This avoids unnecessary resize updates when the scene isn't actually resizing.

@laurentmalka @finetjul This should fix your Z buffer issue. Below is a brief summary of what I think is happening.

Prior to this PR, `vtkOrientationMarkerWidget.updateViewport()` gets called every time the interactor view is updated (which includes mouse movement). `updateViewport()` calls `interactor.render()`, which will of course render the scene. Now consider how vtkOpenGLHardwareSelector works. At a high level, it tells the renderer to do a picking render to populate the picking buffer. When moving the mouse over the scene, we trigger calls to the hardware selector to render the picking buffer. Somehow, during the rendering phase of the hardware selector, the interactor `onModified` callbacks are being invoked, and so as a result a regular render happens in the middle of the picking render. (To be more specific, this is a result of onModified listeners being immediately called when a setter is invoked.)

### DIFF
--- a/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
+++ b/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
@@ -19,8 +19,12 @@ function vtkOrientationMarkerWidget(publicAPI, model) {
 
   const previousCameraInput = [];
   const selfRenderer = vtkRenderer.newInstance();
+  const resizeObserver = new ResizeObserver((entries) => {
+    if (entries.length === 1) {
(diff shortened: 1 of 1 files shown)


finetjul:
LGTM (especially if that fixes the https://github.com/Kitware/vtk-js/pull/1633)

laurentmalka:
It indeed seems to fix the Z buffer issue and the flickering when picking

PR #1645, opened 2020-12-03
improvement(linewidget): LineWidget rotation and visibility features

ocrossi:
go from dynamic to static elements in the state of LineWidget
SVG layer solve position bug when multiple line Widgets in the same scene
add ghost sphere handle (visible only when mouseover)
isDragging variable linked to widget state


ocrossi:
@finetjul

Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js
```diff
@@ -11,6 +11,22 @@ function vtkSVGLandmarkRepresentation(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkSVGLandmarkRepresentation');
 
+  publicAPI.setDy = (val) => {
+    model.dy = val;
+  };
+
+  publicAPI.getDy = () => {
+    return model.dy;
+  };
```
finetjul, line 20: use macro.setGet to create those methods

Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js
```diff
@@ -11,6 +11,22 @@ function vtkSVGLandmarkRepresentation(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkSVGLandmarkRepresentation');
 
+  publicAPI.setDy = (val) => {
```
finetjul, line 14: dy is probably not a good variable name (not self explicit enough)

Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js
```diff
@@ -11,6 +11,22 @@ function vtkSVGLandmarkRepresentation(publicAPI, model) {
...
+    model.dy = val;
+  };
+
+  publicAPI.getDy = () => {
+    return model.dy;
+  };
+
+  publicAPI.setFontProperties = (fontProperties) => {
```
finetjul, line 22: use macro.setGet

Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js
```diff
@@ -45,15 +60,32 @@ function vtkSVGLandmarkRepresentation(publicAPI, model) {
...
-        Object.keys(model.textProps || {}).forEach((prop) =>
-          text.setAttribute(prop, model.textProps[prop])
-        );
-        text.setAttribute('x', x);
-        text.setAttribute('y', y);
-        text.textContent = texts[i];
-
-        root.appendChild(text);
+        const splitText = texts[i].split('\n');
+        let j = 0;
```
finetjul, line 64: not needed here.
Simply do: 
```
splitText.forEach((subText, j) => {...
```

Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js
```diff
@@ -87,6 +119,16 @@ export function extend(publicAPI, model, initialValues = {}) {
 
   macro.setGet(publicAPI, model, ['circleProps', 'textProps', 'name']);
 
+  model.fontProperties = null;
+  if (initialValues.fromLineWidget === true) {
+    // this allows for different offset values when there are multiple instances of this class
+    model.dx = model.textProps.dx;
+    model.dy = model.textProps.dy;
+    model.fromLineWidget = true;
```
finetjul, line 127: not needed, initialValues.fromLineWidget is already in model since `Object.assign(model, DEFAULT_VALUES,initialValues)`

Sources/Widgets/Widgets3D/LineWidget/Constants.js
```diff
@@ -11,6 +11,7 @@ export const HandleBehavior = {
 export const HandleRepresentationType = {
   // 3D handles
   SPHERE: 'sphere',
+  GHOST_SPHERE: 'ghost_sphere',
```
finetjul, line 14: please explain what this is

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -99,6 +105,34 @@ export default function widgetBehavior(publicAPI, model) {
     );
   }
 
+  publicAPI.hideGhostSpheres = () => {
+    if (model.handle1Shape === HandleRepresentationType.GHOST_SPHERE) {
+      model.representations[0].getActors()[1].getProperty().setOpacity(0.01);
```
finetjul, line 110: seems like a hack, please explain

Sources/Widgets/Widgets3D/LineWidget/state.js
```diff
@@ -15,7 +15,7 @@ export default function generateState() {
     })
     .addStateFromMixin({
       labels: ['handle1'],
-      mixins: ['origin', 'color', 'scale1', 'visible'],
+      mixins: ['origin', 'color', 'scale1', 'visible', 'manipulator'],
```
finetjul, line 18: what is it for ?

Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js
ocrossi: the problem with set/get macro is to my mind that it modifies data stored only in default values. My problem here was to allow multiple instances of SVGLandmarkRepresentation to have different values of Dy for each instance (and avoid a bug of moving all instances at once when only one instance was to be moved)

Sources/Widgets/Widgets3D/LineWidget/state.js
ocrossi: not sure, but manipulators are used in OFSEP (behavior similar to intensityWidget). Should i remove it?

Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js
jourdain: Not true. Data are stored inside `model` that is specific to each instance.

finetjul:
@ocrossi I just pushed a cleanup commit. Can you please explain how I can I reproduce the `setDy` error in the VTK.JS LineWidget example ?

Sources/Filters/Sources/Arrow2DSource/index.js
```diff
@@ -70,7 +70,24 @@ function vtk6PointsArrow(publicAPI, model) {
   points[16] = model.height * 0.1 - thickOp - offsetOp - centerOffsetOp;
   points[17] = 0.0;
 
-  const cells = Uint8Array.from([6, 0, 1, 2, 3, 4, 5]);
+  const cells = Uint8Array.from([
```
floryst, line 73: Here's a neat trick to make this more "prettier": put `// prettier-ignore` on the previous line, and then you can format this array like the following:

```
// prettier-ignore
const cells = Uint8Array.from([
  3, 0, 1, 5,
  3, 1, 4, 5,
 ...
```

ocrossi: Sweet, way cleaner thank you!

finetjul:
image

ocrossi:
@finetjul

ocrossi:
@finetjul

Sources/Widgets/Representations/ConeHandleRepresentation/index.js
```diff
@@ -6,11 +6,14 @@ import vtkPixelSpaceCallbackMapper from 'vtk.js/Sources/Rendering/Core/PixelSpac
 import vtkActor from 'vtk.js/Sources/Rendering/Core/Actor';
```
finetjul, line 6: is ConeHandleRepresentation really needed ? Can't it be merged into ArrowHandleRepresentation ?

Sources/Widgets/Representations/ConeHandleRepresentation/index.js
```diff
@@ -6,11 +6,14 @@ import vtkPixelSpaceCallbackMapper from 'vtk.js/Sources/Rendering/Core/PixelSpac
...
 import { ScalarMode } from 'vtk.js/Sources/Rendering/Core/Mapper/Constants';
+import { RenderingTypes } from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
 
 function vtkConeHandleRepresentation(publicAPI, model) {
   model.classHierarchy.push('vtkConeHandleRepresentation');
 
+  const superClass = { ...publicAPI };
   // --------------------------------------------------------------------------
   // internal polydata dataset
   // --------------------------------------------------------------------------
```
finetjul, line 19: Add comments (line 60) to explain what those Display Actors/Mappers are. Please do the same in the other related representations

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -152,6 +173,28 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+      const displayRotation = mat3.create();
+      mat3.fromMat4(displayRotation, displayMatrix);
+      const rotation = mat3.create();
+      mat3.multiply(rotation, viewMatrix, displayRotation);
+      vec3.transformMat3(scale3, scale3, rotation);
+
+      typedArray.direction.set(rotation, 9 * i);
       typedArray.scale[i] =
         scaleFactor *
         (!state.isVisible || state.isVisible() ? 1 : 0) *
```
finetjul, line 200: What is the difference between `state.isVisible` and `handleVisibility` ?

finetjul, line 200: No need to take care of "visibility" in requestData, just in `updateActorVisibility` should be enough

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -168,6 +211,29 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+    handleVisible = false
+  ) => {
+    superClass.updateActorVisibility(
+      renderingType,
+      widgetVisible,
+      ctxVisible,
+      handleVisible
+    );
+    if (model.fromLineWidget) {
+      const visibility = model.handleVisibility;
```
finetjul, line 228: use the existing behavior of `visibilityFlagArray` (see WidgetRepresentation)

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -62,6 +60,12 @@ function vtkLineWidget(publicAPI, model) {
 
   function initializeHandleRepresentations() {
     HandleRepresentation[0] = shapeToRepresentation[model.handle1Shape];
```
finetjul, line 62: ```HandleRepresentation[0] = shapeToRepresentation[model.handle1Shape] || vtkSphereHandleRepresentation;```

finetjul: HandleRepresentation is a constant, do not modify it.
Create a mapping function instead of a variable ?

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -100,6 +104,10 @@ function vtkLineWidget(publicAPI, model) {
               scaleInPixels: true,
               /* to detect arrow type in ArrowHandleRepresentation, mandatory */
               handleType: model.handle1Shape,
+              /* To detect in handleRepresentation the origin of LineWidget, mandatory */
+              fromLineWidget: true,
+              handleVisibility: model.handle1Visibility,
+              handleCameraOrientation: model.handle1CameraOrientation,
```
finetjul, line 110: handleCameraOrientation -> faceCamera

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -100,6 +104,10 @@ function vtkLineWidget(publicAPI, model) {
               scaleInPixels: true,
               /* to detect arrow type in ArrowHandleRepresentation, mandatory */
               handleType: model.handle1Shape,
+              /* To detect in handleRepresentation the origin of LineWidget, mandatory */
+              fromLineWidget: true,
```
finetjul, line 108: should not be needed

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -108,15 +116,20 @@ function vtkLineWidget(publicAPI, model) {
...
+              handleVisibility: model.handle2Visibility,
+              handleCameraOrientation: model.handle2CameraOrientation,
             },
           },
           {
             builder: vtkSVGLandmarkRepresentation,
             initialValues: {
               showCircle: false,
               isVisible: false,
+              fromLineWidget: true,
```
finetjul, line 132: fromLineWidget -> offsetText (or a name that explains what it does )

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -108,15 +116,20 @@ function vtkLineWidget(publicAPI, model) {
...
+              handleVisibility: model.handle2Visibility,
+              handleCameraOrientation: model.handle2CameraOrientation,
             },
           },
           {
             builder: vtkSVGLandmarkRepresentation,
             initialValues: {
               showCircle: false,
               isVisible: false,
```
finetjul, line 131: is it really useful to set that variable?  I don't see where it is being used

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -108,15 +116,20 @@ function vtkLineWidget(publicAPI, model) {
...
+              fromLineWidget: true,
             },
             labels: ['SVGtext'],
           },
```
finetjul, line 135: Where is 'moveHandle' defined ?

finetjul: Btw, there is no handle displayed before placing the 2nd handle

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -132,39 +145,20 @@ function vtkLineWidget(publicAPI, model) {
   // --- Public methods -------------------------------------------------------
 
   publicAPI.getDistance = () => {
-    const nbHandles =
-      model.widgetState.getHandle1List().length +
-      model.widgetState.getHandle2List().length;
-    if (nbHandles !== 2) {
+    const nbHandles = model.widgetState.getNbHandles();
```
finetjul, line 148: in the example, distance is still not refreshed before placing 2nd point.

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -132,39 +145,20 @@ function vtkLineWidget(publicAPI, model) {
...
-  };
-
-  publicAPI.updateHandleFromUI = (input, handleId) => {
-    if (handleId === 1) {
-      model.handle1Shape = input;
-    } else if (handleId === 2) {
-      model.handle2Shape = input;
-    }
+  publicAPI.setHandleShape = (handleId, shape) => {
+    model[`handle${handleId}Shape`] = shape;
```
finetjul, line 161: Use an array instead:
model.handle${handleId}Shape -> model.shapes[handleId] = shape

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -132,39 +145,20 @@ function vtkLineWidget(publicAPI, model) {
...
     initializeHandleRepresentations();
     publicAPI.getRepresentationsForViewType(0);
```
finetjul, line 163: Why do you need it ? I do not see how a "getter" can change/set anything

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -210,8 +209,13 @@ export function extend(publicAPI, model, initialValues = {}) {
     'manipulator',
     'handle1Shape',
     'handle2Shape',
+    'handle1Visibility',
+    'handle2Visibility',
+    'handle1CameraOrientation',
+    'handle2CameraOrientation',
     'positionOnLine',
```
finetjul, line 216: Can you please explain when a variable should go into `model.abc` and when it should go to `state.abc` ?

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -122,44 +228,23 @@ export default function widgetBehavior(publicAPI, model) {
...
-      SVGLayerText.setOrigin(
-        calculateTextPosition(model, model.widgetState.getPositionOnLine())
-      );
       calcTextPosWithLineAngle();
       moveHandle.setVisible(true);
     } else {
-      isDragging = true;
+      model.widgetState.setIsDragging(true);
       model.openGLRenderWindow.setCursor('grabbing');
       model.interactor.requestAnimation(publicAPI);
```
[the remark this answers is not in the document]
ocrossi: the isDragging variable is meant to describe the user moving a fully initialized widget. GrabFocus is called when the user initializes the widget. This way the grabFocus method remains as concise as it could be

Sources/Widgets/Representations/ConeHandleRepresentation/index.js
ocrossi: The coneHandle is a 3d Glyph. It can be merged with arrowHandle but would break the 3d/2d distinction in arrowHandle vs other glyph/handle representations

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
ocrossi: state.isVisible refers to the variable from the widget state tree. handleVisibility comes from abstractWidget and is used as a general purpose flag to manage visibility across all widgets

ocrossi: this code is not from me (this mecanism is inherited from spherHandleRep the blueprint i used to create this file). The issue here is to link the visibility of the points directly to the state of the widget (to my mind). It allows the widget state to override every other visibility change (e.g handleVisibility from abstractWidget) by operating directly on the scale of an object.

Sources/Widgets/Widgets3D/LineWidget/index.js
ocrossi: all of these variables operate on the handleRepresentation level

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
ocrossi: incredible! makes the whole code way more readable and avoids to overload updateActorVisibility!

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
finetjul: Does it still work without this line ?

Sources/Widgets/Widgets3D/LineWidget/index.js
finetjul: I do not see where `isVisible` is being used

Sources/Widgets/Widgets3D/LineWidget/index.js
ocrossi: this function is used in example to change handles from UI.

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
ocrossi: It doesn't. This isVisible variable is used by the state tree to display handles when user adds it to the scene. Since I opted for static handles, the process is to generate a widget when the VTK scene loads, hide the handles to the VTK scene, so we have handles loaded but not positionned yet, and then display them to the scene (here allow them to have a scale different from 0) when the user grabs focus and clicks on the scene. The problem with this line is that it is common to all the handleRepresentation used by lineWidget. It wouldn't make sense to remove it here and leave it in sphereHandle IMO. The last part however would most likely be a breaking change for other widgets.

Sources/Widgets/Widgets3D/LineWidget/index.js
ocrossi: I can understand the point of not modifying constant variable. What i don't understand is the point of using a mapping function instead.  Plus i think i messed it up by putting HandleRepresentation in constants. I think adding the handleRepresentation tab in the scope of this object would be better.

Sources/Widgets/Widgets3D/LineWidget/index.js
ocrossi: The state.abc variables can be accessed easily within the behavior of a widget, and be shared to all the parts of the widget. The model.abc variables are set when the widget initializes and are difficult to use from the widgetProp. So for me a variable should be in the initial values if its value is needed on initialization, and a it should be placed in the state tree if it changes in the lifetime of the widget. For instance positionOnLine should be in default values since it is needed on initialization, and  it should also be in state tree since the user can interact with it to change the text position

Sources/Widgets/Widgets3D/LineWidget/index.js
ocrossi: Why do i need an array here? Is it for readability or performances purposes? Shouldn't I use a setter function instead?

Sources/Widgets/Widgets3D/LineWidget/index.js
finetjul: Thank you.

I think variables should not be both in LineWidget state AND in LineWidget model.
My understanding is LineWidget state for "display"/graphical properties and LineWidget model for "behavioral" properties.

So I believe you can clean those up.

Sources/Widgets/Widgets3D/LineWidget/index.js
finetjul: Truth is, if I comment it, it still works in the example. So my understanding is it should be removed. Do not keep useless function

Sources/Widgets/Widgets3D/LineWidget/index.js
finetjul: for readability. Performance is rarely a good reason in JS.

Btw, I think this should be stored in the widget state instead of the model.

Sources/Widgets/Representations/ConeHandleRepresentation/index.js
finetjul: I feel that a ConeHandleRepresentation will not be used by any other widget. But only as an "arrow" handle. 
Eventually we could create a "3D arrow handle representation" but is there a need ? Can't ArrowHandleRepresentation support 3D shapes ?

Sources/Widgets/Widgets3D/LineWidget/index.js
ocrossi: WDYM by "behaviorial properties"? Properties not prone to modification?

finetjul: behaviorial properties: properties that impacts the interaction with the widget.
E.g. `scaleOnRightClick`, a boolean property could carry the information to change the scale of a handle when right clicking a handle.

Sources/Widgets/Widgets3D/LineWidget/index.js
ocrossi: This was linked to the state tree

Sources/Widgets/Widgets3D/LineWidget/index.js
ocrossi: so basically properties that would change somehow the state tree? Sorry but i still don't get it. A behavioral property would be a variable meant to indicate whether the state should change from one behavior to an other?

ocrossi:
@finetjul

Sources/Widgets/Widgets3D/LineWidget/index.js
finetjul: If a property is used only by behavior.js then it is a behavioral property. 
If a property is used by a representation then it is a display property.
`handle1Visibility, handle2Visibility, handle1CameraOrientation, handle2CameraOrientation, positionOnLine` are therefore state properties and should not be added in `LineWidget/index.js`

Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js
```diff
@@ -14,48 +14,64 @@ function vtkSVGLandmarkRepresentation(publicAPI, model) {
...
+          }
+          const splitText = texts[i].split('\n');
+          const newlineOffset =
+            model.fontProperties != null && model.fontProperties.fontSize
+              ? model.fontProperties.fontSize
+              : 15;
+          splitText.forEach((subText, j) => {
+            const text = createSvgElement('text');
+            Object.keys(model.textProps || {}).forEach((prop) => {
+              if (model.fromLineWidget === true && prop === 'dy') {
```
finetjul, line 54: I remember in a comment to rename `fromLineWidget`.
That variable should not define the context where SVGLandmarkRepresentation is instantiated from. Instead it should be explicit on the behavior to expect:
e.g. `newLines`.

I do not see a case where we do not want that variable to be set to True, do you ?

Sources/Widgets/Widgets3D/LineWidget/Constants.js
```diff
@@ -13,7 +13,8 @@ export const HandleRepresentationType = {
   SPHERE: 'sphere',
   CUBE: 'cube',
   CONE: 'cone',
-  NONE: 'sphere',
+  // none is a sphere handle always invisible even on mouseover
+  NONE: 'voidSphere',
```
finetjul, line 17: should be the first choice of the list.
'voidSphere' -> 'none'

Sources/Widgets/Widgets3D/LineWidget/Constants.js
```diff
@@ -22,9 +23,15 @@ export const HandleRepresentationType = {
   CIRCLE: 'circle',
 };
 
+export const HandleVisibilityBehavior = {
```
finetjul, line 26: can you please explain why you need an enum for a boolean property ?

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -1,19 +1,30 @@
...
-
 export default function widgetBehavior(publicAPI, model) {
   model.classHierarchy.push('vtkLineWidgetProp');
-  let isDragging = null;
+
+  const moveHandle = model.widgetState.getMoveHandle();
+  const handle1 = model.widgetState.getHandle1();
+  const handle2 = model.widgetState.getHandle2();
+  const handle1Visibility = model.widgetState.getHandle1Visibility();
+  const handle2Visibility = model.widgetState.getHandle2Visibility();
```
finetjul, line 24: I believe it should instead be `model.widgetState.getHandle1().isVisible()`

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -26,70 +37,79 @@ export default function widgetBehavior(publicAPI, model) {
...
-    } else {
-      const modifier = bv === 1 ? 1 : -1;
-      bv -= 1;
-      const handle1Pos = model.widgetState.getHandle1().getOrigin();
-      const handle2Pos = model.widgetState.getHandle2().getOrigin();
-      vtkMath.subtract(handle1Pos, handle2Pos, Direction);
-      vtkMath.multiplyScalar(Direction, modifier);
-    }
-    model.representations[bv].getGlyph().setDirection(Direction);
+  function calcTextPosWithLineAngle() {
```
finetjul, line 77: calcTextPosWithLineAngle -> updateTextPosition()

+ add documentation on this function to explain what's going on

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -26,70 +37,79 @@ export default function widgetBehavior(publicAPI, model) {
...
+  // --------------------------------------------------------------------------
+
   /*
+   * DetectOffsetDirectionForTextPosition
    * check for handle 2 position in comparison to handle 1 position
    * and sets text offset to not overlap on the line representation
    */
-  function calcTextPosWithLineAngle() {
+
+  function detectOffsetDirectionForTextPosition() {
```
finetjul, line 64: detect -> get

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -26,70 +37,79 @@ export default function widgetBehavior(publicAPI, model) {
...
-    model.representations[bv].getGlyph().setDirection(Direction);
+  function calcTextPosWithLineAngle() {
+    const dySign = detectOffsetDirectionForTextPosition();
+    const textPropsCp = { ...model.representations[2].getTextProps() };
+    textPropsCp.dy = dySign * Math.abs(textPropsCp.dy);
+    model.representations[2].setTextProps(textPropsCp);
   }
 
+  publicAPI.setText = (text) => {
+    if (model.widgetState.getText()) {
```
finetjul, line 85: why getText() could evaluate to False ?

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -26,70 +37,79 @@ export default function widgetBehavior(publicAPI, model) {
...
+    model.interactor.render();
+  };
+
+  publicAPI.updateTextPosition = (positionOnLine) => {
+    model.widgetState.setPositionOnLine(positionOnLine);
+    updateTextPosition(model, positionOnLine);
+    model.interactor.render();
+  };
+
+  publicAPI.setPositionOnLine = macro.chain(
```
finetjul, line 97: who/where calls `setPositionOnLine` ?

Maybe you should instead observe when `widgetState.PositionOnLine` is modified

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -99,7 +119,54 @@ export default function widgetBehavior(publicAPI, model) {
     );
   }
 
-  // set in public to update handle  Direction when handle change in UI
+  function orientFirstHandleBeforeSecondHandlePlacement(callData) {
+    const handle1Pos = model.widgetState.getHandle1().getOrigin();
+    const WorldMousePos = publicAPI.computeWorldToDisplay(
```
finetjul, line 124: WorldMousePos  -> worldMousePos

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -99,7 +119,54 @@ export default function widgetBehavior(publicAPI, model) {
     );
   }
 
-  // set in public to update handle  Direction when handle change in UI
+  function orientFirstHandleBeforeSecondHandlePlacement(callData) {
+    const handle1Pos = model.widgetState.getHandle1().getOrigin();
```
finetjul, line 123: handle1Pos -> handle1Origin

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -99,7 +119,54 @@ export default function widgetBehavior(publicAPI, model) {
...
+    const mousePos = publicAPI.computeDisplayToWorld(
+      model.renderer,
+      callData.position.x,
+      callData.position.y,
+      WorldMousePos[2]
+    );
+    vtkMath.subtract(
+      model.widgetState.getHandle1().getOrigin(),
+      mousePos,
+      Direction
```
finetjul, line 139: Do not modify a Constant. That will not go well if you have multiple widgets.

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -99,7 +119,54 @@ export default function widgetBehavior(publicAPI, model) {
...
+    );
+  }
+
+  function orientHandlesWithCompleteWidget(modifier) {
+    const handle1Pos = model.widgetState.getHandle1().getOrigin();
+    const handle2Pos = model.widgetState.getHandle2().getOrigin();
+    vtkMath.subtract(handle1Pos, handle2Pos, Direction);
+    vtkMath.multiplyScalar(Direction, modifier);
```
finetjul, line 147: you need a function that computes a direction between 2 points. one function that calls it with p1 and p2, and one function that calls it with p1 and mouse.

This is called code factorization (to not duplicate code)

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -99,7 +119,54 @@ export default function widgetBehavior(publicAPI, model) {
...
+    }
+    if (
+      (bv === 0 && publicAPI.isHandleGlyph2D(model.handle1Shape)) ||
+      (bv === 1 && publicAPI.isHandleGlyph2D(model.handle2Shape))
+    ) {
+      model.representations[bv].setOrientation(Direction);
+    } else {
+      model.representations[bv].getGlyph().setDirection(Direction);
+    }
+  }
```
finetjul, line 168: Please cleanup (make it more simple) that whole function.

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -109,6 +176,69 @@ export default function widgetBehavior(publicAPI, model) {
     }
   };
 
+  publicAPI.toggleHandleVisibility = (bv) => {
```
finetjul, line 179: please explain what this function is supposed to do. I feel it can be written in 4 lines.

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -109,6 +176,69 @@ export default function widgetBehavior(publicAPI, model) {
...
+    if (bv === HandleVisibilityBehavior.TOGGLE_ON) {
+      if (handle1Visibility === false && handle1.getActive()) {
+        model.representations[0].setHandleVisibility(true);
+        model.handle1ToHide = true;
```
finetjul, line 183: is it used only here ? if so, why add it into model ?

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -109,6 +176,69 @@ export default function widgetBehavior(publicAPI, model) {
...
+      model.handle1ToHide = false;
+      model.representations[0].setHandleVisibility(handle1Visibility);
+    }
+    if (model.handle2ToHide === true) {
+      model.handle2ToHide = false;
+      model.representations[1].setHandleVisibility(handle2Visibility);
+    }
+  };
+
+  publicAPI.isHandleGlyph2D = (handleShape) => {
```
finetjul, line 201: move it next to `isHandleOrientable`

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -109,6 +176,69 @@ export default function widgetBehavior(publicAPI, model) {
...
+    return (
+      handleShape === HandleRepresentationType.ARROWHEAD3 ||
+      handleShape === HandleRepresentationType.ARROWHEAD4 ||
+      handleShape === HandleRepresentationType.ARROWHEAD6 ||
+      handleShape === HandleRepresentationType.STAR
+      // TODO: implement rotation for circleHandle
+    );
+  };
+
+  publicAPI.setRotationHandleToFaceCamera = () => {
```
finetjul, line 211: a set should take a `parameter`

finetjul, line 211: instead of having to call it manually, maybe you should observe when the shape is modified.

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -109,6 +176,69 @@ export default function widgetBehavior(publicAPI, model) {
...
+      publicAPI.isHandleGlyph2D(model.handle2Shape)
+    ) {
+      model.representations[1].setViewMatrix(
+        Array.from(model.camera.getViewMatrix())
+      );
+    }
+  };
+
+  publicAPI.placeHandle = (handle, nb) => {
+    model.widgetState.setNbHandles(nb);
```
finetjul, line 231: what if
``` 
placeHandle(...,1)
placeHandle(...,2)
placeHandle(...,1)
``` 
?
you would think you have only 1 placed handle ?

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -1,19 +1,30 @@
...
-  let isDragging = null;
+
+  const moveHandle = model.widgetState.getMoveHandle();
+  const handle1 = model.widgetState.getHandle1();
+  const handle2 = model.widgetState.getHandle2();
+  const handle1Visibility = model.widgetState.getHandle1Visibility();
+  const handle2Visibility = model.widgetState.getHandle2Visibility();
+  const handle1FaceCamera = publicAPI.getWidgetState().getHandle1FaceCamera();
+  const handle2FaceCamera = publicAPI.getWidgetState().getHandle2FaceCamera();
+  const SVGLayerText = model.widgetState.getText();
```
finetjul, line 27: call `model.widgetState.getText()` each time you need it. Do not store it in this variable

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -99,7 +119,54 @@ export default function widgetBehavior(publicAPI, model) {
...
+      (bv === 0 && publicAPI.isHandleGlyph2D(model.handle1Shape)) ||
+      (bv === 1 && publicAPI.isHandleGlyph2D(model.handle2Shape))
+    ) {
+      model.representations[bv].setOrientation(Direction);
+    } else {
+      model.representations[bv].getGlyph().setDirection(Direction);
+    }
+  }
+
   publicAPI.setHandleDirection = () => {
```
finetjul, line 170: setHandleDirection -> updateHandleDirections()

Sources/Widgets/Widgets3D/LineWidget/Constants.js
```diff
@@ -13,7 +13,8 @@ export const HandleRepresentationType = {
   SPHERE: 'sphere',
```
finetjul, line 13: I do not see the difference between HandleBehavior and widgetState.nbHandles. Don't you have some duplicated information ?

Please review the usefulness of HandleBehavior

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -219,7 +329,7 @@ export default function widgetBehavior(publicAPI, model) {
   // --------------------------------------------------------------------------
 
   publicAPI.handleLeftButtonRelease = () => {
-    if (isDragging && model.pickable) {
+    if (model.widgetState.getIsDragging() && model.pickable) {
```
finetjul, line 332: isDragging does not belong to the widget state. It is a behavior property. It can go into model.isDragging

Sources/Widgets/Widgets3D/LineWidget/state.js
```diff
@@ -45,7 +45,14 @@ export default function generateState() {
...
     })
-    .addField({ name: 'positionOnLine', initialValues: 0 })
-    .addField({ name: 'nbHandles', initialValues: 0 })
+    .addField({ name: 'positionOnLine', initialValue: 0 })
+    .addField({ name: 'nbHandles', initialValue: 0 })
+    .addField({ name: 'isDragging', initialValue: false })
+    .addField({ name: 'handle1Shape', initialValue: 'sphere' })
+    .addField({ name: 'handle2Shape', initialValue: 'sphere' })
+    .addField({ name: 'handle1Visibility', initialValue: true })
+    .addField({ name: 'handle2Visibility', initialValue: true })
```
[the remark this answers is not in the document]
ocrossi: handleVisibility is a behavior to hide handles to the user while still being able to interact with it. There is a isVisible managed by the state tree but this functionnality operates on the size of the handle (if i understood it correctly that is). This means that the handleVisibility from widget state prevents any interaction when set to false. If i understood correctly the handleVisibility would be a good case of behavioral property?

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -236,17 +346,24 @@ export default function widgetBehavior(publicAPI, model) {
       model.widgetManager.enablePicking();
       model.interactor.render();
     }
-    isDragging = false;
+    if (
+      model.widgetState.getIsDragging() === false &&
+      (!model.activeState || !model.activeState.getActive())
+    ) {
+      publicAPI.setRotationHandleToFaceCamera();
```
finetjul, line 353: I do not understand why adding a condition to call it.

Sources/Widgets/Widgets3D/LineWidget/state.js
```diff
@@ -45,7 +45,14 @@ export default function generateState() {
         origin: [0, 0, 0],
       },
     })
-    .addField({ name: 'positionOnLine', initialValues: 0 })
-    .addField({ name: 'nbHandles', initialValues: 0 })
+    .addField({ name: 'positionOnLine', initialValue: 0 })
+    .addField({ name: 'nbHandles', initialValue: 0 })
+    .addField({ name: 'isDragging', initialValue: false })
```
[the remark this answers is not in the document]
ocrossi: isDragging and nbHandles are in the state tree to be used outside of the scope of the behavior function of VTK.js.
positionOnLine is in state tree to allow user to interact with it in the UI of lineWidget example.

Sources/Widgets/Representations/CubeHandleRepresentation/index.js
```diff
@@ -41,6 +44,13 @@ function vtkCubeHandleRepresentation(publicAPI, model) {
   // Generic rendering pipeline
   // --------------------------------------------------------------------------
 
+  model.displayMapper = vtkPixelSpaceCallbackMapper.newInstance();
+  model.displayActor = vtkActor.newInstance();
+  model.displayActor.setMapper(model.displayMapper);
+  model.displayMapper.setInputConnection(publicAPI.getOutputPort());
+  publicAPI.addActor(model.displayActor);
+  model.alwaysVisibleActors = [model.displayActor];
```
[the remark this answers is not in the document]
ocrossi: Every handle used by the LineWidget should have a displayActor/displayMapper to allow the grab when not visible feature.
Should i create a new MR for this?

Sources/Widgets/Widgets3D/LineWidget/state.js
```diff
@@ -45,7 +45,14 @@ export default function generateState() {
...
+    .addField({ name: 'handle1Shape', initialValue: 'sphere' })
+    .addField({ name: 'handle2Shape', initialValue: 'sphere' })
```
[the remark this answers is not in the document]
ocrossi: related to the shapeMixin upgrade

Sources/Widgets/Widgets3D/LineWidget/behavior.js
ocrossi: this function is called in the UI

Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js
ocrossi: for the normal behavior of SVGLandmarkRepresentation as it is in PolyLineWidget example the variable should not be evaluated to true

Sources/Widgets/Widgets3D/LineWidget/behavior.js
finetjul: Problem is, some public functions are declared in index.js and some in behavior.js.
It is not very developer friendly.

Sources/Widgets/Widgets3D/LineWidget/state.js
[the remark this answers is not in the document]
ocrossi: Related to shapeMixin upgrade

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -168,6 +224,19 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+    widgetVisible = true,
+    ctxVisible = true
+  ) => {
+    superClass.updateActorVisibility(
+      renderingType,
+      widgetVisible,
+      ctxVisible,
+      model.handleVisibility
+    );
+  };
```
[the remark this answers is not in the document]
ocrossi: Related to shapeMixin upgrade

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -176,6 +245,9 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
 
 const DEFAULT_VALUES = {
   defaultScale: 1,
+  orientation: [0, 0, 0],
+  viewMatrix: mat4.create(),
+  handleVisibility: true,
```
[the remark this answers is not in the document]
ocrossi: Related to shapeMixin upgrade

Sources/Widgets/Widgets3D/LineWidget/behavior.js
ocrossi: Related to shapeMixin upgrade

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -176,6 +245,9 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
 
 const DEFAULT_VALUES = {
   defaultScale: 1,
+  orientation: [0, 0, 0],
+  viewMatrix: mat4.create(),
```
jourdain, line 249: You should not have that `mat4.create()` inside the defaultValues. You should instead put something like below in the extend function.

```
model.viewMatrix = new Float64Array(16);
```

And just leave that line in defaults but commented out.

ocrossi: why is it not recommanded to do so?

jourdain: because the same typed array will be shared across all widget instances

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -128,60 +136,59 @@ export default function widgetBehavior(publicAPI, model) {
...
   }
 
-  function orientHandlesWithCompleteWidget(modifier) {
+  function orientFirstHandleBeforeSecondHandlePlacement(callData) {
     const handle1Origin = model.widgetState.getHandle1().getOrigin();
-    const handle2Origin = model.widgetState.getHandle2().getOrigin();
-    vtkMath.subtract(handle1Origin, handle2Origin, direction);
-    vtkMath.multiplyScalar(direction, modifier);
+    const mousePos = computeMousePosition(handle1Origin, callData);
+    direction = computeDirectionWithTwoPoints(handle1Origin, mousePos);
```
finetjul, line 158: return the return value of "computeDirectionWithTwoPoints" and use the returned value instead of using "global" variable" `direction`

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -192,32 +199,32 @@ export default function widgetBehavior(publicAPI, model) {
         handle1.getActive()
       ) {
         model.representations[0].setHandleVisibility(true);
-        model.handle1ToHide = true;
+        handleToHide[0] = true;
```
finetjul, line 202: declare that variable locally in `toggleHandleVisibility`

ocrossi: the issue about this is that i will loose track of handles to hide when i exit the scope of this function

finetjul: We will see the usefulness of this variable once the work on the mixin is done

finetjul:
Superseded by #1757

PR #1651, opened 2020-12-09
fix(JSZipDataAccessHelper): Downgrade JSZip version

boucaud:
The latest versions of JSZip seem to have broken the JSZipDataAccessHelper.

Calling JSRoot.file().async() returns a promise that never resolves. This breaks .vtkjs loading functionality.

People having the same issue:

https://github.com/Stuk/jszip/issues/724

https://github.com/Stuk/jszip/issues/712

Downgrading to 3.2.0 fixes the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "blueimp-md5": "2.18.0",
     "commander": "6.2.0",
     "gl-matrix": "3.3.0",
-    "jszip": "3.5.0",
+    "jszip": "3.2.0",
     "pako": "1.0.11",
     "seedrandom": "3.0.5",
     "shelljs": "0.8.4",
--- a/Utilities/ExampleRunner/template-config.js
+++ b/Utilities/ExampleRunner/template-config.js
@@ -1,5 +1,6 @@
-var path = require('path');
-var vtkBasePath = path.resolve('.');
+const path = require('path');
+
+const vtkBasePath = path.resolve('.');
 
 const settings = require('../../webpack.settings.js');
 
@@ -47,6 +48,7 @@ module.exports = {
     },
     fallback: {
       fs: false,
+      stream: require.resolve('stream-browserify')
     },
   },
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -14447,9 +14447,9 @@
       }
     },
     "jszip": {
-      "version": "3.5.0",
-      "resolved": "https://registry.npmjs.org/jszip/-/jszip-3.5.0.tgz",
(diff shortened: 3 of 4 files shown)


boucaud:
@floryst What do you think ?

@finetjul @laurentmalka fyi.

jourdain:
Do we need `stream` as an added dependency?

boucaud:
> Do we need `stream` as an added dependency?

Without it we get this error:

```
ERROR in ../node_modules/jszip/lib/readable-stream-browser.js 9:0-34
Module not found: Error: Can't resolve 'stream' in '/work/repo/total/vtk-js/node_modules/jszip/lib'
```

floryst:
That should have been covered by the `stream-browserify` dependency I added, but I guess webpack isn't picking up on it. I did a quick test and the following `webpack.common.js` patch should work without having to pull in the extra stream dependency.

```diff
diff --git a/webpack.common.js b/webpack.common.js
index 44dd277ee5..5b2dd6aa8c 100644
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -97,6 +97,7 @@ const baseConfig = {
     alias: {
       'vtk.js': __dirname,
     },
+    fallback: { "stream": require.resolve("stream-browserify") },
   },
   module: {
     rules: configureVtkRules(),
@@ -148,6 +149,7 @@ const liteConfig = merge(
           'Sources/Rendering/Core/ColorTransferFunction/LiteColorMaps.json'
         ),
       },
+      fallback: { "stream": require.resolve("stream-browserify") },
     },
   },
   baseConfig,

```

package.json
```diff
@@ -32,10 +32,11 @@
     "blueimp-md5": "2.18.0",
     "commander": "6.2.0",
     "gl-matrix": "3.3.0",
-    "jszip": "3.5.0",
+    "jszip": "3.2.0",
     "pako": "1.0.11",
     "seedrandom": "3.0.5",
     "shelljs": "0.8.4",
+    "stream": "0.0.2",
```
floryst, line 39: See if https://github.com/Kitware/vtk-js/pull/1651#issuecomment-741909256 allows us to remove this stream dependency.

floryst (changes requested):

package.json
boucaud: I tried this, and also tried with `readable-stream`, it doesn't seem to help.

boucaud: Note: I am testing with `npm run example -- SceneExplorer` because it uses the JSZipDataAccessHelper. `npm run build` and other examples do seem to work. No luck with that example though.

boucaud: Fixed it. I also had to change the example template.

floryst (approved):

PR #1656, opened 2020-12-11
fix(VolumeMapper): Avoid unnecessary recreation of the shader when the actor changes

swederik:
Hi everyone,

We've been evaluating the rendering performance in some soon-to-be-public work on top of VTK.js. One of the odd things that came up is that inside the ShaderCache, the md5 hash of the combined vertex, fragment, and geometry shaders is used as the key to avoid the creation of duplicate programs.

That all sounds fine in theory, but it appears that for things like the VolumeMapper, this md5 hash call was very slow, upwards of 5 milliseconds. We are using multiple VTK.js viewports and noticed that these md5 calls were over 50% of the entire CPU time during the rendering. With this change we have dramatically improved performance in our examples (7 viewports synced with volume mappers), but I think it will improve performance in general.

So we thought about maybe replacing the md5 hash with some faster hashing algorithm, but then realized we can just skip it entirely and use the combined string as the key for the lookup, and then the 'does this exist' check `Object.keys(model.shaderPrograms).indexOf(result)` happens very quickly.

I realize that storing these huge strings (1400+ characters) in the object key feels a bit ugly, but I can't find a downside other than it being slightly weird during debugging. I notice that ShaderProgram has a setter for md5Hash as well, but it doesn't appear to be used anywhere.

What do you think? Is there some other use for the hash that we are missing?

Thanks!

### DIFF
--- a/Sources/Rendering/OpenGL/VolumeMapper/index.js
+++ b/Sources/Rendering/OpenGL/VolumeMapper/index.js
@@ -26,6 +26,19 @@ import vtkVolumeFS from 'vtk.js/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl';
 
 const { vtkWarningMacro, vtkErrorMacro } = macro;
 
+// TODO: Do we want this in some shared utility? Shouldwe just use lodash.isEqual
+function arrayEquals(a, b) {
+  if (a.length !== b.length) {
+    return false;
+  }
+  for (let i = 0; i < a.length; ++i) {
+    if (a[i] !== b[i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
 // ----------------------------------------------------------------------------
 // vtkOpenGLVolumeMapper methods
 // ----------------------------------------------------------------------------
@@ -376,6 +389,74 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
       needRebuild = true;
     }
 
+    const numComp = model.scalarTexture.getComponents();
+    const iComps = actor.getProperty().getIndependentComponents();
+    let usesProportionalComponents = false;
+    const proportionalComponents = [];
+    if (iComps) {
+      // Define any proportional components
+      for (let nc = 0; nc < numComp; nc++) {
+        proportionalComponents.push(actor.getProperty().getOpacityMode(nc));
+      }
+
+      if (proportionalComponents.length > 0) {
+        usesProportionalComponents = true;
+      }
+    }
+
+    const ext = model.currentInput.getExtent();
+    const spc = model.currentInput.getSpacing();
+    const vsize = vec3.create();
+    vec3.set(
+      vsize,
+      (ext[1] - ext[0]) * spc[0],
+      (ext[3] - ext[2]) * spc[1],
+      (ext[5] - ext[4]) * spc[2]
+    );
+
+    const maxSamples =
+      vec3.length(vsize) / model.renderable.getSampleDistance();
+
+    const state = {
+      interpolationType: actor.getProperty().getInterpolationType(),
+      useLabelOutline: actor.getProperty().getUseLabelOutline(),
+      numComp,
+      usesProportionalComponents,
+      iComps,
+      maxSamples,
+      useGradientOpacity: actor.getProperty().getUseGradientOpacity(0),
+      blendMode: model.renderable.getBlendMode(),
+      averageIPScalarMode: model.renderable.getAverageIPScalarRange(),
+      proportionalComponents,
+    };
+
+    // We only need to rebuild the shader if one of these variables has changed,
+    // since they are used in the shader template replacement step.
+    if (
+      !model.previousState ||
+      model.previousState.interpolationType !== state.interpolationType ||
+      model.previousState.useLabelOutline !== state.useLabelOutline ||
+      model.previousState.numComp !== state.numComp ||
+      model.previousState.usesProportionalComponents !==
+        state.usesProportionalComponents ||
+      model.previousState.iComps !== state.iComps ||
+      model.previousState.maxSamples !== state.maxSamples ||
+      model.previousState.useGradientOpacity !== state.useGradientOpacity ||
+      model.previousState.blendMode !== state.blendMode ||
+      !arrayEquals(
+        model.previousState.averageIPScalarMode,
+        state.averageIPScalarMode
+      ) ||
+      !arrayEquals(
+        model.previousState.proportionalComponents,
+        state.proportionalComponents
+      )
+    ) {
+      model.previousState = { ...state };
+
+      return true;
+    }
+
     // has something changed that would require us to recreate the shader?
     if (
       cellBO.getProgram() === 0 ||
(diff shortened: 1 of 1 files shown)


jourdain:
I like the idea and what about the following code instead?

```
publicAPI.getShaderProgram = (vertexCode, fragmentCode, geometryCode) => {
  const key = `${vertexCode}${fragmentCode}${geometryCode}`;
  let sps = model.shaderPrograms[key];

  if (!sps) {
    sps = vtkShaderProgram.newInstance();
    sps.setContext(model.context);
    sps.getVertexShader().setSource(vertexCode);
    sps.getFragmentShader().setSource(fragmentCode);
    if (geometryCode) {
      sps.getGeometryShader().setSource(geometryCode);
    }
    model.shaderPrograms[key] = sps;
  }

  return sps;
};
```

jourdain:
FYI @martinken

martinken:
The idea behind the md5 is that it should be rarely computed. I don;t recall in vtk.js if that is the case but in C++ it works like this.

- Check some values to see if we need to rebuild the shader code
- if we need to rebuild the shader code (should be rare) then
-- rebuild the shader strings, compute the MD5 call the shader cache
- else just check the md5 which is very fast

I suspect the problem is that vtk.js is being overly aggresive in thinking it needs to recompute the shader.

getNeedToRebuildShaders should really only return true if something has changed that requires the shaders to be rebuilt. My quick look at the code seems like it is being overly aggressive. I'm not sure in your example what is causing this to be true, but that is probably what we should fix. If we fix that it will be even faster as the shader strings will not even be recomputed which is also expensive.

The problem with using the shader string as a key is that for 1000's of mappers all using the same shader this change will cause them all to do a full string compare of the entire string for every mapper for every frame. 

The mess in the c++ code (which has more features so vtk.js should be simpler)  Note that the actor mtime is not checked against. Basically it is just more specific that what we have in vtk.js.

```
  // shape of input data changed?
  float factor, offset;
  this->GetCoincidentParameters(ren, actor, factor, offset);
  unsigned int scv = (this->CurrentInput->GetPointData()->GetNormals() ? 0x01 : 0) +
    (this->HaveCellScalars ? 0x02 : 0) + (this->HaveCellNormals ? 0x04 : 0) +
    ((cam->GetParallelProjection() != 0.0) ? 0x08 : 0) + ((offset != 0.0) ? 0x10 : 0) +
    (this->VBOs->GetNumberOfComponents("scalarColor") ? 0x20 : 0) +
    (vtkOpenGLRenderer::SafeDownCast(ren)->GetUseSphericalHarmonics() ? 0x40 : 0) +
    (actor->GetProperty()->GetCoatStrength() > 0.0 ? 0x80 : 0) +
    (actor->GetProperty()->GetAnisotropy() > 0.0 ? 0x100 : 0) +
    ((this->VBOs->GetNumberOfComponents("tcoord") % 4) << 9);

  if (cellBO.Program == nullptr || cellBO.ShaderSourceTime < this->GetMTime() ||
    cellBO.ShaderSourceTime < actor->GetProperty()->GetMTime() ||
    cellBO.ShaderSourceTime < actor->GetShaderProperty()->GetShaderMTime() ||
    cellBO.ShaderSourceTime < this->LightComplexityChanged[&cellBO] ||
    cellBO.ShaderSourceTime < this->SelectionStateChanged ||
    cellBO.ShaderSourceTime < renderPassMTime || cellBO.ShaderChangeValue != scv)
  {
    cellBO.ShaderChangeValue = scv;
    return true;
  }
```

swederik:
Thanks! Yes I ended up digging down that path too since I filed this. Ken,
you are right, removing the actor mTime check and making things more
explicit (only checking fields that mutate the shader e.g. num components)
improved the performance a lot since it doesn't get to this path as often.
I'll update the PR with that when I get a chance. The only issue with that
is that it needs to be updated for each different mapper. I haven't checked
if this issue is also present for other mappers. In our case actor mTime is
changing due to a lookup table change (window/level), so using it for the
shader rebuild check doesn't make sense.

swederik:
Okay I've updated the PR to gather the variables that are used it the shader template replacement step and check them against previous settings. That seems like a more explicit way to determine if we need to rebuild the shader. I've also removed the actor Mtime check.

The easiest way to test it is to load up the VolumeMapper example:

```
npm run example -- VolumeMapper
```

and paste something like this into the browser console:

```javascript
actor.getProperty().getScalarOpacity(0).setRange(0,1)
```

The next time you try to move the volume (e.g. rotate it a bit), you will hit the `getNeedToRebuildShaders` check, which should now return false. Prior to this PR, it would return true, and we would be rebuilding the shader for all actor modifications.

Let me know if you want me to change anything. I wasn't sure what you wanted to do with arrayEquals, so I put it at the top of the file for now.

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -334,14 +347,81 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
     // has something changed that would require us to recreate the shader?
     if (
       cellBO.getProgram() === 0 ||
       needRebuild ||
       model.lastHaveSeenDepthRequest !== model.haveSeenDepthRequest ||
       !!model.lastZBufferTexture !== !!model.zBufferTexture ||
       cellBO.getShaderSourceTime().getMTime() < publicAPI.getMTime() ||
-      cellBO.getShaderSourceTime().getMTime() < actor.getMTime() ||
       cellBO.getShaderSourceTime().getMTime() < model.renderable.getMTime() ||
       cellBO.getShaderSourceTime().getMTime() < model.currentInput.getMTime()
```
swederik, line 426: Maybe we can also remove this line?

thewtex:
Is this ready for rebase / merge?

Sources/Rendering/OpenGL/VolumeMapper/index.js
swederik: Removed it, resolving this

swederik:
Sorry been swamped with other stuff. I've removed the currentInput mTime check and rebased on master / squashed the commits. It should be fine to re-review now.

Thanks!

floryst:
@martinken please take another look :)

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -376,16 +389,82 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
+      model.previousState.blendMode !== state.blendMode ||
+      !arrayEquals(
+        model.previousState.averageIPScalarMode,
+        state.averageIPScalarMode
+      ) ||
+      !arrayEquals(
+        model.previousState.proportionalComponents,
+        state.proportionalComponents
+      )
+    ) {
```
martinken, line 454: hey @jourdain @swederik Can the above block of code be simplified as something like 
```
if (model.previousState === state)
```

My knowledge of JS is rough but it seems like it could be a simple object to object comparison maybe?

martinken: or @floryst , really anyone with more of a clue than me :-)

jourdain: Unfortunately `model.previousState === state` won't be enough as it will always be false (not same ref between calls).

But if everything is serializable we could do `JSON.stringify(model.previousState) !== JSON.stringify(state)`

martinken:
LGTM pending rebase etc +1

swederik:
Looks like it can be squashed and merged without me needing to rebase. Thanks for the reviews @jourdain and @martinken!

PR #1660, opened 2020-12-15
fix(colortransferfunction): fix discretized CTF returning incorrect values

boucaud:
A discretized ColorTransferFunction would return incorrect values (0, 0, 0) when the discretized `x` was exactly equal to the beginning of the mapping range.

This fixes the issue.

### DIFF
--- a/Sources/Rendering/Core/ColorTransferFunction/test/testColorTransferFunction.js
+++ b/Sources/Rendering/Core/ColorTransferFunction/test/testColorTransferFunction.js
@@ -10,6 +10,8 @@ import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
 import vtkDataArray from 'vtk.js/Sources/Common/Core/DataArray';
 import vtkPolyData from 'vtk.js/Sources/Common/DataModel/PolyData';
 
+import { areEquals } from 'vtk.js/Sources/Common/Core/Math';
+
 import baseline from './testColorTransferFunction.png';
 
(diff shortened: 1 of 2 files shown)


boucaud:
@finetjul @jourdain WDYT ?

jourdain:
seems ok

finetjul:
LGTM.

It would be great if a unit test could be added to ensure it won't suffer from regression in future

boucaud:
> It would be great if a unit test could be added to ensure it won't suffer from regression in future

I've added a test.

Sources/Rendering/Core/ColorTransferFunction/test/testColorTransferFunction.js
```diff
@@ -118,3 +120,41 @@ test.onlyIfWebGL('Test Interpolate Scalars Before Colors', (t) => {
...
+    [1, 0, 0],
+    [1, 0, 0],
+    [1, 0, 0],
+    [0.5, 0.5, 0],
+    [0, 1, 0],
+    [0, 0.5, 0.5],
+    [0, 0, 1],
+    [0, 0, 1],
+    [0, 0, 1],
+  ];
```
finetjul, line 147: does this test fail without your fix ? for what index ? 0 or 1 ?

boucaud: It fails for 2 and 3.

PR #1665, opened 2020-12-18
Add HttpDataSetSeriesReader

laurentmalka:


laurentmalka:
@boucaud 

I still need to find a better way to provide timesteps to the setUpdateTimeStep function and I need to document this class, but this should be good enough for you to test

Sources/IO/Core/HttpSceneLoader/index.js
```diff
@@ -301,8 +302,131 @@ function loadHttpDataSetReader(item, model, publicAPI) {
 
 // ----------------------------------------------------------------------------
 
+function loadHttpDataSetSeriesReader(item, model, publicAPI) {
```
boucaud, line 305: Can you try to factorize some of the duplicated code between `loadHttpDataSetReader` and this ?

Sources/IO/Core/HttpDataSetSeriesReader/example/index.js
```diff
@@ -0,0 +1,52 @@
...
+const reader = vtkHttpDataSetSeriesReader.newInstance({ fetchGzip: true });
+reader.setUrl(`${__BASE_PATH__}/data/animation.vtp/0`).then(() => {
+  let index = 0;
+  setInterval(() => {
+    const timeSteps = reader.getTimeSteps();
+    reader.setUpdateTimeStep(timeSteps[index]);
+    index = (index + 1) % timeSteps.length;
+    renderer.resetCamera();
+    renderWindow.render();
+  }, 1000);
```
boucaud, line 32: I would make sense to use the actual timestep duration for this

Sources/IO/Core/HttpDataSetSeriesReader/index.js
```diff
@@ -0,0 +1,214 @@
...
+    const timeSteps = publicAPI.getTimeSteps();
+    const length = timeSteps.length;
+    if (length >= 1) {
+      return [timeSteps[0], timeSteps[length - 1]];
+    }
+    return [];
+  };
+
+  publicAPI.setUpdateTimeStep = (timeStep) => {
+    // TODO find a better way to check for timeSteps
```
boucaud, line 149: @finetjul what do you think would be the best approach here ? snap to the closest timeStep, only accept valid timesteps, use indices instead, something else ?

boucaud (commented):
I think some kind array memoization mechanism in `HttpDataSetReader`, or similar, to avoid reading duplicate data between each timestep would be very useful.

Sources/IO/Core/HttpDataSetSeriesReader/index.js
```diff
@@ -0,0 +1,214 @@
...
+// Global methods
+// ----------------------------------------------------------------------------
+
+function processDataSet(publicAPI, model, dataset, resolve, reject, loadData) {
+  model.readers = [];
+
+  const pendingPromises = [];
+  dataset.series.forEach((timeStep) => {
+    const newReader = HttpDataSetReader.newInstance({
+      fetchGzip: model.fetchGzip,
```
boucaud, line 22: Add `dataAccessHelper: model.dataAccessHelper`

Sources/IO/Core/HttpDataSetSeriesReader/index.js
finetjul: I think the best is to have the same behavior as VTK C++ (which I do not know)

boucaud: VTK C++ seems to leave this choice to the reader implementation.

The UPDATE_TIME_STEP key usually returns a double value, so that leaves out the indices option.

I looked at vtkExtractTimeSteps, and it clamps the input value to the timestep range, then has a configurable option to either use the previous, next or nearest timestep.

For vtkGLTFReader, I think we chose to interpolate, because we could.

I think what makes sense here is to always snap to the previous value. We can do nearest instead, but that seems less consistent.

finetjul: I agree with you, snapping to the previous value seems to make more sense.

finetjul: @laurentmalka  please mark this conversation as resolved if you fixed it in your following work.

Sources/IO/Core/HttpSceneLoader/index.js
```diff
@@ -168,141 +169,146 @@ function initializeVolumeComponents(components) {
...
-    dataSetLODsLoader.setSceneItem(sceneItem);
-    dataSetLODsLoader.setBaseUrl(sourceLODs.baseUrl);
-    dataSetLODsLoader.setFiles(sourceLODs.files);
-
-    if (model.startLODLoaders) {
-      dataSetLODsLoader.startDownloads();
+    mapper.setInputConnection(source.getOutputPort());
+
+    source
+      .setUrl([model.baseURL, item.httpDataSetReader.url].join('/'), {
```
boucaud, line 274: This will cause an error with a httpDataSetSeriesReader

boucaud (changes requested):
We don't need do add the whole animation example (i.e we only need animation.vtp/0, and to move data/ in there). We could also get rid of the color arrays to take less space.

Sources/IO/Core/HttpSceneLoader/index.js
boucaud: `      .setUrl([model.baseURL, item[item.type].url].join('/'), {` ?

boucaud:
LGTM, though I don't know if there's a specific process to add example binaries.

@finetjul @jourdain WDYT ?

jourdain:
If you want to push sample data, you should do it [here](https://github.com/kitware/vtk-js-datasets)

laurentmalka:
This PR requires https://github.com/Kitware/vtk-js-datasets/pull/8 to be merged for the HttpDataSetSeries example to work.

Sources/IO/Core/HttpDataSetSeriesReader/index.js
```diff
@@ -0,0 +1,224 @@
...
+
+      // Remove the file in the URL
+      const path = url.split('/');
+      path.pop();
+      model.baseURL = path.join('/');
+    }
+
+    model.compression = options.compression;
+
+    return publicAPI.updateMetaData(!!options.loadData);
```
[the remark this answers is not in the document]
finetjul: `options.loadData == true` ? (or simply `options.loadData`) ?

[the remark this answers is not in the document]
jourdain: `options.loadData == ...` is bad, `options.loadData` is much better.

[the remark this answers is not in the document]
jourdain: BTW `!!` is a common patten in JS if you really need a bool.

PR #1672, opened 2020-12-20
fix(RenderWindowInteractor): Handle initeraction outside of browser w…

jourdain:
…indow


floryst (approved):
Seems reasonable to me. Not sure if there are cases where this is undesirable. Only thing I can think of involves widget handles being dragged off-canvas, and thus may be unretrievable.

PR #1677, opened 2021-01-02
feat(vtkvolumemapper): Clipping volume by planes

gandis0713:
I have rendered a clipped volume using something like vtkInteractorStyleMPRSlice in the hardware accelerator environment.
Including this way, how about using a clip plane to clip the volume?


martinken:
I'm not 100% sure on the transformations of the world crop plane values into the coordinate system used by the ray bounding code. Been a while since I was in that code. I would suggest (if you haven;t already) doing a quick check with a volume with 1 10 10 spacing and maybe an actor scale of 10 1 10, or similar,  and make sure it all still works as expected. Just seeing idxNormalMatrix makes me think index coordinates while worldtoview is view coordinates. But again, been a while and I didn't dig into it to really track it down.

But as long as it tests out OK for you then it's a +1

martinken:
BTW the general approach and code all looks great!

gandis0713:
> I'm not 100% sure on the transformations of the world crop plane values into the coordinate system used by the ray bounding code. Been a while since I was in that code. I would suggest (if you haven;t already) doing a quick check with a volume with 1 10 10 spacing and maybe an actor scale of 10 1 10, or similar, and make sure it all still works as expected. Just seeing idxNormalMatrix makes me think index coordinates while worldtoview is view coordinates. But again, been a while and I didn't dig into it to really track it down.
> 
> But as long as it tests out OK for you then it's a +1

gandis0713:
> I'm not 100% sure on the transformations of the world crop plane values into the coordinate system used by the ray bounding code. Been a while since I was in that code. I would suggest (if you haven;t already) doing a quick check with a volume with 1 10 10 spacing and maybe an actor scale of 10 1 10, or similar, and make sure it all still works as expected. Just seeing idxNormalMatrix makes me think index coordinates while worldtoview is view coordinates. But again, been a while and I didn't dig into it to really track it down.
> 
> But as long as it tests out OK for you then it's a +1

I have checked your suggestion. If the direction of Volume is not [1, 0, 0, 0, 1, 0, 0, 0, 1], applying idxNormalMatrix to planeNormal does incorrect. So it seems correct to use the camera's NormalMatrix to align the normal orientation.
Thank you so much for your advice.

jourdain:
@martinken is it ready to merge? I'll let @floryst merge it when it will be convenient for him before/after his work on the build system.

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -779,6 +822,40 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
+        const clipPlanePos = clipPlanes[i].getOrigin();
+
+        vec3.transformMat3(
+          clipPlaneNormal,
+          clipPlaneNormal,
+          keyMats.normalMatrix
+        );
+
+        vec3.transformMat4(clipPlanePos, clipPlanePos, model.modelToView);
+
```
martinken, line 845: I believe here model.modelToView should instead be keyMats.wcvc as the clip planes are specified in world coordinates.

martinken:
I think there is one other tiny change that needs to be made and then it is good to go.

Sources/Rendering/OpenGL/VolumeMapper/index.js
gandis0713: It is correct to be transformed to world coordinates. I have reflected what you commented on.

floryst:
@gandis0713 if you can re-base your commits onto the latest vtk.js master branch, then I can merge.

gandis0713:
@floryst I rebased branch into master. Could you please check it?

floryst:
Thanks!

PR #1678, opened 2021-01-04
Remove kw web suite

### LINKED ISSUE #1669: vtk.js doc publish is broken
The parallel build of the examples is not fully done by the time kw-doc tries to generate some web pages and the full process fails. (assumption)

That's what is happening with travis on the master branch. We manage to publish release because we moved that step before the examples build and web site publication.

This issue can be reproduced locally by running `npm run doc:www` or even `npm run doc -- -m` and checking how many example were actually built by running `find ./Documentation/build-tmp/public/examples | grep .js$ | wc -l` which should give you 105...

It seems that the last successful build of the examples are from `2020-10-26`.

@finetjul @floryst

floryst:
This removes the kw-web-suite dependency, as dealing with the one-level indirection was time consuming and unnecessary.

This contains a breaking change.

Other changes:
- fix some eslint complaints
- remove size-limit, since it wasn't working. We have webpack-bundle-analyzer anyways
- add eslint scripts
- update terser config
- Fixes #1669 by pulling in the latest kw-doc.

Things left to do:
- [x] update packages
- [x] squash commits
- [x] Fix kw-doc (blocked on https://github.com/Kitware/kw-doc/pull/11)


.eslintrc.js
```diff
@@ -55,4 +55,8 @@ module.exports = {
     es6: true,
     browser: true,
   },
+  ignorePatterns: [
+    // ignore old Actor example
+    '**/example_/*.js',
+  ],
```
jourdain, line 61: We should probably fix those old example lint wise...

Sources/Interaction/Style/InteractorStyleUnicam/example/index.js
```diff
@@ -83,12 +83,10 @@ function updateWorldUpVec() {
...
-    const coordinates = coordinateElements.map((item) => {
-      return Number.parseFloat(item.value);
-    });
-    const validCoordinates = !coordinates.some((value) => {
-      return Number.isNaN(value);
-    });
+    const coordinates = coordinateElements.map((item) =>
+      Number.parseFloat(item.value)
+    );
+    const validCoordinates = !coordinates.some((value) => Number.isNaN(value));
```
jourdain, line 89: I know it was not your code but... `const validCoordinates = !coordinates.some(Number.isNaN);`

jourdain: FYI @finetjul

package.json
```diff
@@ -41,9 +41,27 @@
...
+    "eslint": "7.17.0",
+    "eslint-config-airbnb": "18.2.1",
+    "eslint-config-prettier": "7.1.0",
+    "eslint-import-resolver-webpack": "0.13.0",
+    "eslint-loader": "4.0.2",
+    "eslint-plugin-import": "2.22.1",
+    "eslint-plugin-jsx-a11y": "6.4.1",
+    "eslint-plugin-prettier": "3.3.0",
+    "eslint-plugin-react": "7.22.0",
+    "hson-loader": "2.0.0",
```
jourdain, line 62: Do we use hson somewhere?

package.json
```diff
@@ -53,34 +71,53 @@
     "karma-tap-pretty-reporter": "4.2.0",
     "karma-webpack": "git://github.com/ryanclark/karma-webpack#eec6dfc",
     "kw-doc": "2.0.0",
-    "kw-web-suite": "11.0.0",
+    "moment": "2.29.1",
```
jourdain, line 74: Do we use moment?

package.json
```diff
@@ -53,34 +71,53 @@
...
     "node-pre-gyp": "0.16.0",
+    "parallel-webpack": "2.6.0",
```
jourdain, line 76: parallel-webpack is used in kw-doc. I'm not sure we need it here?

package.json
```diff
@@ -53,34 +71,53 @@
...
+    "postcss-loader": "4.1.0",
+    "prettier": "2.2.1",
     "process": "0.11.10",
+    "raw-loader": "4.0.2",
     "request": "2.88.2",
     "resemblejs": "3.2.5",
+    "semantic-release": "17.3.1",
+    "shader-loader": "1.3.1",
```
jourdain, line 84: Should we have the worker/shader loaders in plain dependencies rather than dev? So consumer will benefit from them directly or should we document the fact that they need to add those to their project?

package.json
```diff
@@ -53,34 +71,53 @@
...
     "stream-browserify": "3.0.0",
+    "string-replace-loader": "3.0.1",
+    "style-loader": "2.0.0",
     "tap-markdown": "1.2.1",
     "tap-spec": "5.0.0",
     "tape": "5.0.1",
     "tape-catch": "1.0.6",
+    "terser-webpack-plugin": "5.0.3",
```
jourdain, line 92: is it still the way to go to minify es6+?

package.json
```diff
@@ -53,34 +71,53 @@
...
+    "url-loader": "4.1.1",
+    "webpack": "5.11.1",
+    "webpack-bundle-analyzer": "4.3.0",
+    "webpack-cli": "4.3.1",
+    "webpack-dashboard": "3.2.1",
+    "webpack-dev-server": "3.11.1",
+    "webpack-merge": "5.7.3",
+    "webpack-notifier": "1.12.0",
+    "worker-loader": "3.0.7",
```
jourdain, line 101: should `worker-loader` be in the main dependencies? (see previous comment/question)

package.json
```diff
@@ -53,34 +71,53 @@
...
     "wslink": "0.1.15",
     "xml2js": "0.4.23"
   },
   "peerDependencies": {
     "wslink": "0.1.15",
     "shader-loader": "^1.3.1",
-    "worker-loader": "^2.0.0"
+    "worker-loader": "^3.0.0"
```
jourdain, line 108: Or I guess having things in devDep (shader-loader, worker-loader) and tagging them here might be a better way to go?

jourdain (commented):
Small set of questions/comments

package.json
floryst: I think this is used in pdb loading when running tests. Installed it when `npm test` complained.

package.json
floryst: We use it in [`webpack.prod.js`](https://github.com/Kitware/vtk-js/blob/master/webpack.prod.js#L3).

package.json
floryst: I'll need to look into this further. kw-doc should bring it along, but running `npm run doc` complained about missing it.

package.json
floryst: Right now we have them included as peer dependencies. Probably worthwhile to stick it in the main dependencies.

.eslintrc.js
floryst: These look to be old examples. See: [Actor](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/Actor/example_/index.js), [BoundingBox](https://github.com/Kitware/vtk-js/blob/master/Sources/Common/DataModel/BoundingBox/example_/index.js). How about we remove them instead, or move example usage code into `api.md`?

package.json
floryst: Turns out webpack v5 comes with terser bundled by default, so I'll make that change.

package.json
floryst: I'm adding them to dependencies.

floryst:
Should this be a major version bump or minor version bump?

.eslintrc.js
jourdain: Remove the actor but the boundingBox one should be replace to a test if not already the case. In other words, they should be removed but we should make sure we have a test for BoundingBox

package.json
jourdain: ok. kw-doc may depend on kw-web-suite which may introduce that issue.
We could migrate some of those deps in kw-doc to be proper rather than depend again on kw-web-suite

package.json
jourdain: up to you, but it might be good enough to be in the dep to prevent user error/complain

jourdain:
Go with Major...

jourdain:
Seems reasonable if things are working but you may need to update/fix kw-doc before doing it...

package.json
floryst: Yep, kw-doc depends on kw-web-suite. I'll go update that first.

floryst:
Update: running `npm run example -- GeometryViewer` works (the `-mf` switches don't work with the vtk.js example runner) and `npm run doc` works after updating kw-doc.

jourdain:
What about `npm run doc -- -mf GeometryViewer` ?

floryst:
Oh it was `npm run doc`, not `npm run example -- -mf`. I'll investigate.

floryst:
Examples now build successfully.

jourdain (approved):

PR #1620, opened 2020-11-08
refactor(reslicecursor): rename reslice cursor view types

finetjul:
Remove medical meaning to variable names.

BREAKING CHANGE: Rename of ViewTypes

PR #1679, opened 2021-01-06
fix(reslicecursorwidget): use correct plane name between coronal (XZ)…

laurennlam:
… and sagittal (YZ) view

@finetjul

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/Constants.js
+++ b/Sources/Widgets/Core/WidgetManager/Constants.js
@@ -3,8 +3,8 @@ export const ViewTypes = {
   GEOMETRY: 1,
   SLICE: 2,
   VOLUME: 3,
-  YZ_PLANE: 4, // Coronal
-  XZ_PLANE: 5, // Sagittal
+  YZ_PLANE: 4, // Sagittal
+  XZ_PLANE: 5, // Coronal
   XY_PLANE: 6, // Axial
 };
 
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
@@ -45,7 +45,7 @@ table.appendChild(trLine2);
 const viewAttributes = [];
 const widget = vtkResliceCursorWidget.newInstance();
 widget.getWidgetState().setOpacity(0.6);
-const sliceTypes = [ViewTypes.YZ_PLANE, ViewTypes.XZ_PLANE, ViewTypes.XY_PLANE];
+const sliceTypes = [ViewTypes.XZ_PLANE, ViewTypes.YZ_PLANE, ViewTypes.XY_PLANE];
 
 for (let i = 0; i < 3; i++) {
   const element = document.createElement('td');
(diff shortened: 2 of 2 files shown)


Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -196,9 +196,9 @@ reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
       const reslice = obj.reslice;
       let viewType = ViewTypes.XY_PLANE;
```
finetjul, line 197: `const viewType = slliceTypes[i]` ?

laurennlam: Indeed, seems better.

Sources/Widgets/Core/WidgetManager/Constants.js
```diff
@@ -3,8 +3,8 @@ export const ViewTypes = {
   GEOMETRY: 1,
   SLICE: 2,
   VOLUME: 3,
-  YZ_PLANE: 4, // Coronal
-  XZ_PLANE: 5, // Sagittal
+  XZ_PLANE: 4, // Coronal
+  YZ_PLANE: 5, // Sagittal
```
finetjul, line 7: I would prefer to not change that order if possible. i.e.
```
X = 0 = YZ_PLANE
Y = 1 = XZ_PLANE
Z = 2 = XY_PLANE
```

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -57,7 +57,7 @@ function vtkResliceCursorWidget(publicAPI, model) {
     // at least completely cover the viewed region
     const planeSource = vtkPlaneSource.newInstance();
 
-    if (viewType === ViewTypes.XZ_PLANE) {
+    if (viewType === ViewTypes.YZ_PLANE) {
```
finetjul, line 60: according to the below code, it should be XZ_PLANE

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -89,7 +89,7 @@ function vtkResliceCursorWidget(publicAPI, model) {
         bounds[3] - offset[1],
         center[2]
       );
-    } else if (viewType === ViewTypes.YZ_PLANE) {
+    } else if (viewType === ViewTypes.XZ_PLANE) {
```
finetjul, line 92: According to the below code, it is YZ_PLANE

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: XZ plane means that the plane is constructed with X and Z axis. So, the constructed plane is Y plane. We want the associated Y normal.

finetjul: isn't it what the code does ?

laurennlam: Ok. I didn't check the correct lines.
But I inverted these two lines because there was a mistake when we moved from medical view name to plane name.
cf https://github.com/Kitware/vtk-js/pull/1620
I just fixed it.

PR #1680, opened 2021-01-06
fix(shapewidget): upgrade placePoint methods to not propagate Interac…

ocrossi:
…tionEvents

### DIFF
--- a/Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
@@ -131,7 +131,6 @@ export default function widgetBehavior(publicAPI, model) {
       }
 
       publicAPI.updateShapeBounds();
-      publicAPI.invokeStartInteractionEvent();
 
       if (model.visibleOnFocus) {
         model.shapeHandle.setVisible(true);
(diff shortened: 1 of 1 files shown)


ocrossi:
@finetjul clean PR

finetjul:
LGTM.

Please note that the ShapeWidget example is broken (not due to your fix though)

finetjul:
@agirault

ocrossi:
after more testing some problems emerge from this hotfix. I'm adding a WIP tag to the PR and dig further

agirault:
Hi @ocrossi! What's the reason for not propagating those events?

ocrossi:
The reason would be to force the placePoint functions to only place points and to leave the interaction propagation part to the handleLeftButtonPress and handleLeftButtonRelease. This comes from a case where you want to place a widget with a set of coordinates without having the user interacting with it.

finetjul:
> This comes from a case where you want to place a widget with a set of coordinates without having the user interacting with it.

said differently, Octave does not want to trigger * Interaction events when points are placed programmatically.

floryst:
If the event handlers (`*Press` and `*Release`) handle emitting interaction events, then this looks good to me.

agirault:
Understood thank you @ocrossi and @finetjul. That's good information to have: could you add that explanation to the commit message?

PR #1683, opened 2021-01-07
fix(ResliceCursorWidget): setup correct updateMethodName when scrolling

laurennlam:
Setup the updateMethodName when scrolling (mouse click + move and mouse scrolling) in order to be consistent.
The user can need this out of the Reslice Cursor Widget.

@finetjul

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
@@ -20,6 +20,29 @@ export default function widgetBehavior(publicAPI, model) {
   let isDragging = null;
   let isScrolling = false;
 
+  // Reset "updateMethodName" attribute when no actors are selected
+  // Useful to update 'updateMethodeName' to the correct name which
+  // will be TranslateCenter by default
+  publicAPI.resetUpdateMethod = () => {
+    if (model.representations.length !== 0) {
+      model.representations[0].getSelectedState();
+    }
+  };
+
+  publicAPI.startScrolling = (newPosition) => {
+    if (newPosition) {
+      model.previousPosition = newPosition;
+    }
+    isScrolling = true;
+    publicAPI.resetUpdateMethod();
+    publicAPI.startInteraction();
+  };
+
+  publicAPI.endScrolling = () => {
+    isScrolling = false;
+    publicAPI.endInteraction();
+  };
+
   publicAPI.updateCursor = () => {
     switch (model.activeState.getUpdateMethodName()) {
       case InteractionMethodsName.TranslateCenter:
@@ -52,9 +75,7 @@ export default function widgetBehavior(publicAPI, model) {
       model.widgetState.getScrollingMethod() ===
       ScrollingMethods.LEFT_MOUSE_BUTTON
     ) {
-      isScrolling = true;
-      model.previousPosition = callData.position;
-      publicAPI.startInteraction();
+      publicAPI.startScrolling(callData.position);
     } else {
       return macro.VOID;
     }
@@ -83,10 +104,9 @@ export default function widgetBehavior(publicAPI, model) {
 
   publicAPI.handleLeftButtonRelease = () => {
     if (isDragging || isScrolling) {
-      publicAPI.endInteraction();
+      publicAPI.endScrolling();
     }
     isDragging = false;
-    isScrolling = false;
     model.widgetState.deactivate();
   };
 
@@ -95,9 +115,7 @@ export default function widgetBehavior(publicAPI, model) {
       model.widgetState.getScrollingMethod() ===
       ScrollingMethods.RIGHT_MOUSE_BUTTON
     ) {
-      model.previousPosition = calldata.position;
-      isScrolling = true;
-      publicAPI.startInteraction();
+      publicAPI.startScrolling(calldata.position);
     }
   };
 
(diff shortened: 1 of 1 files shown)


Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -188,6 +188,14 @@ export default function widgetBehavior(publicAPI, model) {
   };
 
   publicAPI.translateCenterOnCurrentDirection = (nbSteps, renderer) => {
+    // As we don't select a representation actor, the update method name is
+    // not set (normally set by the ResliceCursorContextRepresentation)
+    const methodName = InteractionMethodsName.TranslateCenter;
```
finetjul, line 193: Where (what file) is InteractionMethodsName.TranslateCenter defined ?

laurennlam: vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/Constants

laurennlam: There are also used in ResliceCursorContextRepresentation

finetjul: (my bad, my master was not up to date)

finetjul: and why do you need to set `updateMethodName` ?

laurennlam: Because if you don't click on a representation actor (line, sphere), then this method is empty. Or, we definitely apply an action related to an InteractionMethodsName.
If the user wants to know which action is currently managed, then he can use : `widget.getWidgetState().getUpdateMethodName()`. Then, he can know if we are actually rotatitin/translating axis/translating center.
It's interesting when we want to define different interaction/rendering (example: behaviors of focal point)

finetjul: Understood, if it is "interesting", that means it is not a "fix" but "feat" ?

How do you know it is a center translation and not an axis translation?

And if it is an axis translation, then you would overwrite it with center translation?

laurennlam: I think it's more a feat because during interaction, this attribute should always be filled. And when we scroll, it wasn't.
And it's a fix, because if you make a rotation and then scroll, the `updateMethodName` will keep 'rotate'. Why ? because the only place it's overwrite is in the `ResliceContextRepresentation`. And if you don't pick an actor, then you never go into this class.

I know it's a center translation because I put this piece of code in `translateCenterOnCurrentDirection` method, which is only called when we scroll along slices (mouse scroll or mouse click + move).

laurennlam:
@finetjul Does it look better ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -20,6 +20,15 @@ export default function widgetBehavior(publicAPI, model) {
   let isDragging = null;
   let isScrolling = false;
 
+  // Update some widget state attributes when no actors are selected
+  // Usefull to update 'updateMethodeName' to the correct name which
```
finetjul, line 24: Useful

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -68,6 +77,7 @@ export default function widgetBehavior(publicAPI, model) {
     }
     if (isScrolling) {
       if (model.previousPosition.y !== callData.position.y) {
+        publicAPI.updateState();
```
finetjul, line 80: why not call it on `handleLeftButtonPress` instead?

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -20,6 +20,15 @@ export default function widgetBehavior(publicAPI, model) {
   let isDragging = null;
   let isScrolling = false;
 
+  // Update some widget state attributes when no actors are selected
```
finetjul, line 23: Reset "updateMethodName"

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -20,6 +20,15 @@ export default function widgetBehavior(publicAPI, model) {
...
+  // Usefull to update 'updateMethodeName' to the correct name which
+  // will be TranslateCenter by default
+  publicAPI.updateState = () => {
```
finetjul, line 26: I think the function name is misleading. It carries a message that it is an update(), therefore it is safe to call it anywhere.

However, it does not take into account the mouse position, so it can't be called anywhere. It "resets" the updateMethodName state.

Therefore I suggest instead: 
`updateState` -> `resetState` or `resetUpdateMethod`.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
laurennlam: I just updated where I call this method. We don't want to call it each time we click (it's already done by the representation) but only when we scroll.
So, I called this method when the attribute `isScrolling` is set to true.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -134,6 +145,7 @@ export default function widgetBehavior(publicAPI, model) {
       ScrollingMethods.MIDDLE_MOUSE_BUTTON
     ) {
       isScrolling = true;
+      publicAPI.resetUpdateMethod();
```
finetjul, line 148: why not also in `handleStartMouseWheel` ?

laurennlam: Error.

Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -97,6 +107,7 @@ export default function widgetBehavior(publicAPI, model) {
     ) {
       model.previousPosition = calldata.position;
       isScrolling = true;
+      publicAPI.resetUpdateMethod();
       publicAPI.startInteraction();
```
finetjul, line 111: it looks like you can factorize some code:
```
startScrolling(newPosition) {
if (newPosition) {
  model.previousPosition = newPosition;
}
isScrolling = True;
publicAPI.resetUpdateMethod();
publicAPI.startInteraction();
}
```

laurennlam: Done.
Before you'll ask, I can't set the `isScrolling` to true in `handleStartMouseWheel` because, if you set it to true, then you scroll and move the mouse at the same time, it will try to access to `handleMouseMove` and will potentially failed, especially if you scroll right after loading data, because model.previousPosition doesn't exist.

laurennlam: Does it look better ?

PR #1686, opened 2021-01-08
chore(github): Use custom github publish token

floryst:
I'm pretty sure the default github app token's permissions don't work
for committing to gh-pages branch.


floryst:
FYI @jourdain hopefully this makes doc:publish work again.

PR #1689, opened 2021-01-13
fix(ResliceCursorWidget): Correct reslice cursor interactions

laurennlam:
@finetjul

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
@@ -214,14 +214,15 @@ reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
                 .getWidgetState()
                 .getActiveViewName();
               const currentViewName = getViewPlaneNameFromViewType(viewType);
+              const keepFocalPointPosition =
+                activeViewName !== currentViewName && canUpdateFocalPoint;
               updateReslice({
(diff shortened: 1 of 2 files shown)


Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -214,14 +215,21 @@ reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
                 .getWidgetState()
                 .getActiveViewName();
               const currentViewName = getViewPlaneNameFromViewType(viewType);
+              const currentUpdateMethodName = widget
+                .getWidgetState()
+                .getUpdateMethodName();
+              const keepFocalPointPosition =
+                activeViewName !== currentViewName &&
+                canUpdateFocalPoint &&
```
finetjul, line 223: shouldn't `canUpdateFocalPoint` be already set to false when `currentUpdateMethodName !== InteractionMethodsName.RotateLine` ?

laurennlam: Good point. After rechecking comments, the behavior class wasn't correct. 
With this correction, we have now:
- translation (axe, center): move reslice cursor center
- rotation: don't move reslice cursor center

finetjul: I like that

PR #1691, opened 2021-01-14
fix(ResliceCursorWidget): Fix view up of resliced image when multiply…

laurennlam:
… rotations

@finetjul 
I didn't found out yet how to avoid the short flip when applying 180° rotation. This bug has not been created by this MR. It existed before.
Still in progress.

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
@@ -8,6 +8,7 @@ import vtkPlane from 'vtk.js/Sources/Common/DataModel/Plane';
 import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
 
 import { ViewTypes } from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
+import { defaultViewUpFromViewType } from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/Constants';
 
 import vtkMatrixBuilder from 'vtk.js/Sources/Common/Core/MatrixBuilder';
 
@@ -255,4 +256,56 @@ export function updateState(widgetState) {
     pdLength,
     yRotationLength
   );
+}
+
+/**
+ * First rotate planeToTransform to match targetPlane normal.
+ * Then rotate around targetPlane normal to preserve world plane "up" vector (i.e. Origin->p2 ).
+ * There is an infinite number of options to rotate a plane normal to another. Here we attempt to
+ * do it by preserving Origin, P1 and P2 to be constrained within the volume bounds.
+ * @param {vtkPlaneSource} planeToTransform
+ * @param {vtkPlane} targetPlane
+ * @param {ViewTypes} viewType
+ */
+export function transformPlane(planeToTransform, targetPlane, viewType) {
+  const defaultViewUp = defaultViewUpFromViewType[viewType];
+  const rotatedNormal = targetPlane.getNormal();
+  const rotatedOrigin = targetPlane.getOrigin();
+  // Apply rotation onto plane (i.e. origin, p1, p2)
+  planeToTransform.setNormal(...rotatedNormal);
+  // TBD: isn't it a no-op ?
+  planeToTransform.setCenter(...rotatedOrigin);
+
+  const rotatedOrig = planeToTransform.getOrigin();
+  const rotatedPoint1 = planeToTransform.getPoint1();
+  const rotatedPoint2 = planeToTransform.getPoint2();
+
+  // Compute local view up of transformed plane
+  const rotatedViewUp = vtkMath.subtract(rotatedPoint2, rotatedOrig, [0, 0, 0]);
+  const rotatedPlane = vtkPlane.newInstance({
+    normal: rotatedNormal,
+    origin: rotatedOrigin,
+  });
+  // Project the default viewup we want to fit on
+  const projectedDefaultViewUp = [0, 0, 1];
+  rotatedPlane.projectVector(defaultViewUp, projectedDefaultViewUp);
+
+  vtkMath.normalize(projectedDefaultViewUp);
+  vtkMath.normalize(rotatedViewUp);
+
+  const rotationAngle = vtkMath.angleBetweenVectors(
+    projectedDefaultViewUp,
+    rotatedViewUp
+  );
+
+  // Compute the new plane points
+  const transform = vtkMatrixBuilder
+    .buildFromRadian()
+    .rotate(rotationAngle, rotatedNormal);
+  transform.apply(rotatedOrig);
+  transform.apply(rotatedPoint1);
+  transform.apply(rotatedPoint2);
+  planeToTransform.setOrigin(rotatedOrig);
+  planeToTransform.setPoint1(rotatedPoint1);
+  planeToTransform.setPoint2(rotatedPoint2);
 }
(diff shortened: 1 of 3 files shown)


Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -21,13 +22,17 @@ import { vec4, mat4 } from 'gl-matrix';
...
 
 // ----------------------------------------------------------------------------
 // Factory
 // ----------------------------------------------------------------------------
 
 function vtkResliceCursorWidget(publicAPI, model) {
   model.classHierarchy.push('vtkResliceCursorWidget');
+  defaultViewUpFromViewType[ViewTypes.YZ_PLANE] = [0, 0, 1]; // Sagittal
+  defaultViewUpFromViewType[ViewTypes.XZ_PLANE] = [0, 0, 1]; // Coronal
+  defaultViewUpFromViewType[ViewTypes.XY_PLANE] = [0, 1, 0]; // Axial
```
finetjul, line 35: can't this go in Constants ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -399,10 +404,46 @@ function vtkResliceCursorWidget(publicAPI, model) {
...
+    // Compute the new plane points
+    const transform = vtkMatrixBuilder
+      .buildFromRadian()
+      .rotate(rotationAngle, newNormal);
+    transform.apply(rotatedOrig);
+    transform.apply(rotatedP1);
+    transform.apply(rotatedP2);
+    planeSource.setOrigin(rotatedOrig);
+    planeSource.setPoint1(...rotatedP1);
+    planeSource.setPoint2(...rotatedP2);
```
finetjul, line 446: it would be great to move all this in a separate function for readability, reusability and testing purpose

finetjul: Additionally, it would be great to document a bit more because it is not very obvious

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -399,10 +404,46 @@ function vtkResliceCursorWidget(publicAPI, model) {
...
+    const newNormal = plane.getNormal();
+    const newOrigin = plane.getOrigin();
     // Apply rotation onto plane (i.e. origin, p1, p2)
-    planeSource.setNormal(...plane.getNormal());
+    planeSource.setNormal(...newNormal);
     // TBD: isn't it a no-op ?
-    planeSource.setCenter(...plane.getOrigin());
+    planeSource.setCenter(...newOrigin);
+
+    const rotatedOrig = planeSource.getOrigin();
```
finetjul, line 415: if it is `rotatedOrig`, then it should be `rotatedNormal` and not `newNormal`

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -399,10 +404,46 @@ function vtkResliceCursorWidget(publicAPI, model) {
...
+    const rotatedP1 = planeSource.getPoint1();
+    const rotatedP2 = planeSource.getPoint2();
```
finetjul, line 417: stay consistent with naming: `rotatedPoint1`

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: It depends if, for any reason, we want to compute it.

finetjul: when we will want to compute it, then we will get it outside of Constants

laurennlam: Sounds good.
Do you know the best way to declare a map with an enum ? Because I tried this but it's not correct :
`const defaultViewUpFromViewType = {
ViewTypes.YZ_PLANE: [0, 0, 1],
ViewTypes.XZ_PLANE: [0, 0, 1],
ViewTypes.XY_PLANE: [0, 1, 0]
};`

finetjul:
```
const defaultViewUpFromViewType = { [ViewTypes.YZ_PLANE]: [0, 0, 1], [ViewTypes.XZ_PLANE]: [0, 0, 1], [ViewTypes.XY_PLANE]: [0, 1, 0] };
```

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -256,3 +257,50 @@ export function updateState(widgetState) {
     yRotationLength
   );
 }
+
+export function correctPlaneViewUp(
+  initialPlaneSource,
+  computedPlane,
+  viewType
+) {
```
finetjul, line 265: it does not only correct view up, it also rotate the view plane.
Suggestion: 
```
/**
 * First rotate planeToTransform to match targetPlane normal.
 * Then rotate around targetPlane normal to preserve world plane "up" vector (i.e. Origin->p2 ).
 * There is an infinite number of options to rotate a plane normal to another. Here we attempt to
 * do it by preserving Origin, P1 and P2 to be constrained within the volume bounds.
 * @param {*} planeToTransform
 * ...
 */
export function transformPlane(planeToTransform, targetPlane, viewType)
````

File: Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js [hunk 6]
```diff
@@ -256,3 +257,50 @@ export function updateState(widgetState) {
...
+
+  // Compute the new plane points
+  const transform = vtkMatrixBuilder
+    .buildFromRadian()
+    .rotate(rotationAngle, rotatedNormal);
+  transform.apply(rotatedOrig);
+  transform.apply(rotatedPoint1);
+  transform.apply(rotatedPoint2);
+  initialPlaneSource.setOrigin(rotatedOrig);
+  initialPlaneSource.setPoint1(...rotatedPoint1);
```
finetjul, line 304: why do you sometime use the spread operator and sometime not ? be consistent

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -256,3 +257,50 @@ export function updateState(widgetState) {
...
+  const rotatedPoint2 = initialPlaneSource.getPoint2();
+
+  // Compute local view up of transformed plane
+  const rotatedViewUp = vtkMath.subtract(rotatedPoint2, rotatedOrig, [0, 0, 0]);
+  const rotatedPlane = vtkPlane.newInstance({
+    normal: rotatedNormal,
+    origin: rotatedOrigin,
+  });
+  // Project the default viewup we want to fit on
+  const projectedDefaultViewUp = [0, 0, 0];
```
finetjul, line 285: does not seem like a meaningful default value -> `[0, 0, 1]`

laurennlam:
@finetjul , if it's LGTY, you can merge.
I didn't manage to reproduce the issue whereas I found out the issue.
The problem is that, for any reason, if we set a normal to a vtkPlaneSource which is rotated about 180° from the planeSourceNormal, then sometimes the computed matrix is the identity. Thats why sometimes the image seems to be flipped.
But as soon as I found the origin of the problem, I didn't manage to reproduce it... So maybe we can merge as it is now and see if we can reproduce it.

PR #1695, opened 2021-01-15
vtk.js ESM build with rollup

### LINKED ISSUE #1693: Fix vtk.js ES module usage
Right now, vtk.js ES usage requires the library consumer to bring in asset and source preprocessors in order to build their apps against vtk.js. This is alleviated by us providing webpack rules that consumers can use, but that only makes sense in the case that they use webpack to bundle their application. Userse of other bundlers (parcel, rollup, etc.) are left to figure out what preprocessors are needed (e.g. for shaders and workers) in order to consume vtk.js.

The goal is to allow library consumers to not have to deal with shader-loader, raw-loader, or any other preprocessing specific to vtk.js. Instead of having consumers configure their webpack to include vtk.js rules so that imports work, we can simplify the process by not having them include vtk.js-specific rules. To do this, we can use rollup to generate ES modules for each vtk.js file. Below is an example of what this will look like from the developer's perspective.

```js
// note no more "/Sources/"
import vtkRenderer from 'vtk.js/Rendering/Core/Renderer';
```

Webpack config:
```
module.exports = {
  ...
  module: {
    // don't need to bring in vtk.js rules anymore!
    rules: [ { test: /\.js$/, loader: 'babel-loader' } ],
  },
};
```

In the end, the library consumer should only have to deal with their build deps, and not vtk.js build deps as well.

This was borne out of discussions with @jourdain, and there will be a PR shortly to test this.

floryst:
Since webpack v5 does not (yet) support ESM library output, we use
rollup to generate the ESM build of vtk.js.

We still intend to keep the umd build of vtk.js for use as a script tag drop-in.

Closes #1693 

This will be a beta release for testing.


jourdain (approved):
LGTM

PR #1696, opened 2021-01-19
fix(ResliceCursorWidget): Setup correct viewUp for Z view

laurennlam:
Fix the view up in order to fit with Slicer
@finetjul


finetjul:
@laurennlam what data did you use to setup those value ?

laurennlam:
our client data.

PR #1697, opened 2021-01-20
[WIP] ShapeMixin

ocrossi:


Sources/Widgets/Core/StateBuilder/shapeMixin.js
```diff
@@ -0,0 +1,18 @@
+import macro from 'vtk.js/Sources/macro';
+
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {
+  shapeName: '',
```
finetjul, line 6: name it 'shape' instead of 'shapeName'

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -57,7 +71,17 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
       [HandleRepresentationType.ARROWHEAD6]: {
         initialValues: { shape: 'arrow6points' },
       },
+      [HandleRepresentationType.CONE]: {
+        initialValues: { shape: 'cone' },
```
finetjul, line 75: add a `builder` prop:
```
...
[HandleRepresentationType.CONE]: {
  builder: vtkConeSource,
  initialValues: {shape: 'cone'},
}
```
then:
```
 const rep = representationToSource[model.handleType];
return rep.builder.newInstance(rep.initalValues);
```

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -168,6 +228,20 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+    renderingType = RenderingTypes.FRONT_BUFFER,
+    widgetVisible = true,
+    ctxVisible = true
+  ) => {
+    const state = publicAPI.getRepresentationStates()[0];
+    superClass.updateActorVisibility(
+      renderingType,
+      widgetVisible,
+      ctxVisible,
+      state.isVisible()
```
finetjul, line 242:
```
updateActorVisibility = (
    renderingType = RenderingTypes.FRONT_BUFFER,
    widgetVisible = true,
    ctxVisible = true,
    handleVisible = true
  ) => {
    const state = publicAPI.getRepresentationStates()[0];
    superClass.updateActorVisibility(
      renderingType,
      widgetVisible,
      ctxVisible,
      handleVisible && state.isVisible()
    );
```

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -176,16 +250,22 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
 
 const DEFAULT_VALUES = {
   defaultScale: 1,
+  orientation: [0, 0, 0],
```
finetjul, line 253: how about setting a meaningful initial value ?

Sources/Widgets/Representations/ConeHandleRepresentation/index.js
```diff
@@ -137,18 +145,37 @@ function vtkConeHandleRepresentation(publicAPI, model) {
...
   vtkHandleRepresentation.extend(publicAPI, model, initialValues);
-  macro.setGet(publicAPI, model, ['glyph', 'glyphResolution', 'defaultScale']);
+  macro.setGet(publicAPI, model, [
+    'glyph',
+    'glyphResolution',
+    'defaultScale',
+    'handleVisibility',
+  ]);
   vtkConeHandleRepresentation(publicAPI, model);
 }
```
finetjul, line 180: File to remove ?

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -60,29 +71,32 @@ function vtkLineWidget(publicAPI, model) {
...
   ];
-  model.behavior = widgetBehavior;
-  model.widgetState = stateGenerator();
-  model.widgetState.setPositionOnLine(model.positionOnLine);
-  model.widgetState.setNbHandles(0);
-  initializeHandleRepresentations();
+
+  publicAPI.initializeHandleRepresentations = () => {
+    model.handle1Shape = model.widgetState.getHandle1().getShapeName();
+    model.handle2Shape = model.widgetState.getHandle2().getShapeName();
```
finetjul, line 84: I do not think this is how mix-ins are supposed to be handled.
To me, it shoud be handled inside "ArrowHandleRepresentation::requestData"

ocrossi: about this review, my understanding of the problem is that requestData should have a glyph to display. In order to do so it should know on initialization which is the glyph to display with requestData. How is it possible to request a data not set before?

finetjul: Maybe `requestData` should call `detectArrowShape`

ocrossi: sorry to discuss the review like that but the only solution i came up with is not really elegant. RequestData is looped at each event, like a mouseMove when nothing is selected at all (eg initial state of the scene before the user grabs focus). So to apply detectArrowShape in requestData would mean to set a glyph at null on initialization, then catch it with a bool variable to detect when is the 1st requestData sent. To me it is overkill. Am i missing something here?

finetjul: The idea is to call `detectArrowShape` in `requestData` (or in a function calling requestData) only when the "shape" changes. 
I guess it is fine to keep a variable to know what is the current shape set as input of mapper.
I think it is elegant because you can change the shape without having to reinstantiate the class.

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
ocrossi: for this i think it is important to give it a neutral 1st value because orientation has a direct link to 2 behaviors, set handle direction when moving a handle, and set handle orientation to face camera. I have a limited understanding of the use of matrices in 3d graphics so i don't see what a meaningful value would be.

finetjul: orientation does not play for "facing the camera".
I think by default, it should go orient an arrow from left to right.

ocrossi:
@finetjul new version of the refactored code.
I am currently working on 2 points : 
- refactor handleGlyphRotation
- add event listeners of positionOnLine in UI
apart from that everything discussed in precedent reviews was processed.

ocrossi:
@finetjul

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -168,6 +250,40 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+    if (shapeName !== currShape) {
+      if (firstShape === false) {
+        firstShape = true;
+      } else {
+        model.orientation = [1, 0, 0];
+        model.viewMatrix = mat4.create();
+      }
+      shapeName = currShape;
+      model.pipelines.arrow.glyph = detectArrowShape(shapeName);
+      model.glyph = model.pipelines.arrow.glyph;
```
finetjul, line 267: `model.glyph` was used to be a shortcut to `model.pipelines.arrow.glyph`.
Let's get rid of it because it does not help readability

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -22,10 +30,14 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkArrowHandleRepresentation');
 
+  const superClass = { ...publicAPI };
   // --------------------------------------------------------------------------
   // Internal polydata dataset
   // --------------------------------------------------------------------------
 
+  let firstShape = false;
+  let shapeName = '';
```
finetjul, line 39: shapeName -> currentShape

finetjul: and move it into model.currentShape (no accessor)

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -22,10 +30,14 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkArrowHandleRepresentation');
 
+  const superClass = { ...publicAPI };
   // --------------------------------------------------------------------------
   // Internal polydata dataset
   // --------------------------------------------------------------------------
 
+  let firstShape = false;
```
finetjul, line 38: get rid of it. it should not be needed. If shapeName is empty it means it is a first shape

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -74,24 +108,33 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
   model.displayMapper.setInputConnection(publicAPI.getOutputPort());
   publicAPI.addActor(model.displayActor);
   model.alwaysVisibleActors = [model.displayActor];
+  model.pipelines = {
+    arrow: {
+      source: publicAPI,
+      glyph: vtkSphereSource.newInstance(),
```
finetjul, line 114: if you have a valid glyph, then must update `model.currentShape` accordingly

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -125,8 +168,39 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+    displayOrientation[2] = 0;
+    if (shapeName === HandleRepresentationType.CONE) {
+      displayOrientation = [0, 1, 0];
+    }
+    const baseDir =
+      shapeName === HandleRepresentationType.CIRCLE ? [1, 0, 0] : [0, 1, 0];
+    displayOrientation =
+      shapeName === HandleRepresentationType.CIRCLE
+        ? [0, 0, 1]
+        : displayOrientation;
```
finetjul, line 188: those last 9 lines are hard to read. Please clean it up.

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -168,6 +250,40 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+  publicAPI.requestData = (inData, outData) => {
+    const currShape = publicAPI
+      .getRepresentationStates(inData[0])[0]
+      .getShape();
+    if (shapeName !== currShape) {
+      if (firstShape === false) {
+        firstShape = true;
+      } else {
+        model.orientation = [1, 0, 0];
+        model.viewMatrix = mat4.create();
```
finetjul, line 263: do not reset orientation nor viewMatrix when changing shape

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
ocrossi: here i dont have access to the state of my representation yet. The trick i used was to set a default shape that would be overriden by the 1st call of requestData. The solution i originally came up with was to include model shape in initialValues of my representation in the lineWidget/index.js level. Should i use this solution?

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
ocrossi: the fact of not reseting orientation and viewMatric can create some annoying effects when the user changes handles. For instance cube or cone would inherit the former arrow orientation which is not what was originally planned in the behavior.js file (e.g function isHandleGlyph2D). What is the best solution here: leaving reset parameters as it is on this comment or generalize the faceCamera behavior to all glyphs?

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
ocrossi: after some tests it seems that model.glyph is used to link glyph to the setGet glyph below, which i'm using to orient the cone handle to match line direction. I think it should not be removed.

finetjul: ok to keep it

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
finetjul: faceCamera does not depend on the shape.

Maybe it is more a matter of ignoring faceCamera when the shape does not make sense

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -125,8 +168,52 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+    const displayMatrix = vtkMatrixBuilder
+      .buildFromDegree()
+      .rotateFromDirections(baseDir, displayOrientation)
+      .getMatrix();
+    const displayRotation = mat3.create();
+    const rotation = mat3.create();
+    if (
+      model.currShape !== HandleRepresentationType.CONE &&
+      model.currShape !== HandleRepresentationType.CUBE &&
+      model.currShape !== HandleRepresentationType.SPHERE
```
finetjul, line 206: create a convenient function `isCurrentShape3D()` or something like that

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -125,8 +168,52 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+    mat3.invert(viewMatrixInv, viewMatrix);
+
+    let displayOrientation = vec3.create();
+    let baseDir = [0, 1, 0];
+    switch (model.currShape) {
+      case HandleRepresentationType.CONE:
+        displayOrientation = [0, 1, 0];
+        break;
+      case HandleRepresentationType.CIRCLE:
+        displayOrientation = [0, 0, 1];
```
finetjul, line 184: why do you bother with displayorientation for a circle ? can't you simplify it to be the same than the cone ?

ocrossi: The circle source is oriented not to face camera, so i had to reorient it. See CircleSource example to check classic orientation (you have to drag the source to see orientation is on the Y axis)

ocrossi:
@finetjul the upgrades discussed in our previous point are implemented. Should i clean git history and rebase master for upcoming merge?

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
finetjul: Maybe you should expose an orientation parameter to vtkCircleSource.

ocrossi:
@finetjul

Sources/Filters/Sources/CircleSource/index.js
```diff
@@ -62,6 +68,8 @@ const DEFAULT_VALUES = {
   radius: 1.0,
   resolution: 6,
   center: [0, 0, 0],
+  baseOrientation: [0, 1, 0],
```
finetjul, line 71: do not expose publicly

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -125,8 +197,72 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+  };
+
+  // --------------------------------------------------------------------------
+
+  const glyphIsNot3D = () =>
+    model.currentShape !== HandleRepresentationType.CONE &&
+    model.currentShape !== HandleRepresentationType.CUBE &&
+    model.currentShape !== HandleRepresentationType.SPHERE;
+
+  /*
```
finetjul, line 217: /**

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -125,8 +197,72 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+
+  // --------------------------------------------------------------------------
+
+  publicAPI.setDrawFace = (draw) => {
+    model.pipelines.circle.glyph.setFace(draw);
+  };
+
+  // --------------------------------------------------------------------------
+
+  const glyphIsNot3D = () =>
```
finetjul, line 212: `glyphIsNot3D` -> `is2DGlyph`

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -125,8 +197,72 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+    const rotation = mat3.create();
+    if (glyphIsNot3D()) {
+      mat3.fromMat4(displayRotation, displayMatrix);
+      mat3.multiply(rotation, viewMatrix, displayRotation);
+      vec3.transformMat3(scale3, scale3, rotation);
+    }
+    return rotation;
+  }
+
+  function handleGlyphRotation(scale3) {
```
finetjul, line 249: handleGlyphRotation -> `getGlyphRotation`

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -74,29 +134,41 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+      mapper: vtkGlyph3DMapper.newInstance({
+        orientationArray: 'direction',
+        scaleArray: 'scale',
+        colorByArrayName: 'color',
+        scalarMode: ScalarMode.USE_POINT_FIELD_DATA,
+      }),
+      actor: vtkActor.newInstance(),
+    },
+  };
+  model.currentShape = 'sphere';
```
finetjul, line 150: set in DEFAUT_VALUES

Sources/Widgets/Representations/ArrowHandleRepresentation/index.js
```diff
@@ -125,8 +197,72 @@ function vtkArrowHandleRepresentation(publicAPI, model) {
...
+      .buildFromDegree()
+      .rotateFromDirections(baseDir, displayOrientation)
+      .getMatrix();
+    return displayMatrix;
+  }
+
+  function handleGlyphRotationToFaceCamera(scale3, displayMatrix, viewMatrix) {
+    const displayRotation = mat3.create();
+    const rotation = mat3.create();
+    if (glyphIsNot3D()) {
```
finetjul, line 241: move test outside `handleGlyphRotationToFaceCamera`

Sources/Widgets/Widgets3D/LineWidget/helper.js
```diff
@@ -16,3 +18,18 @@ export function updateTextPosition(model) {
   const SVGTextState = model.widgetState.getText();
   SVGTextState.setOrigin(calculateTextPosition(model));
 }
+
+export function getNbHandles(model) {
+  const handle1 = model.widgetState.getHandle1();
+  const handle2 = model.widgetState.getHandle2();
+  const moveHandle = model.widgetState.getMoveHandle();
+  if (handle1.getOrigin().length !== 0 && handle2.getOrigin().length !== 0) {
```
finetjul, line 26: isn't the following working ?
```
if (handle1.getOrigin() && handle2.getOrigin())
```

Sources/Widgets/Widgets3D/LineWidget/helper.js
```diff
@@ -16,3 +18,18 @@ export function updateTextPosition(model) {
...
+    return 2;
+  }
+  if (
+    handle1.getOrigin() === handle2.getOrigin() ||
```
finetjul, line 30: not sure it does what you want

Sources/Widgets/Widgets3D/LineWidget/helper.js
```diff
@@ -16,3 +18,18 @@ export function updateTextPosition(model) {
   const SVGTextState = model.widgetState.getText();
   SVGTextState.setOrigin(calculateTextPosition(model));
 }
+
+export function getNbHandles(model) {
```
finetjul, line 22: add documentation

Sources/Widgets/Widgets3D/LineWidget/state.js
```diff
@@ -1,36 +1,47 @@
...
     .addStateFromMixin({
       labels: ['moveHandle'],
-      mixins: ['origin', 'color', 'scale1', 'visible'],
+      mixins: ['origin', 'color', 'scale1', 'visible', 'shape'],
       name: 'moveHandle',
       initialValues: {
-        scale1: 50,
-        visible: false,
+        scale1: 0,
+        visible: true,
```
finetjul, line 20: should be visible = `false`

Sources/Widgets/Widgets3D/LineWidget/state.js
```diff
@@ -1,36 +1,47 @@
...
       labels: ['handle1'],
-      mixins: ['origin', 'color', 'scale1', 'visible'],
+      mixins: ['origin', 'color', 'scale1', 'visible', 'manipulator', 'shape'],
       name: 'handle1',
       initialValues: {
         scale1: 50,
+        visible: true,
         origin: [],
-        visible: false,
+        shape: '6pointsArrowHead',
```
finetjul, line 33: to remove (sphere by default is ok)

Sources/Widgets/Widgets3D/LineWidget/state.js
```diff
@@ -39,13 +50,14 @@ export default function generateState() {
...
         text: '',
         visible: false,
         origin: [0, 0, 0],
       },
     })
-    .addField({ name: 'positionOnLine', initialValues: 0 })
-    .addField({ name: 'nbHandles', initialValues: 0 })
+    .addStateFromInstance({ name: 'positionOnLine', instance: linePosState })
+    .addField({ name: 'handle1FaceCamera', initialValue: true })
+    .addField({ name: 'handle2FaceCamera', initialValue: true })
```
finetjul, line 61: to move to arrow representation: forceFaceCamera (None let the behavior by default, true forces to face, false forces to not face)

Sources/Widgets/Widgets3D/LineWidget/state.js
```diff
@@ -1,36 +1,47 @@
 import vtkStateBuilder from 'vtk.js/Sources/Widgets/Core/StateBuilder';
 
+const linePosState = vtkStateBuilder
```
finetjul, line 3: Add comment to explain that you want to "observe" it.

Sources/Widgets/Widgets3D/LineWidget/state.js
```diff
@@ -1,36 +1,47 @@
...
+  .createBuilder()
+  .addField({
+    name: 'posOnLine',
+    initialValue: 0,
```
finetjul, line 7: 0.5 by default

package-lock.json
```diff
@@ -24793,7 +24785,6 @@
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-1.1.1.tgz",
       "integrity": "sha512-n/ShnvDi6FHbbVfviro+WojiFzv+s8MPMHBczVePfUpDJLwoLT0ht1l4YwBCbi8pJAveEEdnkHyPyTP/mzRfwg==",
-      "dev": true,
       "requires": {
```
finetjul, line 24788: do not modify package-lock.json

Sources/Widgets/Widgets3D/LineWidget/helper.js
ocrossi: No, since the value of the tab is set [here](https://github.com/ocrossi/vtk-js/blob/a6ae8abd3b3073e664157a7ef3caaa7e2111ff4a/Sources/Widgets/Widgets3D/LineWidget/state.js#L22), you can't check if points are placed this way. Length on the other hand is ok

PR #1703, opened 2021-01-25
Fix initial reslice cursor

laurennlam:
Correct first reslice cursor visualization
@finetjul

### DIFF
--- a/Sources/Interaction/Widgets/ResliceCursor/example/index.js
+++ b/Sources/Interaction/Widgets/ResliceCursor/example/index.js
@@ -1,7 +1,6 @@
 import 'vtk.js/Sources/favicon';
 
-import vtkImageData from 'vtk.js/Sources/Common/DataModel/ImageData';
-import vtkDataArray from 'vtk.js/Sources/Common/Core/DataArray';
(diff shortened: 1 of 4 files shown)


Sources/Interaction/Widgets/ResliceCursor/example/index.js
```diff
@@ -15,118 +14,98 @@ import vtkRenderWindowInteractor from 'vtk.js/Sources/Rendering/Core/RenderWindo
...
+// Define ResliceCursor
 
-const newArray = new Uint8Array(dims[0] * dims[1] * dims[2]);
+const resliceCursor = vtkResliceCursor.newInstance();
 
-let i = 0;
-for (let z = 0; z < dims[2]; z++) {
-  for (let y = 0; y < dims[1]; y++) {
-    for (let x = 0; x < dims[0]; x++) {
-      newArray[i++] = (256 * (i % (dims[0] * dims[1]))) / (dims[0] * dims[1]);
```
finetjul, line 32: This synthetic data could be useful for debug purpose, I would keep it (move into a separate function and do not call it, but keep it)

laurennlam:
Is it possible to merge ?

PR #1705, opened 2021-01-26
fix(ResliceCursorWidget): update how to move center when scrolling

laurennlam:
@finetjul 

Issue:
- rotate sagittal view
- scroll on axial view
Result: both axis are moving instead of one.

ImageSpacing was badly used when computing the new center.
When we are not axis aligned, we cannot use the spacing because it transformed the moving direction.

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
@@ -218,11 +218,17 @@ export default function widgetBehavior(publicAPI, model) {
     const image = model.widgetState.getImage();
     const imageSpacing = image.getSpacing();
 
+    // Use Chebyshev norm
+    // https://math.stackexchange.com/questions/71423/what-is-the-term-for-the-projection-of-a-vector-onto-the-unit-cube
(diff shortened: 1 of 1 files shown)


Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js
```diff
@@ -218,11 +218,28 @@ export default function widgetBehavior(publicAPI, model) {
...
 
+    // Use correct spacing when we are axis aligned
+    // Else, use the nbSteps as the moving factor
+    let movingFactor = nbSteps;
+    let nbZeroComposants = 0;
+    for (let i = 0; i < 3; i++) {
+      if (dirProj[i] === 0) {
+        nbZeroComposants++;
+      }
+    }
```
finetjul, line 229: `const nbZeroComponents = dirProj.filter(comp => comp === 0).length`

finetjul (commented):
To me it does not make sense to consider spacing only when axis aligned. It should be always or never (think of the case where we are VERY close to axis aligned)

laurennlam:
Then, how would you handle this case ? Because it's definitely false the way we did before.

laurennlam:
Should be better now.

PR #1711, opened 2021-01-28
fix(httpsceneloader): properly handle .vtkjs files with no actor

### LINKED ISSUE #1710: Scene Explorer example not working
When trying out Scene Explorer example to load lidar.vtkjs, nothing is rendered:

image

Upon checking console, the following error can be observed: 

image

Is this a known bug with scene explorer?

boucaud:
fixes #1710

### DIFF
--- a/Sources/IO/Core/HttpSceneLoader/index.js
+++ b/Sources/IO/Core/HttpSceneLoader/index.js
@@ -41,9 +41,7 @@ function applySettings(sceneItem, settings) {
   }
 
   if (settings.property) {
-    if (settings.actor) {
-      sceneItem.actor.getProperty().set(settings.property);
-    } else {
+    if (settings.volume) {
       const volumePropertySettings = { ...settings.property };
       delete volumePropertySettings.components;
       sceneItem.volume.getProperty().set(volumePropertySettings);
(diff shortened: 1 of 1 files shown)


boucaud:
@jourdain please review

jourdain:
Did that fix the example with the sample vtkjs file provided?

jourdain:
The code change seems reasonable but just looking at the diff, it is not clear what you've done. But if that works now with the old files, then we can merge once green.

boucaud:
> Did that fix the example with the sample vtkjs file provided?

Yes, I tested with lidar.vtkjs and it works.

The issue is with this scene:

```
        {
            "name": "LIDAR data",
            "httpDataSetReader": {
                "url": "0"
            },
            "mapper": {
                "colorByArrayName": "__CustomRGBColorArray__",
                "colorMode": 0,
                "scalarMode": 3
            },
            "property": {
                "representation": 2,
                "pointSize": 2.0,
                "diffuseColor": [
                    1.0,
                    1.0,
                    1.0
                ]
            },
            "type": "httpDataSetReader",
            "lookupTable": {
                "tableRange": [
                    0.0,
                    51.0
                ],
                "hueRange": [
                    0.5,
                    0
                ]
            }
        }
```

Which does not have an actor object.

When adding support for volumes, I checked for item.actor or item.volume when determining which should be created, so in this case no actor would end up being created.

This MR makes it so an actor is created by default, unless item.volume is present.

PR #1714, opened 2021-02-01
fix(ResliceCursorWidget): Update how to setup reslice plane and impro…

laurennlam:
- Update how the reslice plane is computed (with defined origin, point1 and point2 of the plane)
- Improve reslice cursor widget example with correct color + add 3D view with slices

@finetjul

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
@@ -45,6 +45,8 @@ const widget = vtkResliceCursorWidget.newInstance();
 const widgetState = widget.getWidgetState();
 widgetState.setKeepOrthogonality(true);
 
+const showDebugActors = true;
+
 // ----------------------------------------------------------------------------
 // Define html structure
 // ----------------------------------------------------------------------------
@@ -178,6 +180,7 @@ for (let i = 0; i < 4; i++) {
     const actor = vtkActor.newInstance();
     actor.setMapper(mapper);
     actor.getProperty().setColor(...viewColors[i]);
+    actor.setVisibility(showDebugActors);
     obj.sphereActors.push(actor);
     obj.sphereSources.push(sphere);
   }
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
@@ -313,8 +313,7 @@ reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
     outlineActor.setMapper(outlineMapper);
     view3D.renderer.addActor(outlineActor);
 
-    for (let i = 0; i < viewAttributes.length; i++) {
-      const obj = viewAttributes[i];
+    viewAttributes.forEach((obj, i) => {
       obj.reslice.setInputData(image);
       obj.renderer.addActor(obj.resliceActor);
       view3D.renderer.addActor(obj.resliceActor);
@@ -372,7 +371,7 @@ reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
         resetViewUp: true, // Need to be reset the first time the widget is initialized. Then, can be set to false, so that the camera view up will follow the camera
       });
       obj.renderWindow.render();
-    }
+    });
 
     view3D.renderer.resetCamera();
     view3D.renderer.resetCameraClippingRange();
@@ -383,8 +382,7 @@ reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
 // Define panel interactions
 // ----------------------------------------------------------------------------
 function updateViews() {
-  for (let i = 0; i < viewAttributes.length; i++) {
-    const obj = viewAttributes[i];
+  viewAttributes.forEach((obj, i) => {
     updateReslice({
       viewType: sliceTypes[i],
       reslice: obj.reslice,
@@ -397,7 +395,7 @@ function updateViews() {
       resetViewUp: true,
     });
     obj.renderWindow.render();
-  }
+  });
   view3D.renderer.resetCamera();
   view3D.renderer.resetCameraClippingRange();
 }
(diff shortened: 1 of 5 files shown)


laurennlam:
@finetjul Can you review please ?

laurennlam:
@finetjul It's ready for review.
I added control panel in order to be able to manage the 'keep orthogonality' state so that clients can test it.
In order to show different possibility, I also added checbox to allow/disable rotation and translation.

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -122,8 +165,28 @@ for (let i = 0; i < 3; i++) {
...
+    const sphere = vtkSphereSource.newInstance();
+    sphere.setRadius(10);
+    const mapper = vtkMapper.newInstance();
+    mapper.setInputConnection(sphere.getOutputPort());
+    const actor = vtkActor.newInstance();
+    actor.setMapper(mapper);
+    actor.getProperty().setColor(...viewColors[i]);
+    obj.sphereActors.push(actor);
+    obj.sphereSources.push(sphere);
+  }
```
finetjul, line 183: you might want to create a "Debug" variable and show those spheres only when that Debug variable is `true`. It should be `false` by default.

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -133,7 +196,7 @@ for (let i = 0; i < 3; i++) {
     fontFamily: 'Arial',
     fontColor: 'black',
     fontSizeScale: (res) => res / 2,
-    faceColor: '#0000ff',
+    faceColor: '#ff0000',
```
finetjul, line 199: shouldn't you use `viewColors` instead ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -194,26 +258,33 @@ function updateReslice(
...
     interactionContext.reslice,
     interactionContext.viewType
   );
-  if (modified) {
+  if (obj.modified) {
     // Get returned modified from setter to know if we have to render
     interactionContext.actor.setUserMatrix(
       interactionContext.reslice.getResliceAxes()
     );
+    interactionContext.sphereSources[0].setCenter(...obj.origin);
```
finetjul, line 274: interactionContext.sphereSources.forEach(...)

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -378,6 +380,23 @@ function vtkResliceCursorWidget(publicAPI, model) {
...
+      const p0 = [0, 0, 0];
+      const p1 = [0, 1, 0];
+
+      const worldP0 = renderer.normalizedDisplayToWorld(...p0, aspect);
+      const worldP1 = renderer.normalizedDisplayToWorld(...p1, aspect);
+      const newViewUp = vtkMath.subtract(worldP1, worldP0, [0, 0, 0]);
+      vtkMath.normalize(newViewUp);
+
+      viewUpFromViewType[viewType] = newViewUp;
+    }
```
finetjul, line 399: shouldn't we do that instead in `updateReslicePlane`. 
That way `resetCamera` always apply `viewUpFromViewType`

Btw, `viewUpFromViewType` should probably be stored in the widget state.

finetjul (commented):
It is missing a test to enforce that we do not break anything in the future.

laurennlam:
> It is missing a test to enforce that we do not break anything in the future.

I'm working on it but it take some times to have something consistent. I'd like to merge this MR so that I can deliver a new version to the client.

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
laurennlam: Not sure how to use the foreach as each point has a different value ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: `updateReslicePlane` doesn't have access to the renderer (and so the camera). I'm afraid it will overload the method for nothing.

finetjul:
> I'm working on it but it take some times to have something consistent. I'd like to merge this MR so that I can deliver a new version to the client.

Understood, ok to merge without test then

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
finetjul: Indeed. Nevermind then

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -271,8 +368,60 @@ reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
...
+    view3D.renderer.resetCamera();
+    view3D.renderer.resetCameraClippingRange();
   });
 });
+
+// ----------------------------------------------------------------------------
+// Define panel interactions
+// ----------------------------------------------------------------------------
+function updateViews() {
+  for (let i = 0; i < viewAttributes.length; i++) {
```
finetjul, line 386:
```
viewAttributes.forEach((obj, i) => {
 ...
}
```

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
finetjul: You wouldn't need access to the camera in `updateReslicePlane`.
Indeed the camera view up is always sync-ed with `viewUpFromViewType`. Therefore `updateReslicePlane` can use the `viewUpFromViewType` value to know the current camera view-up value. No ?

laurennlam: I'm not sure to understand why the `updateReslicePlane` would know the current camera view up ?
Moreover, the first time we initialize the widget, we have to compute the view up as it's computed in `updateReslicePlane`. Then, if the user doesn't want to reset the viewup, it will be computed in the `resetCamera`

finetjul: `updateReslicePlane` knows the current camera view up because it is the one that sets the view up (`updateReslicePlane` sets `viewUpFromViewType` and `resetCamera` sets the camera view up from `viewUpFromViewType`).

finetjul (commented):
ok

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: We need to align on this. I don't understand what you want to do.

Issue #1687, opened 2021-01-12
addClippingPlane normals inverted between vtkVolumeMapper and vtkMapper


PaulHax:
Much thanks for #1677, just what I needed.  I'm using `mapper::addClippingPlane` on a scene mixing vtkVolumeMapper and vtkMapper with a shared plane object.  Seems that vtkVolumeMapper clips one direction, vtkMapper the other.  Might be nice to make this consistent?  Not sure what the proper fix would be besides negating these values: https://github.com/Kitware/vtk-js/pull/1677/files#diff-653a0c06ccb1ec727a15802ff1d90b9cb6599f82d58610b2522dcfe483ab7cf1R848-R851

See the volume clipped one way, the plane clipped on the other direction:
![image](https://user-images.githubusercontent.com/16823231/104356055-7acfb080-54d9-11eb-8bf0-75b088affcbb.png)

gandis0713:
I think the conditions to set bounds of the volume by plane normal seem wrong.
So, I modified the conditions and created a PR. https://github.com/Kitware/vtk-js/pull/1715

daker:
@jourdain Fixed in #1715, the issue can be closed

PR #1715, opened 2021-02-01
fix(vtkvolumemapper): fix the bounds of the volume.

gandis0713:
Modify the comparing condition for direction of the plane that sets the bound s of the volume.


daker:
@gandis0713 did you had a chance to look at #1687 ?

gandis0713:
@daker  Yes, I modified after seeing #1687.

martinken:
looks good as our convention is that positive values are kept and negative are discarded.  +1

daker:
@gandis0713 the yarn.lock needs to be removed from PR

PaulHax:
I tried out the branch. This clips the same direction:
```
mapper.addClippingPlane(clipPlane1);
mapperSurface.addClippingPlane(clipPlane1);
```
Thanks for the follow through gandis.

gandis0713:
@daker Thank you for the notice. I removed yarn.lock.

PR #1717, opened 2021-02-03
fix(ShapeWidget & SplineWidget): place widget correctly on ImageSlice

ocrossi:
Add manipulator data update from shapeWidget example, add visibility update for widgets that are not on slice.
Debug painting tool for paintWidget example when it uses shapeWidget data.

### DIFF
--- a/Sources/Widgets/Widgets3D/PaintWidget/example/index.js
+++ b/Sources/Widgets/Widgets3D/PaintWidget/example/index.js
@@ -409,6 +409,7 @@ scene.rectangleHandle.onInteractionEvent(() => {
   const rectangleHandle = scene.rectangleHandle
     .getWidgetState()
     .getRectangleHandle();
+
   painter.paintRectangle(
     rectangleHandle.getOrigin(),
     rectangleHandle.getCorner()
@@ -422,11 +423,17 @@ scene.ellipseHandle.onInteractionEvent(() => {
     .getWidgetState()
     .getEllipseHandle()
     .getOrigin();
-  const scale3 = scene.ellipseHandle
+  const point2 = scene.ellipseHandle
     .getWidgetState()
-    .getEllipseHandle()
-    .getScale3();
-  painter.paintEllipse(center, scale3);
+    .getPoint2Handle()
+    .getOrigin();
+
+  const corner = [
+    center[0] - point2[0],
+    center[1] - point2[1],
+    center[2] - point2[2],
+  ];
+  painter.paintEllipse(center, corner);
 });
 
 initializeHandle(scene.circleHandle);
@@ -436,11 +443,15 @@ scene.circleHandle.onInteractionEvent(() => {
     .getWidgetState()
     .getEllipseHandle()
     .getOrigin();
-  const scale3 = scene.circleHandle
+  const point2 = scene.circleHandle
     .getWidgetState()
-    .getEllipseHandle()
-    .getScale3();
-  painter.paintEllipse(center, scale3);
+    .getPoint1Handle()
+    .getOrigin();
+
+  const radius = vec3.distance(center, point2);
+
+  const corner = [radius, radius, radius];
+  painter.paintEllipse(center, corner);
 });
 
 scene.splineHandle.onStartInteractionEvent(() => {
--- a/Sources/Widgets/Widgets3D/PaintWidget/example/index.js
+++ b/Sources/Widgets/Widgets3D/PaintWidget/example/index.js
@@ -1,6 +1,4 @@
 import 'vtk.js/Sources/favicon';
-
-import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
 import vtkFullScreenRenderWindow from 'vtk.js/Sources/Rendering/Misc/FullScreenRenderWindow';
 import vtkWidgetManager from 'vtk.js/Sources/Widgets/Core/WidgetManager';
 import vtkPaintWidget from 'vtk.js/Sources/Widgets/Widgets3D/PaintWidget';
@@ -152,6 +150,7 @@ function readyAll() {
 function updateControlPanel(im, ds) {
   const slicingMode = im.getSlicingMode();
   const extent = ds.getExtent();
+  document.querySelector('.slice').setAttribute('min', extent[slicingMode * 2]);
   document
     .querySelector('.slice')
     .setAttribute('max', extent[slicingMode * 2 + 1]);
@@ -268,30 +267,17 @@ reader
       if (slicingMode > -1) {
         const ijk = [0, 0, 0];
         const position = [0, 0, 0];
-        const normal = [0, 0, 0];
 
         // position
         ijk[slicingMode] = image.imageMapper.getSlice();
         data.indexToWorldVec3(ijk, position);
 
-        // circle/slice normal
-        ijk[slicingMode] = 1;
-        data.indexToWorldVec3(ijk, normal);
-        vtkMath.subtract(normal, data.getOrigin(), normal);
-        vtkMath.normalize(normal);
-
         widgets.paintWidget.getManipulator().setOrigin(position);
-        widgets.paintWidget.getManipulator().setNormal(normal);
         widgets.rectangleWidget.getManipulator().setOrigin(position);
-        widgets.rectangleWidget.getManipulator().setNormal(normal);
         widgets.ellipseWidget.getManipulator().setOrigin(position);
-        widgets.ellipseWidget.getManipulator().setNormal(normal);
         widgets.circleWidget.getManipulator().setOrigin(position);
-        widgets.circleWidget.getManipulator().setNormal(normal);
         widgets.splineWidget.getManipulator().setOrigin(position);
-        widgets.splineWidget.getManipulator().setNormal(normal);
         widgets.polygonWidget.getManipulator().setOrigin(position);
-        widgets.polygonWidget.getManipulator().setNormal(normal);
 
         painter.setSlicingMode(slicingMode);
 
--- a/Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
@@ -371,7 +371,6 @@ export default function widgetBehavior(publicAPI, model) {
     ) {
       return macro.VOID;
     }
-
     if (!model.point2) {
       // Update orientation to match the camera's plane
       // if the corners are not yet placed
@@ -384,8 +383,6 @@ export default function widgetBehavior(publicAPI, model) {
       model.shapeHandle.setDirection(normal);
       model.manipulator.setNormal(normal);
     }
-    model.manipulator.setOrigin(model.activeState.getOrigin());
-
     const worldCoords = model.manipulator.handleEvent(
       callData,
       model.openGLRenderWindow
--- a/Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
+++ b/Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
@@ -140,6 +140,7 @@ function readyAll() {
 function updateControlPanel(im, ds) {
   const slicingMode = im.getSlicingMode();
   const extent = ds.getExtent();
+  document.querySelector('.slice').setAttribute('min', extent[slicingMode * 2]);
   document
     .querySelector('.slice')
     .setAttribute('max', extent[slicingMode * 2 + 1]);
@@ -276,10 +277,38 @@ reader
       }
     );
 
+    const updateWidgetVisibility = (widget, slicePos, i, widgetIndex) => {
+      /* testing if the widget is on the slice and has been placed to modify visibility */
+      const widgetVisibility =
+        widget.getWidgetState().getPoint1Handle().getOrigin()[i] ===
+          slicePos[i] ||
+        !scene.widgetManager.getWidgets()[widgetIndex].getPoint1();
+      widget.setVisibility(widgetVisibility);
+    };
+
+    const updateWidgetsVisibility = (position, slicingMode) => {
+      updateWidgetVisibility(widgets.rectangleWidget, position, slicingMode, 0);
+      updateWidgetVisibility(widgets.ellipseWidget, position, slicingMode, 1);
+      updateWidgetVisibility(widgets.circleWidget, position, slicingMode, 2);
+    };
+
     const update = () => {
       const slicingMode = image.imageMapper.getSlicingMode() % 3;
 
       if (slicingMode > -1) {
+        const ijk = [0, 0, 0];
+        const position = [0, 0, 0];
+
+        // position
+        ijk[slicingMode] = image.imageMapper.getSlice();
+        data.indexToWorldVec3(ijk, position);
+
+        widgets.rectangleWidget.getManipulator().setOrigin(position);
+        widgets.ellipseWidget.getManipulator().setOrigin(position);
+        widgets.circleWidget.getManipulator().setOrigin(position);
+
+        updateWidgetsVisibility(position, slicingMode);
+
         scene.rectangleHandle.updateRepresentationForRender();
         scene.ellipseHandle.updateRepresentationForRender();
         scene.circleHandle.updateRepresentationForRender();
@@ -309,6 +338,7 @@ function resetWidgets() {
   scene.rectangleHandle.reset();
   scene.ellipseHandle.reset();
   scene.circleHandle.reset();
+  widgets[activeWidget].setVisibility(true);
   scene.widgetManager.grabFocus(widgets[activeWidget]);
 }
 
@@ -319,6 +349,7 @@ document.querySelector('.slice').addEventListener('input', (ev) => {
 document.querySelector('.axis').addEventListener('input', (ev) => {
   const sliceMode = 'IJKXYZ'.indexOf(ev.target.value) % 3;
   image.imageMapper.setSlicingMode(sliceMode);
+
   setCamera(sliceMode, scene.renderer, image.data);
   resetWidgets();
   scene.renderWindow.render();
--- a/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
@@ -272,9 +272,8 @@ export default function widgetBehavior(publicAPI, model) {
     ) {
       return macro.VOID;
     }
-
-    model.manipulator.setOrigin(model.activeState.getOrigin());
     model.manipulator.setNormal(model.camera.getDirectionOfProjection());
+
     const worldCoords = model.manipulator.handleEvent(
       callData,
       model.openGLRenderWindow
--- a/Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/EllipseWidget/behavior.js
@@ -25,6 +25,7 @@ export default function widgetBehavior(publicAPI, model) {
     const right = model.shapeHandle.getRight();
     const up = model.shapeHandle.getUp();
     const dir = model.shapeHandle.getDirection();
+
     const rightComponent = vec3.dot(diagonal, right);
     const upComponent = vec3.dot(diagonal, up);
     const dirComponent = vec3.dot(diagonal, dir);
--- a/Sources/Widgets/Widgets3D/PaintWidget/example/controlPanel.html
+++ b/Sources/Widgets/Widgets3D/PaintWidget/example/controlPanel.html
@@ -8,7 +8,7 @@
   <tr>
     <td>Slice #</td>
     <td>
-      <input class="slice" type="range" min="1" max="1" step="1">
+      <input class="slice" type="range" min="0" max="0" step="1">
     </td>
   </tr>
   <tr>
--- a/Sources/Widgets/Widgets3D/ShapeWidget/example/controlPanel.html
+++ b/Sources/Widgets/Widgets3D/ShapeWidget/example/controlPanel.html
@@ -2,7 +2,7 @@
   <tr>
     <td>Slice #</td>
     <td>
-      <input class="slice" type="range" min="1" max="1" step="1">
+      <input class="slice" type="range" min="0" max="0" step="1">
     </td>
   </tr>
   <tr>
@@ -22,7 +22,7 @@
         <option name="ellipseWidget" selected>ellipseWidget</option>
         <option name="circleWidget">circleWidget</option>
       </select>
-    </td>
+      </td>
   </tr>
   <tr>
     <td>


ocrossi:
@finetjul

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
```diff
@@ -276,10 +277,62 @@ reader
...
+          widgets.ellipseWidget.setVisibility(ellipseWidgetVisibility);
+          widgets.circleWidget.setVisibility(circleWidgetVisibility);
+        }
+      }
+    };
+
     const update = () => {
       const slicingMode = image.imageMapper.getSlicingMode() % 3;
 
+      scene.renderWindow.render();
```
finetjul, line 311: does not seem right to refresh the render window before updating the widgets

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
```diff
@@ -273,12 +273,12 @@ export default function widgetBehavior(publicAPI, model) {
       return macro.VOID;
     }
 
-    model.manipulator.setOrigin(model.activeState.getOrigin());
     model.manipulator.setNormal(model.camera.getDirectionOfProjection());
     const worldCoords = model.manipulator.handleEvent(
       callData,
       model.openGLRenderWindow
     );
+    model.manipulator.setOrigin(worldCoords);
```
finetjul, line 281: it does not seem to make sense to `handle event` with an out of date origin

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
```diff
@@ -276,10 +277,62 @@ reader
...
+              .getOrigin()[i] === position[i] ||
+            !scene.widgetManager.getWidgets()[1].getPoint1();
+          const circleWidgetVisibility =
+            widgets.circleWidget.getWidgetState().getPoint1Handle().getOrigin()[
+              i
+            ] === position[i] ||
+            !scene.widgetManager.getWidgets()[2].getPoint1();
+          widgets.rectangleWidget.setVisibility(rectangleWidgetVisibility);
+          widgets.ellipseWidget.setVisibility(ellipseWidgetVisibility);
+          widgets.circleWidget.setVisibility(circleWidgetVisibility);
```
finetjul, line 303: seems that it can be factorized... there is a lot of duplicated code here...

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
ocrossi: it is updating the slice, it allows the UI to refresh when the user drags the scroll bar and prevents a refresh when the user moves mouse on the scene after scrolling the bar

finetjul: can't it be done at the end of the update() function ?

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
ocrossi: To me it is the updated value that we  are using here, set in the update function of the example. @laurennlam How would you explain it?

finetjul:
I'm fine if @agirault is ok with the change

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
laurennlam: In the example, when the slice is moved, then the origin of the manipulator is updated with the slice center.
Then, with this new origin, we can compute/project the correct point1 and point 2.
It couldn't work before because we used the point1 of the current state. But this point was never correctly set.

agirault:
Thanks guys. Just give me a bit of time to refresh my memory on this.

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
```diff
@@ -371,29 +371,24 @@ export default function widgetBehavior(publicAPI, model) {
     ) {
       return macro.VOID;
     }
-
+    const normal = model.camera.getDirectionOfProjection();
     if (!model.point2) {
-      // Update orientation to match the camera's plane
-      // if the corners are not yet placed
-      const normal = model.camera.getDirectionOfProjection();
```
agirault, line 378: @ocrossi at first glance I don't agree with this (hence why the comment was here). After the 1st point is placed, the 2nd anchor point should be placed to keep the shape parallel to the camera. However, if the 2nd point is already placed, the shape widget plane shouldn't change based on the camera (ex: if you placed the widget on a slice and rotate the camera, that widget shouldn't be able to start facing another direction than the slice).

Can you help me understand what you're trying to do and what scenario would be fixed by that change?

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
agirault: @floryst can you refresh me on the origin of the manipulator vs origin of the points? I don't quite understand this change. See here: https://github.com/Kitware/vtk-js/pull/1717/files#diff-09c501501761db156d6f7d2d7b93d812c1fccc886aac5a921dbda49b32481362L387-R388

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
```diff
@@ -276,10 +277,52 @@ reader
...
+      for (let i = 0; i < 3; i++) {
+        if (position[i] !== 0) {
+          /* testing if the widget is on the slice and has been placed to modify visibility */
+          updateWidgetVisibility(widgets.rectangleWidget, position, i, 0);
+          updateWidgetVisibility(widgets.ellipseWidget, position, i, 1);
+          updateWidgetVisibility(widgets.circleWidget, position, i, 2);
+        }
+      }
+    };
+
```
agirault, line 298: @ocrossi do I understand correctly that this is to only show the widget representation if it's on the slice it's been initially created?

ocrossi: yes exactly

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
```diff
@@ -276,10 +277,52 @@ reader
...
+        data.indexToWorldVec3(ijk, normal);
+        vtkMath.subtract(normal, data.getOrigin(), normal);
+        vtkMath.normalize(normal);
+
+        widgets.rectangleWidget.getManipulator().setOrigin(position);
+        widgets.rectangleWidget.getManipulator().setNormal(normal);
+        widgets.ellipseWidget.getManipulator().setOrigin(position);
+        widgets.ellipseWidget.getManipulator().setNormal(normal);
+        widgets.circleWidget.getManipulator().setOrigin(position);
+        widgets.circleWidget.getManipulator().setNormal(normal);
```
agirault, line 322: I don't understand why this is necessary. The widget behavior itself should be setting the normal based on the camera, right? Can you explain?

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
ocrossi: the problem encountered here was to place the widget on the good slice. When the user would change slices in the UI the widgetBehavior would not set correct handle position and kept the slice value of when the camera was placed (while initialization of the scene or when you would change slice mode) 
This fix was meant to update the manipulator position from the example when the user changes slices. It fixed this problem. 
I did not consider the problem of changing view while placing widget.

ocrossi (commented):
This was necessary to record the slice change. The normal should change when the camera moves position but the position of the manipulator should be updated when the user changes slice.

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
ocrossi: This was necessary to record the slice change. The normal should change when the camera moves position but the position of the manipulator should be updated when the user changes slice.

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
ocrossi: You are right the orientation should keep the state of the plane facing the camera when the user places point1

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
agirault: > The normal should change when the camera moves position

So no need for `setNormal` here, only `setOrigin`?

> but the position of the manipulator should be updated when the user changes slice.

That sounds to me that the widget behavior should be able to do automatically, and that the application level shouldn't need to do that... I actually thought it did/should already, but clearly, it's not working. Would you agree?

___

_Note: I appreciate your patience in taking the time to explain this to me @ocrossi 👍_

laurennlam: > So no need for setNormal here, only setOrigin?

My guess is that you need to set the normal too, especially if you switch the view type (I, J, K), isn't it ?

>That sounds to me that the widget behavior should be able to do automatically, and that the application level shouldn't need to do that... I actually thought it did/should already, but clearly, it's not working. Would you agree?

About not setting the origin at the application level, I think it's the opposite of what @floryst thinks [here ](https://github.com/Kitware/vtk-js/issues/1718#issuecomment-773486950). But that's definitely a question we need to solves for the next widgets :)

agirault: > My guess is that you need to set the normal too, especially if you switch the view type (I, J, K), isn't it?

I don't think so, it should be handled [here](https://github.com/Kitware/vtk-js/blob/72e12774ea9e2485fa115e9ad9dc15f3e11285ba/Sources/Widgets/Widgets3D/ShapeWidget/behavior.js#L388). Also, as we've noted in both examples, the orientation of those green shape widgets itself does not seem to be the problem.

> About not setting the origin at the application level, I think it's the opposite of what @floryst thinks here. But that's definitely a question we need to solves for the next widgets :)

Ahah thanks for pointing that out @laurennlam, I missed that while catching up with the issue. Looks like we need to have a heated argument with @floryst! 🤜 🤛 Joke aside, I have no preference at the moment on doing this at the application level vs in the library, but from my understanding, all those widgets have been set up to handle things internally until now (through the behavior and manipulator objects), so we'd have to make a bigger switch here. Maybe we should all meet, or I can meet with Forrest first and report back.

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
floryst: Origin of the plane manipulator sets the plane origin. Origin of the points are the coordinates of each point.

Just as a sanity check, is the plane manipulator's origin correct prior to calling [manipulator.handleEvent here](https://github.com/Kitware/vtk-js/pull/1717/files#diff-09c501501761db156d6f7d2d7b93d812c1fccc886aac5a921dbda49b32481362R384)?

In addition, should we be updating the manipulator like this in behavior? [Recent discussions](https://github.com/Kitware/vtk-js/issues/1718#issuecomment-773171223) on this should be considered. If this widget is expected to only ever operate on a plane orthogonal to the view direction, then keeping it is fine.

ocrossi: To answer your question @floryst the manipulator's origin is correct before the 1st call to manipulator.handleEvent.
@finetjul is there a use case which is not using orthogonal planes?

ocrossi:
@agirault i think the work here is finished, apart for the manipulator's role and where it should be updated.
I have a problem understanding the role of the scale3 mixin in these widgets.
From what i red from  the [vtk.js mixin documentation]( https://kitware.github.io/vtk-js/docs/develop_widget.html) scale3 is meant to manage the size of the widget. But from what i have tested [here](https://github.com/Kitware/vtk-js/blob/1d65f61b066a00fdcb2e7da1add8f4c523f2f6ee/Sources/Widgets/Widgets3D/EllipseWidget/behavior.js#L33) it is used to orient the widget. Could you tell me the role of this mixin in the shapeWidget (see [here](https://github.com/Kitware/vtk-js/blob/1d65f61b066a00fdcb2e7da1add8f4c523f2f6ee/Sources/Widgets/Widgets3D/EllipseWidget/state.js#L28) )? I couldn't find relevant examples in the code base.

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
agirault: so we're reverting this?

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
```diff
@@ -371,29 +371,26 @@ export default function widgetBehavior(publicAPI, model) {
...
       const up = model.camera.getViewUp();
       const right = [];
+      normal[0] *= -1;
+      normal[1] *= -1;
+      normal[2] *= -1;
       vec3.cross(right, up, normal);
```
agirault, line 381: why?

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
agirault: So we're reverting this?

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
agirault, line 381: why?

Sources/Widgets/Widgets3D/PaintWidget/example/index.js
```diff
@@ -436,11 +446,15 @@ scene.circleHandle.onInteractionEvent(() => {
...
-    .getEllipseHandle()
-    .getScale3();
-  painter.paintEllipse(center, scale3);
+    .getPoint1Handle()
+    .getOrigin();
+
+  const radius = vec3.distance(center, point2);
+
+  const corner = [radius, radius, radius];
+  painter.paintEllipse(center, corner);
```
agirault, line 457: Why?

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
```diff
@@ -172,7 +173,7 @@ reader
     // default slice orientation/mode and camera view
     const sliceMode = vtkImageMapper.SlicingMode.K;
     image.imageMapper.setSlicingMode(sliceMode);
-    image.imageMapper.setSlice(0);
+    image.imageMapper.setSlice(1);
```
agirault, line 176: Do we have a bug here? Shouldn't the initial index by 0 for image mapper?

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
```diff
@@ -273,8 +273,8 @@ export default function widgetBehavior(publicAPI, model) {
       return macro.VOID;
     }
 
-    model.manipulator.setOrigin(model.activeState.getOrigin());
     model.manipulator.setNormal(model.camera.getDirectionOfProjection());
+
```
agirault, line 277: why?

agirault:
> Yet I have a problem understanding the role of the scale3 mixin in these widgets.
From what i read from the vtk.js mixin documentation scale3 is meant to manage the size of the widget. But from what i have tested here it is used to orient the widget.

I believe it's both: `rightComponent, upComponent, dirComponent` are the projections of the shape's diagonal along the 3 axis, so it's basically used to "stretch" the shape along each axis to set its size, as far as I can tell.

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
agirault: > @finetjul is there a use case which is painting/placing shapeWidgets on non orthogonal planes?

Yes. That was the goal of #1422.

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
ocrossi: Yes there is a problem here. The scroll bar allows user to browse from slice 1 to N. So if you place something on slice 0. it wont be displayed if you change slices and go back to the 1st slice.

Sources/Widgets/Widgets3D/PaintWidget/example/index.js
ocrossi: Here i've tested solutions and found that this one is working. The painting tool was broken due to the change of scale3 value. Since i have a vague comprehension of the impact of this change, i used the computations of the computeScale3 in the painting tool.

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
ocrossi: My bad on this one, these lines were for test purposes and are now removed

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
ocrossi: No, my mistake for leaving this here

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
ocrossi: This is related to the manipulator discussion. The widget works fine (as far as i can tell) without this line and from the discussions i had with @finetjul , new flavored vtk.js widgets should have an update to describe correctly the context in which they are used (e.g imageSlice, 3D scene ...)

ocrossi: ok my bad i didn't read #1422 .

ocrossi: This line plants the widget on the paint example. But when i add it after activeState is set with worldCoords (see [here](https://github.com/Kitware/vtk-js/blob/36575579ca1650c41d8575c830a7ba5edbce7bab/Sources/Widgets/Widgets3D/SplineWidget/behavior.js#L302)), all works fine. My guess is that activeState is not set properly before the 1st call to handleEvent

ocrossi: This line plants the whole widget. I'm not sure why yet, but it seems that setting the plane origin with the activeState as it was uses the value of the state (see [here](https://github.com/Kitware/vtk-js/blob/1d65f61b066a00fdcb2e7da1add8f4c523f2f6ee/Sources/Widgets/Widgets3D/SplineWidget/state.js#L18) to compute the 1st position, which messes all up since it would then use corrupted data

agirault: @floryst can you share your thoughts on this specifically?
```js
model.manipulator.setOrigin(model.activeState.getOrigin())
```

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
agirault: Can we fix the scroll bar instead so that it goes between `0` and `N-1`?
```html
<input type="range" min="-1" max="$(N-1)" ...>
```
or at least converting the `[1;N]` range to `[0;N-1]` before setting it on the mapper? Because `setSlice(N)` shouldn't work either.

Sources/Widgets/Widgets3D/PaintWidget/example/index.js
agirault: I don't understand exactly so I'll test and report back. Thanks for the explanation.

ocrossi: It is not very clear to me, apologies for the shaky explanations. I recommend testing scale3 mixin value before and after the orientation update

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
ocrossi: the max of the slice range is set [here](https://github.com/Kitware/vtk-js/blob/49a5ec41c53b6b4035a208620f26254820ae5cb2/Sources/Widgets/Widgets3D/ShapeWidget/example/index.js#L140), so i think the problem is solved just by setting the range min to 0

agirault: Or to `extent[slicingMode * 2]` (does not technically have to be 0) 🙂

ocrossi: > does not technically have to be 0
Not sure i understood you correctly here. You refer to the max or minimum value when you mention the 0? Because i think the minimum should be 0 on every minimum slice, since the example is using the slice to place the point in space, so the image mapper would'nt be a cube when extended to 3D

finetjul: @ocrossi you might want to consider the following: https://vtk.org/Wiki/VTK/Tutorials/Extents

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
agirault: ping @floryst. Also done here: https://github.com/Kitware/vtk-js/pull/1717/files#diff-09c501501761db156d6f7d2d7b93d812c1fccc886aac5a921dbda49b32481362L387-L388

Thanks!

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
agirault: In the meantime @ocrossi I see you've force pushed some changes: did you test without those `setNormal` in the application `update`, since we believed [this one](https://github.com/Kitware/vtk-js/pull/1717/files#diff-09c501501761db156d6f7d2d7b93d812c1fccc886aac5a921dbda49b32481362R382) should handle it?

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
```diff
@@ -371,10 +371,7 @@ export default function widgetBehavior(publicAPI, model) {
     ) {
       return macro.VOID;
     }
-
     if (!model.point2) {
-      // Update orientation to match the camera's plane
-      // if the corners are not yet placed
```
agirault, line 377: Can you restore those comments (unless they're inaccurate)? All those widgets are pretty confusing already, so I'd rather you add more if possible 😉 Thanks.

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
ocrossi: It works fine without the setNormal from example. The useless lines are removed

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
floryst: I think this should never happen once we have a discussion and implementation on updating manipulators based on context. I don't know how this works in the context of this widget, but usually state is updated based on the current manipulator location, not the other way around. If the activeState.getOrigin() is correct, then this is fine for now, but I would want to move away from this approach of setting manipulator parameters.

ocrossi: @agirault do you think the PR is good for merge then?

ocrossi:
ping @agirault do you think this fix is good to merge?

agirault:
Hey @ocrossi. I haven't had much time (or funding) to look at this, but I blocked some time in my calendar this afternoon to test it. I might invite @floryst to join me if he's interested in discussing some of the details we discussed.

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
```diff
@@ -276,10 +278,45 @@ reader
...
+
+        // position
+        ijk[slicingMode] = image.imageMapper.getSlice();
+        data.indexToWorldVec3(ijk, position);
+
+        // circle/slice normal
+        ijk[slicingMode] = 1;
+        data.indexToWorldVec3(ijk, normal);
+        vtkMath.subtract(normal, data.getOrigin(), normal);
+        vtkMath.normalize(normal);
```
agirault, line 312: What's l300 to l312 used for? Was that to compute the normal but you're not setting it anymore?

If yes, please revert, and note that you could access it in the future using `imageMapper.getSlicingModeNormal()`

Sources/Widgets/Widgets3D/PaintWidget/example/index.js
agirault: FYI: the changes I did here when switching to orientation did not work (my bad), so reverting as you're doing here makes sense for now, though it won't support orientation, so I created an issue to follow-up https://github.com/Kitware/vtk-js/issues/1734.

Resolving this thread in the meantime. Thanks

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
agirault: See other comments @ocrossi, but that specific issue is good: indeed, the state should not set the manipulator, but the contrary should occur. i.e. Manipulator setup + interaction = 2d screen coordinates to 3d world coordinates conversion for all the inner states of the widget.

agirault:
To summarize the discussions and next steps after this PR:
1. while it currently works fine, there are still unknowns to iron out in the future in terms of who should in charge of some properties, app vs library:
    - for the `ShapeWidget`, the normal and the up vectors are currently set in the library based on the camera [here](https://github.com/Kitware/vtk-js/blob/ffa6e50b5844bfeb877822a4c42df7f8ab0fe2b2/Sources/Widgets/Widgets3D/ShapeWidget/behavior.js#L375-L384), so a shape is always aligned with the camera (not just normal but up/right as well) when placed. To me, that makes sense and would be fine staying. After that PR, the user will need to input the origin because the view does not have access to the data content (see bullet point 2 below).
    - if we want to handle things more at the application level, the user could set up things like origin, and normal themselves (as was done [here](https://github.com/Kitware/vtk-js/pull/1717/commits/ffa6e50b5844bfeb877822a4c42df7f8ab0fe2b2#diff-b3e9ab9514537c26fdf04aadbf4511ced2630c5a9d2b73f529db956c5bf09053L285-L297))... but that's not enough: we'd need to add the up/right vector as well, or else [it would stay dependent on the camera](https://github.com/Kitware/vtk-js/blob/ffa6e50b5844bfeb877822a4c42df7f8ab0fe2b2/Sources/Widgets/Widgets3D/ShapeWidget/behavior.js#L378) accessed by the widget inside the library. I don't think using the camera in the widget's behavior is a problem, so I'd prefer leaving it in there.
    - we can continue that discussion at #1718 (though that wasn't originally what the issue was for, so we could also create a new one)
2. one of the main issues before that PR was how the `ShapeWidget` and the `SplineWidget` were resetting their manipulator origin based on the `activeState` origin ([here](https://github.com/Kitware/vtk-js/pull/1717/commits/e968c76bb1d563df54875b671b871441f00b9e91#diff-09c501501761db156d6f7d2d7b93d812c1fccc886aac5a921dbda49b32481362L387) and [here](https://github.com/Kitware/vtk-js/pull/1717/commits/e968c76bb1d563df54875b671b871441f00b9e91#diff-eb1a0809cbb10551a96850267c1a336cabc1b3ce03f01de404e214bbd95752e6L276)). That doesn't seem to make much sense since the `activeState` origin was never set, so removing that makes sense right now. However, should the manipulator be able to pick the actors in the screen to set its origin automatically (and isn't that what `enablePicking` on the widget is supposed to be for)? Then we could leave it all to the library (and not the application) to set the origin, normal, up, etc... (@floryst ?)
3. The `PaintWidget` example (which is more of a `PaintFilter`+ a lot of widgets example) did not work when using the `EllipseWidget` because the `paintEllipse` API does not support using direction vectors for the ellipse. The changes in this PR ([here](https://github.com/Kitware/vtk-js/pull/1717/commits/ffa6e50b5844bfeb877822a4c42df7f8ab0fe2b2#diff-b3e9ab9514537c26fdf04aadbf4511ced2630c5a9d2b73f529db956c5bf09053L442-R450)) revert it to work when the ellipse is axis-aligned but will be invalid for non-axis-aligned data. That can be addressed later on (see #1734)

@ocrossi thanks a lot for the effort 👏 I think if you can do the following we should be good to go:
- give a look at [my last comment ](https://github.com/Kitware/vtk-js/pull/1717#discussion_r579543940)
- _if you feel up for it_: separate your changes into individual commits with descriptive commit messages that explain the changes for:
  - removing `manipulator.setOrigin(activeState.getOrigin)`
  - replacing the inputs of `painter.paintEllipse` 
  - improving the examples (slider min value, visibility of widgets, no need to set the normals)
    - _if you REALLY feel up for it_: fix the orientation of the slices that are not consistent when changing them
- rebase on master

Almost there 💪 Good job.

ocrossi:
Thank you @agirault for all your reviews and insights on this issue, and for the reactivity to this PR (i was rushed by time as you might have guessed). The widget part of VTK.js is much clearer thanks to your advices. The work on commits to track the evolution of the issue is done.
Green light's on?

agirault:
Nice stuff @ocrossi 👏 Just rebase on master and let's merge.

agirault (approved):

PR #1721, opened 2021-02-09
fix(package.json): Use updated karma-webpack

floryst:
Also fixes the version for eslint-webpack-plugin.

This is a fix so we can force-release a new beta release.


jourdain:
lgtm

PR #1724, opened 2021-02-10
fix(rollup): Move all chunks out of Sources/

floryst:


floryst:
This cleans up more of the rollup build output.

PR #1726, opened 2021-02-11
feat(ResliceCursorWidget): Add possibility to programmatically rotate axis

laurennlam:
Add a method which allows developpers to rotate specific plane
around specific plane normal (which should be X, Y or Z, not a defined rotated axis)
Also add a test to avoid regression on viewup computation

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
@@ -118,26 +118,6 @@ export function boundPointOnPlane(p1, p2, bounds) {
   return out;
 }
 
-// Get name of the line in the same plane as the input
-export function getAssociatedLinesName(lineName) {
-  switch (lineName) {
-    case 'AxisXinY':
-      return 'AxisZinY';
-    case 'AxisXinZ':
-      return 'AxisYinZ';
-    case 'AxisYinX':
-      return 'AxisZinX';
-    case 'AxisYinZ':
-      return 'AxisXinZ';
-    case 'AxisZinX':
-      return 'AxisYinX';
-    case 'AxisZinY':
-      return 'AxisXinY';
-    default:
-      return '';
-  }
-}
-
 /**
  * Rotates a vector around another.
  * @param {vec3} vectorToBeRotated Vector to rate
@@ -274,4 +254,68 @@ export function transformPlane(
   );
   planeToTransform.rotate(angle, targetNormal);
   planeToTransform.setCenter(targetCenter);
+}
+
+// Get name of the line in the same plane as the input
+export function getAssociatedLinesName(lineName) {
+  switch (lineName) {
+    case 'AxisXinY':
+      return 'AxisZinY';
+    case 'AxisXinZ':
+      return 'AxisYinZ';
+    case 'AxisYinX':
+      return 'AxisZinX';
+    case 'AxisYinZ':
+      return 'AxisXinZ';
+    case 'AxisZinX':
+      return 'AxisYinX';
+    case 'AxisZinY':
+      return 'AxisXinY';
+    default:
+      return '';
+  }
+}
+
+/**
+ * Get the line name, constructs from the plane name and where the plane is displayed
+ * Example: planeName='X' rotatedPlaneName='Y', then the return values will be 'AxisXinY'
+ * @param {String} planeName Value between 'X', 'Y' and 'Z'
+ * @param {String} rotatedPlaneName Value between 'X', 'Y' and 'Z'
+ * @returns {String}
+ */
+export function getLineNameFromPlaneAndRotatedPlaneName(
+  planeName,
+  rotatedPlaneName
+) {
+  return `Axis${planeName}in${rotatedPlaneName}`;
+}
+
+/**
+ * Extract the plane name from the line name
+ * Example: 'AxisXinY' will return 'X'
+ * @param {String} lineName Should be following this template : 'Axis_in_' with _ a character
+ * @returns {String} Value between 'X', 'Y' and 'Z' or null if an error occured
+ */
+export function getPlaneNameFromLineName(lineName) {
+  const match = lineName.match('([XYZ])in[XYZ]');
+  if (match) {
+    return match[1];
+  }
+  return null;
+}
+
+/**
+ * Get the orthogonal plane name of 'planeName' in a specific 'rotatedPlaneName'
+ * Example: planeName='X' on rotatedPlaneName='Z', then the associated plane name
+ * of 'X' plane is 'Y'
+ * @param {String} planeName
+ * @param {String} rotatedPlaneName
+ */
+export function getAssociatedPlaneName(planeName, rotatedPlaneName) {
+  const lineName = getLineNameFromPlaneAndRotatedPlaneName(
+    planeName,
+    rotatedPlaneName
+  );
+  const associatedLine = getAssociatedLinesName(lineName);
+  return getPlaneNameFromLineName(associatedLine);
 }
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
@@ -6,11 +6,11 @@ import vtkCubeSource from 'vtk.js/Sources/Filters/Sources/CubeSource';
 import vtkCutter from 'vtk.js/Sources/Filters/Core/Cutter';
 import vtkPlane from 'vtk.js/Sources/Common/DataModel/Plane';
 import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
+import vtkMatrixBuilder from 'vtk.js/Sources/Common/Core/MatrixBuilder';
 
 import { ViewTypes } from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
-import { defaultViewUpFromViewType } from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/Constants';
 
-import vtkMatrixBuilder from 'vtk.js/Sources/Common/Core/MatrixBuilder';
+const EPSILON = 0.00001;
 
 /**
  * Fit the plane defined by origin, p1, p2 onto the bounds.
@@ -67,7 +67,6 @@ export function boundPlane(bounds, origin, p1, p2) {
 export function boundPoint(inPoint, v1, v2, bounds) {
   const absT1 = v1.map((val) => Math.abs(val));
   const absT2 = v2.map((val) => Math.abs(val));
-  const epsilon = 0.00001;
 
   let o1 = 0.0;
   let o2 = 0.0;
@@ -80,10 +79,10 @@ export function boundPoint(inPoint, v1, v2, bounds) {
     const absT = useT1 ? absT1 : absT2;
 
     if (inPoint[i] < bounds[i * 2]) {
-      axisOffset = absT[i] > epsilon ? (bounds[2 * i] - inPoint[i]) / t[i] : 0;
+      axisOffset = absT[i] > EPSILON ? (bounds[2 * i] - inPoint[i]) / t[i] : 0;
     } else if (inPoint[i] > bounds[2 * i + 1]) {
       axisOffset =
-        absT[i] > epsilon ? (bounds[2 * i + 1] - inPoint[i]) / t[i] : 0;
+        absT[i] > EPSILON ? (bounds[2 * i + 1] - inPoint[i]) / t[i] : 0;
     }
 
     if (useT1) {
@@ -139,19 +138,13 @@ export function getAssociatedLinesName(lineName) {
   }
 }
 
-export function getViewPlaneNameFromViewType(viewType) {
-  switch (viewType) {
-    case ViewTypes.YZ_PLANE:
-      return 'X';
-    case ViewTypes.XZ_PLANE:
-      return 'Y';
-    case ViewTypes.XY_PLANE:
-      return 'Z';
-    default:
-      return '';
-  }
-}
-
+/**
+ * Rotates a vector around another.
+ * @param {vec3} vectorToBeRotated Vector to rate
+ * @param {vec3} axis Axis to rotate around
+ * @param {Number} angle Angle in radian
+ * @returns The rotated vector
+ */
 export function rotateVector(vectorToBeRotated, axis, angle) {
   const rotatedVector = [...vectorToBeRotated];
   vtkMatrixBuilder.buildFromRadian().rotate(angle, axis).apply(rotatedVector);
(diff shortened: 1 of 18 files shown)


laurennlam:
@finetjul You can start the review. I'm not sure that the test will convince you. Maybe we can talk about it after a first review.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -313,6 +314,43 @@ function vtkResliceCursorWidget(publicAPI, model) {
     publicAPI.modified();
   };
 
+  publicAPI.rotatePlane = (rotatedPlaneName, rotateAngle, rotatedAxisName) => {
```
finetjul, line 317: rotatedPlaneName -> planeToRotate

finetjul, line 317: Please add documentation

finetjul, line 317: rotatedAxisName -> axisOfRotation ? (not sure about that though, it depends what the function documentation will say :-) )

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -313,6 +314,43 @@ function vtkResliceCursorWidget(publicAPI, model) {
...
+        (rotatedPlaneName === 'X' && rotatedAxisName === 'Y') ||
+        (rotatedPlaneName === 'Y' && rotatedAxisName === 'X')
+      ) {
+        associatedRotatedPlaneName = 'Z';
+      } else if (
+        (rotatedPlaneName === 'X' && rotatedAxis === 'Z') ||
+        (rotatedPlaneName === 'Z' && rotatedAxis === 'X')
+      ) {
+        associatedRotatedPlaneName = 'Y';
+      }
```
finetjul, line 337: I thought that you already have a convenient function that gives you the associated plane names. Don't you ?
If not, it might be useful to expose it as a convenient function.

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
```diff
@@ -0,0 +1,373 @@
...
+          obj.renderWindow.render();
+        });
+      }
+
+      /**
+       * Update the reslice view and compare it to ground truthes values
+       * @param {Number} viewId Defines the id of compared view. Id is used to get the correct viewAttributes object
+       * @param {Object} comparisonValues Contains ground truthes values that will be compared (focalPoint, viewUp, origin, point1, point2)
+       */
+      function updateAndCompareView(viewId, comparisonValues) {
```
finetjul, line 225: comparisonValues -> expectedValues

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
```diff
@@ -0,0 +1,373 @@
...
+        );
+        t.ok(vtkMath.areEquals(out.origin, comparisonValues.origin, tolerance));
+        t.ok(vtkMath.areEquals(out.point1, comparisonValues.point1, tolerance));
+        t.ok(vtkMath.areEquals(out.point2, comparisonValues.point2, tolerance));
+      }
+
+      // ----------------------------------------------------------------------
+      // Check initialization
+      // Check X view
+      updateAndCompareView(1, {
```
finetjul, line 261: instead of 1, it would be more readable to specify "YZ_PLANE"

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
```diff
@@ -0,0 +1,373 @@
...
+        origin: [0, 358.59375, 165],
+        point1: [358.59375, 358.59375, 165],
+        point2: [0, 9.466330862652142e-30, 165.00000000000006],
+      });
+
+      // ----------------------------------------------------------------------
+      // Test rotate Z around Y
+      widget.rotatePlane(
+        'X',
+        vtkMath.radiansFromDegrees(44.99999999999998),
```
finetjul, line 289: why not 45 ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
```diff
@@ -0,0 +1,373 @@
...
+        'Y'
+      );
+      updateViews(true);
```
finetjul, line 292: why is it needed ? `updateAndCompareView` will call `updateReslice` on the views also.

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
```diff
@@ -0,0 +1,373 @@
...
+        ],
+      });
+
+      // ----------------------------------------------------------------------
+      // Test rotate Y around Z
+      // Simulate increment of 5, seven times to have 35°
+      for (let i = 0; i < 7; i++) {
+        widget.rotatePlane(
+          'Y',
+          vtkMath.radiansFromDegrees(4.99999999999999),
```
finetjul, line 328: why not 5 ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: The method I have is a method which give me the associated axis (axisXinY or AxisYinZ,...) but not the plane. I'll add a function.

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
laurennlam: Because when I apply a rotation of 45° and then, compute the angle between old and new plane normal, the display angle is 49.99999. I just wanted to be as close as possible of the display value I had. (ame for 4.99999).

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
laurennlam: Indeed, here we don't need it as we test three views. If we had tested less than the three views, then we'll have to update  the whole views. Else, the viewup computation will fail.

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
finetjul: A `tolerance` is used for this exact reason.

44.99999999999998 and 45 are the same numbers, you just have precision errors due to floating point representation.

laurennlam: yes. I just tried and it worked well.

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -309,3 +309,26 @@ export function transformPlane(planeToTransform, targetPlane, viewType) {
...
 }
+
+/**
+ * Get the orthogonal plane name of 'planeName' in a specific 'rotatedPlaneName'
+ * Example: planeName='X' on rotatedPlaneName='Z', then the associated plane name
+ * of 'X' plane is 'Y'
+ * @param {String} planeName
+ * @param {String} rotatedPlaneName
+ */
+export function getAssociatedPlaneName(planeName, rotatedPlaneName) {
```
finetjul, line 320: `getAssociatedLinesName` and `getAssociatedPlaneName` do the same. Plane association should be computed in only one function (no code duplication !).

It is therefore a matter of conversion. e.g. if you had `getLineNameFromPlaneName(planeName, rotatedPlaneName)` and `getPlaneNameFromLineName(lineName)`, then:
```
export function getAssociatedPlaneName(planeName, rotatedPlaneName) {
  const lineName = getLineNameFromPlaneName(planeName, rotatedPlaneName);
  const associatedLineName = getAssociatedLinesName(lineName);
  return  getPlaneNameFromLineName(associatedLineName);
```

You can also decide to do the opposite (it might actually make more sense): keep `getAssociatedPlaneName` but make `getAssociatedLinesName` call it.

finetjul: It is not clear to me when you should use a plane name (X, Y, Z) and when you use a view type (YZ_PLANE, XZ_PLANE, XY_PLANE). It would be great to have a rule and stick to it.

finetjul: Is it `getAssociatedPlaneName`  or `getAssociatedViewPlaneName` ? (see `getViewPlaneNameFromViewType`).

I do not have a preference, but we should be consistent.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -313,6 +315,41 @@ function vtkResliceCursorWidget(publicAPI, model) {
...
+        associatedNormal,
+        rotatedAxis,
+        rotateAngle
+      );
+      model.widgetState[`set${associatedplaneToRotate}PlaneNormal`](
+        newAssociatedNormal
+      );
+    }
+    updateState(model.widgetState);
+    publicAPI.modified();
```
finetjul, line 350: It is not clear to me when modified should be called after calling `updateState`. Sometimes it is called, sometimes not. What's the rule ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
```diff
@@ -0,0 +1,365 @@
...
+  const gc = testUtils.createGarbageCollector(t);
+  t.ok('Widgets', 'ResliceCursorWidgets rendering');
+
+  // --------------------------------------------------------------------------
+  // Create constants
+  // --------------------------------------------------------------------------
+  const viewTypes = [
+    ViewTypes.XZ_PLANE,
+    ViewTypes.YZ_PLANE,
+    ViewTypes.XY_PLANE,
```
finetjul, line 84: I think it would make sense to create some convenient constants in Constant.js:
```
const xyzToViewType = [ 
  ViewTypes.YZ_PLANE,
  ViewTypes.XZ_PLANE,
  ViewTypes.XY_PLANE
];
const viewTypeToXYZ = { 
  ViewTypes.YZ_PLANE: 0,
  ViewTypes.XZ_PLANE: 1,
  ViewTypes.XY_PLANE: 2
];
const viewTypeToViewName = { 
  ViewTypes.YZ_PLANE: "X",
  ViewTypes.XZ_PLANE: "Y"),
  ViewTypes.XY_PLANE: "Z"
];
const viewNameToViewType = { 
  "X": ViewTypes.YZ_PLANE,
  "Y": ViewTypes.XZ_PLANE,
  "Z": ViewTypes.XY_PLANE
];
```

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
```diff
@@ -0,0 +1,365 @@
...
+          viewType,
+          reslice: viewObj.reslice,
+          actor: viewObj.resliceActor,
+          renderer: viewObj.renderer,
+          resetFocalPoint: false,
+          keepFocalPointPosition: true,
+          computeFocalpointOffset: false,
+          resetViewUp: false,
+        });
+        const tolerance = 1e-4;
```
finetjul, line 238: declare at the top of the file and uppercase  

Optionally rename to EPSILON (check the rest of the VTK.JS code base to find consistency)

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
```diff
@@ -0,0 +1,365 @@
...
+import vtkResliceCursorWidget from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget';
+import vtkWidgetManager from 'vtk.js/Sources/Widgets/Core/WidgetManager';
+
+import {
+  ViewTypes,
+  CaptureOn,
+} from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
+
+import { getViewPlaneNameFromViewType } from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers';
+
```
finetjul, line 23: Please explain how you got all the expected values.

This will be useful whenever the expected behavior must be changed and the test be updated.

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
laurennlam: To be honest, I thought it was to update the rendering. But it seems to work without. Especially if you update the other views...

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
laurennlam: That's seems risky if anyone, for any reason, change the order of the xyzToViewType. And I don't understand the link between xyzToViewType and viewTypeToXYZ.

finetjul: There is no reason to change the order.

I do not understand what you do not understand :-)
```
xyzToViewType[0] ->  ViewTypes.YZ_PLANE
xyzToViewType[ViewTypes.YZ_PLANE] ->  0
```

laurennlam: Why do we want to use indexes as it's only used in the example ?

finetjul: It is typically a convenient way to deal with axes.

It is actually used in both the example and the test. therefore it makes sense to expose it as conveniency

Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testMultipleRotationPlanes.js
```diff
@@ -0,0 +1,382 @@
...
+        point1: [358.59375, 179.296875, 0],
+        point2: [0, 179.296875, 330],
+      });
+      // Check Z view
+      updateAndCompareView(ViewTypes.XY_PLANE, {
+        focalPoint: [179.296875, 179.296875, 165],
+        viewUp: [0, -1, 1.5851759508024892e-16],
+        origin: [0, 358.59375, 165],
+        point1: [358.59375, 358.59375, 165],
+        point2: [0, 9.466330862652142e-30, 165.00000000000006],
```
finetjul, line 292: `[0, 9.466330862652142e-30, 165.00000000000006]` -> `[0, 0, 165]`

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -309,3 +275,67 @@ export function transformPlane(planeToTransform, targetPlane, viewType) {
...
+ * Example: 'AxisXinY' will return 'X'
+ * @param {String} lineName Should be following this template : 'Axis_in_' with _ a character
+ * @returns {String} Value between 'X', 'Y' and 'Z'
+ */
+export function getPlaneNameFromLineName(lineName) {
+  const match = lineName.match('([XYZ])in[XYZ]');
+  if (match) {
+    return match[1];
+  }
+  return '';
```
finetjul, line 324: I think returning `undefined` or `null` is more appropriate. It carries a message that a problem occured

PR #1727, opened 2021-02-12
feat(vtkimagereslice): create a slab in image reslice

gandis0713:
To create slab in ImageResclie, getCompositeFunc was implemented by referring to vtkGetCompositeFunc implemented in [VTK C++ Implementation](https://github.com/Kitware/VTK/blob/master/Imaging/Core/vtkImageReslice.cxx#L1555).

### DIFF
--- a/Sources/Imaging/Core/ImageReslice/index.js
+++ b/Sources/Imaging/Core/ImageReslice/index.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-param-reassign */
 import { vec4, mat4 } from 'gl-matrix';
 
 import macro from 'vtk.js/Sources/macro';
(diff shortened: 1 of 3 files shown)


finetjul:
Does it work ? 

If so, I do not understand how. Can you please update the ImageReslice example  ?

You can't deal with pointers/typedarray (e.g. tmpPtr) the same way as in C++.

See how pointers/typedarray are used in the rest of the file (e.g. usage of `subarray`, `fill`...)

Sources/Imaging/Core/ImageReslice/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-param-reassign */
```
finetjul, line 1: if you had to add this, it is a sign that something is not right...

gandis0713:
I checked the implementation(second commit) using ResliceCursorWidget Example. currently, It looks normal to me.

Sources/Imaging/Core/ImageReslice/index.js
```diff
@@ -30,6 +31,95 @@ function vtkImageReslice(publicAPI, model) {
...
+    do {
+      let result = subTmpPtr[0] * 0.5;
+      for (let k = n - 1; k !== 0; --k) {
+        subTmpPtr = tmpPtr.subarray(inComponents);
+        result += subTmpPtr[0];
+      }
+      subTmpPtr = tmpPtr.subarray(inComponents);
+      result += subTmpPtr[0] * 0.5;
+      tmpPtr.fill(result * f, 0, 1);
+      tmpPtr = tmpPtr.subarray(1);
```
gandis0713, line 47: According to the C++ implementation, this part seemed necessary.
I think this part seems to require a reassign to tmpPtr.
If there are other way, I will apply them. Or it may have been applied unnecessarily.
@finetjul , Could you please check?

finetjul, line 47: I think it should be simplified into 
```
result += tmpPtr[inComponents] * 0.5;
tmpPtr[0] = result * f;
tmpPtr = tmpPtr.subarray(1);
```

I wonder if there is an advantage to create subarrays when iterating one by one.
Instead you could do something like that (I have not double checked if it is the same as the c++ version):
```
for (let j = 0; j < inComponents; j += 1) {
  let result = tmpPtr[j] * 0.5;
  for (let k = 1; k < n; k += 1) {
     result += tmpPtr[j+k];
  }
  result += tmpPtr[j+n] * 0.5;  
  tmpPtr[j] = result;
}
```

Sources/Imaging/Core/ImageReslice/index.js
```diff
@@ -32,68 +31,45 @@ function vtkImageReslice(publicAPI, model) {
...
     const n = sampleCount - 1;
-    do {
-      let result = subTmpPtr[0] * 0.5;
-      for (let k = n - 1; k !== 0; --k) {
-        subTmpPtr = tmpPtr.subarray(inComponents);
-        result += subTmpPtr[0];
+    for (let i = 0; i < inComponents; i += 1) {
+      let result = tmpPtr[i] * 0.5;
+      for (let j = 1; j < n; j += 1) {
+        result += tmpPtr[i + j];
```
finetjul, line 38: isn't it `tmpPtr[i + j*inComponents]` instead ?

Sources/Imaging/Core/ImageReslice/index.js
```diff
@@ -32,68 +31,45 @@ function vtkImageReslice(publicAPI, model) {
...
-      do {
-        subTmpPtr = tmpPtr.subarray(inComponents);
-        result += subTmpPtr[0];
-      } while (--k);
-      tmpPtr.fill(result * f, 0, 1);
-      tmpPtr = tmpPtr.subarray(1);
-    } while (--m);
+    for (let i = 0; i < inComponents; i += 1) {
+      let result = tmpPtr[i];
+      for (let j = 1; j < sampleCount; j += 1) {
```
finetjul, line 48: what happened to `--n` ?

Sources/Imaging/Core/ImageReslice/index.js
```diff
@@ -32,68 +31,45 @@ function vtkImageReslice(publicAPI, model) {
...
-        subTmpPtr = tmpPtr.subarray(inComponents);
-        result = result > subTmpPtr[0] ? result : subTmpPtr[0];
-      } while (--k);
-      tmpPtr.fill(result, 0, 1);
-      tmpPtr = tmpPtr.subarray(1);
-    } while (--m);
+    for (let i = 0; i < inComponents; i += 1) {
+      let result = tmpPtr[i];
+      for (let j = 1; j < sampleCount; j += 1) {
+        result = result > tmpPtr[i + j] ? result : tmpPtr[i + j];
```
finetjul, line 69: `result = Number.max(result, tmpPtr[i + j])`
or
``` 
if (result > tmpPtr[i + j]) {
  result = tmpPtr[i + j]);
}
```

finetjul (commented):
It would be great to add to the ResliceCursorWidget example a checkbox/slider to control the slab thickness.

Sources/Imaging/Core/ImageReslice/index.js
gandis0713: Can I see why --n need to be?

Sources/Imaging/Core/ImageReslice/index.js
gandis0713: Yes. It must be modified.

Sources/Imaging/Core/ImageReslice/index.js
finetjul: nevermind, you are right, it is not needed.

Sources/Imaging/Core/ImageReslice/index.js
gandis0713: I modified it.

Sources/Imaging/Core/ImageReslice/index.js
gandis0713: I modified it.

Sources/Imaging/Core/ImageReslice/index.js
```diff
@@ -30,6 +30,72 @@ function vtkImageReslice(publicAPI, model) {
...
   let optimizedTransform = null;
 
+  function getImageResliceSlabTrap(tmpPtr, inComponents, sampleCount, f) {
+    const n = sampleCount - 1;
+    for (let i = 0; i < inComponents; i += 1) {
+      let result = tmpPtr[i] * 0.5;
+      for (let j = 1; j < n; j += 1) {
+        result += tmpPtr[i + j * inComponents];
+      }
+      result += tmpPtr[i + n] * 0.5;
```
finetjul, line 40: should also be `i + n * inComponents`

gandis0713:
@finetjul 
I'm trying to add a checkbox/slider to control the slab thickness to the ResliceCursorWidget example.

gandis0713:
I added combobox/slider to control slab mode and number of slices to ResliceCursorWidget example.

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/controlPanel.html
```diff
@@ -18,9 +18,24 @@
...
+        <option id="slabModeMax">MAX</option>
+        <option id="slabModeMean" selected="selected">MEAN</option>
+        <option id="slabModeSum">SUM</option>
+      </select>
     </td>
   </tr>
-
+  <tr>
+    <td>Slab Number of Slices :</td>
+    <td><input id='slabNumber' type="range" min="1" max="10" step="1" value="1" style="width: 100px;"/></td>
```
finetjul, line 33: Can you please increase max value to 100 (or ideally the input volume diagonal length)

gandis0713: I modified that set initial max value to 100. After loading volume, set as input volume dimension diagonal length.

finetjul: This is great ! Thanks, I'm going to merge it. Great job, thanks for your contribution !

Concerning the performance, this is what I feared, as the number of slices increases, the performance get poor. I do not think there is much we can do with ImageReslice (CPU).

Doing it on the GPU would definitely increase performance. The idea would be to replace ImageReslice + ImageMapper with VolumeMapper (which also has slab support).

PR #1729, opened 2021-02-16
feat(vtkrenderwindowinteractor): added mouse enter and mouse leave ev…

gandis0713:
How about adding two events to RenderWindowInteractor to use mouse enter and mouse leave events in InteractorStyle or Widget?

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -28,6 +28,8 @@ const handledEvents = [
   'StartAnimation',
   'Animation',
   'EndAnimation',
+  'MouseEnter',
+  'MouseLeave',
   'StartMouseMove',
   'MouseMove',
   'EndMouseMove',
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -208,8 +210,9 @@ function vtkRenderWindowInteractor(publicAPI, model) {
         model.container[invMethod]('mousemove', publicAPI.handleMouseMove);
       }
 
+      rootElm[method]('mouseenter', publicAPI.handleMouseEnter);
+      rootElm[method]('mouseleave', publicAPI.handleMouseLeave);
```
jourdain, line 214: They should not be added to the document.

jourdain (changes requested):
Since you add/remove the handler in the bind/unbind you should not do it at the document level inside the `interactionRegistration` method.

Sources/Rendering/Core/RenderWindowInteractor/index.js
gandis0713: Removed them in `interactionRegistration`.

jourdain:
Thanks for your contribution...

PR #1736, opened 2021-02-22
Fix reslice cursor rotation

finetjul:


finetjul:
@laurennlam test is failing, can you check the expected values ?

Sources/Common/Core/Math/index.js
```diff
@@ -711,6 +718,26 @@ export function areEquals(a, b, eps = 1e-6) {
 
 export const areMatricesEqual = areEquals;
 
+export function roundNumber(num, digits = 0) {
+  if (!`${num}`.includes('e')) {
+    return +`${Math.round(`${num}e+${digits}`)}e-${digits}`;
+  }
+  const arr = `${num}`.split('e');
+  let sig = '';
+  if (+arr[1] + digits > 0) {
```
laurennlam, line 727: if there is only one 'e', I suppose it will fail ?

laurennlam:
> @laurennlam test is failing, can you check the expected values ?

I think we have to setup new value according to the new implementation as the previous one multiply  error. I'll update the last tests values.

laurennlam:
> > @laurennlam test is failing, can you check the expected values ?
> 
> I think we have to setup new value according to the new implementation as the previous one multiply error. I'll update the last tests values.

In order to be more explicit, the test failed when we apply several times a rotation (which accumulate errors). So, the expected values are now bad with your fix.

finetjul:
I think what could make sense is to do the following:
```
rotate(45)
7x rotate(5)
rotate(-7*5)
-> check here that we have same result than after `rotate(45)`
```

PR #1741, opened 2021-02-23
Add Surface LIC mapper

boucaud:
This is a port of the VTK Surface LIC implementation to vtk-js.

Screenshot of the example:
    
DeepinScreenshot_select-area_20210223134002

### DIFF
--- a/Sources/Rendering/Core/SurfaceLICInterface/Constants.js
+++ b/Sources/Rendering/Core/SurfaceLICInterface/Constants.js
@@ -1,0 +1,22 @@
+export const ContrastEnhanceMode = {
+  NONE: 0,
+  LIC: 1,
+  COLOR: 2,
+  BOTH: 3,
+};
+
+export const NoiseType = {
+  UNIFORM: 0,
+  GAUSSIAN: 1,
+};
+
+export const ColorMode = {
+  BLEND: 0,
+  MULTIPLY: 1,
+};
+
+export default {
+  ColorMode,
+  ContrastEnhanceMode,
+  NoiseType,
+};
--- a/Sources/Rendering/Core/SurfaceLICInterface/index.js
+++ b/Sources/Rendering/Core/SurfaceLICInterface/index.js
@@ -1,0 +1,104 @@
+import macro from 'vtk.js/Sources/macro';
+
+import {
+  ColorMode,
+  ContrastEnhanceMode,
+  NoiseType,
+} from 'vtk.js/Sources/Rendering/Core/SurfaceLICInterface/Constants';
+
+function vtkSurfaceLICInterface(publicAPI, model) {
+  model.classHierarchy.push('vtkSurfaceLICInterface');
+}
+
+const DEFAULT_VALUES = {
+  enableLIC: false,
+  nuberOfSteps: 40,
+  stepSize: 0.25,
+  transformVectors: true,
+  normalizeVectors: true,
+  maskOnSurface: false,
+  maskThreshold: 0.0,
+  maskColor: [0.0, 0.0, 0.0],
+  maskIntensity: 0.0,
+  enhancedLIC: true,
+
+  enhanceContrast: ContrastEnhanceMode.NONE,
+
+  lowLICContrastEnhancementFactor: 0.0,
+  highLICContrastEnhancementFactor: 0.0,
+  lowColorContrastEnhancementFactor: 0.0,
+  highColorContrastEnhancementFactor: 0.0,
+
+  antiAlias: 0,
+  colorMode: ColorMode.BLEND,
+  LICIntensity: 1.0,
+  mapModeBias: 0.0,
+
+  noiseTextureSize: 200,
+  noiseTextureType: NoiseType.GAUSSIAN,
+  noiseGrainSize: 8,
+  noiseImpulseProbability: 0.1,
+  noiseImpulseBackgroundValue: 0.0,
+  noiseGeneratorSeed: 0,
+
+  minNoiseValue: 0.0,
+  maxNoiseValue: 1.0,
+  numberOfNoiseLevels: 2,
+
+  shadersNeedBuilding: true,
+  reallocateTextures: true,
+  rebuildNoiseTexture: false,
+
+  viewPortScale: 1.0,
+};
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  macro.obj(publicAPI, model);
+
+  macro.setGet(publicAPI, model, [
+    'enableLIC',
+    'numberOfSteps',
+    'stepSize',
+    'normalizeVectors',
+    'transformVectors',
+    'maskOnSurface',
+    'maskThreshold',
+    'maskColor',
+    'maskIntensity',
+    'enhancedLIC',
+    'enhanceContrast',
+    'lowLICContrastEnhancementFactor',
+    'highLICContrastEnhancementFactor',
+    'lowColorContrastEnhancementFactor',
+    'highColorContrastEnhancementFactor',
+    'antiAlias',
+    'colorMode',
+    'LICIntensity',
+    'mapModeBias',
+
+    'noiseTextureSize',
+    'noiseTextureType',
+    'noiseGrainSize',
+    'minNoiseValue',
+    'maxNoiseValue',
+    'numberOfNoiseLevels',
+    'noiseImpulseProbability',
+    'noiseImpulseBackgroundValue',
+    'noiseGeneratorSeed',
+    'viewPortScale',
+    'rebuildNoiseTexture',
+  ]);
+
+  // Object methods
+  vtkSurfaceLICInterface(publicAPI, model);
+}
+
+// ----------------------------------------------------------------------------
+
+export const newInstance = macro.newInstance(extend, 'vtkSurfaceLICInterface');
+
+// ----------------------------------------------------------------------------
+
+export default { newInstance, extend };
--- a/Sources/Rendering/Core/SurfaceLICMapper/index.js
+++ b/Sources/Rendering/Core/SurfaceLICMapper/index.js
@@ -1,0 +1,49 @@
+import macro from 'vtk.js/Sources/macro';
+import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
+import vtkSurfaceLICInterface from 'vtk.js/Sources/Rendering/Core/SurfaceLICInterface';
+
+// ----------------------------------------------------------------------------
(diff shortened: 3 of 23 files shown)


boucaud:
@finetjul I'm not sure who should review this. WDYT ?

Sources/Rendering/LICOpenGL/LineIntegralConvolution2D/index.js
```diff
@@ -0,0 +1,563 @@
...
+      texture,
+      size,
+      context,
+      openGLRenderWindow,
+      4
+    );
+
+    let min = values[0];
+    let max = values[0];
+    for (let i = 0; i < values.length; i += 4) {
```
finetjul, line 114: shouldn't you start at `i = 4` ?

Sources/Rendering/LICOpenGL/LineIntegralConvolution2D/index.js
```diff
@@ -0,0 +1,563 @@
...
+          max = val;
+        }
+      }
+    }
+
+    return { min, max };
+  };
+
+  publicAPI.getComponentSelectionProgram = (ids) => {
+    // swizles at 45,46
```
finetjul, line 131: what does it mean ?

Sources/Rendering/LICOpenGL/LineIntegralConvolution2D/index.js
```diff
@@ -0,0 +1,563 @@
...
+    vectorTexture,
+    maskVectorTexture,
+    noiseTexture,
+    openGLRenderWindow,
+    options
+  ) => {
+    model.openGLRenderWindow = openGLRenderWindow;
+    model.context = openGLRenderWindow.getContext();
+    Object.assign(model, options);
+    if (size[0] <= 0.0 || size[1] <= 0) {
```
finetjul, line 191: why test and `0.0` and `0` ?

Sources/Rendering/LICOpenGL/LineIntegralConvolution2D/index.js
```diff
@@ -0,0 +1,563 @@
...
+        model.step1Min = minMax.min;
+        model.step1Max = minMax.max;
+      }
+
+      let { min, max } = minMax;
+      if (max <= min || max > 1.0 || min < 0) {
+        console.error('Invalid color range: ', min, max);
+        min = 0.0;
+        max = 1.0;
+      }
```
finetjul, line 485: to factorize with line 340

Sources/Rendering/LICOpenGL/SurfaceLICInterface/index.js
```diff
@@ -0,0 +1,727 @@
...
+    min,
+    max
+  ) {
+    const N = 2048;
+    const impulseProb = Math.max(0.0, Math.min(1.0, noiseImpulseProbability));
+    const noise = Float32Array.from({ length: length * length }, () => {
+      let val = 0;
+      if (impulseProb === 1.0 || Math.random() > 1.0 - impulseProb) {
+        for (let i = 0; i < N; ++i) {
+          val += Math.random();
```
finetjul, line 129: why not `N * Math.random()` ?

finetjul (commented):
There is no existing file that can be used instead of adding a new file ?

martinken:
Wow, that's awesome!  Two thoughts

1) could you move the constants and related set/get methods into the Rendering/Core LIC class? Then the OpenGL class would get them from the Rendering/Core class. If we have create a webGPU version then it would do the same thing and get the values from the Rendering/Core class.

2) This approach is a bit different in that it requires the app to instantiate an OpenGL specific mapper, for now that is probably fine but normally we let the factory mechanism create the View specific mapper (OpenGL in this case) as part of the build pass.

Sources/Rendering/LICOpenGL/LineIntegralConvolution2D/index.js
```diff
@@ -0,0 +1,563 @@
...
+      context,
+      openGLRenderWindow,
+      4
+    );
+
+    let min = values[0];
+    let max = values[0];
+    for (let i = 0; i < values.length; i += 4) {
+      // Make sure the current pixel is inside the rendered geometry (g === 0)
+      if (values[i - 0 + 1] === 0.0) {
```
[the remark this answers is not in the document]
boucaud: values.length is always a multiple of 4, so this should be fine. (I'm removing that -0 though)

jourdain:
Any reason to create a new directory `Sources/Rendering/LICOpenGL` rather than using the existing one `Sources/Rendering/OpenGL`? 

The `OpenGL` directory aims to capture the full `WebGL` backend, therefore having more than one, can make things confusing especially once Ken will add a new one... 

I get why it was done in the first place, but now that we put it back to the mainstream we should make sure it blends in. ;-)

Sources/Rendering/LICOpenGL/LineIntegralConvolution2D/index.js
boucaud: I'm not sure why, wouldn't that skip the first pixel ?

Sources/Rendering/LICOpenGL/LineIntegralConvolution2D/index.js
boucaud: That was a debugging comment, I removed it.

Sources/Rendering/LICOpenGL/LineIntegralConvolution2D/index.js
finetjul: first pixel is already handled line 112 and 113

boucaud: I see. That's actually a mistake, as the first pixel should not be handled if it's out of the geometry. I replaced with MAX_VALUE and MIN_VALUE.

Sources/Rendering/LICOpenGL/SurfaceLICInterface/index.js
boucaud: Don't we need more than 1 sample to estimate a Gaussian distribution ?

boucaud:
> There is no existing file that can be used instead of adding a new file ?

I didn't find one. This new file is from the VTK test data.

> 1. could you move the constants and related set/get methods into the Rendering/Core LIC class? Then the OpenGL class would get them from the Rendering/Core class. If we have create a webGPU version then it would do the same thing and get the values from the Rendering/Core class.

Do you think adding a `vtkSurfaceLICInterface class in Core will all the set/get parameters, but that doesn't handle anything WebGL, then a vtkOpenGLSurfaceLICInterface that handles the WebGL resources, as it does currently, could be a good way to do this ?

Moving all these parameters directly to the core mapper complicates things if there are multiple mapper instances to configure, and currently SurfaceLICInterface allows for sharing of WebGL resources.

> 2. This approach is a bit different in that it requires the app to instantiate an OpenGL specific mapper, for now that is probably fine but normally we let the factory mechanism create the View specific mapper (OpenGL in this case) as part of the build pass.

My understanding is that this is currently not the case (vtkSurfaceLICMapper is a vtkMapper that is not OpenGL specific, and vtkOpenGLSurfaceLICMapper is already instantiated by the ViewNodeFactory). Did I miss something ?

> Any reason to create a new directory `Sources/Rendering/LICOpenGL` rather than using the existing one `Sources/Rendering/OpenGL`?

Actually, I did this because that's the way directories are organised in the VTK implementation, it makes sense to move these classes to Rendering/OpenGL, maybe in a subdirectory.

Sources/Rendering/LICOpenGL/SurfaceLICInterface/index.js
finetjul: https://stackoverflow.com/questions/25582882/javascript-math-random-normal-distribution-gaussian-bell-curve

martinken:
> > 1. could you move the constants and related set/get methods into the Rendering/Core LIC class? Then the OpenGL class would get them from the Rendering/Core class. If we have create a webGPU version then it would do the same thing and get the values from the Rendering/Core class.
> 
> Do you think adding a `vtkSurfaceLICInterface` class in Core will all the set/get parameters, but that doesn't handle anything WebGL, then a `vtkOpenGLSurfaceLICInterface` that handles the WebGL resources, as it does currently, could be a good way to do this ?

If the interface class (and it's properties) are intended to be shared between mappers then sure, that works. 

> > 1. This approach is a bit different in that it requires the app to instantiate an OpenGL specific mapper, for now that is probably fine but normally we let the factory mechanism create the View specific mapper (OpenGL in this case) as part of the build pass.
> 
> My understanding is that this is currently not the case (`vtkSurfaceLICMapper` is a `vtkMapper` that is not OpenGL specific, and `vtkOpenGLSurfaceLICMapper` is already instantiated by the `ViewNodeFactory`). Did I miss something ?

As it is, I believe you have to set an interface on the mapper and the interface class only exists in OpenGL. That is the break. The goal is that examples should only reference the OpenGLRenderWindow and all other opengl classes should be hidden. It is a data-view model where rendering core holds the data and OpenGL is a view. Examples add an OpenGL view by adding an OpenGLRenderWindow as a view on RenderWindow. RenderWindow supports multiple views so we could also add a WebGPURenderWindow as a view.  All the other view specific logic is ideally hidden if that makes sense.

martinken:
And if you make the interface class Core (without any opengl code) then that solves the issue.

boucaud:
> And if you make the interface class Core (without any opengl code) then that solves the issue.

I have pushed the change. Does this look better ?

jourdain:
The file layout seems much better. Thx!

martinken:
@boucaud that addresses my suggestions, thanks!

Sources/Rendering/LICOpenGL/SurfaceLICInterface/index.js
boucaud: That link says:

> Javascript Math.random is not a Normal Distribution

finetjul: yes, and it tells you how to generate a normal distribution from a uniform distribution.

boucaud: so should I use the Box-Muller transform ? This central limit estimation seemed sufficient to me.

finetjul: my bad, I did not notice it was CLT.
I think you can leave it as is. It would be great to keep the original (C++) comments. That helps to read the code :-)

I was also about to suggest to make things a bit more "JS" like but I think it is better to leave it close to the original code.

PR #1742, opened 2021-02-23
Wheel event position

floryst:
Adds screen position information to wheel event call data.
WheelEvent inherits from MouseEvent, and so has position information.

Fix case: scrolling on a slice with an active widget handle doesn't correctly position the handle, because positioning the widget doesn't happen on wheel event.

Also, move from Object.assign to object spread operator for conciseness.

FYI @agirault @finetjul, if either of you ran into the aforementioned widget issue before.


jourdain (approved):
LGTM

finetjul:
LGTM

PR #1755, opened 2021-02-25
fix(ImageReslice): Improve image reslice performances

finetjul:
especially when the number of slab is high.
Avoid usage of subarray that is a bit slow.


finetjul:
@gandis0713 it should improve things when slab is large.

gandis0713:
@finetjul ,
Thank you for improving performance.

PR #1756, opened 2021-02-25
fix(ResliceCursorWidget): Fix reslice cursor lines directions

finetjul:
When keep-orthogonality is false, the lines were not correct.


finetjul:
@laurennlam

laurennlam:
LGTM

PR #1757, opened 2021-02-26
Line widget final

finetjul:


finetjul:
@ocrossi can you give a try ?

Issue #1708, opened 2021-01-27
Incompatibility with DistanceWidget and InteractorStyleMPRSlice


wtchnm:
As discussed in https://github.com/Kitware/vtk-js/issues/1673, DistanceWidget doesn't work well with InteractorStyleMPRSlice.
The same problem can be reproduced with LineWidget, AngleWidget and probably with the other 3D widgets.

POC: https://codesandbox.io/s/vtkjs-distance-widget-with-interactor-style-mpr-slice-hzqzo.

floryst:
Clarification: the linked PR doesn't address the underlying issue of InteractorMPRSlice. The goal is to instead create a more suitable example involving the VolumeMapper's clipping planes that has much better support for widgets, and deprecate the InteractorMPRSlice.

aylward:
Should use VolumeMapper's clipping planes instead.

Issue #1673, opened 2020-12-21
Regression in version 15.3.2


wtchnm:
Hello!
PR #1639 introduces a breaking change to widgets rendered in a different renderer than the renderer that captured the mouse move event.

![vtk js](https://user-images.githubusercontent.com/37726261/102791109-6295c700-435b-11eb-9c69-eb2fd476a1fc.gif)

What would be the best way to solve this problem?

floryst:
I'll take a look.

floryst:
Do you happen to have a minimum PoC that reproduces this issue? It will make verifying changes easier.

wtchnm:
Yes.

V15.3.1: https://codesandbox.io/s/vtkjs-poc-1531-p07ti.
V15.3.2: https://codesandbox.io/s/vtkjs-poc-1532-wtcur.

floryst:
Is there any reason that you need to put the measurements in a separate renderer from the main scene? The widgets shouldn't be occluded by regular scene objects when picking, so placing measurement widgets in the main renderer should be fine.

The measurement renderer has interactivity set to false, so all mouse move events will reference the base renderer. From an API perspective, I would say that widgets in a non-interactive layer shouldn't be interactive anyways.

If you require a separate measurements renderer, then we can discuss further. However, if you only wanted to ensure picking of widgets do not interfere with regular scene objects, then I would ask you to consider removing the measurements renderer all together and just attach the widgets to the base renderer. 

Let me know what you try and the results. Or if you require a measurements renderer, I'm interested in hearing more about that use-case.

wtchnm:
Sorry for the delay.
Unfortunately, the distance widget [doesn't work well](https://codesandbox.io/s/vtkjs-distance-widget-with-interactor-style-mpr-slice-hzqzo) with the MPR slice interactor style, so that was the solution I found.

floryst:
Thanks, that makes sense. I suspect that reverting #1639 isn't the best solution here, since that functionality makes sense.

This issue also manifests in the InteractiveOrientationWiget, so I begin figuring out a better way to support this functionality.

wtchnm:
I agree with you.

I will open a new issue that describes the real problem.

Issue #1636, opened 2020-11-19
GPU memory redundancy in single context.


gandis0713:
I added multiple vtkRenderers to one vtkRenderWindow to implement multiple views for rendering volume. vtkRenderer has different viewport areas. However, It seems that the same object(vtkImageData) is duplicated in single context(vtkOpenGLRenderWindow) it is rendered with multiple renderers. (GPU memory redundancy.)

Here is what I checked.
vtkViewNodes create vtkViewNodes corresponding to the rendering objects they have and have them as children.
For example, vtkOpenGLRenderer seems to have vtkCamera and vtkVolume as Objects and created vtkOpenGLCamera and vtkOpenGLVolume as vtkViewNode corresponding to each Object.
However, even though vtkViewNodes have the same Object, vtkViewNodes corresponding to Objects are each newly created.
In the case of my vtkOpenGLRenderers, even though they have the same vtkVolume, each vtkOpenGLVolume corresponding to the vtkVolume has been created.
And vtkOpenGLVolume that was created individually creates and has vtkOpenGLVolumeMapper corresponding to vtkVolumeMapper.
Eventually a vtkOpenGLVolumeMapper that have vtkOpenGLTexture is created as many as vtkRenderer. So It seems it uses GPU memory in redundancy.

How do I solve the GPU memory that I use redundantly?

floryst:
@martinken

jourdain:
Can you create `1` mapper and `n` actors/volumes ?

gandis0713:
I modified vtkOpenGLRenderWindow and vtkOpenGLVolumeMapper so that vtkOpenGLTexture is not duplicated and then test them. Testing confirmed that GPU memory was not used redundantly. 

However I am wondering if there are other ways I haven't checked. Also, I am concerned that these personal modifications will not be able to respond updating vtk.

gandis0713:
Is there any other way to not redundantly use GPU memory? Or do you have any advice for me?

PR #1760, opened 2021-02-26
Revert #1639

### LINKED ISSUE #1749: Interactive orientation widget example broken
Hello,

the example of the interactive orientation widget does not seem to work:
https://kitware.github.io/vtk-js/examples/InteractiveOrientationWidget.html

click on the widget has no effect

floryst:
After some discussion, I have decided to revert #1639 until we get a clearer picture of what that PR was solving. @gandis0713 Could you please provide some more context, and ideally some reproducing example on codepen or codesandbox that uses an older version of vtk.js (e.g. v15.3.0) that didn't have your PR? While the idea makes sense, there might be a different solution to the problem you are experiencing.

For context surrounding the discussions, the WidgetManager is associated with a renderer, yet listens to events from the Interactor. This is necessarily by design; events only occur at the RenderWindow level. The WidgetManager then listens to these top-level events and processes these events in its own renderer.  I would argue that the widget manager should process events regardless of the poked renderer, as events are global in nature and the widget manager processes these events separately.

Take the case of a renderer on top of another renderer. If the top renderer contains widgets, then it should be the case that those widgets are interactive. But what if the top renderer is marked as _non-interactive_? A "non-interactive renderer" currently means a renderer that provides no interaction, meaning no InteractorStyles will trigger for it. However, the WidgetManager processes events separately, and only cares about events on widgets. Restricting the widget manager to only process events in its own renderer suddenly restricts widgets operating in overlayed, "transparent" renderers. This functionality is desired, e.g. in the case of the InteractiveOrientationWidget, which uses a non-interactive "transparent" renderer, yet also has a widget which has interactivity.

This fixes #1749 and #1673.

Partially addresses #1708.


gandis0713:
@floryst ,
In the past, when configuring multiple views, I tried to divide one window into viewports. However, it seems that there is a problem that the GPU memory is used redundantly (#1636). So I don't currently use it that way. Instead, I am using customized ResliceCursorWidget, So I couldn't constantly check the code I was modified.
I have checked what you have described and I fully agree. Thank you for notifying me.

PR #1766, opened 2021-03-08
fix(piecewisegaussianwidget): terminate leaking histogram web workers

PaulHax:
The nifty Piecewise Gaussian Widget abandons web worker instances with each call to `setDataArray`


thewtex (approved):
@PaulHax thanks!

PR #1767, opened 2021-03-08
Very early cut at WebGPU support

martinken:
Currently only handles some polydata with point
normals or point colors.

Works on Chrome Canary when WebGPU is enabled. You have to
set --enable-unsafe-webgpu there is a test target named
test:webgpu that will specificly try to run tests using
chrome canary.

From an application POV replacing your OpenGLRenderWindow with
a WebGPURenderWindow shoudl be the only change required :-)

To be done
------------------------------------------------------------
- texture maps (with cache, same idea as Vulkan)
- cell data support
- free unused buffers/the cache holds on to them IIRC
- PBR lighting to replace the simple model currently coded
- image display
- volume rendering
- transparency
- actor matrix support
- sphere/stick mappers
- post render operations/framebuffers


.eslintrc.js
```diff
@@ -2,6 +2,7 @@ var prettierConf = require('./prettier.config.js');
 
 module.exports = {
   extends: ['airbnb', 'prettier'],
+  parser: 'babel-eslint',
   rules: {
```
martinken, line 6: @jourdain sanity check here

Sources/Rendering/Core/RenderWindow/index.js
```diff
@@ -34,6 +37,18 @@ function vtkRenderWindow(publicAPI, model) {
...
+    if (options && options.api) {
+      return model.apiSpecificViews[options.api]();
+    }
+
+    if (typeof navigator !== 'undefined' && !!navigator.gpu) {
+      return model.apiSpecificViews['WebGPU']();
+    }
+    return model.apiSpecificViews['OpenGL']();
+  }
+
```
martinken, line 51: And thoughts on this file

webpack.dev.js
```diff
@@ -50,7 +50,7 @@ module.exports = [
     devServer: configureDevServer(port + 1),
     plugins: [
       new webpack.HotModuleReplacementPlugin(),
-      new DashboardPlugin(dashboard.setData),
+      // new DashboardPlugin(dashboard.setData),
     ],
```
martinken, line 54: Turned this off as on my terminals it was always a mess

package.json
```diff
@@ -121,6 +121,7 @@
     "test:headless": "karma start ./karma.conf.js --browsers ChromeHeadlessNoSandbox --single-run",
     "test:debug": "karma start ./karma.conf.js --no-single-run",
     "test:firefox": "karma start ./karma.conf.js --browsers Firefox",
+    "test:webgpu": "karma start ./karma.conf.js --browsers ChromeWebGPU --no-single-run",
     "test:firefox-debug": "karma start ./karma.conf.js --browsers Firefox --no-single-run",
```
martinken, line 125: And sanity check here

.eslintrc.js
jourdain: [This](https://www.npmjs.com/package/@babel/eslint-parser) might be more appropriate. 
What do you think @floryst?

Sources/Rendering/Core/RenderWindow/index.js
jourdain: I would write it that way

```
publicAPI.newAPISpecificView = (options={}) => {
  const defaultAPI = navigator.gpu ? 'WebGPU' : 'OpenGL';
  return model.apiSpecificViews[options.api || defaultAPI]
}
```

But also I would not put the `apiSpecificViews` inside the models. This should be private.

jourdain: And you could have `const defaultAPI = navigator.gpu ? 'WebGPU' : 'OpenGL';` be at the top so it get only evaluated at load time.

`const DEFAULT_API = navigator.gpu ? 'WebGPU' : 'OpenGL';`

jourdain: Unless we want to configure it. In which case it could be a field in the model of the RW.

Sources/Rendering/Core/RenderWindow/index.js
```diff
@@ -104,6 +119,7 @@ function vtkRenderWindow(publicAPI, model) {
 // ----------------------------------------------------------------------------
 
 export const DEFAULT_VALUES = {
+  apiSpecificView: null,
```
jourdain, line 122: Don't store it in the defaults and/or model.

Sources/Rendering/Core/RenderWindow/index.js
```diff
@@ -118,9 +134,14 @@ export function extend(publicAPI, model, initialValues = {}) {
 
   // Build VTK API
   macro.obj(publicAPI, model);
+
+  model.apiSpecificViews = [];
```
jourdain, line 138:
```
const VIEW_CONSTRUCTORS = Object.create(null); // same as {}

publicAPI.registerViewConstructor = (name, constructor) => VIEW_CONSTRUCTORS[name] = constructor;
publicAPI.newAPISpecificView = (name, initialValues={}) => {
   return VIEW_CONSTRUCTORS[name || DEFAULT_API](initialValues);
}
```

Sources/Rendering/Core/RenderWindow/index.js
```diff
@@ -118,9 +134,14 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+  model.apiSpecificViews['OpenGL'] = vtkOpenGLRenderWindow.newInstance;
+  model.apiSpecificViews['WebGPU'] = vtkWebGPURenderWindow.newInstance;
```
jourdain, line 140: Those calls should happen outside of that class.

Sources/Rendering/Core/RenderWindow/index.js
```diff
@@ -118,9 +134,14 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+
   macro.setGet(publicAPI, model, ['interactor', 'numberOfLayers', 'views']);
   macro.get(publicAPI, model, ['neverRendered']);
-  macro.getArray(publicAPI, model, ['renderers']);
+  macro.getArray(publicAPI, model, ['apiSpecificViews', 'renderers']);
```
jourdain, line 144: I don't think we want to have access to those constructor as a map.

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -72,17 +71,17 @@ function vtkFullScreenRenderWindow(publicAPI, model) {
   model.renderer = vtkRenderer.newInstance();
   model.renderWindow.addRenderer(model.renderer);
 
-  // OpenGlRenderWindow
-  model.openGLRenderWindow = vtkOpenGLRenderWindow.newInstance();
-  model.openGLRenderWindow.setContainer(model.container);
-  model.renderWindow.addView(model.openGLRenderWindow);
+  // apiSpecificRenderWindow
+  model.apiSpecificRenderWindow = model.renderWindow.newAPISpecificView();
```
jourdain, line 75: We could use the url parameter extractor to select the implementation. That way, we can run the same example but based on the URL pick one implementation vs the other.

Sources/Rendering/Core/RenderWindow/index.js
martinken: You would do a static to query the available APIs (to populate a menu) ?

webpack.dev.js
jourdain: Sounds good to me. that was an experiment. We should remove it forever... ;-)

package.json
jourdain: LGTM

Sources/Rendering/Core/RenderWindow/index.js
jourdain: Also thinking that VIEW_CONSTRUCTORS and registerViewConstructor should be static.

Sources/Rendering/Core/RenderWindow/index.js
jourdain: Yes we can list the registered keys as a static method

Examples/Applications/GeometryViewer/index.js
```diff
@@ -282,6 +283,7 @@ function createPipeline(fileName, fileContents) {
   function updateColorBy(event) {
     const [location, colorByArrayName] = event.target.value.split(':');
     const interpolateScalarsBeforeMapping = location === 'PointData';
+    const interpolateScalarsBeforeMapping = false;
```
jourdain, line 286: Do you want to remove that for now?

martinken: bah yeh

Examples/Applications/GeometryViewer/index.js
```diff
@@ -360,9 +361,11 @@ function createPipeline(fileName, fileContents) {
...
 
   // Manage update when lookupTable change
-  lookupTable.onModified(() => {
-    renderWindow.render();
-  });
+  // too heavy handed, causes multiple buffer pushes as opposed
+  // to defering them to when all changes are done
+  // lookupTable.onModified(() => {
+  //   renderWindow.render();
+  // });
```
jourdain, line 368: We should register the listener with a priority of -1 and debounce the call to render.

jourdain (approved):

martinken:
@jourdain good to merge?

PR #1770, opened 2021-03-09
fix(package): Manually create npm package

floryst:
The create-package script constructs a package separately from the
project root.
In particular, this is designed to help webpack v4 support vtk.js ESM,
since webpack v4 does not process the package.json exports field.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,6 +28,8 @@ jobs:
         run: npm run validate
       - name: Test
         run: npm run test:headless
+      - name: Create package folder
+        run: npm run release:create-package
       - name: Release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


jourdain:
I was wondering if we should publish the new esm build to `@kitware/vtk`?
That will explain the possible different layout. 

What do you think?

Also, I'm not sure how to test/validate your PR. 

Maybe one missing directory is the `Examples` one.

jourdain (approved):
I trust you on that one... ;-)

floryst:
I wasn't sure if we wanted to export the Examples directory in the first place.

As for `@kitware/vtk`, I think that's a good idea. Do we have that org already?

jourdain:
Yes I'm the owner of the @kitware org ;-)

jourdain:
Let me see if I can add you along with the robot... ;-)

jourdain:
Also which one should we do `@kitware/vtk` or `@kitware/vtk.js`? Knowing that we may publish one day some WebAssembly part of VTK/C++.

floryst:
I'm closing this in favor of publishing the ESM build to `@kitware/vtk.js`.

PR #1775, opened 2021-03-10
fix(cellarray): fix cell array data type

### LINKED ISSUE #1772: CellArray default value is ignored
The following should return 'Uint32Array' instead of 'Float32Array'.
```
const cells = vtkCellArray.newInstance({ values: [3, 0, 1, 2] });
cells.getDataType();
```

finetjul:
When providing an array at construction time, default cell array UNSIGNED_INT data type was lost

fix #1772


jourdain:
LGTM

PR #1776, opened 2021-03-10
docs(ts): initial typescript definitions

### LINKED ISSUE #466: Typescript type definitions
Are typescript type definitions planned for vtk.js? (as per available packages for [three.js](https://threejs.org/), for example https://www.npmjs.com/package/@types/three)

daker:
First set of typescript definitions

fix #466

### DIFF
--- a/Sources/Rendering/Core/Camera/index.d.ts
+++ b/Sources/Rendering/Core/Camera/index.d.ts
@@ -1,0 +1,302 @@
+import VtkObject from '../../../macro';
+
+export interface vtkCamera extends VtkObject {
+    /**
+     * 
+     */
+    orthogonalizeViewUp(): void;
+
+    /**
+     * 
+     * @param x 
+     * @param y 
+     * @param z 
+     */
+    setPosition(x: number, y: number, z: number): void;
+
+    /**
+     * 
+     * @param x 
+     * @param y 
+     * @param z 
+     */
+    setFocalPoint(x: any, y: any, z: any): void;
+
+    /**
+     * 
+     * @param d 
+     */
+    setDistance(d: number): void;
+
+    /**
+     * ----------------------------------------------------------------------------
+     * This method must be called when the focal point or camera position changes
+     */
+    computeDistance(): void;
+
+    /**
+     * ----------------------------------------------------------------------------
+     * Move the position of the camera along the view plane normal. Moving
+     * towards the focal point (e.g., > 1) is a dolly-in, moving away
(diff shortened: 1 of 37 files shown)


daker:
@jourdain i'll add more in Rendering/Core if you are ok with the first batch

Sources/Rendering/Core/Camera/index.d.ts
```diff
@@ -0,0 +1,302 @@
+import VtkObject from '../../../macro';
```
jourdain, line 1: can you use the absolute path? Or is it not possible as the alias are not understood by ts?

daker: Yes i can use the absolute path

Sources/Rendering/Core/Camera/index.d.ts
```diff
@@ -0,0 +1,302 @@
...
+     */
+    setPosition(x: number, y: number, z: number): void;
+
+    /**
+     * 
+     * @param x 
+     * @param y 
+     * @param z 
+     */
+    setFocalPoint(x: any, y: any, z: any): void;
```
jourdain, line 23: should they be number too? 

Also those methods can support both `xyz[]` and `x, y, z`

daker: Yes i'll do another pass for the param and the setter/getter

jourdain:
I think this is a great first step and thanks for doing that. It seems that the key part is that `tsconfig.json` at the root.

For some of the arguments we should probably create some generic types Vec3Float and Vec3FloatSpread if possible to illustrate the API in a better fashion. Also I bet those types will be widely use across the various classes.

For the comment on top, we should indeed improve them through time. The key part that I would like to nail down is getting the proper types / pattern used.

Maybe we can also use better name for our internal types like LightInitialValues rather than IValues. Since with inheritance we may want to provide a union of InitialValues? 

Feel free to argue, since I'm not sure what I'm talking about. ;-)

Thanks anyway as this is going to be great!

Sources/Rendering/Core/Camera/index.d.ts
jourdain: Also more of those get generated. (Look at [setGetArray](https://github.com/Kitware/vtk-js/blob/master/Sources/macro.js#L484-L485) macro to see what I mean).

daker:
@jourdain i did add more for Rendering/Core, if you and other devs are ok we can merge then add more gradually.

daker:
Any news on this?

PR #1781, opened 2021-03-13
refactor(boundingbox): Redesign BoundingBox and Box

### LINKED ISSUE #1779: vtkBoundingBox vs vtkBox
It is stated that vtkBoundingBox is a lightweight class because it does not derive from vtkObject (probably a bad copy/paste from the VTK documentation). This is not true though, it calls `macro.obj`. .

Now, I also notice that vtkBox provides almost the same features of vtkBoundingBox. Moreover, vtkBox has a vtkBoundingBox in its model.

Don't you think those 2 classes duplicate each other ?

I like the idea of having a lightweight class that can be used in time-critical parts. What do you think ? Should we make vtkBoundingBox a simple container of static functions to manipulate bounds (like it is a bit already)?

finetjul:
BREAKING CHANGE: vtkBoundingBox is not longer instantiable.
vtkBox.intersectBox() is moved to vtkBoundingBox.intersectBox()
computeCornerPoints() parameter order has changed.

fix #1779

### DIFF
--- a/Sources/Common/DataModel/BoundingBox/api.md
+++ b/Sources/Common/DataModel/BoundingBox/api.md
@@ -1,117 +1,103 @@
-vtkBoundingBox maintains a 3D axis aligned bounding box.
-It is very lite weight and many of the member functions are in-lined so its
-very fast. It is not derived from vtkObject so it can be allocated on the stack
+vtkBoundingBox maintains a collection of function to manipulate 3D axis
+aligned bounding box.
+It is very lite weight and the static functions are very fast.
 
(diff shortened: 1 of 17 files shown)


finetjul:
To be merged in the next major release (no rush)

jourdain:
I'll see if I can add to it as I don't like some of what was done as it complexify its usage. I understand the performance issue and we should enable it but we are lacking some useful usage.

jourdain:
Add-on available [here](https://github.com/Kitware/vtk-js/pull/1784)

Issue #1777, opened 2021-03-11
vtkOpenGLCellArrayBufferObject: cellCounters return negative value and cause range error


gaoxinyu18:
As the image and codes (from file `vtk.js\Sources\Rendering\OpenGL\CellArrayBufferObject\index.js`) display below, the `caboCount` value is nagative here and cause a range error,
The `caboCount` is a sum of the `cellCounters` return value, but in some case the `numPoints`  is less than `2` or even equal to `0`,
Thus the `caboCount` may be negative finally.
```js
const cellCounters = {
      // easy, every input point becomes an output point
      anythingToPoints(numPoints, cellPts) {
        return numPoints;
      },
      linesToWireframe(numPoints, cellPts) {
        return (numPoints - 1) * 2;
      },
      polysToWireframe(numPoints, cellPts) {
        return numPoints * 2;
      },
      stripsToWireframe(numPoints, cellPts) {
        return numPoints * 4 - 6;
      },
      polysToSurface(npts, cellPts) {
        if (npts < 3) {
          return 0;
        }
        return (npts - 2) * 3;
      },
      stripsToSurface(npts, cellPts, offset) {
        return (npts - 2) * 3;
      },
    };

    let func = null;
    let countFunc = null;
    if (outRep === Representation.POINTS || inRep === 'verts') {
      func = cellBuilders.anythingToPoints;
      countFunc = cellCounters.anythingToPoints;
    } else if (outRep === Representation.WIREFRAME || inRep === 'lines') {
      func = cellBuilders[`${inRep}ToWireframe`];
      countFunc = cellCounters[`${inRep}ToWireframe`];
    } else {
      func = cellBuilders[`${inRep}ToSurface`];
      countFunc = cellCounters[`${inRep}ToSurface`];
    }

    const array = cellArray.getData();
    const size = array.length;
    let caboCount = 0;
    for (let index = 0; index < size; ) {
      caboCount += countFunc(array[index], array);
      index += array[index] + 1;
    }
````
![vtk-error](https://user-images.githubusercontent.com/9973510/110762411-babdd480-828b-11eb-9d34-5a55080be365.png)

gaoxinyu18:
@martinken it works for my case after upgrade vtk-js to v16.13.0, thanks~

martinken:
@gaoxinyu18  Awesome, I'll close this then. Sorry for the long wait.

PR #1783, opened 2021-03-15
feat(WebGPU): support for wireframe, point and cellData

martinken:
WebGPU should now handle
- cellScalar coloring
- wireframe and point representations

Fix an issue with degeerate cell rendering in both WebGPU and WebGL
re https://github.com/Kitware/vtk-js/issues/1777


martinken:
@jourdain if you want to take a quick look

jourdain:
LGTM

PR #1784, opened 2021-03-15
1779 vtkboundingbox redesign

jourdain:


finetjul:
It's fine by me.

It's still a major BREAKING CHANGE. To be merged only in next major branch.

jourdain:
We could have made it without a breaking change... But because of bbox delete, it is a breaking change... ;-)

PR #1786, opened 2021-03-16
feat(linewidget): enlarge line and make it draggable

### LINKED ISSUE #1593: LineWidget wishlist
**Coming upgrades for the Line Widget**
These are the next features to be developed on the Line Widget, ranked by priority
- [x] Rotate the 2D handles with the camera
- [x] Implement None Handle to match an invisible sphere
- [x] Add UI
- [x] implement Right to Left/ Left to Right behavior for SVG Layer
- [ ] implement pickable poly line to move the widget all in one drag
- [x] keep steady size during handle mouse over
- [x]  Add a Disk handle
- [x] Hide Poly Line behind the handle
- [x] add rotation feature to CircleHandle
- [x] resolve bug text doesn't move when there is a circle handle
- [ ] size of handles to fix when scaleInPixels is deactivated

finetjul:
fix #1593

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/vdom.js
+++ b/Sources/Widgets/Core/WidgetManager/vdom.js
@@ -35,6 +35,10 @@ export function render(vnode) {
     const key = keys[i];
     node.setAttribute(key, vnode.attrs[key]);
   }
+  // TODO: support removing event listener (e.g. use snabbdom)
+  Object.keys(vnode.eventListeners).forEach((key) => {
+    node.addEventListener(key, vnode.eventListeners[key]);
+  });
 
   if (vnode.textContent) {
     node.textContent = vnode.textContent;
--- a/Sources/Widgets/Representations/PolyLineRepresentation/index.js
+++ b/Sources/Widgets/Representations/PolyLineRepresentation/index.js
@@ -2,9 +2,11 @@ import macro from 'vtk.js/Sources/macro';
 import vtkActor from 'vtk.js/Sources/Rendering/Core/Actor';
 import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
 import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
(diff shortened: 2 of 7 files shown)


finetjul:
@floryst @jourdain how can I make the SVG[Landmark]Representation pickable and draggable ?

floryst:
SVG picking is something that I've had in consideration for a while. Right now, the best way to do this is to have an invisible, pickable representation in the 3D view that corresponds to the SVG rep. To support SVG picking/mapping to a widget state, we need a few things first:
- move from my vdom impl to one that supports event listeners (e.g. snabdom supports events I believe)
- Maintain the map between SVG rep and a widget (sub)state

finetjul:
Thanks Forest, I think the invisible rect will be the way to go for me.

Sources/Widgets/Core/WidgetManager/vdom.js
```diff
@@ -35,6 +35,9 @@ export function render(vnode) {
     const key = keys[i];
     node.setAttribute(key, vnode.attrs[key]);
   }
+  Object.keys(vnode.eventListeners).forEach((key) => {
+    node.addEventListener(key, vnode.eventListeners[key]);
+  });
```
floryst, line 40: At some point I want to move away from this vdom implementation to something like [snabbdom](https://github.com/snabbdom/snabbdom). That way we get proper handling of event listeners, including proper removal of event listeners. Also so we don't have to maintain our own vdom.

For now, this is fine, but the current vdom doesn't address event listener removal when the incoming virtual node doesn't have the same set of listeners.

finetjul: gotcha. I'll add a note in the code and merge it for the time being.

finetjul:
> I think the invisible rect will be the way to go for me.

Btw, the invisible rect approach does not work because we couldn't know the size of the rect for the rect.

Issue #1751, opened 2021-02-24
SVGLandmarkRepresentation are shared amongst every instance


agirault:
When using LineWidget, all the text svg representations use the same position.

https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js#L84

Might occur on other SVG representations.

jourdain:
Indeed that is the [issue](https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/SVG/SVGLandmarkRepresentation/index.js#L69-L78)

jourdain:
Objects share their references... You need a function to create object or put them into the model at the execution of `vtkSVGLandmarkRepresentation()` function.

finetjul:
> Objects share their references... You need a function to create object or put them into the model at the execution of `vtkSVGLandmarkRepresentation()` function.

It seems that it is a recurring error. How about changing `DEFAULT_VALUES` into a function ? e.g.
```
const defaultValues = () => {...};
...
Object.assign(model, defaultValues(), initialValues);
```

jourdain:
I wanted to do that but I was worried to do the refactor work everywhere. But this could be done on case by case...

jourdain:
at that point you can also just define it as

```
function defaultValues(initialValue) {
   return {
       a, b, c, 
       ...initialValue,
   };
}
```

PR #1790, opened 2021-03-19
Cube axes

martinken:


Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,791 @@
...
+];
+
+// some shared temp variables to reduce heap allocs
+const ptv3 = vec3.create();
+const pt2v3 = vec3.create();
+const tmpv3 = vec3.create();
+const tmp2v3 = vec3.create();
+const xDir = vec3.create();
+const yDir = vec3.create();
+const invmat = mat4.create();
```
finetjul, line 65: My understanding is that `mat4.create()` brings precision errors, it's better to store matrice elements on 64 bits. -> 
```
const invmat = new Float64Array(16);
mat4.initialize(invmat);
```

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,791 @@
...
+    let numLines = 0;
+    numPts += 8; // always start with the 8 cube points
+
+    // count edgesToDraw
+    let numEdgesToDraw = 0;
+    for (let e = 0; e < 12; e++) {
+      if (edgesToDraw[e] > 0) {
+        numEdgesToDraw++;
+      }
+    }
```
finetjul, line 151: If performance is not an issue:
```
const numEdgesToDraw = edgesToDraw.filter(e => e > 0).length;
```

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,791 @@
...
+
+    // did something significant change? If so rebuild a lot of things
+    if (facesChanged || boundsChanged || force) {
+      // compute the edges to draw
+      // for each drawn face, mark edges, all single mark edges we draw
+      const edgesToDraw = [];
+      for (let e = 0; e < 12; e++) {
+        edgesToDraw[e] = 0;
+      }
```
finetjul, line 347:
```
const edgesToDraw = new Array(12).fill(0);
```

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,791 @@
...
+    }
+
+    if (array.length !== 3) {
+      throw new RangeError(
+        `Invalid number of values for array setter (AxesLabels)`
+      );
+    }
+
+    let changeDetected = false;
+    model.axesLabels.forEach((item, index) => {
```
finetjul, line 685: shouldn't it be a `some` function instead of `forEach` ?

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,791 @@
...
+    model.tmPolyData.getPolys().setData(polys, 1);
+    model.tmPolyData.getPolys().modified();
+    model.tmPolyData.modified();
+  };
+
+  publicAPI.getActors = () => {
+    const actors = [];
+    actors.push(model.actor);
+    actors.push(model.pixelActor);
+    actors.push(model.tmActor);
```
finetjul, line 654: why not simply 
```
const actors = [model.actor, model.pixelActor, model.tmActor];
```

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,791 @@
...
+];
+const edgeAxes = [0, 1, 0, 1, 0, 1, 0, 1, 2, 2, 2, 2];
+const faceAxes = [
+  [1, 2],
+  [1, 2],
+  [0, 2],
+  [0, 2],
+  [0, 1],
+  [0, 1],
+];
```
finetjul, line 56: should probably be moved in a Constants.js file

martinken: I don't want to export these ever (if they get exported it should be from a cell type class). So should they still be in a constants file? Just not sure what the motivation is for that case.

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,830 @@
...
+  tickTextStyle: {
+    fontColor: 'white',
+    fontStyle: 'normal',
+    fontSize: 14,
+    fontFamily: 'serif',
+    strokeColor: 'black',
+    strokeSize: 1,
+    lineSpace: 0.2,
+  },
+};
```
finetjul, line 749: You can't have objects in `DEFAULT_VALUES`, see https://github.com/Kitware/vtk-js/issues/1751#issuecomment-785411007 for an alternative to `DEFAULT_VALUES`

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,830 @@
...
+
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  // Inheritance
+  vtkActor.extend(publicAPI, model, initialValues);
+
+  model.dataBounds = [1, -1, 1, -1, 1, -1];
```
finetjul, line 759:
```
model.dataBounds = [...vtkBoundingBox.INIT_BOUNDS];
```

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,830 @@
...
+    for (let z = 0; z < 2; z++) {
+      for (let y = 0; y < 2; y++) {
+        for (let x = 0; x < 2; x++) {
+          points[ptIdx * 3] = model.dataBounds[x];
+          points[ptIdx * 3 + 1] = model.dataBounds[2 + y];
+          points[ptIdx * 3 + 2] = model.dataBounds[4 + z];
+          ptIdx++;
+        }
+      }
+    }
```
finetjul, line 193: isn't `vtkBoundingBox.getCorners()` doing the same ?

Sources/Rendering/Core/CubeAxesActor/index.js
finetjul: I guess it is fine to leave it here if they do not make sense to be exported. @jourdain

Sources/Rendering/Core/CubeAxesActor/index.js
martinken: Ha, I've been bitten by that before.  I lifted that code from label representation so that probably needs to be fixed.

Sources/Rendering/Core/CubeAxesActor/index.js
martinken: Yup

package.json
```diff
@@ -31,6 +31,7 @@
   "dependencies": {
     "blueimp-md5": "2.18.0",
     "commander": "6.2.1",
+    "d3-scale": "3.2.3",
```
finetjul, line 34: according to the CI results, you also need to push package-lock.json

Sources/Rendering/Core/CubeAxesActor/index.js
martinken: I always forget to add the comment before marking resolved. So my comments all disappear. Have to remember, marking resolved doesn't also submit the comment. getCorners is close, it does a z changes fastest as opposed to x. Not sure why. But I tend to use the first axis changes fastest (C array convention) so with i,j,k or x,y,z  i,x will the the fastest changing index. And as kharma would have it, it looks like I wrote the getCorners on BoundingBox.

package.json
martinken: Ah thanks!

Sources/Rendering/Core/CubeAxesActor/index.js
finetjul: I would then suggest to change `vtkBoundingBox.getCorners()`. Corner order does not seem to impact its existing usage and it will be more consistent (also WRT VTK C++).

martinken: Seb has an open MR on bounding box, are you suggesting I edit it now and let him merge the changes?

finetjul: Yes.
I'm happy to do the merge. I prefer that than forgetting to cleanup that code later :-)
@jourdain WDYT ?

martinken: If so I'll probably move some of the constants into there as well since BoundingBox is a reasonable place for box related constants

martinken: Into a constants file I would assume

jourdain:
Is AxisGrid or something a better name?

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,831 @@
...
+            ticks[faceAxes[f][0]].length * 2 + ticks[faceAxes[f][1]].length * 2;
+          numLines +=
+            ticks[faceAxes[f][0]].length + ticks[faceAxes[f][1]].length;
+        }
+      }
+    }
+
+    // now allocate the memory
+    const points = new Float64Array(numPts * 3);
+    const lines = new Uint32Array(numLines * 3);
```
jourdain, line 178: is Uint16 enough?

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,831 @@
...
+// some shared temp variables to reduce heap allocs
+const previousMatrixType = glMatrix.ARRAY_TYPE;
+glMatrix.setMatrixArrayType(Float64Array);
+const ptv3 = vec3.create();
+const pt2v3 = vec3.create();
+const tmpv3 = vec3.create();
+const tmp2v3 = vec3.create();
+const xDir = vec3.create();
+const yDir = vec3.create();
+const invmat = mat4.create();
```
jourdain, line 67: I see the `.create()` idea, but can't we just do `new Float64Array(3)` or  `new Float64Array(16)` instead?

Sources/Rendering/Core/CubeAxesActor/index.js
```diff
@@ -0,0 +1,831 @@
...
+  model.tmMapper = vtkMapper.newInstance();
+  model.tmMapper.setInputData(model.tmPolyData);
+  model.tmTexture = vtkTexture.newInstance();
+  model.tmTexture.setInterpolate(false);
+  model.tmActor = vtkActor.newInstance();
+  model.tmActor.setMapper(model.tmMapper);
+  model.tmActor.addTexture(model.tmTexture);
+  model.tmCanvas = document.createElement('canvas');
+  model.tmContext = model.tmCanvas.getContext('2d');
+  model.tmAtlas = new Map();
```
jourdain, line 792: Ideally we want the model to be serializable. You may want to keep your `Map()` as private (const inside closure).

Sources/Rendering/Core/CubeAxesActor/index.js
martinken: We can, but curious why would we want to? create() is basically their constructor.

Sources/Rendering/Core/CubeAxesActor/index.js
martinken: Yes, Uint16 is plenty, I'll change that up.

martinken:
CubeAxesActor is what it is called in C++ so I just kept it the same.

Sources/Rendering/Core/CubeAxesActor/index.js
jourdain: My take on that is that you don't know what you get depending on their implementation. (I know, you made the call to their API to change the default type). But since their API was meant to work on what ever array you have, it seems less opaque to state what we get natively rather than using an abstraction that hides what we get.

Sources/Rendering/Core/CubeAxesActor/index.js
martinken: Hmm can subclasses access this internal state? e.g. how should I declare things I want to be protected in C++ terminology?

Sources/Rendering/Core/CubeAxesActor/index.js
martinken: OK I'll do it in a different MR and do it for all of vtk.js so we are consistent

Sources/Rendering/Core/CubeAxesActor/index.js
jourdain: What ever you bind to model is ~"protected"

martinken: OK then most/all of these should stay in model right, as a subclass will typically need to access them to add or change much functionality

martinken: Going to merge this and then we can tweak more if we want.

PR #1794, opened 2021-03-23
fix(package): Manually create npm packages

floryst:
This will result in two packages: `vtk.js`, and `@kitware/vtk.js`.

- `vtk.js` will still provide `vtk.js/Sources/*` as well as the UMD build. `vtk.js@beta` no longer contains the ESM build.
- `@kitware/vtk.js` will only provide the ESM build. `@kitware/vtk.js/Sources/*` no longer exists.

A few things to note:
- `build-npm-package.js` generates two packages: the ESM and UMD packages.
- Both packages have customized package.json entries.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -59,5 +59,7 @@ module.exports = {
   ignorePatterns: [
     // ignore old Actor example
     '**/example_/*.js',
+    // ignore js files in utilities
+    'Utilities/**/*.js',
   ],
 };
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,6 +28,8 @@ jobs:
         run: npm run validate
       - name: Test
         run: npm run test:headless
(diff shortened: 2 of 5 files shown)


jourdain (approved):
Looks reasonable

finetjul:
Is `build-npm-package.js` the file to edit to generate custom version of vtk.js and do tree shaking (if needed) ?

floryst:
`build-npm-package.js` does the following:
- create `pkg/esm` and `pkg/umd`.
- Populate the esm folder with the esm build, plus ancillary files. 
- Populate the umd folder with the umd build, plus ancillary files.
- modify the package.json for the esm pkg to point to `@kitware/vtk.js`, and the umd pkg to point to `vtk.js`

Basically, it preps the ESM and UMD package roots so that semantic-release can (hopefully) publish each to their respective package targets. Tree-shaking is not handled by this script.

PR #1795, opened 2021-03-23
fix(reslicecursor): fix reslice cursor when aspect ratio is != 1

finetjul:
displayFocal has 3 elements and not 2


finetjul:
@laurennlam

laurennlam:
LGTM

PR #1796, opened 2021-03-23
chore(Rendering): Update gl-matrix code to use Float64Array

martinken:
The default for gl-matrix is Float32Array but we want Float64Array for
our calculations and a consistent approach.

Removed all .create() methods and replaced with new Float64Array

Added identity calls as needed for matrices.

Cleaned up some cases where more work was being done than we needed.
Still more that could be cleaned up.

Update a couple tests where the improved accuracy resulted in different
test results.

### DIFF
--- a/Sources/Common/DataModel/ImageData/index.js
+++ b/Sources/Common/DataModel/ImageData/index.js
@@ -104,13 +104,13 @@ function vtkImageData(publicAPI, model) {
 
   publicAPI.getPoint = (index) => {
     const dims = publicAPI.getDimensions();
-    const ijk = vec3.fromValues(0, 0, 0);
-    const coords = [0, 0, 0];
 
     if (dims[0] === 0 || dims[1] === 0 || dims[2] === 0) {
       vtkErrorMacro('Requesting a point from an empty image.');
       return null;
     }
+
+    const ijk = new Float64Array(3);
 
     switch (model.dataDescription) {
       case StructuredType.EMPTY:
@@ -157,9 +157,8 @@ function vtkImageData(publicAPI, model) {
         break;
     }
 
-    const vout = vec3.create();
-    publicAPI.indexToWorldVec3(ijk, vout);
-    vec3.copy(coords, vout);
+    const coords = [0, 0, 0];
+    publicAPI.indexToWorld(ijk, coords);
     return coords;
   };
 
@@ -204,13 +203,13 @@ function vtkImageData(publicAPI, model) {
       ex[0], ex[3], ex[5],
       ex[1], ex[3], ex[5]];
 
-    const idx = vec3.fromValues(corners[0], corners[1], corners[2]);
-    const vout = vec3.create();
-    publicAPI.indexToWorldVec3(idx, vout);
+    const idx = new Float64Array([corners[0], corners[1], corners[2]]);
+    const vout = new Float64Array(3);
+    publicAPI.indexToWorld(idx, vout);
     const bounds = [vout[0], vout[0], vout[1], vout[1], vout[2], vout[2]];
     for (let i = 3; i < 24; i += 3) {
       vec3.set(idx, corners[i], corners[i + 1], corners[i + 2]);
-      publicAPI.indexToWorldVec3(idx, vout);
+      publicAPI.indexToWorld(idx, vout);
       if (vout[0] < bounds[0]) {
         bounds[0] = vout[0];
       }
@@ -236,12 +235,7 @@ function vtkImageData(publicAPI, model) {
 
   // Internal, shouldn't need to call this manually.
   publicAPI.computeTransforms = () => {
-    const trans = vec3.fromValues(
-      model.origin[0],
-      model.origin[1],
-      model.origin[2]
-    );
-    mat4.fromTranslation(model.indexToWorld, trans);
+    mat4.fromTranslation(model.indexToWorld, model.origin);
 
     model.indexToWorld[0] = model.direction[0];
     model.indexToWorld[1] = model.direction[1];
@@ -255,12 +249,7 @@ function vtkImageData(publicAPI, model) {
     model.indexToWorld[9] = model.direction[7];
     model.indexToWorld[10] = model.direction[8];
 
-    const scale = vec3.fromValues(
-      model.spacing[0],
-      model.spacing[1],
-      model.spacing[2]
-    );
-    mat4.scale(model.indexToWorld, model.indexToWorld, scale);
+    mat4.scale(model.indexToWorld, model.indexToWorld, model.spacing);
 
     mat4.invert(model.worldToIndex, model.indexToWorld);
   };
@@ -311,35 +300,17 @@ function vtkImageData(publicAPI, model) {
     return true;
   };
 
-  // this is the fast version, requires vec3 arguments
-  publicAPI.indexToWorldVec3 = (vin, vout) => {
-    vec3.transformMat4(vout, vin, model.indexToWorld);
-    return vout;
-  };
-
-  // slow version for generic arrays
   publicAPI.indexToWorld = (ain, aout = []) => {
-    const vin = vec3.fromValues(ain[0], ain[1], ain[2]);
-    const vout = vec3.create();
-    vec3.transformMat4(vout, vin, model.indexToWorld);
-    vec3.copy(aout, vout);
+    vec3.transformMat4(aout, ain, model.indexToWorld);
     return aout;
   };
+  publicAPI.indexToWorldVec3 = publicAPI.indexToWorld;
 
-  // this is the fast version, requires vec3 arguments
-  publicAPI.worldToIndexVec3 = (vin, vout) => {
-    vec3.transformMat4(vout, vin, model.worldToIndex);
-    return vout;
-  };
-
-  // slow version for generic arrays
   publicAPI.worldToIndex = (ain, aout = []) => {
-    const vin = vec3.fromValues(ain[0], ain[1], ain[2]);
-    const vout = vec3.create();
-    vec3.transformMat4(vout, vin, model.worldToIndex);
-    vec3.copy(aout, vout);
+    vec3.transformMat4(aout, ain, model.worldToIndex);
     return aout;
   };
+  publicAPI.worldToIndexVec3 = publicAPI.worldToIndex;
 
   publicAPI.indexToWorldBounds = (bin, bout = []) => {
     const in1 = [0, 0, 0];
@@ -564,13 +535,9 @@ export function extend(publicAPI, model, initialValues = {}) {
   vtkDataSet.extend(publicAPI, model, initialValues);
 
   if (!model.direction) {
-    model.direction = mat3.create();
+    model.direction = mat3.identity(new Float64Array(9));
   } else if (Array.isArray(model.direction)) {
-    const dvals = model.direction.slice(0);
-    model.direction = mat3.create();
-    for (let i = 0; i < 9; ++i) {
-      model.direction[i] = dvals[i];
-    }
+    model.direction = new Float64Array(model.direction.slice(0, 9));
   }
 
   model.indexToWorld = new Float64Array(16);
--- a/Sources/Interaction/Manipulators/MouseCameraUnicamRotateManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseCameraUnicamRotateManipulator/index.js
@@ -86,18 +86,17 @@ function vtkMouseCameraUnicamRotateManipulator(publicAPI, model) {
     cameraFocalPoint[3] = 1.0;
     cameraViewUp[3] = 0.0;
 
-    const transform = mat4.create();
-    mat4.identity(transform);
-    mat4.translate(transform, transform, vec3.fromValues(cx, cy, cz));
-    mat4.rotate(transform, transform, angle, vec3.fromValues(ax, ay, az));
-    mat4.translate(transform, transform, vec3.fromValues(-cx, -cy, -cz));
+    const transform = mat4.identity(new Float64Array(16));
+    mat4.translate(transform, transform, [cx, cy, cz]);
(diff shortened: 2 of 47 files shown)


Sources/Common/DataModel/ImageData/index.js
```diff
@@ -104,7 +104,7 @@ function vtkImageData(publicAPI, model) {
 
   publicAPI.getPoint = (index) => {
     const dims = publicAPI.getDimensions();
-    const ijk = vec3.fromValues(0, 0, 0);
+    const ijk = new Float64Array(3);
```
jourdain, line 107: You can do pre-initialization like that `Float64Array.from([0,0,0])`

Sources/Common/DataModel/ImageData/index.js
```diff
@@ -204,8 +204,8 @@ function vtkImageData(publicAPI, model) {
       ex[0], ex[3], ex[5],
       ex[1], ex[3], ex[5]];
 
-    const idx = vec3.fromValues(corners[0], corners[1], corners[2]);
-    const vout = vec3.create();
+    const idx = new Float64Array([corners[0], corners[1], corners[2]]);
```
jourdain, line 207: Float64Array.from(corners)

Sources/Common/DataModel/ImageData/index.js
```diff
@@ -319,8 +309,8 @@ function vtkImageData(publicAPI, model) {
 
   // slow version for generic arrays
   publicAPI.indexToWorld = (ain, aout = []) => {
-    const vin = vec3.fromValues(ain[0], ain[1], ain[2]);
-    const vout = vec3.create();
+    const vin = new Float64Array([ain[0], ain[1], ain[2]]);
```
jourdain, line 312: Float64Array.from(ain)

Sources/Common/DataModel/ImageData/index.js
martinken: Yes, but native arrays default to 0 so I remeoved it

Sources/Common/DataModel/ImageData/index.js
martinken: Only if I know ain is of length 3, in cases where I knew that I did it. In cases like this I wasn't ready to assume it was a length 3 vector

Sources/Common/DataModel/ImageData/index.js
martinken: Same idea as ain below.

Sources/Interaction/Manipulators/MouseCameraUnicamRotateManipulator/index.js
```diff
@@ -86,18 +86,18 @@ function vtkMouseCameraUnicamRotateManipulator(publicAPI, model) {
...
 
-    const transform = mat4.create();
+    const transform = new Float64Array(16);
     mat4.identity(transform);
-    mat4.translate(transform, transform, vec3.fromValues(cx, cy, cz));
-    mat4.rotate(transform, transform, angle, vec3.fromValues(ax, ay, az));
-    mat4.translate(transform, transform, vec3.fromValues(-cx, -cy, -cz));
+    mat4.translate(transform, transform, new Float64Array([cx, cy, cz]));
+    mat4.rotate(transform, transform, angle, new Float64Array([ax, ay, az]));
+    mat4.translate(transform, transform, new Float64Array([-cx, -cy, -cz]));
```
jourdain, line 93: When you do operation like that, you don't need to rework the array. You can leave [-cx, -cy, -cz] directly. That skip 1 new. 

If you want to prevent any new you could have a `tmp3 = Float64Array(3)` and call `vec3.set(tmp3, -cx, -cy, -cz)`

martinken: Yes I have already edited that locally. I pushed this so I could take a look at it as I find it easier to see the changes here.

Sources/Rendering/Core/Actor/index.js
```diff
@@ -102,18 +102,18 @@ function vtkActor(publicAPI, model) {
...
-        vec3.fromValues(bds[0], bds[2], bds[4]),
-        vec3.fromValues(bds[0], bds[3], bds[4]),
+        new Float64Array([bds[1], bds[3], bds[5]]),
+        new Float64Array([bds[1], bds[2], bds[5]]),
+        new Float64Array([bds[0], bds[2], bds[5]]),
+        new Float64Array([bds[0], bds[3], bds[5]]),
+        new Float64Array([bds[1], bds[3], bds[4]]),
+        new Float64Array([bds[1], bds[2], bds[4]]),
+        new Float64Array([bds[0], bds[2], bds[4]]),
+        new Float64Array([bds[0], bds[3], bds[4]]),
```
jourdain, line 112: Could be written like 

```
const bbox = [
   vec3.set(new Float64Array(3), bds[1], bds[3], bds[5]),
   vec3.set(new Float64Array(3), bds[1], bds[2], bds[5]),
   ...
]
```

jourdain: But I'm assuming this is not called in a tight loop, so it doesn't matter. Otherwise you should probably re-use those Float64Arrays...

martinken: It should call BoundingBox.getCorners probably as that is what it is doing and do it on one single Float64Array(24)

Sources/Common/DataModel/ImageData/index.js
finetjul: I do not understand the purpose of this function (or to be more precise, the purpose of `worldToIndexVec3`).
shouldn't it be simply
```
publicAPI.indexToWorld = (ain, aout = []) => {
 vec3.transformMat4(aout , ain, model.indexToWorld);
}
```
Javascript numbers are 64 bits by default, no ? Even on 32 bit machines, no? So that means that regular `Array`s are 64bits.

Why bother creating intermediate 64 bit arrays ?

Sources/Rendering/Core/ImageSlice/index.js
```diff
@@ -88,19 +89,11 @@ function vtkImageSlice(publicAPI, model) {
...
-        vec3.fromValues(bds[1], bds[2], bds[5]),
-        vec3.fromValues(bds[0], bds[2], bds[5]),
-        vec3.fromValues(bds[0], bds[3], bds[5]),
-        vec3.fromValues(bds[1], bds[3], bds[4]),
-        vec3.fromValues(bds[1], bds[2], bds[4]),
-        vec3.fromValues(bds[0], bds[2], bds[4]),
-        vec3.fromValues(bds[0], bds[3], bds[4]),
-      ];
+      const bbox = [];
+      vtkBoundingBox.getCorners(bds, bbox);
```
finetjul, line 93: it would seem that it would be easy to change the API of `vtkBoundingBox.getCorners` to allow the following:
```
const bbox = vtkBoundingBox.getCorners(bds);
```

Sources/Common/DataModel/ImageData/index.js
martinken: @jourdain @finetjul I did another pass and improved about another 50 places that were doing more allocs than needed.  If you can please take a look to check that I didn't do something wrong. There are a ton of other places in vtk.js that can be improved in terms of heap allocs etc.

More broadly I'm not 100% sure why we are using Float64 at all for these small arrays (3,4,9,16). WebGL takes Float32 so we are converting form Float64. We definitely want double calcs everywhere but Float64 and JS numbers are both double. From what I can find there is some belief that native arrays can be optimized more easily than generic arrays, but other than that I'm not sure if there is any difference for these small vector/matrices.

Sources/Rendering/Core/ImageSlice/index.js
martinken: Yes, but I don't want to lose focus on the topic here which was removing the use of vec3.create() as it was using Float32. We are drifting a bit. Personally I think it should have an additional signature that takes a Float64Array(24) as the current version requires 9 heap allocs. And generally I think it is best to pass in the return argument that way we can potentially avoid one heap alloc if the same memory is used every time (not the case above but it could be). 

But again, the goal of this MR was to remove vec3.create(); and friends due to their use of Float32. Personally I think many of the matrix allocs in ImageReslice should be class conts, there is no reason to realloc them every time. That would be an easy change, but the ImageReslice code is using Float64 and so for the purposes of this MR it is fine. And there are a lot of similar cases I saw throughout the code. Heap allocs === garbage collection === slow rendering and we'll have to keep our eye on them and clean them up where we can. Just looking at the diffs this topic should remove a ton of heap allocs, upgrade a lot of calcs to double, and add a few heap allocs due to the now required conversion back to Float32 in some cases for WebGL. But the use of double is the right move throughout vtk.js.

finetjul: I Agree, it might not be the focus of this MR.

I did not mean to remove the output as parameter but make it optional (slight decrease in perfs though) and return the output:  `export function getCorners(bounds, corners=[]) {... return corners;}`

I like your `Float64Array(24)` suggestion.

Sources/Common/DataModel/ImageData/index.js
```diff
@@ -564,10 +547,11 @@ export function extend(publicAPI, model, initialValues = {}) {
   vtkDataSet.extend(publicAPI, model, initialValues);
 
   if (!model.direction) {
-    model.direction = mat3.create();
+    model.direction = new Float64Array(9);
+    mat3.identity(model.direction);
   } else if (Array.isArray(model.direction)) {
     const dvals = model.direction.slice(0);
-    model.direction = mat3.create();
+    model.direction = new Float64Array(9);
```
finetjul, line 554: `model.direction = new Float64Array(model.direction)`
or, if you fear that direction.length > 9:
`model.direction = new Float64Array(model.direction.slice(0, 9))`

Sources/Widgets/Widgets3D/ImageCroppingWidget/helpers.js
```diff
@@ -6,10 +6,9 @@ export const AXES = ['-', '=', '+'];
 // ----------------------------------------------------------------------------
 
 export function transformVec3(ain, transform) {
-  const vin = vec3.fromValues(ain[0], ain[1], ain[2]);
-  const vout = vec3.create();
-  vec3.transformMat4(vout, vin, transform);
-  return [vout[0], vout[1], vout[2]];
+  const vout = new Float64Array(3);
+  vec3.transformMat4(vout, ain, transform);
+  return vout;
```
finetjul, line 11: `vout` could be added as a function parameter:  `transformVec3(ain, transform, vout=[])`

Sources/Common/DataModel/ImageData/index.js
finetjul: I briefly looked, it looks fine to me. Thanks !

I think we should decide of a convention: when should we instantiate Float64Array when a JS array?

I'm surprised you did not run into troubles with `macro.setArray`, it does not support well typed arrays.

jourdain (approved):
LGTM

Sources/Common/DataModel/ImageData/index.js
finetjul: There is actually a small issue: `vtkImageData.indexToWorldVec3` is now doing the same than `vtkImageData.indexToWorld`. One function must be removed. Same with `worldToIndex`.

martinken: There were a number of methods that had

const foo = Float64Array(3); 
...
return [[foo[0], foo[1], foo[2]];

that I changed to just return foo. I noticed that done in a few other places. That worried me a bit but it all seemed to work and I'm not sure why it wouldn't as my knowledge of JS type issues is weak. 

WRT Array versus Float64Array from what I can see TypedArrays are the fastest option, BUT the JS compilers are smart and as long as you keep a JS array homogeneous it basically makes it a TypedArray under the hood. My gut is that if we know an array should only be used for storing one type, we should explicitly create it as that type. But interested in what @jourdain is thinking as well.

finetjul:
I mostly agree with you on the convention, it will be some kind of a hybrid solution. Few remarks:

> * Use standard JS arrays [] when:
>   * you know the initial set of values (i.e.: [-1, 0, 0])

What if you know that you want an identity 4x4 matrix. I don't see us typing the 16 values, should we have helper functions/constants ? (currently we do `new Float64Array(16) + mat4.indentity`

>   * you want to dynamically grow its size

for all those points, vectors and matrices, size is typically fix.

>   * used inside a function

Indeed, for some reason, I find it more readable to default-initialize function parameters with `= []` rather than `= new Float64Array(3)`.

> * Use Typed Array when:
>   * When shared inside a class instance/module rather than function (vec3, mat4....)

Will it be fine with (json) serialization ?

>   * When those variables are meant to not grow

Sources/Common/DataModel/ImageData/index.js
jourdain: returning foo directly make sense. 

I share your opinion of Use TypedArray when `used for storing one type` and the size is known.

jourdain: Except maybe when you already know its content like `[0.5, 1, -1]`. I would not convert that to `new Float64Array([0.5, 1, -1])`.

jourdain:
`new Float64Array(16) + mat4.indentity` => `mat4.identity(new Float64Array(16))`

Sources/Common/DataModel/ImageData/index.js
martinken: Yup, thanks!

jourdain:
> Use Typed Array when:
> When shared inside a class instance/module rather than function (vec3, mat4....)
> Will it be fine with (json) serialization ?

Yes it will be fine in serialization but I was thinking private const var inside the instance rather than things in the model.

martinken:
Planning on merging this commit unless you guys see something bad, tests passed locally but there may still be a gotcha in there so if larger consumers have an issue let me know.

jourdain:
Go for it, I did not see anything obvious

PR #1799, opened 2021-03-25
fix(CubeAxes): various fixes to CubeAxesActor

martinken:
- It was recomputing the text atlas far too often
basically every camera change as when the camera changes it
was calling update with the camera as an argument which
made the force argument true.

- It was redrawing the grid lines twice every time

- Exposed an issue with missing default value for lastAspectRatio

- Now works with WebGPU

- minor updates to WebGPU due to recent changes in the spec

- added a faceVisibilityAngle where you can specify what angle
a face must be towards the viewer before it is visibile. Defaults
to 8 degrees.

### DIFF
--- a/Sources/Rendering/WebGPU/PixelSpaceCallbackMapper/index.js
+++ b/Sources/Rendering/WebGPU/PixelSpaceCallbackMapper/index.js
@@ -1,0 +1,64 @@
+import macro from 'vtk.js/Sources/macro';
+import vtkViewNode from 'vtk.js/Sources/Rendering/SceneGraph/ViewNode';
+
+// ----------------------------------------------------------------------------
+// vtkWebGPUPixelSpaceCallbackMapper methods
+// ----------------------------------------------------------------------------
+
+function vtkWebGPUPixelSpaceCallbackMapper(publicAPI, model) {
+  model.classHierarchy.push('vtkWebGPUPixelSpaceCallbackMapper');
+
+  publicAPI.opaquePass = (prepass, renderPass) => {
+    model.WebGPURenderer = publicAPI.getFirstAncestorOfType(
+      'vtkWebGPURenderer'
+    );
+    model.WebGPURenderWindow = model.WebGPURenderer.getParent();
+    const aspectRatio = model.WebGPURenderer.getAspectRatio();
+    const camera = model.WebGPURenderer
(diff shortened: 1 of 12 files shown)


Sources/Rendering/WebGPU/PixelSpaceCallbackMapper/index.js
```diff
@@ -0,0 +1,74 @@
...
+    );
+  };
+
+  publicAPI.queryPass = (prepass, renderPass) => {
+    if (prepass) {
+      if (model.renderable.getUseZValues()) {
+        renderPass.setDepthRequested(true);
+      } else {
+        renderPass.setDepthRequested(true);
+      }
```
jourdain, line 42: Why does the if/else call the same method?

Should this be something like `renderPass.setDepthRequested(model.renderable.getUseZValues())` with maybe a `!!` infront to make sure it is a boolean?

martinken: Because it is a straight copy of the OpenGL one :-) Maybe @scottwittenburg knows but likely a bug.

martinken: I think I know the correct logic and will give it a try

martinken: The wording on the renderPass method is not great. It should be renderPass.requestDepth(); as you only call it when you want depth values, otherwise you leave it alone. For WebGPU it will probably be removed when I get to code that uses it as with WebGPU I always use a texture backed depth buffer.

martinken: Actually I'm going to change that now as it isn't quite right.

scottwittenburg: Sorry I don't recall, but it looks wrong obviously.  I'm guessing if we're not using z values, then it was supposed to be depth requested = false. 🤷

martinken: Thanks @scottwittenburg   I think I have it fixed now, fingers crossed.

martinken:
@jourdain maybe take another look now.

jourdain (approved):
LGTM

PR #1800, opened 2021-03-25
feat(svgrepresentation): add picking support to SVG representations

finetjul:


Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -441,6 +442,23 @@ function vtkWidgetManager(publicAPI, model) {
 
   publicAPI.updateSelectionFromXY = (x, y) => {
     if (model.pickingEnabled) {
+      // First pick SVG representation
+      for (let i = 0; i < model.widgets.length; ++i) {
+        const widget = model.widgets[i];
+        const hoveredSVGReps = widget
+          .getRepresentations()
+          .filter((r) => r.isA('vtkSVGRepresentation') && r.getHover() != null);
```
floryst, line 450: Linter will likely complain about `!=`. Use `!==`.

Sources/Widgets/SVG/SVGRepresentation/index.js
```diff
@@ -149,14 +142,19 @@ function vtkSVGRepresentation(publicAPI, model) {
   publicAPI.render = () => {
     throw new Error('Not implemented');
   };
+
+  /**
+   * Return the whole widget state.
+   */
+  publicAPI.getSelectedState = (prop, compositeID) => model.inputData[0];
```
floryst, line 149: At some point do we want to try associating the prop/compositeID with the appropriate state? e.g. hovering over an SVG landmark rep to select the underlying handle state.

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -481,11 +499,19 @@ function vtkWidgetManager(publicAPI, model) {
...
-      return {};
+    let widget = null;
+    let representation = null;
+
+    if (propsWeakMap.has(prop)) {
+      const props = propsWeakMap.get(prop);
+      widget = props.widget;
+      representation = props.representation;
+    } else {
+      const selectionProps = model.selections[0].getProperties();
```
floryst, line 510: Can `model.selections` have zero length here, or are we always guaranteed to have at least one element?

floryst (changes requested):

Sources/Widgets/SVG/SVGRepresentation/index.js
finetjul: Eventually I think it could, yes.
It does not make sense for the LineWidget (tested with) though.

Sources/Widgets/Core/WidgetManager/index.js
finetjul: we are sure model.selections is not empty. First thing checked in `getSelectedData()`

Sources/Widgets/Core/WidgetManager/index.js
finetjul: linter does not complain.

I thought it was fine to use `==`/`!=` when comparing to `null`. That way it works for both `undefined` and `null`.

I can do `!==` if you prefer though. your call.

Sources/Widgets/SVG/SVGRepresentation/index.js
finetjul: Ah, I understand now. You were right, I could use the default implementation. I just had to switch `prop` and `compositeId`

Sources/Widgets/Core/WidgetManager/index.js
floryst: Ah, if you want to cover both `undefined` and `null` (and just those two) as being falsy, then yes you can use the non-strict version.

floryst (approved):
LGTM

PR #1812, opened 2021-03-31
Fix label representation

finetjul:


finetjul:
@luciemac

floryst (approved):
LGTM. While not necessarily relevant to this PR, are there other old-style widgets that don't handle HiDPI screens correctly?

finetjul:
Thanks @floryst . I am not aware of other old-style widget issues.

PR #1813, opened 2021-03-31
Interactor first renderer

jourdain:
This provide the first render in the callback of interactor as well. As some actions should be performed on that specific renderer in the style.


jourdain:
@martinken feel free to merge if you are good with that...

As a second stage we should have some style manipulator make use of the firstRenderer if need be.

martinken:
Sure +1

PR #1817, opened 2021-04-01

### LINKED ISSUE #466: Typescript type definitions

daker:
This should fix the `vtkMatrixBuilder` and `setData` definitions

fix #466

### DIFF
--- a/Sources/macro.d.ts
+++ b/Sources/macro.d.ts
@@ -56,6 +56,18 @@ export enum TYPED_ARRAYS {
   Int32Array,
 }
 
+export type TypedArray =
+    | number[]
+    | Uint32Array
+    | Uint16Array
+    | Uint8Array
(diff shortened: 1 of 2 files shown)


Sources/macro.d.ts
```diff
@@ -56,6 +56,18 @@ export enum TYPED_ARRAYS {
   Int32Array,
 }
 
+export type TypedArray =
+    | number[]
```
finetjul, line 60: @daker any reason why `number[]` is part of `TypedArray` ?

daker: i think you are right, it should not be part of it

PR #1819, opened 2021-04-02
feat(PLYReader): add basic support for face texture

robbie-li:
For PLY file with face texture, currently vtk.js cannot support the texture coordinates.

I add basic support to convert face texture to points texture.

### DIFF
--- a/Sources/IO/Geometry/PLYReader/index.js
+++ b/Sources/IO/Geometry/PLYReader/index.js
@@ -130,9 +130,9 @@ function postProcess(buffer, elements) {
   const normalsArray = new Float32Array(nbVerts * 3);
 
   const hasColor = buffer.colors.length > 0;
-  const hasVertTCoods = buffer.uvs.length > 0;
+  const hasVertTCoords = buffer.uvs.length > 0;
   const hasNorms = buffer.normals.length > 0;
-  const hasFaceTCoods = buffer.faceVertexUvs.length > 0;
+  const hasFaceTCoords = buffer.faceVertexUvs.length > 0;
 
   for (let vertIdx = 0; vertIdx < nbVerts; vertIdx++) {
     let a = vertIdx * 3 + 0;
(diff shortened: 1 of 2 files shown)


finetjul:
Would be great if you can update the example to show the usage of texture on PLY.

robbie-li:
> Would be great if you can update the example to show the usage of texture on PLY.

I add the example, you can select one PLY with one PNG/JPG together.

Sources/IO/Geometry/PLYReader/index.js
```diff
@@ -157,6 +163,27 @@ function postProcess(buffer, elements) {
     }
   }
 
+  if (hasFaceTCoods) {
```
finetjul, line 166: What if you have both `hasVertTCoords` and `hasFaceTCoods`?

robbie-li: I have never seen such type of PLY.
Do you have any suggestion, use VertTCoords or use FaceTCoords?

daker: I don't think we can have such case

robbie-li: Anyway, I prefer to take the VertTCoords in this case.

Sources/IO/Geometry/PLYReader/index.js
```diff
@@ -157,6 +163,25 @@ function postProcess(buffer, elements) {
     }
   }
 
+  if (!hasVertTCoords && hasFaceTCoords) {
+    // don't use array.shift, because buffer.indices will be used later
+    let idxVerts = 0;
+    let idxCoord = 0;
+    for (let faceIdx = 0; faceIdx < nbFaces; ++faceIdx) {
+      const nbFaceVerts = buffer.indices[idxVerts++];
+      const texcoods = buffer.faceVertexUvs[idxCoord++];
```
finetjul, line 172: texcoods -> texcoords

Sources/IO/Geometry/PLYReader/index.js
```diff
@@ -157,6 +163,25 @@ function postProcess(buffer, elements) {
...
+      if (nbFaceVerts * 2 === texcoods.length) {
```
finetjul, line 173: if it is false, shouldn't you increment idxVerts anyway ?

robbie-li: Yes, you are right.

finetjul:
LGTM

daker:
@jourdain can you please merge if it's ok for you?

PR #1821, opened 2021-04-05
feat(ScalarBar): add ScalarBarActor

martinken:
Add a scalar bar actor that when given a scalarToColors
instance (or subclass such as lookuptable) will draw
an annotation showing the colors, and ranges of the data.

Also draws a title as specified by the user.

If automated is true then it will try to scale and position
itself automatically to look nice.


jourdain (approved):
LGTM

PR #1824, opened 2021-04-06
docs(ts): add dox script to generate docs from typescript definitions

daker:
This pull request adds the ability to generate the documentation from .d.ts using `dox` module


jourdain:
The CI is failing as you need to push the lock file as well.

daker:
@jourdain can you check again?

jourdain:
I guess I'll see how it looks like on the deployed doc ;-)
Thanks for all your hard work, I'm eager to see the result of your work...

PR #1826, opened 2021-04-06
docs(ts): update api docs

daker:
This pull request update the api docs using the `doc:generate-api`


daker:
@jourdain not bad as a start https://kitware.github.io/vtk-js/api/Rendering_Core_Camera.html

PR #1833, opened 2021-04-08
feat(animationproxymanager): add animation manager, and animation proxies

boucaud:
An implementation of animations in the proxy system, with support for the vtkTimeStepBasedAnimationHandler.

The goal is integration into Paraview Glance.

### DIFF
--- a/Sources/IO/Core/HttpSceneLoader/index.js
+++ b/Sources/IO/Core/HttpSceneLoader/index.js
@@ -274,9 +274,7 @@ function defineLoadFuctionForReader(type) {
       .setUrl([model.baseURL, item[item.type].url].join('/'), {
         loadData: true,
       })
-      .then(() => {
-        publicAPI.invokeReady();
-      });
+      .then(() => publicAPI.invokeReady());
 
     applySettings(sceneItem, item);
     model.scene.push(sceneItem);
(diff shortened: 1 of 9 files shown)


boucaud:
@jourdain what do you think of the API ?

Also, I'm not sure if an animation manager instance should be directly composed in the ProxyManager, or if they should stay separate. What is you opinion ?

@finetjul FYI

Sources/IO/Core/HttpSceneLoader/index.js
```diff
@@ -307,8 +305,10 @@ function loadTimeStepBasedAnimationHandler(data, model) {
     scene: model.scene,
     originalMetadata: model.metadata,
     applySettings,
-    renderer: model.renderer,
   });
+  if (model.renderer) {
```
jourdain, line 309: Should we have both `model.animationHandler && model.renderer` in the if?

jourdain:
I would create the animation proxy from the proxy manager calling `pxm.create(group, name)`

jourdain:
Looks reasonable otherwise

boucaud:
> I would create the animation proxy from the proxy manager calling `pxm.create(group, name)`

I've updated the example to do it this way.

PR #1837, opened 2021-04-08
feat(MouseBoxSelectorManipulator): Add Box selector inside manipulators

jourdain:


jourdain:
If possible I would like to merge it on Friday morning... But I'm worried about breaking the doc ;-)

PR #1839, opened 2021-04-09
Fix widgets

finetjul:
Different fixes for the widgets.


Sources/Widgets/Representations/PolyLineRepresentation/index.js
```diff
@@ -45,6 +45,21 @@ function vtkPolyLineRepresentation(publicAPI, model) {
...
+   * Change the line/tube thickness.
+   * @param {number} lineThickness
+   */
+  function applyLineThickness(lineThickness) {
+    let scaledLineThickness = lineThickness;
+    if (publicAPI.getScaleInPixels()) {
+      const center = vtkBoundingBox.getCenter(
+        model.internalPolyData.getBounds()
+      );
+      scaledLineThickness *= publicAPI.getPixelWorldHeightAtCoord(center);
```
floryst, line 58: Does this also work if the line is not perpendicular to the viewing direction? Seems like the line would look like a perspective-projected cylinder to me.

Sources/Interaction/Widgets/AbstractWidget/index.js
```diff
@@ -54,6 +54,9 @@ function vtkAbstractWidget(publicAPI, model) {
       model.interactor.render();
     }
   };
+
+  publicAPI.isDragable = () =>
```
floryst, line 58: "dragable" -> "draggable"

Sources/Widgets/Core/AbstractWidgetFactory/api.md
```diff
@@ -41,6 +41,10 @@ Sets visibility for all associated view widgets.
 
 Sets pickable flag for all associated view widgets.
 
+## setDragable(bool)
```
floryst, line 44: dragable -> draggable

floryst (commented):
I didn't closely examine the animation additions, but from a pass-over it seems reasonable.

I looked around in VTK and saw that they also use the "dragable" spelling. However, I don't see "dragable" as some alternative spelling of "draggable".

Sources/Widgets/Representations/PolyLineRepresentation/index.js
finetjul: you are correct, it looks like a perspective-projected cylinder in that case.

I do not have anything better to suggest though.

finetjul:
> I looked around in VTK and saw that they also use the "dragable" spelling. However, I don't see "dragable" as some alternative spelling of "draggable".

Problem is, `Rendering/Core/Prop` declares a 'dragable' property. Do you suggest to rename it everywhere ?

finetjul:
I merged the MR, but we can still iterate on those discussions, no problem. I just needed the fixes to be merged for the time being. Thanks.

PR #1843, opened 2021-04-11

daker:
Add more fixes to the typescript definitions


floryst:
I only did a cursory glance, but looks good to me.

PR #1844, opened 2021-04-11
chore(build): Generate api docs at build step

daker:
This will help generate the API docs at build time rather than pushing them each time.

@floryst


floryst:
LGTM, thanks!

PR #1847, opened 2021-04-12
chore(WebGPU): add a UBO class and use it

martinken:
Add a Uniform Buffer Object class for WebGPU to manage UBOs
and make it easier for people to extend mappers etc by adding
additional fields to the UBOs used in a pipeline.

The UBO class handles memory alignment issues but currently
only supports basic types such as vec/mat not arrays or
structs as members.

### DIFF
--- a/Sources/Common/DataModel/Cell/test/testCell.js
+++ b/Sources/Common/DataModel/Cell/test/testCell.js
@@ -27,7 +27,7 @@ test('Test vtkCell initialize without pointsIds', (t) => {
 
 test('Test vtkCell initialize with pointsIds', (t) => {
   const points = vtkPoints.newInstance();
-  points.setData([0, 0, 0, 2, 0, 0, 2, 2, 0]);
+  points.setData(Float64Array.from([0, 0, 0, 2, 0, 0, 2, 2, 0]));
 
   const cell = vtkCell.newInstance();
 
--- a/Sources/Filters/General/TriangleFilter/index.js
+++ b/Sources/Filters/General/TriangleFilter/index.js
@@ -73,10 +73,10 @@ function vtkTriangleFilter(publicAPI, model) {
     const dataset = vtkPolyData.newInstance();
     dataset
       .getPoints()
-      .setData(macro.TYPED_ARRAYS[pointsDataType].from(newPoints));
+      .setData(macro.newTypedArrayFrom(pointsDataType, newPoints));
     dataset
       .getPolys()
-      .setData(macro.TYPED_ARRAYS[cellsDataType].from(newCells));
+      .setData(macro.newTypedArrayFrom(cellsDataType, newCells));
 
     outData[0] = dataset;
   };
--- a/Sources/Rendering/WebGPU/Types/index.js
+++ b/Sources/Rendering/WebGPU/Types/index.js
@@ -1,3 +1,5 @@
+import { vtkErrorMacro } from 'vtk.js/Sources/macro';
+
 // ----------------------------------------------------------------------------
 // vtkWebGPUDevice static functions
 //
@@ -10,34 +12,59 @@
 //  - texture types such as rgba32float
 // ----------------------------------------------------------------------------
 
+// see https://gpuweb.github.io/gpuweb/#enumdef-gpuvertexformat
+// for possible formats
 function getByteStrideFromBufferFormat(format) {
-  if (!format) return 0;
-  let numComp = 1;
-  if (format.substring(format.length - 2) === 'x4') numComp = 4;
-  if (format.substring(format.length - 2) === 'x3') numComp = 3;
-  if (format.substring(format.length - 2) === 'x2') numComp = 2;
+  if (!format || format.length < 5) return 0;
 
-  let typeSize = 4;
-  if (numComp > 1) {
-    if (format.substring(format.length - 3, format.length - 1) === '8x') {
-      typeSize = 1;
-    }
-    if (format.substring(format.length - 4, format.length - 1) === '16x') {
-      typeSize = 2;
-    }
-  } else {
-    if (format.substring(format.length - 1) === '8') typeSize = 1;
-    if (format.substring(format.length - 2) === '16') typeSize = 2;
+  // options are x2, x3, x4 or nothing
+  let numComp = 1;
+  if (format[format.length - 2] === 'x') {
+    numComp = format[format.length - 1];
   }
+
+  const sizeStart = numComp === 1 ? format.length - 1 : format.length - 3;
+  // options are 8, 16, 32 resulting in 8, 6, 2 as the last char
+  // plugged into the formula below gives 1, 2, 4 respectively
+  const num = Number(format[sizeStart]);
+  if (Number.isNaN(num)) {
+    vtkErrorMacro(`unknown format ${format}`);
+    return 0;
+  }
+  const typeSize = 5 - num / 2;
   return numComp * typeSize;
 }
 
+// see https://gpuweb.github.io/gpuweb/#enumdef-gpuvertexformat
+// for possible formats
 function getNativeTypeFromBufferFormat(format) {
-  if (!format) return null;
-  if (format.substring(0, 7) === 'float32') return 'Float32Array';
-  if (format.substring(0, 6) === 'snorm8') return 'Int8Array';
-  if (format.substring(0, 6) === 'unorm8') return 'Uint8Array';
-  return '';
+  if (!format || format.length < 5) return 0;
+
+  // raw types are Uint Int or Float as follows
+  let result;
+  if (format[0] === 'f') {
+    result = 'Float';
+  } else if (format[0] === 's') {
+    result = 'Int';
+  } else if (format[0] === 'u') {
+    result = 'Uint';
+  } else {
+    vtkErrorMacro(`unknown format ${format}`);
+    return undefined;
+  }
+
+  // options are 8, 16, 32 resulting in 8, 6, 2 as the last char
+  // plugged into the formula below gives 1, 2, 4 respectively
+  const base = format.split('x')[0];
+  const num = Number(base[base.length - 1]);
+  if (Number.isNaN(num)) {
+    vtkErrorMacro(`unknown format ${format}`);
+    return undefined;
+  }
+  result += 8 * (5 - num / 2);
+  result += 'Array';
+
+  return result;
 }
 
 function getByteStrideFromShaderFormat(format) {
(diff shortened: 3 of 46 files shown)


Sources/Rendering/WebGPU/Types/index.js
```diff
@@ -0,0 +1,74 @@
...
+// WebGPU currently tends to use multiple type formats:
+//  - buffer types such as float32x4
+//  - shader types suchs as vec4<f32>
+//  - texture types such as rgba32float
+// ----------------------------------------------------------------------------
+
+function getByteStrideFromBufferFormat(format) {
+  if (!format) return 0;
+  let numComp = 1;
+  if (format.substring(format.length - 2) === 'x4') numComp = 4;
```
jourdain, line 16: could this be replaced by `const numComp = Number(format.split('x').pop());` ?
It will be less efficient if called often though.

Sources/Rendering/WebGPU/Types/index.js
```diff
@@ -0,0 +1,74 @@
...
+      typeSize = 1;
+    }
+    if (format.substring(format.length - 4, format.length - 1) === '16x') {
+      typeSize = 2;
+    }
+  } else {
+    if (format.substring(format.length - 1) === '8') typeSize = 1;
+    if (format.substring(format.length - 2) === '16') typeSize = 2;
+  }
+  return numComp * typeSize;
```
jourdain, line 32: Hum it event looks like it could be written like 

```
return format.split('x').map(Number).reduce((a,b) => a*b, 1/8);
```

Sources/Rendering/WebGPU/Types/index.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function getNativeTypeFromBufferFormat(format) {
+  if (!format) return null;
+  if (format.substring(0, 7) === 'float32') return 'Float32Array';
+  if (format.substring(0, 6) === 'snorm8') return 'Int8Array';
+  if (format.substring(0, 6) === 'unorm8') return 'Uint8Array';
+  return '';
```
jourdain, line 40: Should those test be done like `if (format.indexOf('float32') === 0) return 'Float32Array';` ? 
That would prevent any object creation.

jourdain: unless the format string could be long hence the substring?

jourdain: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith could be a better match

Sources/Rendering/WebGPU/Types/index.js
```diff
@@ -0,0 +1,74 @@
...
+function getNativeTypeFromShaderFormat(format) {
+  if (!format) return null;
+  let type = format.substring(format.length - 3, format.length);
+  if (format[format.length - 1] === '>') {
+    type = format.substring(format.length - 4, format.length - 1);
+  }
+  if (type === 'f32') return 'Float32Array';
+  if (type === 'i32') return 'Int32Array';
+  if (type === 'u32') return 'Uint32Array';
+  return '';
```
jourdain, line 66: do you need to be spot on or can you just use the `includes`? 
You can also create static object { f32: 'Float32Array', ... } and do the lookup that way rather than a set of if...

Sources/Rendering/WebGPU/UniformBuffer/index.js
```diff
@@ -0,0 +1,338 @@
...
+      model.sendDirty = false;
+    }
+  };
+
+  publicAPI.createView = (type) => {
+    if (type in model === false) {
+      if (!model.arrayBuffer) {
+        model.arrayBuffer = new ArrayBuffer(model.sizeInBytes);
+      }
+      model[type] = new window[type](model.arrayBuffer);
```
jourdain, line 217: You should not use `window[type]` but `TYPED_ARRAYS[type]` instead from `import { TYPED_ARRAYS } from 'vtk.js/Sources/macro';`.

The reason being that `window` does not exist outside of a browser. That way we can support some usage of vtk.js inside node/deno and else...

Sources/Rendering/WebGPU/UniformBuffer/index.js
```diff
@@ -0,0 +1,338 @@
...
+    publicAPI.sortBufferEntries();
+
+    let result = `[[block]] struct ${model.name}Struct\n{\n`;
+    for (let i = 0; i < model.bufferEntries.length; i++) {
+      const entry = model.bufferEntries[i];
+      result += `[[offset(${entry.offset})]] ${entry.name}: ${entry.type};\n`;
+    }
+    result += `};\n[[binding(${model.binding}), group(${model.group})]] var<uniform> ${model.name}: ${model.name}Struct;`;
+    return result;
+  };
```
jourdain, line 285: It is usually better / faster to use an array and just join it with `\n` at the end than doing a `+=` on string.

```
const lines = [
   `[[block]] struct ${model.name}Struct`,
   '{',
    model.bufferEntries.map(({offset, name, type}) => `[[offset(${offset})]] ${name}: ${type}`).join('\n'),
   '};',
   `[[binding(${model.binding}), group(${model.group})]] var<uniform> ${model.name}: ${model.name}Struct;`
];
return lines.join('\n');
```

Sources/Rendering/WebGPU/Types/index.js
martinken: That doesn't work because an input like float32x4 splits to float32, 4 and float32 fails Number. 

I'll rework it to be a bit shorter and do less heap allocs.

Sources/Rendering/WebGPU/UniformBuffer/index.js
```diff
@@ -0,0 +1,338 @@
...
+
+  // Internal objects
+  model.bufferEntryNames = new Map();
+  model.bufferEntries = [];
+
+  model.sendTime = {};
+  macro.obj(model.sendTime, { mtime: 0 });
+
+  model.sendDirty = true;
+  model.sortDirty = true;
```
jourdain, line 317: should they be part of the defaultValues?

Sources/Rendering/WebGPU/UniformBuffer/index.js
```diff
@@ -0,0 +1,338 @@
...
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  // Build VTK API
+  macro.obj(publicAPI, model);
+
+  // Internal objects
+  model.bufferEntryNames = new Map();
```
jourdain, line 310: ideally no`Map()` in model. Can this just be a const fully private? (in function rather than extend)?
Otherwise for this type of classes we don't care as we won't ask for `.getState()`...

jourdain (approved):
In general LGTM

Sources/Rendering/WebGPU/UniformBuffer/index.js
martinken: OK I just copied it form DataArray. Should I change DataArray as well?

Sources/Rendering/WebGPU/Types/index.js
jourdain: I see, but if not called often, the overhead is not the end of the world. ;-)

You can use some pre-compiled regexp for the extraction of those 2 numbers ;-)

Sources/Rendering/WebGPU/UniformBuffer/index.js
jourdain: yes if you can. The `TYPED_ARRAYS` was added later when some folks used vtk.js on node and it was not working...

Sources/Rendering/WebGPU/UniformBuffer/index.js
martinken: Hmm, Map is a really useful class to me and using an Object as a map seems like bad practice for a few reasons. While I can make Maps private that really kills the ability for people to subclass and do things that require access to that member. So... since this will continue to come up, how would you like me to implement protected Maps such that they are not private?

jourdain: I'm not against using `Map()`, I'm just against binding it to model if not mandatory (accessible for sub-classes).

jourdain: you can create a model._protected = { ... all the protected objects }

jourdain: scratch that... Just put it inside the model with a `_` as prefix. I guess the idea would be that things in the model with an `_` should not have a getter and should not be serialized. That way we can update the `getState` to take into account that aspect.

jourdain: The linter will bother you on that but you can turn it off for that part specifically

Sources/Rendering/WebGPU/Types/index.js
martinken: Reworked to support the current full list of types.

Sources/Rendering/WebGPU/Types/index.js
martinken: includes works in this context, I'll go that way.

Sources/Rendering/WebGPU/UniformBuffer/index.js
martinken: OK that was a large change. There were A LOT of window[dataType] uses in the code (which could have been changed when TYPED_ARRAYS was added). Plus some folks were passing Array as a data type which window[Array] worked but TYPED_ARRAYS[Array] does not. We really do not want non-native arrays in DataArray so I made some changes to handle that case as well.

Sources/Rendering/WebGPU/UniformBuffer/index.js
martinken: Nah, they are the protected flavor of public.

Sources/Rendering/WebGPU/UniformBuffer/index.js
martinken: If we are going to use that as a convention in vtk.js then we should add it to our .eslintrc.js yes?

jourdain: yes let's add that to our eslintrs

Sources/Common/DataModel/Cell/test/testCell.js
```diff
@@ -27,7 +27,7 @@ test('Test vtkCell initialize without pointsIds', (t) => {
 
 test('Test vtkCell initialize with pointsIds', (t) => {
   const points = vtkPoints.newInstance();
-  points.setData([0, 0, 0, 2, 0, 0, 2, 2, 0]);
+  points.setData(new Float64Array([0, 0, 0, 2, 0, 0, 2, 2, 0]));
```
jourdain, line 30: Float64Array.from([..]) ?

Sources/Filters/General/TriangleFilter/index.js
```diff
@@ -73,10 +73,10 @@ function vtkTriangleFilter(publicAPI, model) {
     const dataset = vtkPolyData.newInstance();
     dataset
       .getPoints()
-      .setData(macro.TYPED_ARRAYS[pointsDataType].from(newPoints));
+      .setData(macro.macro.newTypedArrayFrom(pointsDataType, newPoints));
     dataset
       .getPolys()
-      .setData(macro.TYPED_ARRAYS[cellsDataType].from(newCells));
+      .setData(macro.macro.newTypedArrayFrom(cellsDataType, newCells));
```
jourdain, line 79: macro.macro

PR #1848, opened 2021-04-12
fix(macro): add typedarray support to setArray macro function

### LINKED ISSUE #1832: ImageCroppingWidget partially not working
**Describe the bug**

In ImageCroppingWidget faces and edges handles do not work. Only corners handles produce an effect.

This can be seen in ImageCroppingWidget example (https://kitware.github.io/vtk-js/examples/ImageCroppingWidget.html ) as long as locally using 17.5.0 vtk.js version.

finetjul:
typed array can be passed as unique argument

fix #1832


jourdain:
LGTM

PR #1853, opened 2021-04-14
chore(Rendering): Remove some OpenGL Dependencies

martinken:
Start removing places where people explicitly used
OpenGL so that the code can be more rendering API
agnostic.

Also fix one test that was not releasing its resources.

### DIFF
--- a/Examples/Widgets/Box/BoxWidget.js
+++ b/Examples/Widgets/Box/BoxWidget.js
@@ -55,7 +55,7 @@ function widgetBehavior(publicAPI, model) {
       model.manipulator.setNormal(model.camera.getDirectionOfProjection());
       const worldCoords = model.manipulator.handleEvent(
         callData,
-        model.openGLRenderWindow
+        model.apiSpecificRenderWindow
       );
 
       if (worldCoords.length) {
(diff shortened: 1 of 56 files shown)


martinken:
@floryst @jourdain  this passes the tests, but... I'm not sure how much the widgets are tested in the base tests or if these changes will impact glance etc. So please take a look.

jourdain:
At a first glance that looks reasonable... But a true widget validation should be performed.

jourdain:
Some examples are currently broken because of an update method name?

`const openGLRenderWindow = fullScreenRenderer.getOpenGLRenderWindow();`

martinken:
Yes from a prior commit, but I'll update this one to fix it for those two examples.

martinken:
@jourdain can you take a look at how I changed testCategoricalColors.js  I want to change all the tests but wanted to run it by you first.

jourdain:
Your change to `testCategoricalColors.js` looks fine to me. Maybe put the import `import 'vtk.js/Sources/Rendering/Misc/RenderingAPIs';` inside the root test.js file instead. So one location for all the tests...

martinken:
I might move it there in the future. As I will likely move some tests in OpenGL to RenderingCore in the future as well. I haven;t gotten to the point of trying to really run the tests on WebGPU but when I do I'll probably do another round of cleanup.

martinken:
Manually ran and played with five of the widget examples and they seemed to work. So merging. If someone runs into an issue let me know. If needed we can add a widget.getOpenGLRenderWindow method for backwards compatibility that just returns the apiSpecificWindow, if there are external codes that require it and don't want to update.

PR #1859, opened 2021-04-19
docs(ts): add more typescript definitions

daker:


daker:
@jourdain more TS definitions, i am starting to cleanup api.md since it's generated at build step

PR #1860, opened 2021-04-19
feat(WebGPU): add hardware selector support

martinken:
add supporrt for actor/composite picking in WebGPU

As WebGPU is a more asynchronous API this required creating
a new API for selections that is async. In this case the
way to do selections is

const srcData = await publicAPI.getSourceDataAsync(fx1, fy1, fx2, fy2);
if (srcData) {
  return srcData.generateSelection(fx1, fy1, fx2, fy2);
}

In the call to getSourceData you can get the entire window and
then make multiple calls to generate selection on the result.

Note that generateSelection is called on the result, not on the
selector as there may be multiple selections in flight at a time
so the state of the selector may not be correct for a specific
selection. All required state for generateSelection is kept in
the returned srcData object.

Reworked the WebGL hardware selector to support the new API
in addition to the old API.

Rewrote the hardware selector example to use the new API.

Left the other couple cases of using the hardware selector
using the old API for now as making them async could be a
breaking change.


martinken:
@jourdain as usual there is a ton of code here but if you can take a look it's probably worth it.

Sources/Rendering/Core/HardwareSelector/example/index.js
```diff
@@ -151,13 +150,12 @@ renderWindow.render();
...
-const hardwareSelector = vtkOpenGLHardwareSelector.newInstance({
-  captureZValues: true,
-});
+const hardwareSelector = apiSpecificRenderWindow.getSelector();
+hardwareSelector.setCaptureZValues(true);
 hardwareSelector.setFieldAssociation(
   FieldAssociations.FIELD_ASSOCIATION_POINTS
 );
-hardwareSelector.attach(openGLRenderWindow, renderer);
+hardwareSelector.attach(apiSpecificRenderWindow, renderer);
```
jourdain, line 158: Should the `hardwareSelector` already have a reference to `apiSpecificRenderWindow` since it was created from it? Could we just attach/setRenderer?

martinken: Yes we could. That API is old so I just left it as it wast for backwards compatibility and the off chance someone explicitly creates a hardware selector bypassing window.getHardwareSelector

Sources/Rendering/WebGPU/HardwareSelector/index.js
```diff
@@ -0,0 +1,441 @@
...
+import vtkWebGPUBuffer from 'vtk.js/Sources/Rendering/WebGPU/Buffer';
+import vtkWebGPUHardwareSelectionPass from 'vtk.js/Sources/Rendering/WebGPU/HardwareSelectionPass';
+import vtkSelectionNode from 'vtk.js/Sources/Common/DataModel/SelectionNode';
+import vtkDataSet from 'vtk.js/Sources/Common/DataModel/DataSet';
+
+const { SelectionContent, SelectionField } = vtkSelectionNode;
+const { FieldAssociations } = vtkDataSet;
+const { vtkErrorMacro } = macro;
+
+function getInfoHash(info) {
```
jourdain, line 12: looks familiar? Are some of those functions common between implementations?

Sources/Rendering/Core/HardwareSelector/example/index.js
jourdain: Should we have `setView`, `setRenderer` and `attach(view, renderer)` (as-is) while deprecating the `attach` one?

martinken: Yes, that makes more sense. Or we could have renderer be an argument to getSourceDataAsync as what it does is specify what renderer to do the selection on. Which do you think is better?

jourdain: I think it will be less error prone to pass the renderer at the time you call `getSourceDataAsync`. That way the user never needs to remember to bind/attach/set anything...

Sources/Rendering/WebGPU/HardwareSelector/index.js
martinken: Some commonality but something like a raytracer would probably not have them. I'm trying to keep the Rendering\Core part to be just the public API. So any common code would have to land somewhere else.

Sources/Rendering/Core/HardwareSelector/example/index.js
jourdain: I'm not a big fan of async in the method name... I don't know if we need to be explicit in the name that we return a promise...

Sources/Rendering/WebGPU/HardwareSelector/index.js
jourdain: Sure it could be in `Rendering/Core/HardwareSelector/helpers.js` if this is more than just that function. I didn't double check but if the `convert` one could be shared and maybe others? 
If not, then don't worry about it.

Sources/Rendering/Core/HardwareSelector/example/index.js
martinken: Well we should have a convention but I'm fine either way. I just followed WebGPU's approach which appends Async onto any async methods. So feel free to make a call and let me know and I'll update to match.

jourdain: If WebGPU does that, I'm fine keeping the same structure.

PR #1861, opened 2021-04-20
Fix vtkTriangle intersect with line

laurennlam:
@finetjul


finetjul:
I am not sure `betweenPoints` brings much value because it can easily be computed from `t`, but ok to add it though.

It is missing documentation in `api.md` (which has an incorrect documentation for `t` btw)

`outObj` may not return `betweenPoints` in the second part of the function (with line intersection).

laurennlam:
> I am not sure `betweenPoints` brings much value because it can easily be computed from `t`, but ok to add it though.

Yes, but as it's already computed in the vtkPlane.intersectWithLine, why don't using it.

> It is missing documentation in `api.md` (which has an incorrect documentation for `t` btw)
> 
> `outObj` may not return `betweenPoints` in the second part of the function (with line intersection).
On my project, it's quite important. But I suppose that it depends on when you consider that a vtkTriangle is intersected.
Before the vtkPlane.intersectWithLine has been updated, we considered an intersection only if the intersection was between points. But the implementation has changed. Now vtkPlane.intersectWithLine returns True for intersection even if the intersection is out of between points... That's why it seems to me very important to transfer this data out of vtkTriangle.intersectWithLine.
WDYT ?

Sources/Common/DataModel/Triangle/index.js
```diff
@@ -51,7 +51,7 @@ function vtkTriangle(publicAPI, model) {
 
   publicAPI.getCellDimension = () => 2;
   publicAPI.intersectWithLine = (p1, p2, tol, x, pcoords) => {
-    const outObj = { subId: 0, t: 0, intersect: -1 };
+    const outObj = { subId: 0, t: 0, intersect: -1, betweenPoints: false };
```
finetjul, line 54: why not compute `betweenPoints` in the line line intersection ?

finetjul: which btw, it is also missing `t` in `outObj`

Sources/Common/DataModel/Triangle/index.js
```diff
@@ -69,6 +69,7 @@ function vtkTriangle(publicAPI, model) {
     if (n[0] !== 0 || n[1] !== 0 || n[2] !== 0) {
       // Intersect plane of triangle with line
       const plane = vtkPlane.intersectWithLine(p1, p2, pt1, n);
```
finetjul, line 71: it is too bad that `vtkPlane` returns a key named `intersection` and not `intersect`.
That would be more consistent with `vtkLine`, `vtkTriangle`, `vtkCellPicker`.

@jourdain WDYT is it worth it to rename `intersection` into `intersect` (or inversely)? That would be a breaking change though.

Sources/Common/DataModel/Triangle/index.js
laurennlam: I just adapted the vtkTriangle for a quick fix because du vtkPlane.intersectWithLine has been updated a long time ago.
I need a quick fix for a client, I didn't investigate more about it.
The same question will come back. When you call "intersectWithLine" what do you expect in output if the intersection is between point or not ?

Sources/Common/DataModel/Line/index.js
```diff
@@ -166,6 +176,7 @@ function vtkLine(publicAPI, model) {
       // one of the parametric coords must be outside 0-1
       if (outObj.t < 0.0) {
         outObj.t = 0.0;
+        outObj.betweenPoints = isBetweenPoints(outObj.t);
```
finetjul, line 179: I wouldn't recompute `betweenPoints` here

Sources/Common/DataModel/Line/index.js
```diff
@@ -176,6 +187,7 @@ function vtkLine(publicAPI, model) {
       }
       if (outObj.t > 1.0) {
         outObj.t = 1.0;
+        outObj.betweenPoints = isBetweenPoints(outObj.t);
```
finetjul, line 190: I wouldn't recompute `betweenPoints` here

Sources/Common/DataModel/Line/index.js
laurennlam: I think it's mandatory because if the value is set to 0.0, then it's considered between points. Or, as outObj.t is inferior to zero right before, the betweenPoint value would probably be false before entering in if.

Sources/Common/DataModel/Line/index.js
laurennlam: Same as the previous comment.

Sources/Common/DataModel/Line/index.js
finetjul: fair enough, I would do the following however:
```
const intersect = intersection(p1, p2, a1, a2, u, v);
outObj.t = u[0];
outObj.betweenPoints = (t >= 0.0 && t <= 1.0);
pcoords[0] = v[0];
...

if (outObj.t < 0.0) {
  outDistance = distanceToLine(p1, a1, a2, x)
  if (outDistance.distance <= tol * tol) {
    outObj.t = 0.0;
    outObj.betweenPoints = true; // intersection is near  p1
    outObj.intersect = 1;
  }
  return outObj;
}
if (outObj.t > 1.0) {
  outDistance = distanceToLine(p2, a1, a2, x)
  if (outDistance.distance <= tol * tol) {
    outObj.t = 1.0;
    outObj.betweenPoints = true; // intersection is near p2
    outObj.intersect = 1;
  }
  return outObj;
}
if (pcoords[1] < 0.) {
  ...
    outObj.betweenPoints = false; // intersection is outside [p1, p2]
  ...
}
if (pcoords[1] < 0.) {
  ...
    outObj.betweenPoints = false; // intersection is outside [p1, p2]
  ...
}

```

Sources/Common/DataModel/Line/index.js
```diff
@@ -132,9 +132,18 @@ function vtkLine(publicAPI, model) {
...
+  function isBetweenPoints(t) {
+    return t >= 0.0 && t <= 1.0;
+  }
+
   publicAPI.getCellDimension = () => 1;
   publicAPI.intersectWithLine = (p1, p2, tol, x, pcoords) => {
-    const outObj = { intersect: 0, t: Number.MIN_VALUE, subId: 0 };
+    const outObj = {
+      intersect: 0,
+      t: Number.MIN_VALUE,
```
finetjul, line 143: plane initializes `t` to Number.MAX_VALUE, shouldn't we do the same ?

Sources/Common/DataModel/Triangle/index.js
```diff
@@ -51,7 +51,7 @@ function vtkTriangle(publicAPI, model) {
 
   publicAPI.getCellDimension = () => 2;
   publicAPI.intersectWithLine = (p1, p2, tol, x, pcoords) => {
-    const outObj = { subId: 0, t: 0, intersect: -1 };
+    const outObj = { subId: 0, t: 0, intersect: 0, betweenPoints: false };
```
finetjul, line 54: how about `t: Number.MAX_VALUE`

PR #1877, opened 2021-04-25
docs(ts): add Misc typescript definitions

daker:
This introduces types.ts which will hold common types and reference the gl-matrix types. The next step is to use those types in index.d.ts directly instead of the importing them.

### Usage

```js
///<reference path='$path_to_types.ts' />
```


jourdain:
LGTM

Issue #848, opened 2018-08-22
Issue with tubeFilter


ghost:
Hi,

I get an issue when i try to tubeFilter some vtp files.

![issue_tubing](https://user-images.githubusercontent.com/40494123/44451792-0092ee80-a5f5-11e8-9ceb-8931d3577b5c.png)

As you can see, some tubes are not drawn. I guess it's because the lines are not "smooth enough" (the derivate is not continuous enough), but maybe you already got this issue in the past.

Have a good day.

jourdain:
@sankhesh

sankhesh:
@SchLagOs The tube filter does not add any curvature/bend or smart smoothing at the intersection of two tubes. Try removing intermediate points from the input polylines based on a curvature threshold.

SchLagOs:
@sankhesh thank you for your answer. However, due to process which must be very accurate, i can't remove points from the input polylines. Is there an other way to solve this ?

sankhesh:
@SchLagOs The better (and right) way to solve it would be to add capabilities to the tube filter. Ideally, you want a tubular end to each tube based on the parameters computed for the next or prior tube.

sankhesh:
Another possible solution, and this one wouldn't require modifications to the tube filter, is to pass the input polylines through a connectivity filter and reduce each loop to be a single cell (line) with the same number of points.

jourdain:
Closing as it could be handled by pre-processing user data to match vtkTubeFilter expectation

Issue #1546, opened 2020-07-29
Tube filter glitch


EmileSonneveld:
Hi, the tube filter in VTK.js behaves different than in c++.
In the following image, green is from VTK.js and gray from Paraview 5.8.0
![image](https://user-images.githubusercontent.com/2298426/88802001-c223e800-d1aa-11ea-87fc-e2e90b26c933.png)

Here is the code I used to reproduce in VTK.js: (the polydata line is embedded in the JS)
[tube-code.txt](https://github.com/Kitware/vtk-js/files/4994937/tube-code.txt)

Is this a problem with my poly data? code? vtk.js?
Thank you so much :)
Emile

EmileSonneveld:
@sankhesh do you know more? I saw that you contributed a lot to this :)

sankhesh:
@EmileSonneveld I'll take a look and let you know

sankhesh:
@EmileSonneveld I extracted the data from the text file you uploaded into a vtp file. Essentially just copying everything from the `VTKFile` section to the vtp file. 

Next, I ran it through the [tubesviewer](https://kitware.github.io/vtk-js/examples/TubesViewer/index.html) vtk-js application and paraview. The results are as in the screenshots below:

### TubesViewer
![Screenshot from 2020-08-04 14-53-48](https://user-images.githubusercontent.com/936354/89333381-dfe9c500-d662-11ea-84a8-736f0c4acb32.png)

### ParaView
![paraview-tube](https://user-images.githubusercontent.com/936354/89333400-e7a96980-d662-11ea-9765-4e18586173ff.png)

There is definitely an issue where the last line/cell is ignored by the vtk-js tube filter but I don't see the spike(?) that you show in the picture.

EmileSonneveld:
That spike is indeed mysterious. I had 2 exports with the spike on a different location. IIRC, I got it when exporting to STL, but not for VTI.

sankhesh:
>> IIRC, I got it when exporting to STL,

That would indicate an issue with the STL export and not tube filter.

EmileSonneveld:
Any fix possible for this one? Or an easy workaround?

sankhesh:
Since the issue is in the STL export code, you'd have to debug that. Or use a different geometry export format (e.g. OBJ).

EmileSonneveld:
The problem occurs before the STL export. STL export could have a second issue, but that can only be investigated after the tube filter is fixed.

[...]

PR #1880, opened 2021-04-27
fix(tubefilter): fix incorrectly sized output array in TubeFilter

jbeezley:
When drawing multiple lines, the output array is too small.  This causes segments to be missing from the scene.

A minimal example is
```js
const lines = vtk({
  vtkClass: 'vtkPolyData',
  points: { 
    vtkClass: 'vtkPoints',
    dataType: 'Float32Array',
    numberOfComponents: 3,
    values: [ 
      0,0,0,
      1,1,0,
      1,1,1,
      0,1,1,
    ] 
  },
  lines: { 
    vtkClass: 'vtkCellArray',
    dataType: 'Uint16Array',
    values: [ 
      2, 0, 2,
      2, 1, 3,
    ],
  },
});

const tubeFilter = vtkTubeFilter.newInstance({numberOfSides: 3, capping: false, sidesShareVertices: false});
tubeFilter.setInputData(lines);
```
where the second line is missing on current master.

Not sure, but this could be related to #848 as well.


jbeezley:
Actually, now that I look at this more, it still isn't correct when lines have varying number of points.

daker:
Is this related to #1546 ?

jbeezley:
Indeed it is.

jbeezley:
The logic is more complicated than I originally thought.  In the latest version, I'm calculating the output array size that is consistent with how it is being used in `generatePoints`.  I don't have great insight in how precisely the code works, but playing around with this version appears to fix the issue.

jourdain:
You could use dynamic arrays and just convert them at the end

jbeezley:
There could be a significant performance hit from not preallocating the arrays.  Also, since I don't completely understand the code, I would rather avoid making more changes than necessary.

jourdain:
You can also slightly over allocate and make a view at the end. (In case that help to handle the "unknown part")

jourdain:
@jbeezley Are you good with your changes? Should we merge it?

jbeezley:
Yes, I'm happy with it.

PR #1888, opened 2021-04-30
fix(WebGPU): fix glyph3dmapper add customize example

martinken:
Fix an issue with coloring of Glyphs

Add an example of customizing vtk.js and it's WebGPU
backend by creating a new mapper subclass.


martinken:
@jourdain can you do a sanity check on https://github.com/Kitware/vtk-js/pull/1888/files#diff-d2f0850d07639f616caa008e1041b463c815673178d813a39843cf29acb81a3b specifically where I am creating two new classes using a more compact approach. It works for me but I'm not sure if there is a gotcha with distributions or something. Trying to keep the example compact but yet still show how to do it.

jourdain:
Looks great! No issue on my end...

jourdain:
Do we want to add an API on ViewNodeFactory to better test backend?

```
vnfactory.getClassName().includes('WebGPU')
```

vs 

```
vnfactory.isBackend('webgpu`) // case insensitive?
```

martinken:
Yes, good idea, I'll add that to the factory.

jourdain:
Or even have the backend encoded in the classes so when you register them, they get only added when matching backend?

Actually, it seems that this could be independent of the backend. You just register them to the proper implementation internally and automatically.

jourdain:
Basically, once I implement the "Profiles" like I was talking about to enable better tree shaking, this could be improved.
But I guess I have a question. Do you see a usage where you want several view with the same backend not sharing the same class mapping? Or would it be fine if those were global by backend?

jourdain:
I'll see if I can tackle that part sometime this afternoon. Profile + better registration

martinken:
I was going to bring the profiles up. The more I think about it the more I believe we need it and need to get it right. One of VTK's issues right now is dealing with remote modules versus in tree code. I think for vtk.js we need a good first class approach to extending it without just adding more and more into the main tree.  etc. Ideally the core of vtk.js would become fairly static and limited in size and we would have packages that extend it etc.

martinken:
Going to merge as is and we can easily refine as we think about profiles/modules more etc.

PR #1889, opened 2021-04-30
Enable tree shaking on rendering

### LINKED ISSUE #1865: Tree-shaking improvement
We have an app that uses vtk.js to render meshes and the bundle is produced by webpack, the analyze shows that some pieces of vtk.js are included in the bundle even if they are not used or called in the app.

So basically `OpenGL/RenderWindow` calls `OpenGL/ViewNodeFactory` which instantiate classes even if they are not used the app, i can list :

* vtkGlyph3DMapper 
* vtkPixelSpaceCallbackMapper
* vtkSphereMapper
* vtkStickMapper
* vtkSurfaceLICMapper
* vtkVolumeMapper

is there something we can do in vtk.js to improve this part?

Also i can see the import of vtkMath which is 20kb 

```js
import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
```
instead of the individual functions

```js
import { XYZ } from 'vtk.js/Sources/Common/Core/Math';
```

https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/ViewNodeFactory/index.js#L48-L89

tree-shaking

jourdain:
This provide a better mechanism to pick and choose which part of the rendering you want to include.

fix #1865


jourdain:
Still WIP - Do not merge...

jourdain:
FYI @floryst @finetjul @daker

PR #1890, opened 2021-05-01
fix(generator): replace async by promise

### LINKED ISSUE #1882: regeneratorRuntime is not defined
In version VTK v17.15.0, running VTK gives an error about regeneratorRuntime is missing.
Eventually, it can be fixed by installing the module, but I was wondering whether this was meant to happen?

**Screenshots**
image

***vtk.js version:***
v17.15.0,

jourdain:
This should allow to prevent issue #1882

fix #1882

### DIFF
--- a/Sources/Rendering/Core/HardwareSelector/index.js
+++ b/Sources/Rendering/Core/HardwareSelector/index.js
@@ -14,21 +14,20 @@ function vtkHardwareSelector(publicAPI, model) {
   // must be called at least once before calling generateSelection. In
   // raster based backends this method will capture the buffers. You can
   // call this once and then make multiple calls to generateSelection.
-  publicAPI.getSourceDataAsync = async (renderer, fx1, fy1, fx2, fy2) => {};
+  publicAPI.getSourceDataAsync = (renderer, fx1, fy1, fx2, fy2) => {};
 
-  publicAPI.selectAsync = async (renderer, fx1, fy1, fx2, fy2) => {
-    const srcData = await publicAPI.getSourceDataAsync(
-      renderer,
-      fx1,
-      fy1,
-      fx2,
-      fy2
-    );
-    if (srcData) {
-      return srcData.generateSelection(fx1, fy1, fx2, fy2);
-    }
-    return [];
-  };
+  publicAPI.selectAsync = (renderer, fx1, fy1, fx2, fy2) =>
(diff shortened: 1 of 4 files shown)


jourdain:
@martinken I did not tried it but I wanted to see if we can handle it without the nicer async syntax which introduce some vtk.js consumption burden. 

@floryst feel free to comment on it...

martinken:
Is there a reason not to use aync? Just trying to understand the logic here. As of today it is supported on 95% of browsers and the 5% it isn't supported on tend to also not support WebGL. It is a 4 year old standard and most of what I have read says it is how asynchronos code should be done addressing issues in the promise API.  If not today when do you see us supporting it? I just hate converting 4 year old APIs to 6 year old APIs.

floryst:
Here are the stats on async/await:

- https://caniuse.com/mdn-javascript_operators_await
- https://caniuse.com/async-functions

It's reasonable to assume that we can safely use async/await in most browsers, save for IE 11.

I think the problem arises when someone consumes vtk.js. right now, it's up to the user to dictate how they process source files with async/await. Either they transform with babel, this requiring them to import regenerator-runtime in their app (this would have to be a docs update explicitly outlining this), or they don't.

The decision to fix this will depend on how we want to handle async/await. Do we shift it towards the user and make a doc note mentioning regenerator-runtime, or do we only use Promise-based code in vtk.js? I'm inclined to pick the former, since consuming vtk.js using the source files approach does necessitate some environment work on the user's part anyways.

If there is a different solution for this, I'm all ears.

jourdain:
My take was to see how much async we use and how "bad" it will look with promises.  In general, it is easy to create a method that return a single promise, but it is not as easy to cross sync them without async. When I did that pass, most of the code was trivial to convert and it could have been fine to use promise rather than async. But 1-2 methods was bad and async show its strength. 

Anyway, I was ok to stick to promise to ease the consumption of vtk.js assuming it was trivial. But with that conversion, it shows that it is not that obvious and I think we just need to document what vtk.js is using and which of the latest spec we are using so consumers knows which preset and part of the latest spec vtk.js is using.

So, I think we can close that PR but as a next step, we need to document the various advanced language elements that vtk.js use and explain the babel runtime pieces the user needs to import.

@floryst am I right to think that the ESM build don't have any of that problem?

PR #1891, opened 2021-05-04
feat(WebGPU): add actor matrix support

martinken:
Should now handle actor matrices. Also by default performs
shifting into to handle floating point resolution issues.

Moved a test from OpenGL to Core as it didn't need OpenGL


Sources/Rendering/Core/Coordinate/api.md
```diff
@@ -20,15 +20,47 @@ The coordinate systems in vtk are as follows:
 
   POSE - world coords translated and rotated to the camera position and view direction
 
+  STABILIZED - used by rendering backends to deal with floating point resolution issues. Similar to world coordinates but with a translation to try to get the
```
jourdain, line 23: api.md are now generated (which mean we should start removing them from the repo).

The documentation should be added in the [TypeScript file](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/Coordinate/index.d.ts#L180-L210) instead.

jourdain: @daker is that true that the api.md are not auto-generated? If so, should we start removing them to prevent any confusion?

daker: Documentation needs to go in .d.ts file if it's present otherwise it needs to go in api.md, i will remove the api.md from folders where we have the .d.ts, i'll make a PR to cleanup this and sorry about the confusion.

jourdain: np, I just wanted to make sure. Can you grab the doc that Ken added here and put it inside the ts?

daker: Sure, i'll fix that.

PR #1893, opened 2021-05-04
docs(ts): cleanup the api.md files

daker:
This commit cleanup the api.md that are generated and leaves the ones that don't have .d.ts yet


agirault:
Should we update this to reflect the current documenting strategy? Or is this done/explained somewhere else already? https://github.com/Kitware/vtk-js/blob/master/CONTRIBUTING.md#updating-documentation

jourdain:
Here might be better
https://kitware.github.io/vtk-js/docs/develop_class.html#API-documentation

agirault:
Then the CONTRIBUTING.md doc should link to that, so we don't duplicate it. I personally never look at https://kitware.github.io/vtk-js/docs/ since I assume it's all in Github. Are we spread out too much across the github.io doc and the .md docs? It seems some of it is here, some of it there, some of it in both.

jourdain:
For me `CONTRIBUTING.md` explain the workflow and tools. While `https://kitware.github.io/vtk-js/docs/develop_class.html` explain how to write code for vtk.js. And in general when you write code you provide the doc with it.

daker:
@agirault for now, if you have an index.d.ts inside the folder then you will not find the api.md since it's generated at build time using the generate-api-docs.js script, if we don't have the index.d.ts we have two options 1) create the index.d.ts and put the documen each method/func 2) continue updating the api.md until we complete the migration to index.d.ts to not break haxo build.

agirault:
> for now, if you have an index.d.ts inside the folder then you will not find the api.md since it's generated at build time using the generate-api-docs.js script

@daker yes, sorry for the confusion, I was not referring to the API docs, but the high-level doc we write, like those:
- overview
  - [in README.md](https://github.com/Kitware/vtk-js/blob/master/README.md#introduction)
  - [in github.io](https://kitware.github.io/vtk-js/docs/)
- requirements
  - [in README.md](https://github.com/Kitware/vtk-js/blob/master/README.md#requirements)
  - [in github.io](https://kitware.github.io/vtk-js/docs/develop_requirement.html)
- contributing
  - [in CONTRIBUTING.md](https://github.com/Kitware/vtk-js/blob/master/CONTRIBUTING.md)
  - [in github.io](https://kitware.github.io/vtk-js/docs/misc_contributing.html)
- testing
  - [in CONTRIBUTING.md](https://github.com/Kitware/vtk-js/blob/master/CONTRIBUTING.md#tests)
  - [in github.io](https://kitware.github.io/vtk-js/docs/develop_test.html)
- reporting bugs
  - [in README.md](https://github.com/Kitware/vtk-js#reporting-bugs)
  - [in CONTRIBUTING.md](https://github.com/Kitware/vtk-js/blob/master/CONTRIBUTING.md#reporting-issues)
  - [in github.io](https://kitware.github.io/vtk-js/docs/misc_contributing.html#Reporting-Issues)
- writing documentation  <- not accurate anymore!
  - [in CONTRIBUTING.md](https://github.com/Kitware/vtk-js/blob/master/CONTRIBUTING.md#updating-documentation)
  - [in github.io (1)](https://kitware.github.io/vtk-js/docs/misc_contributing.html#Updating-Documentation)
  - [in github.io (2)](https://kitware.github.io/vtk-js/docs/develop_class.html#API-documentation)

I was trying to find a good place to reference/document your work using `index.d.ts` instead of a manual `api.md`. I suggest we update those last 3 links I referenced just above, but we should probably think of a way to consolidate github.io and github repo docs, since some of them are updated, some aren't, so their just duplicates which can easily not match. I can open an issue for that.

PR #1895, opened 2021-05-06
fix(httpdatasetreader): when caching array data, differentiate by class

boucaud:
A clear and concise description of what the problem was and how this pull request solves it.

The reader caches array instances to avoid having to load and instanciate the same array multiple times. Arrays are
cached using a hash of the binary data they contain.

This means a vtkPoints and vtkDataArray sharing the same data could be cached under the same index.

This case can happen quite often with empty meshes, as all arrays will share the same hash.

This fix appends the class type to the caching key to avoid these conflicts.

### DIFF
--- a/Sources/IO/Core/HttpDataSetReader/index.js
+++ b/Sources/IO/Core/HttpDataSetReader/index.js
@@ -144,7 +144,7 @@ function vtkHttpDataSetReader(publicAPI, model) {
 
   // Internal method to fetch Array
   function fetchArray(array, options = {}) {
-    const arrayId = array.ref.id;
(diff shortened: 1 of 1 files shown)


boucaud:
@finetjul this fixes the issue we discussed with empty animations not loading. WDYT ?

jourdain:
The cache is returning vtkDataArray rather than TypedArray? Hence the issue your are seeing?

boucaud:
What happens is that when an empty vtkPoints is first loaded then cached. Then if there are empty vtkDataArrays to be created next, the reader will use the cached vtkPoints data to create them, which will result in an error.

PR #1898, opened 2021-05-07
doc(Examples): add missing profile import

jourdain:


jourdain:
FYI @floryst

jourdain:
That should fix the widget example when missing the sphere (Glyph) rendering.

PR #1899, opened 2021-05-07
fix(polyfill): fix regeneratorRuntime issue

### LINKED ISSUE #1882: regeneratorRuntime is not defined

jourdain:
Update build configuration to automatically add polyfill based on what it used inside vtk.js

fix #1882

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,8 @@
   "presets": [
     ["@babel/preset-env", {
       "debug": false,
-      "useBuiltIns": false
+      "useBuiltIns": "usage",
+      "corejs": "3.12.0"
     }]
   ],
   "env": {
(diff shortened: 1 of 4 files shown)


jourdain:
FYI @daker 

But we are still testing that to make sure that is the right solution with @floryst

floryst:
I think this still won't work. Trying to get babel to parse the `.babelrc` from `node_modules/vtk.js/.babelrc` isn't something that I think can work, and a brief test seems to indicate such. I think instead we need to update the docs to require users to use `useBuiltIns: "usage"`.

jourdain:
go for it...

PR #1900, opened 2021-05-07
feat(mouserangemanipulator): adds a scale parameter to mouse listener…

### LINKED ISSUE #1894: Add scale property on the `MouseRangeManipulator` to control its speed
### Problem

The scroll can be too slow with `MouseRangeManipulator` (ex: slicing)

### Suggested solution 

Add a `scale` property on the listener objects of the `MouseRangeManipulator` to be consumed in `processDelta` to control the rate of change. Keeping the `scale` optional/defaulting to `1` should make it backward compatible.

WillCMcC:
…s to scale the delta

Adds a scale parameter to mouseRangeManipulator with a default value of 1. In processDelta, this
scale is applied to the step value. This allows users to achieve variable scroll speeds at a single
step value.

re #1894

-A clear and concise description of what the problem was and how this pull request solves it.

fix #1894

This contribution is funded by [Sirona](https://sironamedical.com).

### DIFF
--- a/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
@@ -24,53 +24,97 @@ function vtkMouseRangeManipulator(publicAPI, model) {
   //-------------------------------------------------------------------------
   function processDelta(listener, delta) {
     const oldValue = listener.getValue();
-    let value = oldValue + delta + incrementalDelta.get(listener);
 
+    // Apply scale and cached delta to current delta
+    const newDelta = delta * listener.scale + incrementalDelta.get(listener);
+    let value = oldValue + newDelta;
+
+    // Compute new value based on step
     const difference = value - listener.min;
     const stepsToDifference = Math.round(difference / listener.step);
     value = listener.min + listener.step * stepsToDifference;
     value = Math.max(value, listener.min);
     value = Math.min(value, listener.max);
 
-    // Check if value will change or if we need to store
-    // the delta to append at the next iteration
     if (value !== oldValue) {
+      // Update value
       listener.setValue(value);
       incrementalDelta.set(listener, 0);
-    } else {
-      const v = incrementalDelta.get(listener);
-      incrementalDelta.set(listener, v + delta);
+    } else if (
+      (value === listener.min && newDelta < 0) ||
+      (value === listener.max && newDelta > 0)
+    ) {
       // Do not allow incremental delta to go past range
-      if (
-        (value === listener.min && incrementalDelta.get(listener) < 0) ||
-        (value === listener.max && incrementalDelta.get(listener) > 0)
-      ) {
-        incrementalDelta.set(listener, 0);
-      }
+      incrementalDelta.set(listener, 0);
+    } else {
+      // Store delta for the next iteration
+      incrementalDelta.set(listener, newDelta);
     }
   }
 
   // Public API methods
+
+  // min:number = minimum allowable value
+  // max:number = maximum allowable value
+  // step:number = value per step -- smaller = more steps over a given distance, larger = fewer steps over a given distance
+  // getValue:fn = function that returns current value
+  // setValue:fn = function to set value
+  // scale:number = scale value is applied to mouse event to allow users accelerate or decelerate delta without emitting more events
   //-------------------------------------------------------------------------
-  publicAPI.setHorizontalListener = (min, max, step, getValue, setValue) => {
+  publicAPI.setHorizontalListener = (
+    min,
+    max,
+    step,
+    getValue,
+    setValue,
+    scale = 1
+  ) => {
     const getFn = Number.isFinite(getValue) ? () => getValue : getValue;
-    model.horizontalListener = { min, max, step, getValue: getFn, setValue };
+    model.horizontalListener = {
+      min,
+      max,
+      step,
+      getValue: getFn,
+      setValue,
+      scale,
+    };
     incrementalDelta.set(model.horizontalListener, 0);
     publicAPI.modified();
   };
 
   //-------------------------------------------------------------------------
-  publicAPI.setVerticalListener = (min, max, step, getValue, setValue) => {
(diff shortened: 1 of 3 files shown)


WillCMcC:
@agirault took a first pass at this with an optional param, passed in last, defaulting to 1. Let me know what you think and what changes you'd like to see here!

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -52,25 +52,60 @@ function vtkMouseRangeManipulator(publicAPI, model) {
...
+  ) => {
     const getFn = Number.isFinite(getValue) ? () => getValue : getValue;
-    model.verticalListener = { min, max, step, getValue: getFn, setValue };
+    model.verticalListener = {
+      min,
+      max,
+      step,
+      getValue: getFn,
+      scale,
+      setValue,
```
agirault, line 92: put `scale` at the bottom for consistency?

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -28,7 +28,7 @@ function vtkMouseRangeManipulator(publicAPI, model) {
 
     const difference = value - listener.min;
     const stepsToDifference = Math.round(difference / listener.step);
-    value = listener.min + listener.step * stepsToDifference;
+    value = listener.min + (listener.step + listener.scale) * stepsToDifference;
```
agirault, line 31: I don't think this works:
- if you use an integer scale, then you will always skip some values, even if you try to go slowly.
  - Ex: you want to go from slice 1 to slice 2 faster, not go from slice 1 to slice 3 at the same speed as you went from slice 1 to slice 2 while skipping slice 2.
- if you use a floating-point scale, then your `value` doesn't follow the stepping mechanism.
  - Ex: you could be at slice 1.5

I think applying the scale on this line is similar to just changing the `step` value. Instead, I'd recommend trying updating the delta value directly above (input from the mouse event that we want to accelerate or decelerate):
```js
const oldValue = listener.getValue()
const newDelta = delta * listener.scale + incrementalDelta.get(listener)
let value = oldValue + newDelta
```

If this works, you'd want to make sure to set `incrementalDelta` to `newDelta` if you haven't changed value ([on line 42](https://github.com/Kitware/vtk-js/pull/1900/commits/89453ba478ea295ddbc29cc7a15918ce8c16f09c#diff-e5dc7da166e6ad75f440a37272ace4a141e12d8495d59ddb22cb6d174d5d3b99R42))

agirault (changes requested):
Hey @WillCMcC, great to see you here 👋 Thanks for the contribution!

So... which year did you come from?


<sup><sub>Hoping you've seen that show, or my joke will fall flat.</sub></sup>

I left a couple of comments below but haven't tested, so feel free to correct any of my assumptions below.

It would be ideal if you could update the example / API docs / tests... if those exist. If not, you're more than welcome to take a stab at adding them, but I understand if that's not high on your priority list. It would help to maintain the changes, if not we'll look into it eventually.

Thanks again!

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
WillCMcC: Got it, I'll take those suggestions and then validate it and report back! I truly did steal 100% of my swag from Jonas didn't I 😂 . What can I say... _Jonas, Ich bin du_

WillCMcC:
Ok I made the changes that you requested. I played with the example locally and it appears to be behaving as you expected. I'm validating with my eyes so if you can think of something to log out to validate I'll post screenshots. Tests pass! 

I added some comments to the public API methods that will hopefully help the next user. Im unsure if my language is precise enough so let me know if those are helpful or confusing. 

I've updated the example as well so that horizontalListener has scale of 2 to demonstrate how to use that parameter.

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -39,7 +40,14 @@ function vtkMouseRangeManipulator(publicAPI, model) {
...
       const v = incrementalDelta.get(listener);
-      incrementalDelta.set(listener, v + delta);
+
+      const updatedDelta = v + delta;
+
+      if (updatedDelta === oldValue && updatedDelta !== newDelta) {
+        incrementalDelta.set(listener, newDelta);
+      } else {
+        incrementalDelta.set(listener, updatedDelta);
+      }
```
agirault, line 50: Why not just `incrementalDelta.set(listener, newDelta)` ?

agirault: not tested:
```js
} else if ((value === listener.min && newDelta < 0) || (value === listener.max && newDelta > 0)) {
  // Do not allow incremental delta to go past range
  incrementalDelta.set(listener, 0);
} else {
  // Store delta for the next iteration
  incrementalDelta.set(listener, newDelta);
}
```

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -35,26 +35,17 @@ function vtkMouseRangeManipulator(publicAPI, model) {
 
     // Check if value will change or if we need to store
     // the delta to append at the next iteration
+
     if (value !== oldValue) {
+      // Update value
       listener.setValue(value);
+      incrementalDelta.set(listener, 0); // TODO: could still store partial delta that's ignored due to the rounding?
+    } else if (value === listener.max && newDelta > 0) {
```
agirault, line 43: Missing min & negative newDelta?

agirault (approved):
@WillCMcC: I took the liberty of fixing [that](https://github.com/Kitware/vtk-js/pull/1900#discussion_r631279107) + some improvements to the example + squashed, rebased, and forced push on the branch. I'll merge once the CI shows green. Thanks a lot for your work 👏

WillCMcC:
@agirault Thank you for humoring me as I put this together, and your guidance throughout 🙏

PR #1903, opened 2021-05-10
fix: add missing dataAccessHelper imports in examples

boucaud:
### Context

Individual DataAccessHelper files now need to be imported in order to be registered.

### Changes

Add imports for JSZipDataAccesHelper in the animation examples.
### Results

These examples work again

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


boucaud:
@jourdain WDYT ?

jourdain:
Just rename your commit to be `doc(Example): Fix missing import in examples` as there is no need to create a new version for that.

boucaud:
> Just rename your commit to be `doc(Example): Fix missing import in examples` as there is no need to create a new version for that.

Done

PR #1906, opened 2021-05-11
fix(RenderWindowInteractor): add/remove event listeners on same object

scottwittenburg:
### Context

Fix console errors coming from `vtkRenderWindowInteractor` after unbinding and rebinding events.

### Changes

#1858 Changed the `bindEvents` method in the RWI to add key event listeners on the `document` rather than the `body`.  This change fixes the `unbindEvents` methods to match, i.e. it uses the `document` rather than the `body` to `removeEventListener`.


scottwittenburg:
@jourdain Thanks.  I guess you can ignore my request for your review :wink: I must have requested it just as you were merging.

PR #1907, opened 2021-05-11
fix(RenderWindow): Support custom sizes for screenshots

scottwittenburg:
Currently the screenshot capture functionality captures the render window image as it appears on the screen.  However, some applications could benefit from the ability to specify an arbitrary size when capturing an image and have it rendered at full resolution rather than resampled.

I haven't tried this with WebGPU, but just provided an implementation the same as that for OpenGL.

### DIFF
--- a/Sources/Rendering/WebGPU/RenderWindow/index.js
+++ b/Sources/Rendering/WebGPU/RenderWindow/index.js
@@ -12,6 +12,13 @@ import vtkRenderWindowViewNode from 'vtk.js/Sources/Rendering/SceneGraph/RenderW
 
 const { vtkErrorMacro } = macro;
 // const IS_CHROME = navigator.userAgent.indexOf('Chrome') !== -1;
+const SCREENSHOT_PLACEHOLDER = {
+  position: 'absolute',
+  top: 0,
+  left: 0,
+  width: '100%',
+  height: '100%',
+};
 
 // ----------------------------------------------------------------------------
 // vtkWebGPURenderWindow methods
@@ -244,7 +251,10 @@ function vtkWebGPURenderWindow(publicAPI, model) {
     publicAPI.invokeImageReady(screenshot);
   }
 
-  publicAPI.captureNextImage = (format = 'image/png') => {
+  publicAPI.captureNextImage = (
+    format = 'image/png',
+    { resetCamera = false, size = null, scale = 1 } = {}
+  ) => {
     if (model.deleted) {
       return null;
     }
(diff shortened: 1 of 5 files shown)


Sources/Rendering/WebGPU/RenderWindow/index.js
```diff
@@ -244,19 +244,96 @@ function vtkWebGPURenderWindow(publicAPI, model) {
...
+            const pscale = camera.getParallelScale();
+            model.activeCamParamMap.set(ren, {
+              focalPoint: [fp[0], fp[1], fp[2]],
+              position: [pos[0], pos[1], pos[2]],
+              pScale: pscale,
+            });
+            // Without this resetCamera(), we can end up showing only a subset
+            // of the prop(s) depending on the difference between the aspect
+            // ratios of the original size and the custom capture size.
+            ren.resetCamera();
```
jourdain, line 332: This should be optional and off by default.

Sources/Rendering/WebGPU/RenderWindow/index.js
```diff
@@ -244,19 +244,96 @@ function vtkWebGPURenderWindow(publicAPI, model) {
...
+            model.resizedScreenshotReplacement.remove();
+            model.resizedScreenshotReplacement = null;
+            model.activeCamParamMap = null;
+            model.originalSize = null;
+          }
+          resolve(imageURL);
+        } else {
+          // Create a stand-in image overlay while we resize and render
+          const tmpImg = document.createElement('img');
+          tmpImg.width = model.size[0];
```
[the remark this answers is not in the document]
scottwittenburg: Ah right, I think probably not needed.  Will fix.

Sources/Rendering/WebGPU/RenderWindow/index.js
scottwittenburg: Yeah, I was wondering about that.  It's kind of best effort anyway.  To really allow you to get what you want in your screenshot, we'd probably have to call you back before we render and let you compute/set what you want.  

But you think if we just make it optional, the way it is now could be good enough?

Sources/Rendering/WebGPU/RenderWindow/index.js
```diff
@@ -244,19 +244,96 @@ function vtkWebGPURenderWindow(publicAPI, model) {
...
+          // resize it
+          model.size[0] = model.customCaptureSize[0];
+          model.size[1] = model.customCaptureSize[1];
+          model.customCaptureSize = null;
+
+          // process the resize
+          publicAPI.modified();
+          // save camera parameters before resetCamera so we can restore
+          // them later
+          model.activeCamParamMap = new WeakMap();
```
jourdain, line 318: I would not use a weak map but just rely on the order of the renderer because it really happen in a non-interactive setup.

You should use a private object on the model. 

```
model._screenshot = {
    originalSize: [],
    imagePaceHolder: ...,
    cameras: [],
};
```

Sources/Rendering/WebGPU/RenderWindow/index.js
```diff
@@ -244,19 +244,96 @@ function vtkWebGPURenderWindow(publicAPI, model) {
...
+            height: '100%',
+          };
+          tmpImg.src = imageURL;
+          model.resizedScreenshotReplacement = model.el.appendChild(tmpImg);
+
+          // save the dimensions of the main canvas
+          model.originalSize = [model.size[0], model.size[1]];
+
+          // hide it
+          model.canvas.style.display = 'none';
```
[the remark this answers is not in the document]
scottwittenburg: Reversing it is done in the `updateWindow` method as a result of calling `publicAPI.modified()`.

Sources/Rendering/WebGPU/RenderWindow/index.js
```diff
@@ -244,19 +244,96 @@ function vtkWebGPURenderWindow(publicAPI, model) {
...
+          model.size[0] = model.customCaptureSize[0];
+          model.size[1] = model.customCaptureSize[1];
+          model.customCaptureSize = null;
+
+          // process the resize
+          publicAPI.modified();
+          // save camera parameters before resetCamera so we can restore
+          // them later
+          model.activeCamParamMap = new WeakMap();
+          model.renderable.getRenderers().forEach((ren) => {
```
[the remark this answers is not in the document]
scottwittenburg: Nice!  Thx.

Sources/Rendering/WebGPU/RenderWindow/index.js
```diff
@@ -244,19 +244,96 @@ function vtkWebGPURenderWindow(publicAPI, model) {
...
+          model.resizedScreenshotReplacement = model.el.appendChild(tmpImg);
+
+          // save the dimensions of the main canvas
+          model.originalSize = [model.size[0], model.size[1]];
+
+          // hide it
+          model.canvas.style.display = 'none';
+
+          // resize it
+          model.size[0] = model.customCaptureSize[0];
```
jourdain, line 310:
```
model._screenshot.originalSize = model.size;
model.size = model._screenshot.size;

[...]

model.size = model._screenshot.originalSize;
```

Sources/Rendering/WebGPU/RenderWindow/index.js
```diff
@@ -414,6 +491,7 @@ const DEFAULT_VALUES = {
   containerSize: null,
   renderPasses: [],
   notifyStartCaptureImage: false,
+  resizedScreenshotReplacement: null,
```
jourdain, line 494: nothing should be added to the defaults.

Sources/Rendering/WebGPU/RenderWindow/index.js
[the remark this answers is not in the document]
jourdain: you may have to clone the arg depending if you get a ref or a copy of each array.

`clone = (obj) => JSON.parse(JSON.stringify(obj))`

Sources/Rendering/WebGPU/RenderWindow/index.js
```diff
@@ -414,6 +504,7 @@ const DEFAULT_VALUES = {
...
```
jourdain, line 507: Should that line be removed?

scottwittenburg: Weird, I thought I did.

scottwittenburg:
@jourdain Let me know if I should at least document the duplication between the `OpenGL` and `WebGPU` implementations or something.  Or if there's anything else you think this needs.  Otherwise maybe we can target merging it tomorrow?  Thanks!

scottwittenburg:
@jourdain Do you think this is still missing something?  If just rebase and force-push the branch and the checks are still green, could we merge it?

scottwittenburg:
Thanks @jourdain :+1:

Issue #1867, opened 2021-04-23
ES Module documentation


floryst:
Documentation is required for ES module usage of `@kitware/vtk.js` package.

PR #1911, opened 2021-05-14
docs: Overhaul intro docs to use the ESM build

floryst:
### Context

This PR updates the documentation to use the new `@kitware/vtk.js` ESM build. The old webpack-dependent approach is still supported, but its use in new projects is discouraged.

Fixes issue #1867 

### Changes

This only introduces documentation changes.

### DIFF
--- a/Documentation/content/docs/intro_vtk_as_es6_dependency.md
+++ b/Documentation/content/docs/intro_vtk_as_es6_dependency.md
@@ -1,243 +1,76 @@
 title: Using vtk.js as an ES6 dependency
 ---
 
-This guide illustrates how to build an application using vtk.js as a dependency using a modern toolsets such as Webpack, NPM.
+This guide illustrates how to consume vtk.js as an ES dependency.
 
-## Creation of the Project structure
+## Starting a new vtk.js project
+
+There are several ways to start a new vtk.js project. Below lists a few ways you can begin using vtk.js with your favorite project build tool.
+
+- [vtk.js from scratch (with webpack)](./vtk_vanilla.html)
+- [vtk.js with React](./vtk_react.html)
+- [vtk.js with vue](./vtk_vue.html)
+
+## Adding vtk.js to an existing project
+
+You can add vtk.js to an existing project as follows:
 
 ```sh
-$ mkdir MyWebProject
-$ cd MyWebProject
-$ npm init
+$ npm install --save @kitware/vtk.js
+```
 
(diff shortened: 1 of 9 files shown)


Documentation/content/docs/intro_vtk_as_es6_dependency.md
```diff
@@ -1,243 +1,76 @@
...
-```
+## Adding vtk.js to an existing project
 
-Then install and save your dependencies
+You can add vtk.js to an existing project as follows:
 
 ```sh
-$ npm install vtk.js --save
-$ npm install kw-web-suite --save-dev
+$ npm install --save @kitware/vtk.js
```
jourdain, line 19: Is the save automatic now?

Documentation/content/docs/vtk_react.md
```diff
@@ -0,0 +1,161 @@
...
+## Using vtk.js in your app
+
+To add a minimal vtk.js example to your app, replace `src/App.js` with the following contents.
+
+```js src/App.js
+import { useState, useRef, useEffect } from 'react';
+
+import '@kitware/vtk.js/Rendering/Profiles/Geometry';
+
+import vtkFullScreenRenderWindow from '@kitware/vtk.js/Rendering/Misc/FullScreenRenderWindow';
```
jourdain, line 33: This is good, but it might be interesting to share/include a React component that map to a vtkRenderWindow...
Maybe we can bundle those in the code without importing react or what not? (Same for vue?)

jourdain (approved):
LGTM we should merge it sooner rather than later

Documentation/content/docs/intro_vtk_as_es6_dependency.md
floryst: I think save is automatic, but I'll keep the flag here. I don't know which versions of npm have it.

Documentation/content/docs/vtk_react.md
floryst: Or we could just point people to check out `react-vtk-js`. But yeah, having extra docs for that could be a useful thing down the road. For this tutorial, I'm less concerned about it, but it is an option.

jourdain: agree, for the current step it is good.

jourdain: It is just that I don't want users use the fullscreenRW in production when their goal is to have the 3D not in fullscreen.

PR #1913, opened 2021-05-14
Unique class hierarchy

### LINKED ISSUE #1910: Multiple inheritance problem
### High-level description

Widgets do not list 'vtkProp' in their class hierarchy despite 'deriving' from it.
Here is the class hierarchy of LineWidget:
```js
['vtkObject', 'vtkInteractorObserver', 'vtkAbstractWidget', 'vtkLineWidgetProp']
```

### Detailed behavior

Here is the constructor of AbstractWidget
```js
vtkProp.extend(publicAPI, model, initialValues);
vtkInteractorObserver.extend(publicAPI, model, initialValues);
```
and here is vtkInteractorObserver constructor:
```js

  // Object methods
  macro.obj(publicAPI, model);
```
`macro.obj()` resets `model.classHierarchy`  that was previously initialized by `vtkProp.extend`

### Expected behavior

Not sure what to expect in a linear class hierarchy when doing multiple inheritance.

Should it be:
```
['vtkObject', 'vtkProp', 'vtkObject', 'vtkInteractorObserver', 'vtkAbstractWidget', 'vtkLineWidgetProp']
```
Or I guess we can be a bit smarter:
```js
['vtkObject', 'vtkProp', 'vtkInteractorObserver', 'vtkAbstractWidget', 'vtkLineWidgetProp']
```

Or maybe it should not be a flat list ? But that might break few things...

### Steps to reproduce

Instantiate a lineWidget and look at its hierarchy:
```js
const widget = vtkLineWidget.newInstance();
const instance = widgetManager.addWidget(widget, ViewTypes.SLICE);
console.log(instance.get(['classHierarchy']))
```

### Context ###
I found a `visibility` property in my line widget `model` and could not understand why because I only looked at the model class hierarchy.

floryst:
### Context

Fixes #1910 

### Changes

- Introduces `ClassHierarchy`, which subclasses Array. This is used to remove duplicates only when items are added via `.push`.
- Fixes unnecessary class hierarchy additions to AbstractWidget

### Results

model.classHierarchy should now not have any duplicates. Any existing code using classHierarchy should continue to operate correctly.

### Testing

I tested this by modifying the `PolyLineWidget` example to output the class hierarchy of the widget via `viewWidget.get('classHierarchy').classHierarchy`. The result should be the following:

```
ClassHierarchy(5) ["vtkObject", "vtkProp", "vtkInteractorObserver", "vtkAbstractWidget", "vtkPolyLineWidgetProp"]
```

### DIFF
--- a/Sources/Common/Core/ClassHierarchy/index.js
+++ b/Sources/Common/Core/ClassHierarchy/index.js
@@ -1,0 +1,7 @@
+export default class ClassHierarchy extends Array {
+  push(...args) {
+    // no perf issue since args.length should be small
+    const newArgs = args.filter((arg) => !this.includes(arg));
+    return super.push(...newArgs);
+  }
+}
--- a/Sources/Widgets/Core/AbstractWidgetFactory/index.js
+++ b/Sources/Widgets/Core/AbstractWidgetFactory/index.js
@@ -31,10 +31,15 @@ function vtkAbstractWidgetFactory(publicAPI, model) {
         camera,
       } = extractRenderingComponents(renderer);
       const widgetModel = {};
-      const widgetPublicAPI = {
+      const widgetPublicAPI = {};
+
+      macro.obj(widgetPublicAPI, widgetModel);
+      Object.assign(widgetPublicAPI, {
         onWidgetChange: publicAPI.onWidgetChange,
-      };
-      Object.assign(widgetModel, model, {
+      });
+      Object.assign(widgetModel, {
+        widgetState: model.widgetState,
+        manipulator: model.manipulator,
         viewType,
         renderer,
         camera,
--- a/Sources/macro.js
+++ b/Sources/macro.js
@@ -1,4 +1,5 @@
 import vtk from './vtk';
+import ClassHierarchy from './Common/Core/ClassHierarchy';
 
 let globalMTime = 0;
 
@@ -214,7 +215,12 @@ export function obj(publicAPI = {}, model = {}) {
   if (!Number.isInteger(model.mtime)) {
     model.mtime = ++globalMTime;
   }
-  model.classHierarchy = ['vtkObject'];
+
+  if (!('classHierarchy' in model)) {
+    model.classHierarchy = new ClassHierarchy('vtkObject');
+  } else if (!(model.classHierarchy instanceof ClassHierarchy)) {
+    model.classHierarchy = ClassHierarchy.from(model.classHierarchy);
+  }
 
   function off(index) {
     callbacks[index] = null;


jourdain (approved):
LGTM

finetjul:
It fixes the issue indeed. Thanks !

Few comments though:
* a unit test would be welcomed :-)
* It's too bad that `safeArrays` duplicates arrays for some keys.
* If some `publicApi` functions were overriden, they might be reset when calling `macro.obj` again. :-/
* `depth` in `getClassName` does not really make sense anymore.

finetjul (commented):
It fixes the issue indeed. Thanks !

Few comments though:
* a unit test would be welcomed :-)
* It's too bad that `safeArrays` duplicates arrays for some keys.
* If some `publicApi` functions were overriden, they might be reset when calling `macro.obj` again. :-/
* `depth` in `getClassName` does not really make sense anymore.

finetjul (commented):
It fixes the issue indeed. Thanks !

Few comments though:
* a unit test would be welcomed :-)
* It's too bad that `safeArrays` duplicates arrays for some keys.
* If some `publicApi` functions were overriden, they might be reset when calling `macro.obj` again. :-/
* `depth` in `getClassName` does not really make sense anymore.

floryst:
- +1 I'll make a unit test for ClassHierarchy

I'm not understanding what you mean by `safeArrays` duplicating arrays for some keys. The point of `safeArrays` is to make a clone of all arrays. This is partially due to the way default values are populated, and partially to clone array-based initial values that are passed in.

The `publicAPI` override issue is a valid concern. The issue lies in how we evaluate the parent classes.

In the following class hierarchy, A and B likely use `macro.obj`. Given that we call `extend` on parent classes before calling our own class init, execution happens in depth-first ordering. That is, we call `extend` in roughly the following order: A, C, B, D, E (assuming E calls C.extend before D.extend).

```
A B
| |
C D
|/
E
```

This means we call `macro.obj()` in B *after* potentially making overrides in C. Note this applies to *all* macro functions that add to the public API, which means we need to avoid overwriting functions that already exist on `publicAPI`. I don't know what else might be impacted by this behavior.

One solution is to rewrite all instances of `publicAPI.method = () => ...` to be `setMethod(publicAPI, 'method', () => ...`, where `setMethod` will only set `publicAPI['method']` *if* there is no method there already.

finetjul:
In your example, for an array variable `V` passed as initial-value, it will be cloned twice: once  when `A` calls `macro.obj`, and once when `B` calls `macro.obj`.

What you suggest `setMethod` would work but as is it prevents override (and sometimes we do want to override in subclasses).

I wonder if instead, those `extend` functions (including `macro.ob`) should be no-op as a whole if we try to declare a class that already exists in `model.classHierarchy`. Maybe we could call `model.classHierarchy.push()` earlier than now:
```
extend(publicAPI, model, initialValues) {
if (macro.pushClassName(model, 'vtkMyClass'))
  return;
)
Object.assign(...);
subclass.extend(...);
...
vtkMyClass(...);
}
```
```
macro.pushClassName(model, className) {
if model.classHierarchy.includes('className') {
  return true;
}
model.classHierarchy.push(className);
return false;
}
```

floryst:
What we can do for `macro.obj` is to see if `classHierarchy` exists, and if so then don't set any other function (e.g. `isA`, `getClassName`, etc.) That should also reduce `safeArray` calls to just 1.

The idea is that `setMethod` should be only used in `macro.js`, meaning user-defined classes can still override. The only overrides it protects against are publicAPI methods defined in `macro.js` helpers.

I think your idea is a good approach for classes that end up multiple times in an inheritance hierarchy. For macros, we will still need to have some means of checking if it should be applied (e.g. `macro.obj`), since we don't want to overwrite everything. (Example: `macro.obj` could be called twice as in my example, so we don't want the second `macro.obj` to invoke.)

finetjul:
> That should also reduce safeArray calls to just 1.

If `safeArray()` is called only once, we need 2 things:
 - have all DEFAULT_VALUES create new arrays (by using the previously mentionned [function to set default values](https://github.com/Kitware/vtk-js/issues/1751#issuecomment-785411007))
 - and do not use `initial_values` after `safeArrays` is called, here is the reason (reusing your ABCDE example):
```
E::extend(publicAPI, model, initial_values) {
  Object.assign(model, E_DEFAULT_VALUES, initial_values);
  C.extend(publicAPI, model, initial_values); // macro.obj will call `safeArrays` on all E/C/A_DEFAULT_VALUES and `initial_values` values
  D.extend(publicAPI, model, initial_values); // macro.obj will not call safeArrays
}
...
D::extend(publicAPI, model, initial_values) {
  Object.assign(model, D_DEFAULT_VALUES, initial_values); // Here, `initial_values` resets the copies made by the first call to `safeArrays`
  B.extend(publicAPI, model, initial_values); // ... macro.obj will not call safeArrays
}
```
So we should do instead:
```
E::extend(publicAPI, model, initial_values) {
  Object.assign(model, E_DEFAULT_VALUES, initial_values);
  C.extend(publicAPI, model, initial_values); // macro.obj will call `safeArrays` on all E/C/A_DEFAULT_VALUES and `initial_values` values
  D.extend(publicAPI, model, {}); // macro.obj will not call safeArrays
}
```
So that would be something like :
```
E::extend(publicAPI, model, initial_values) {
  C.extend(publicAPI, model, defaultValues(initial_values)); // macro.obj will call `safeArrays` on all E/C/A_DEFAULT_VALUES and `initial_values` values
  D.extend(publicAPI, model); // macro.obj will not call safeArrays
}
```

> The idea is that setMethod should be only used in macro.js
What other function in `macro.js` do you have in mind (other than `macro.obj` because for `macro.obj` we can check `classHierarchy` ) ?

> I think your idea is a good approach for classes that end up multiple times in an inheritance hierarchy. For macros, we will still need to have some means of checking if it should be applied
My idea would also apply to `macro.js`:
```
function obj(publicAPI, model) {
if (macro.pushClassName(model, 'vtkObject'))
  return;
)
safeArrays(model);
...
```

floryst:
> have all DEFAULT_VALUES create new arrays (by using the previously mentionned function to set default values)

We either need this, or use some deep object merge approach. In both cases, there will be a lot of small updates to every class.

> What other function in macro.js do you have in mind (other than macro.obj because for macro.obj we can check classHierarchy ) ?

There are several macro.js mixins where it would be nice to know if it's been applied already. For example, take `macro.proxy()`. If classes A and B uses `macro.proxy()`, and class C overrides a proxy method like `publicAPI.setKey()`, then when B's constructor is called the `publicAPI.setKey()` method gets reset.

With regards to what this PR addresses, I will merge for now. The rest we can address in separate PRs, as those changes will likely be larger in scope. Also,

PR #1914, opened 2021-05-16
docs(ts): use relative import of typescript definitions

daker:
### Context
The use of absolute import of TS defs only works with the umd and not esm build, also the use "vtk.js/Sources/..." makes writing .d.ts harder since the typescript transpiler can't resolve that path.

### Changes
This commit is changing the import method to make both the umd & esm build work using the same definitions.

### Results
image

_All my contributions are funded by my employer._


floryst:
Looks good! I'll briefly test this out as well and then we will be good to merge.

Sources/Common/Core/CellArray/index.d.ts
```diff
@@ -1,5 +1,5 @@
-/// <reference path='../../../types.ts' />
-import vtkDataArray from "vtk.js/Sources/Common/Core/DataArray";
+import { TypedArray } from "../../../types";
+import vtkDataArray from "../../Core/DataArray";
```
floryst, line 2: I think all of your relative paths in this commit need to remove a single ".." traversal. So, this should be:

```
import { TypedArray } from "../../types";
import vtkDataArray from "../Core/DataArray";
```

Prior to the change: 

image

After removing a ".." traversal:

image

floryst (changes requested):
The single comment might apply to all changes.

Sources/Common/Core/CellArray/index.d.ts
daker: The UMD build does have folders structures any idea why it's not the case for the ESM build ?

image

floryst: Ah right, that will explain the inconsistency.

The ESM build pulled the JS files out of the folder due to import issues with consuming projects that use Rollup. Specifically, rollup wouldn't resolve stuff like `import vtkDataArray from '@kitware/vtk.js/Common/Core/DataArray'` into using `Common/Core/DataArray/index.js`. I'd have to double check to see if this is now doable in rollup, since having consistent paths for the d.ts files would be nice. The fallback would be to rewrite paths in the d.ts files when generating the ESM build.

floryst: For now, I will make changes to the ESM package build script so it can rewrite the d.ts files to reference the appropriate path. @daker, if you want to take a shot at that, you can check out the npm script `release:create-packages`. Otherwise, I can merge this now and make the build script changes later.

daker: Yes just merge for now so i can continue adding more, i will also check the npm script

daker:
@floryst you can merge it for now, i have made change to the script to rewrite the imports, once i clean the code i'll create a PR

PR #1925, opened 2021-05-20
 fix(setAverageIPScalarRange)

### LINKED ISSUE #1924: AverageIPScalar not working
### High-level description
AverageIPScalar not behaving as expected.

I've been trying to use the AverageIPScalar so limit the values rendered with and average blend mode. I haven't had any luck getting this running. 

### Steps to reproduce
In the VolumeMapperBlendModes example add:

`mapper.setAverageIPScalarRange(0.0, 0.9);`

The volume is now not visible when selecting average blend mode.

### Expected behavior
The volume would average all the values between 0.0 and 0.9

GeorgeFuller:
### Context
fix #1924
https://discourse.vtk.org/t/averageipscalarrange/5544/4

### Changes
averageIPScalarRange working for single component volumes (from some angles).
averageIPScalarRange can now accept float values.

### Results
Previously setting the range to a float: 
`mapper.setAverageIPScalarRange(0.0, 0.9);`
would result in a blank canvas on a volume and shader errors.

Currently: Adjusting the max slider between 0.0 and 1.0 results in this: 
averageIPScalarTest

### Testing
run 
`npm run example -- VolumeMapperBlendModes`
switch the drop down to Average Intensity.

### DIFF
--- a/Sources/Rendering/OpenGL/VolumeMapper/index.js
+++ b/Sources/Rendering/OpenGL/VolumeMapper/index.js
@@ -232,18 +232,31 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
     ).result;
 
     const averageIPScalarRange = model.renderable.getAverageIPScalarRange();
+    let min;
+    let max;
 
-    // TODO: Adding the .0 at the end feels hacky
+    // If min or max is not already a float.
+    // make them into floats for glsl
+    if (Number.isInteger(averageIPScalarRange[0])) {
+      min = `${averageIPScalarRange[0]}.0`;
+    } else {
+      min = `${averageIPScalarRange[0]}`;
+    }
+    if (Number.isInteger(averageIPScalarRange[1])) {
+      max = `${averageIPScalarRange[1]}.0`;
+    } else {
+      max = `${averageIPScalarRange[1]}`;
+    }
     FSSource = vtkShaderProgram.substitute(
       FSSource,
       '//VTK::AverageIPScalarRangeMin',
-      `${averageIPScalarRange[0]}.0`
+      min
     ).result;
 
     FSSource = vtkShaderProgram.substitute(
       FSSource,
       '//VTK::AverageIPScalarRangeMax',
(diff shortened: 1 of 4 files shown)


Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -232,18 +232,31 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
+    if (Number.isInteger(averageIPScalarRange[0])) {
+      min = `${averageIPScalarRange[0]}.0`;
+    } else {
+      min = `${averageIPScalarRange[0]}`;
+    }
+    if (Number.isInteger(averageIPScalarRange[1])) {
+      max = `${averageIPScalarRange[1]}.0`;
+    } else {
+      max = `${averageIPScalarRange[1]}`;
+    }
```
daker, line 249: Why not use [toFixed](https://caniuse.com/?search=toFixed)() instead?

something like this 

```js
const a = averageIPScalarRange[0];
const b = averageIPScalarRange[1];

min = (Number.isInteger(a)) ? a.toFixed(1).toString() : (a.toString());
max = (Number.isInteger(b)) ? b.toFixed(1).toString() : (b.toString());
```
ex : 
```js
let i = 1
(Number.isInteger(i)) ? i.toFixed(1).toString() : (i.toString());
"1.0"
i = 1.25
(Number.isInteger(i)) ? i.toFixed(1).toString() : (i.toString());
"1.25"
```

GeorgeFuller: Oh I like that, looks clean and works well! I changed it to this. Never used toFixed before

GeorgeFuller:
I added a couple of changes to the volume shader which now has this behaving as I'd expect it to.

martinken (approved):
LGTM

martinken:
Any reason not to merge this?

martinken:
But other than that any reason not to merge?

jourdain:
I don't think so..

PR #1928, opened 2021-05-21
fix(surfacelicmapper): backport LIC fixes from VTK

boucaud:
See https://gitlab.kitware.com/vtk/vtk/-/merge_requests/7535 and
https://gitlab.kitware.com/vtk/vtk/-/merge_requests/7793

### Context

Some recent fixes have been added to VTK regarding the surface LIC mapper, specifically issues with face culling and normals not being properly loaded sometimes.

I reproduced related issues with the vtk.js mapper, which this PR fixes.

### Changes

Light complexity should always be > 0, as we always need normals when rendering LIC.

Handle property settings for culling when needed. Disable culling when rendering to quads, then restore to initial state.

### Results

The mapper will now work when polyDatas have their own normals, instead of showing an initial black render.

Face culling will be supported by the mapper.

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


boucaud:
@finetjul

boucaud:
@finetjul

martinken:
LGTM +1

PR #1929, opened 2021-05-21
docs(ImageCroppingWidget): Use clipping planes

floryst:
### Changes

Update ImageCroppingWidget to use the volume clip planes.

### Results

Cropping still works, but we no longer use the ImageCropFilter for this example.

### Testing

Nothing here for now.

### DIFF
--- a/Sources/Widgets/Widgets3D/ImageCroppingWidget/example/index.js
+++ b/Sources/Widgets/Widgets3D/ImageCroppingWidget/example/index.js
@@ -1,3 +1,5 @@
+import { vec3, quat, mat4 } from 'gl-matrix';
+
 import 'vtk.js/Sources/favicon';
 
 // Load the rendering pieces we want to use (for both WebGL and WebGPU)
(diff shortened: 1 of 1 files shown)


Sources/Widgets/Widgets3D/ImageCroppingWidget/example/index.js
```diff
@@ -125,25 +127,53 @@ actor.getProperty().setDiffuse(0.7);
...
-    window.asdf = widget;
     const cropState = widget.getWidgetState().getCroppingPlanes();
     cropState.onModified(() => {
-      cropFilter.setCroppingPlanes(cropState.getPlanes());
+      const planes = getCroppingPlanes(image, cropState.getPlanes());
+      mapper.removeAllClippingPlanes();
+      planes.forEach((plane) => {
+        mapper.addClippingPlane(plane);
+      });
+      mapper.modified();
```
floryst, line 176: `addClippingPlane` does not call `publicAPI.modified()`. @jourdain would it be worth it to update `addClippingPlane` to call modified? The only perf case I can think of: not calling `modified` on every add means we don't trigger a render when adding multiple planes.

jourdain: We could have a debounced modified call too?

jourdain: maybe with an optional argument (batch=true)

thewtex:
nice!

Sources/Widgets/Widgets3D/ImageCroppingWidget/example/index.js
floryst: For now I will not bother with making changes to `addClippingPlane`. However, I will make an issue to reference this potential change.

PR #1931, opened 2021-05-24
fix(ImageMapper): Apply ClippingPlanes

thewtex:


Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -509,6 +559,35 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
+        macro.vtkErrorMacro('OpenGL has a limit of 6 clipping planes');
+        numClipPlanes = 6;
+      }
+      const planeEquations = [];
+      for (let i = 0; i < numClipPlanes; i++) {
+        const planeEquation = [];
+        console.log('imagemat', model.imagemat);
+        console.log('getmatrix', actor.getMatrix());
+        model.renderable.getClippingPlaneInDataCoords(
+          // actor.getMatrix(),
```
thewtex, line 576: @martinken do you know what is the correct matrix to use here?

martinken: From my read it should be actor.getMatrix(). The code is not documented well (probably by me) but it is a copy of C++. The shader code is using the vertexMC which is in model coordinates. The clipPlanes are in world coordinates so you need a matrix to convert the clip planes from world to the data's (model) coordinate system. That matrix would be WorldToModel which is the inverse of ModelToWorld. But the clip planes are basically normals, and when transforming normals we use the inverse transpose. So we can do some short cuts and it basically it becomes ModelToWorld which is the actor's matrix (maybe the multiplication code is doing a transpose as well.

thewtex: @martinken thanks for the information! :book: :books: 

That direction worked -- thanks for the explanation.

It turns out that we also have to tack on the world to index matrix from the image. Inverse of:

https://github.com/Kitware/vtk-js/blob/9962dd8926d90ff5b0de2dff25499725dd670e56/Sources/Rendering/OpenGL/ImageMapper/index.js#L518-L520

I validated in a test with a polyData mapper:

testImageCroppingPlanes

martinken:
LGTM +1

PR #1937, opened 2021-05-26
Fix ResliceCursorWidget example

laurennlam:
Add missing import following updates in the HttpDataSetReader
@finetjul


jourdain:
FYI, this was fixing an example. Examples are not part of releases, therefore it is not necessary to create a new release with a `fix()` since nothing has changed "in the release". It would have been better to do a `doc(Example): Fix ResliceCursorWidget example`.

PR #1943, opened 2021-05-28
fix(package): rewrite TS import path

daker:
### Context
The UMD build does have folders structure, while ESM build does not, this cause the relative import to not work correctly with the ESM build. The import path should relative in dev (to let TS transpiler work) and absolute when packaged.

https://github.com/Kitware/vtk-js/pull/1914#discussion_r639090476
https://github.com/Kitware/vtk-js/pull/1914#discussion_r639778439

### Changes
Change has been made the build script to rewrite the import path to '@kitware/vtk.js/...' for the ESM build and 'vtk.js/Sources/...' for the UMD build.

### Results
TS intellisense and typing should work for ESM and UMD builds.

### Testing
Check the import paths for each build

**ESM**
image

**UMD**
image

_This contribution is funded by my employer._

### DIFF
--- a/Utilities/ci/build-npm-package.js
+++ b/Utilities/ci/build-npm-package.js
@@ -2,6 +2,7 @@ const fs = require('fs-extra');
 const path = require('path');
 const glob = require('glob');
 
+const importRegex = /(?:import|from) ['"]([^'"]*)['"]/g;
 const rootdir = 'pkg';
 
 // assumes src and dst are directories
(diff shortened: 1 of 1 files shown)


daker:
@floryst can i have a first review?

jourdain:
@floryst can you review and merge if it is ok?

floryst:
I'm not quite a fan of the excluded modules approach, since that requires updating the build script whenever some other module is suddenly used. Another approach would be to see if an import started with a dot (../, ./) and only rewrite those paths. How does that sound?

An alternative would be to change the imports back to absolute imports, and rewrite vtk.js/sources to @kitware/vtk.js.

FYI I have limited internet connectivity right now, so my response time has been limited. @jourdain if neither of the options above seem promising, then you can merge.

jourdain:
Sorry @floryst, enjoy your vacations!

daker:
@jourdain i did take the first option to let TS work correctly for errors checking and intellisense when writing the defs.

PR #1946, opened 2021-05-31
fix(reslicecursor): keep correct viewUp on rotation

laurennlam:
### Context
Setup correct viewup on old ResliceCursor widget in order to be consistent with the new one

@finetjul


finetjul:
Does it behave the same as in the new reslice cursor widget ?

laurennlam:
Yes. Here is a capture of the old updated widget:
image


Here is the new one:
image

Co concerning the view up, it's ok. There are still one difference, but t concerns the camera position, not the view up as you can see in the view with green/red lines

finetjul:
Understood.

What's the issue with the camera position ?

laurennlam:
It seems to be a sign issue but I didn't investigate it.

PR #1947, opened 2021-05-31
fix(ResliceCursor): Setup correct initial normal to axis view

laurennlam:
@finetjul , I finnaly found out the last issue about camera position on the old reslice cursor widget.


finetjul:
You rock !

laurennlam:
Now, here what we have:
Corrected old reslice cursor widget :
image

New reslice cursor widget:
image

PR #1948, opened 2021-06-02
doc(box): fix the Box widget example

boucaud:
### Context

This example still used getOpenGLRenderWindow

### Changes

Use getApiSpecificRenderWindow instead

### Results

The example works again
### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


finetjul:
If you fix an example, it should not be a `fix` but a `doc` (because it should not bump the release version)

PR #1951, opened 2021-06-04
Addtive Blending Mode

GeorgeFuller:
### Context
We had a need for an additive blend mode. Here's a simple implementation using the existing average blend mode.

### Changes
Added in an additive blend mode
Renamed AverageIPScalar to ipScalar, since it is now used for average and additive.
Added in a depreciation warning for AverageIPScalar and a handler to pass it through to ipScalar

### Results
additive

### Testing
`npm run example -- VolumeMapperBlendModes`
Then select additive from the dropdown.


jourdain:
@martinken I let you comment and merge if the changes make sense.

martinken:
LGTM +1

martinken (approved):

PR #1952, opened 2021-06-07
test(testtubefilter): fix rng consistency

boucaud:
Seeding the rng in global scope meant that any call to Math.random() in previous tests would alter the random sequence produced in testTubeFilter, thus altering the results.

This MR makes sure that Math.random is properly seeded right before the
test is executed.

### Context

The addition of testPaintEllipse caused this test to fail (because testPaintEllipse creates a worker, which causes a call to Math.random).

### Changes

Make sure that vtkMath.random is initialized at the right time in testTubeFilter.

### Results

testTubeFilter passes again.
### Testing

  - **vtk.js**: master
  - **OS**:  linux
  - **Browser**: chromium 91.0.4472.77

### DIFF
--- a/Sources/Filters/General/TubeFilter/test/testTubeFilter.js
+++ b/Sources/Filters/General/TubeFilter/test/testTubeFilter.js
@@ -29,10 +29,13 @@ test('Test vtkTubeFilter instance', (t) => {
   t.end();
 });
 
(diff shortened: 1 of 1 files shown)


boucaud:
@finetjul @jourdain WDYT ?

Issue #1935, opened 2021-05-25
View becomes completely black when attempting to pan in InteractorStyleUnicam example


tamaynard:
### High-level description

When attempting to run the InteractorStyleUnicam demo (https://kitware.github.io/vtk-js/examples/InteractorStyleUnicam.html), if you attempt to pan by clicking on the cone rather than the background, the view becomes completely blank, including the orientation marker, if enabled.

### Steps to reproduce

1. Go to https://kitware.github.io/vtk-js/examples/InteractorStyleUnicam.html
2. Click within the boundaries of the displayed cone object
3. Without releasing the mouse button, move the mouse

### Detailed behavior

Once the view has been blanked it is impossible to recover without a page refresh. Also there doesn't appear to be any relevant info/messages in the browser's debug console that would indicate a possible source of this issue.

### Expected behavior
Clicking on an object in the scene to pan should have the same effect as clicking on the background

finetjul:
@laurentmalka an idea of what's going on ?

daker:
I think this was fixed, not sure why it wasn't closed by the bot @jourdain

martinken:
Hmm yes, it should be fixed, and thought the bot would have closed it. Closing it now.

PR #1953, opened 2021-06-07
fix(Interaction): fix the unicam interactor

martinken:
fix issue when using the unicam style and picking on top of an object.

Also clean up some small code typos that had no impact.

Fixes issue https://github.com/Kitware/vtk-js/issues/1935


jourdain:
+1

PR #1957, opened 2021-06-16
fix(vtkTriangleFilter): ignore duplicated last points before triangulating cells

boucaud:
### Context

The SplineWidget example is currently broken (surface area missing a triangle).

This is due to a change that caused the vtkTriangleFilter to stop supporting cells that duplicate the last point when triangulating.

### Changes

If a cell is found to have its last point duplicated, ignore this last point when triangulating the cell.

### Results

The SplineWidget example is fixed


boucaud:
@finetjul

PR #1958, opened 2021-06-17
feat(Interaction): Add world up vector option to trackball rotate man…

tamaynard:
…ipulator

### Context
This brings over the world up vector feature of the unicam interactor into the manipulator interactor. It works in exactly the same way in that it forces the camera to rotate about the specified vector for the azimuth component of the total rotation, The API is the same as well, except it has to be set on the rotate manipulator instead.

### Testing
Call setUseWorldUpVec on the rotate manipulator with a parameter of true to enable the feature. By default the world up vector is [0, 1, 0], which can be changed by calling setWorldUpVec.


jourdain:
Is that ready or are you still working on it?

tamaynard:
Functionally, it's good to go. It passed tests and behaves as expected. Only thing I could think of would be style issues. If there are let me know and I can fix them.

PR #1959, opened 2021-06-20
fix(DracoReader): Compatible with Draco 1.4.0

yarous224:
### Context

Since Draco 1.4.0, Emscripten has been updated to V2, this causes the Draco codec modules to return a promise instead of the module directly.
[Draco Release https://github.com/google/draco/releases/tag/1.4.0](https://github.com/google/draco/releases/tag/1.4.0)

### Changes

- [ ] Use `Promise.resolve` to be compatible with both new and old versions.


jourdain:
Can you run `npm run reformat`

yarous224:
I reformated and force-pushed the code.

PR #1980, opened 2021-06-22
Add option to use camera focal point as center of

tamaynard:
…rotation to trackball rotate manipulator

### Testing
Call setUseFocalPointAsCOR on the rotate manipulator with a parameter of true to enable the feature.


jourdain:
Another approach is in your app to update the center of rotation when the focal-point of the camera change...

Also I'm not keen with the name `useFocalPointAsCOR` maybe `useFocalPointAsCenterOfRotation` it is (too) long but at least it is self explanatory. ;-)

tamaynard:
You're correct @jourdain, this was just more of a convenience option so that if you want to rotate about whatever you're looking at in the scene you could just set this option and have the desired behaviour automatically applied to the camera. Of course for finer-grained controls your approach would make more sense.

PR #1984, opened 2021-07-04
fix(macro): global object to globalThis

thewtex:
`global` is not necessarily defined in browsers:

  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

Update in macro.js.

### DIFF
--- a/Sources/macro.js
+++ b/Sources/macro.js
@@ -1,4 +1,4 @@
-import vtk from './vtk';
+import vtk, { vtkGlobal } from './vtk';
 import ClassHierarchy from './Common/Core/ClassHierarchy';
 
 let globalMTime = 0;
@@ -33,14 +33,14 @@ consoleMethods.forEach((methodName) => {
   fakeConsole[methodName] = noOp;
 });
 
-global.console = console.hasOwnProperty('log') ? console : fakeConsole;
+vtkGlobal.console = console.hasOwnProperty('log') ? console : fakeConsole;
 
 const loggerFunctions = {
   debug: noOp, // Don't print debug by default
-  error: global.console.error || noOp,
-  info: global.console.info || noOp,
-  log: global.console.log || noOp,
-  warn: global.console.warn || noOp,
+  error: vtkGlobal.console.error || noOp,
+  info: vtkGlobal.console.info || noOp,
+  log: vtkGlobal.console.log || noOp,
+  warn: vtkGlobal.console.warn || noOp,
 };
 
 export function setLoggerFunction(name, fn) {
--- a/Sources/vtk.js
+++ b/Sources/vtk.js
@@ -1,3 +1,7 @@
+import globalThisShim from 'globalthis';
+
+export const vtkGlobal = globalThisShim(); // returns native globalThis if compliant
+
 const factoryMapping = {
   vtkObject: () => null,
 };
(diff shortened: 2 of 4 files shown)


thewtex:
@floryst @jourdain I encountered the `global not defined` issue in the more rare context of ES modules and iframes.

finetjul:
Can't we think of a better variable name for `shimmedGlobalThis` ? `vtkGlobal` or maybe `macro.global` (eventually we will also have macro.window ?)

package-lock.json
```diff
@@ -7524,7 +7524,6 @@
       "version": "1.1.3",
       "resolved": "https://registry.npmjs.org/define-properties/-/define-properties-1.1.3.tgz",
       "integrity": "sha512-3MqfYKj2lLzdMSf8ZIZE/V+Zuy+BgD6f164e8K2w7dgnpKArBDerGYpM46IYYcjnkdPNMjPk9A6VFB8+3SKlXQ==",
-      "dev": true,
```
finetjul, line 7527: This is a genuine question: why get rid of "dev" ? It shouldn't have been here in the first place ?

thewtex:
> Can't we think of a better variable name for shimmedGlobalThis ? vtkGlobal or maybe macro.global (eventually we will also have macro.window ?)

Sure, we can use `vtkGlobal`. Note that this variable is not exported, though.

package-lock.json
thewtex: This file is auto-generated by `npm` and updates with their policies.

finetjul:
It looks good to me but I don't know much about that "global" issue.

Yet, I wonder if we shouldn't wrap `globalThisShim()` into a `macro.js` function to make the dependency less sensitive to future changes.

thewtex:
> Yet, I wonder if we shouldn't wrap globalThisShim() into a macro.js function to make the dependency less sensitive to future changes.

Yes, if we find that we need it in more contexts, we could make it available in `macro.js`. The examples use `global`, but this shim is not required there. In general, we want to avoid using `global` unless necessary as a best practice. So, we can avoid making it easy to do the wrong thing. At this point, it may be a cart before the hoarse. :shopping_cart: :hoarse:

agirault:
@jourdain @floryst ?

Sources/macro.js
```diff
@@ -1,6 +1,9 @@
+import globalThisShim from 'globalthis';
 import vtk from './vtk';
 import ClassHierarchy from './Common/Core/ClassHierarchy';
 
+const vtkGlobal = globalThisShim(); // returns native globalThis if compliant
```
jourdain, line 5: can you `export` it so it could be used outside?

Sources/vtk.js
```diff
@@ -1,3 +1,7 @@
+import globalThisShim from 'globalthis';
+
+const vtkGlobal = globalThisShim(); // returns native globalThis if compliant
```
jourdain, line 3: I would rather use the one from macro once exported. I don't want to spread our dependency to `globalthis` especially if we replace it down the road.

Sources/macro.js
thewtex: Sure

Sources/vtk.js
thewtex: Ok

Sources/macro.js
thewtex: I had to export from `vtk.js` becase `macro.js` depends on it.

jourdain:
LGTM

Issue #1985, opened 2021-07-06
Drawing Rectangle image using `ImageToImageData` 


EJShim:
### High-level description
`vtkImageHelpr.ImageToImageData` Only works with squre image

### Steps to reproduce
Make non-square imagedata

```
let image = new Image()
image.width = 512;
image.height = 128;

let imageData = vtkImageHelper.ImageToImageData(image)

console.log(imageData.getPointData().getScalars().getData()) // output : 0, 0, 0, 0....
```

### Detailed behavior
in ImageToImageData function, ctx.drawImage uses only image.width
 ctx.drawImage(image, -image.width / 2, -image.width / 2);

### Expected behavior
the code should be changed to use image.height
ctx.drawImage(image, -image.width / 2, -image.height/ 2);

### Environment
- **vtk.js**: 18.1.4

EJShim:
#1986

PR #1986, opened 2021-07-06
Should be able to convert non-squared image into `vtkimagedata`

EJShim:
### Context
`vtkImageHelper.ImageToImageData` only properly works with `Image` in squared size (width == height = true)

### Changes
in `ImageToImageData` changed ctx.drawImage parameter to use image.height
- [ x ] Documentation and TypeScript definitions were updated to match those changes

### Results

Rectangle-sized image can be converted into vtk.js imagedata

image
image

### Testing
Try convert (512,128) sized image into vtkImageData using `vtkImageHelper.ImageToImageData`, render usign `vtkImageSlice`.
Originally, it displays nothing, and this change resolves the problem.

- [ x] This change adds or fixes unit tests
  - **vtk.js**: latest master
  - **OS**: all

### DIFF
--- a/Sources/Common/Core/ImageHelper/index.js
+++ b/Sources/Common/Core/ImageHelper/index.js
@@ -58,7 +58,7 @@ function imageToImageData(
   ctx.translate(canvas.width / 2, canvas.height / 2);
   ctx.scale(flipX ? -1 : 1, flipY ? -1 : 1);
   ctx.rotate((rotate * Math.PI) / 180);
-  ctx.drawImage(image, -image.width / 2, -image.width / 2);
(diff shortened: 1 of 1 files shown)


thewtex:
Thanks for the contribution @EJShim ! 

Could a check for `width` and `height` please be added to the tests?

agirault:
Hi @jourdain. Thanks for approving that while you're on a break. Note that @thewtex was still reviewing this and expecting the `Testing` section of the PR to be addressed, so we should try to be aware of that before merging.

@EJShim thanks a lot for the contribution, and congrats 👍 Even though this is merged, we'd more than welcome the test that @thewtex mentioned if you had time to provide it. TY!

agirault:
Fixes #1985

PR #1992, opened 2021-07-08
feat(reslicecursorwidget): set rotation handle position distance to c…

finetjul:
…enter in screen space


Sources/Widgets/Representations/ResliceCursorContextRepresentation/index.js
```diff
@@ -296,33 +311,39 @@ function vtkResliceCursorContextRepresentation(publicAPI, model) {
...
-    },
-    Line: {
-      factor: -1.5,
-      offset: -1.5,
+function defaultValues(initialValues) {
+  return {
+    axis1Name: '',
+    axis2Name: '',
+    coincidentTopologyParameters: {
+      Point: {
```
luciemac, line 319: Shouldn't "Point", "Line", "Polygon" be an enum (or use an existing enum)?

finetjul: Such enum does not exist (and it is not the topic of this PR :-) )

Issue #1936, opened 2021-05-25
Incorrect volume mapper bounds  


sedghi:
### High-level description

Volume mapper bounds are incorrect

### Steps to reproduce

This is a bug for rendering the first and last slice.

 It is visible on any sample data (headsq.vti etc). Also, I have created a synthetic dicom data set with long stripes in each slice along the y-axis, which change position for each slice (illustration below).
In addition, if we manually change the intensity values of the first and last slice in `headsq.vti`, we can see the cut at the edges.
 

### Expected behavior

Bounds should be to the edge of the volume.

### Environment
- **vtk.js**:  17.xx, 18.xx
- **OS**:  MacOS
- **Browser**:  Chrome 90.0.4430.212

#### Code Sandbox link

https://codesandbox.io/s/volume-rendering-bounds-exp9h?file=/src/index.js

headsq.vti with the changed intensity value for the first and last slice
- As seen, the slice thickness is not consistent and the last slice is thinner (as if bounds are forcing it to get cut)
- If interpolation is nearest, the last slice is gone

#### With linear interpolation
![image](https://user-images.githubusercontent.com/7490180/119558297-3994ba80-bd6f-11eb-9f5f-c0ebc56ee546.png)

With nearest interpolation
![image](https://user-images.githubusercontent.com/7490180/119559871-38648d00-bd71-11eb-832e-6d2850865e2c.png)

This becomes more clear if we intentionally increase the slice thickness (below screenshot is from local running of demo, not in sandbox)
![image](https://user-images.githubusercontent.com/7490180/119562346-3a7c1b00-bd74-11eb-95ea-7d89cd1343f4.png)

Test dicom files can be downloaded from [my dropbox (9kb)-10 slices](https://www.dropbox.com/s/eng1kkqkl5ec49d/Archive.zip?dl=0) and can be dropped into paraview glance for volume rendering as well

Synthetic data illustration
![image](https://user-images.githubusercontent.com/7490180/119561408-1d931800-bd73-11eb-9728-74c334a59c05.png)

Paraview glance volume rendering on the left, with the slice rendering on the right

![image](https://user-images.githubusercontent.com/7490180/119563109-2be23380-bd75-11eb-8ffb-3e87b137c892.png)

sedghi:
I have tried 3D Slicer for volume rendering and I'm getting different renderings depending on CPU, GPU or Multi-volume. 

I recorded my screen and you can see it here

[...]

Issue #1884, opened 2021-04-29
Parallel projection does not render the first slice of 3D image


sedghi:
**Describe the bug**

`vtkInteractorStyleMPRSlice` and Parallel projection does not render the very first slice. 

***Live example***

https://codesandbox.io/s/parallel-projection-y9w1q

**Expected behavior**

When turning on parallel projection for `vtkInteractorStyleMPRSlice` it should render all slices

**Platform:**

***Device:***

- [x] Desktop
- [ ] Mobile

***OS:***

- [ ] Windows
- [ ] MacOS
- [ ] Linux
- [ ] Android
- [ ] iOS

***Browsers Affected:***

- [x] Chrome
- [ ] Firefox
- [ ] Microsoft Edge
- [ ] Safari
- [ ] Opera
- [ ] Brave
- [ ] IE 11

***vtk.js version:***
16.11.2

floryst:
I would recommend using the VolumeMapper clipping planes for this, but I have yet to write up an example. I can look into the issue here anyways since it probably relates to the VolumeMapper clipping planes approach.

sedghi:
@floryst Thanks for your comment and investigation in advance

floryst:
Linking to #1872

sedghi:
@floryst I'm wondering if this is related to how bounds are computed for the images which is not the exact bounds but from middle of the first slice to the middle of the last slice 

see 
https://discourse.vtk.org/t/bounding-box-of-vtkimagedata/2178/2

Issue #1969, opened 2021-06-22
Add tests harness for rendering & interaction


agirault:
### Need

[Have better test coverage](https://kitware.github.io/vtk-js/coverage/home.html), focusing on rendering and interactivity capabilities.

### Feature

Could look into Percy or Selenium for regression testing.

thewtex:
For UI testing, I have heard good things about [tiako](https://taiko.dev/).

floryst:
Another goal: unify tests and examples, i.e. examples should also be tests.

PR #1993, opened 2021-07-08
fix(VolumeMapper): extend the physical bounds of volumeMapper to match the imageData

sedghi:
### Intro
VolumeMapper renders incorrect physical bounds of the imageData. 
This has been reported [here](https://github.com/Kitware/vtk-js/issues/1936) and [here](https://github.com/Kitware/vtk-js/issues/1884) and discussion [here](https://discourse.slicer.org/t/incorrect-volume-rendering-bounds-first-and-last-slice/17827) with screenshots and screen recordings of why it is incorrect bounds of the image. This will cause incorrect rendering of imageData for the first and last slice. 

Note: 
I have created test data with 11 slices of DICOM files; you can download DICOM files from [here](https://github.com/sedghi/dicom-test/tree/master/src/data). Also the .vti file for the same series is put in the same repo and can be loaded `vtkHttpDataSetReader` in vtk.js via `reader.setUrl('https://sedghi.github.io/dicom-test/src/data/synthetic_dicom_11Slices_1mm.vti')` which is used in the new created demo

### Changes

- Physical bounds of the volumeMapper are extended by half voxel in all directions, and Origin is modified 
- A new `vOriginPlusHalfVoxelVC` is added for the shader to compute the ray intersection correctly inside the `getRayPointIntersectionBounds`
- Indices inside `computeIndexSpaceValues` logic of shader is modified to include the half voxel addition

### Results

The following renders the same volume reported in the issues + two spheres inside the space at the two corners of the volume: one at (0,0,0), which is the Origin, and another one on the other end. 

Before: 

The 11 slice volume is rendered incorrectly and lacks half voxels.

ezgif com-gif-maker (6)

After: 

You can see that the 11 slice volume is rendered correctly and the position of the spheres are correct 

ezgif com-gif-maker (7)

### Comments:

In imageData, worldToIndex is actually a worldToContinuousIndex which we have in itk and simpleITK, but we lack in vtk.js. It makes sense to have a separate method for worldToIndex and worldToContinuousIndex because of the following reason:

Assume that that Origin is set at (0,0,0); before, no volume were rendered left and bottom and back (x,y,z) of the Origin (see movie in the "before" above), so worldToIndex was calculated fine and always > 0.0 

After the proposed change, the same worldToIndex is valid; however, since the points to the left, bottom and back of the Origin are rendered as the first voxel (similar for the other end), those will have negative indices (less than 0 but greater than -0.5) which should be the case as the imagePositionPatient is always the center of the first voxel and itk and simpleITK works the same way.  

As a result, this might be confusing for people using the worldToIndex and seeing negative values, which is correct theoretically. I didn't include the worldToContinousIndex in this PR, but It makes sense to have two separate methods.

### Testing

  - **vtk.js**:  master
  - **OS**:  MacBook Pro (13-inch, 2020)
  - **Browser**:   91.0.4472.114 (Official Build) (x86_64)


agirault:
Thanks, @sedghi. That's a big change so let's make sure we're all aligned on the impact, and if this makes sense for vtk.js. For non medical-data, that convention might not be accurate.

@martinken this is directly related to some of the discussion we had recently. We might need to talk more in depth about the impact this would have, and how it related to VTK C++ as well.

martinken:
This fix (which is awesome) relies on our current definition of ImageData which I now am convinced is incorrect. Pixel values should be CellData not PointData in VTK terminology and everything else should be modified to deal with that. I think in vtk.js we should make this change as it is earlier in the process and we can do it.

This MR keeps the image data as point data but treats it as cells. Basically it assumes the point data is really cell data and adjust the rendering for it. I think that is fine for now so +1 but in the future we really should look at changing image data or creating a new one for vtk.js to have the data as cell data.

agirault:
Thanks @martinken.

@floryst @jourdain: this is big change, should that be marked as BREAKING_CHANGE, and should we think of ways to communicate that accordingly to not surprise users of the change?

@floryst @thewtex @finetjul: I believe this could have a big impact on some of your projects as well, you might want to test/review this with some of your applications that leverage imagedata before making that switch?

All: what kind of test should we add here to effectively capture this behavior and ensure it remains correct?

finetjul:
@agirault this commit is fine to me and my projects. It only impacts 3D views which are not really used in medical applications.

Concerning a good test, I think a front facing screenshot of the synthetic 3D volume with pixel positions (represented as sphere as in @sedghi gif animations) should be fine.

agirault:
>  It only impacts 3D views which are not really used in medical applications.

@finetjul you bring up a good point: will we have a mismatch between the VolumeMapper and the ImagerMapper? Shouldn't we stay consistent here? + @martinken 

> Concerning a good test, I think a front-facing screenshot of the synthetic 3D volume with pixel positions (represented as sphere as in @sedghi gif animations) should be fine.

Are you suggesting an image regression test à la VTK C++? That would require #1969 afaik.

___

@sedghi we have an internal dev meeting at Kitware this Friday, I'm thinking we'll discuss some of those points in person so we can agree on the best path forward to use your changes. Thanks again.

thewtex:
This LGTM, thanks for the contribution, @sedghi !

> @finetjul you bring up a good point: will we have a mismatch between the VolumeMapper and the ImagerMapper? Shouldn't we stay consistent here?

I was also wondering about the ImageMapper.

@martinken I see the conceptual attractiveness of `CellData` versus `PointData`, and I agree that it could be preferred. But, the backwards compatibility challenges and the ensuing potential confusion outweighs these benefits, in my humble opinion. Also, we then have to consider what to do about `CellType`, `CellPoints`, which adds overhead / confusion, without added value.

finetjul:
> > @finetjul you bring up a good point: will we have a mismatch between the VolumeMapper and the ImagerMapper? Shouldn't we stay consistent here?
> 
> I was also wondering about the ImageMapper.

I just want to make sure I get it right. I do not think we had the same issue between 2D and 3D. Currently in VTK.JS (vtkImageSlice) we have:
image

And @martinken 's suggestion is that:
image

I think I do not mind either way, maybe it is more a matter of offering convenient methods to switch representations.

thewtex:
@finetjul, @sedghi @martinken are looking to correctly put the center of a voxel in the center of its rendered representation. It does not look like that is the case in those drawings.

finetjul:
> @finetjul, @sedghi @martinken are looking to correctly put the center of a voxel in the center of its rendered representation. It does not look like that is the case in those drawings.

My bad, I just updated my drawings.

martinken:
The volume mapper and image mapper can be made to render either approach in a way we think is correct. The problem comes when a more generic filter calls the getCell interface on ImageData and processes off of that. If you use point centered data the ImageMapper can be made to render it correctly, but calling getCell on all the cells and then rendering them results in something different. We see this in ParaView IIRC. For example the image mapper might draw 4, 1mm pixels as having a total width of 4mm while the getCell interface would return only 3 cells each 1mm in size. This is where moving to a cell centerd data approch fixes issues as then the getCell method can produce consistent results with the ImageMapper and Volume Mapper. 

But these issues are beyond this MR. The changes in this MR are fine IMO for now. +1

thewtex:
Logging notes from discussion with @finetjul: as @martinken rightly notes with `getCell`, we also want to ensure `getBounds` is consistent.

agirault:
Hi @sedghi,

Here is the update from our developer meeting at Kitware.

Before we release this change, we'll want to:
- [ ] understand what that means for the bounds, then choose and define a clear convention
  - representation outside of the bounds could be an issue with picking, intersection...
  - we could change the bounds computation instead to include the whole representation
- [ ] Add consistency across the ImageMapper and the VolumeMapper
- [ ] Properly document our convention, and how it compares to VTK C++, ITK, to lower confusions
- [ ] Add regression testing on screenshots

However, this is already a very high-quality effort & PR, and a lot of us agreed that there is a very good net benefit to your changes, so we'd like to release your changes on the beta-release branch soon: @floryst will be helping with that at some point.

We'll meet internally at Kitware again to plan efforts to tackle the items above. Do let us know if you're interested in joining some of those discussions and helping with any of the actionable items.

Thank you again!

floryst:
I've changed the merge target to the `beta` branch. I can merge this shortly and we can test this out on the beta release channel.

floryst:
@sedghi I just noticed that you've committed a `yarn.lock` file. It would be helpful to remove that, since we currently already have a `package-lock.json` and we do not have infra to keep the two in sync (if that's even trivially possible). @swederik since you mentioned @sedghi is on vacation, do you have edit rights to modify the commits?

swederik:
@floryst Yep I should be able to do that. Right now I only seem to have read access, but I pinged him to give me edit access as well.

sedghi:
Hi @agirault I'm back to work.
Is there necessary to meet and discuss changes anymore after the merge or we are good for now?
Also, I was wondering when does these changes getting out of beta and into the master? Is there task that I can do to facilitate it?

Thanks

floryst:
We will need address the issues in https://github.com/Kitware/vtk-js/pull/1993#issuecomment-883396606 before we release on master. This is a fairly significant shift in how vtk.js handles image data, so we want to ensure consistency before we make the release.

Issue #1994, opened 2021-07-09
gzip decompression not working for binary files


aerogt3:
When reading binary gzipped files (in my case STL), there's no support for the compression option. fetchArray, fetchArray, fetchText all contain handling for compression, but fetchBinary does not. I have tried endlessly to implement this myself at the line linked below, but can't figure it out. 

https://github.com/Kitware/vtk-js/blob/b1a9fe4dff2ab37c7b5319412d5357f425fe89e8/Sources/IO/Core/DataAccessHelper/HttpDataAccessHelper.js#L38

PR #1995, opened 2021-07-10
implement decompression pass for gzipped binary files

aerogt3:
When reading binary gzipped files (in my case STL), the compression option has no impact. fetchArray, fetchArray, fetchText all contain handling for compression, but fetchBinary does not.

Was a pretty simple fix to include this for binary as well.

### DIFF
--- a/Sources/IO/Core/DataAccessHelper/HttpDataAccessHelper.js
+++ b/Sources/IO/Core/DataAccessHelper/HttpDataAccessHelper.js
@@ -29,13 +29,24 @@ function openAsyncXHR(method, url, options = {}) {
 }
 
 function fetchBinary(url, options = {}) {
+  if (options && options.compression && options.compression !== 'gz') {
+    vtkErrorMacro('Supported algorithms are: [gz]');
+    vtkErrorMacro(`Unkown compression algorithm: ${options.compression}`);
+  }
+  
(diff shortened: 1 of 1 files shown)


aerogt3:
#1994

jourdain:
The code change looks good. 
Just need to fix:
- [ ] Commit message to follow semantic release
- [ ] Create only 1 commit rather than a merge + commit

aerogt3:
I think I've fixed the first item?

I can't figure out how to modify this request to accomplish #2. Do I need to close this one and create another request with only a commit?

jourdain:
You can rewrite history with Git and force push it here. But if you are not that familiar with git, you can create a clean new PR with only your changes and proper commit message format. Either approach would be fine. 

Thanks for contributing to vtk.js

jourdain:
BTW, you did not fixed the commit message only the PR title/description which does not affect your commit.

aerogt3:
I don't even know what npm is, and I have barely used github before. I'd like to help, but I just don't know how to accomplish what's been asked.

The changes I made are like 4-5 lines, it's super simple. If someone who understands these tools wants to make the changes the way they are required, great. Unfortunately, navigating all these tools and conventions is beyond my knowledge.

PR #1997, opened 2021-07-12
WIP Fix #1996 and add debounced function cancellation

### LINKED ISSUE #1996: macro.shouldUpdate doesn't check for deleted output
### High-level description

vtk filters fail to update when the filter output dataset is deleted via `.delete()`.

### Steps to reproduce

- Create a vtkTrivialProducer, which is a 1-1 macro.algo that outputs the input dataset
- set the input to some vtk dataset
- get the filter output, and call `getOutputData().delete()`
- set the input to some new vtk dataset
- `getOutputData()` returns the old deleted dataset

### Detailed behavior

`macro.shouldUpdate` checks if it should update when `output.getMTime()` is stale. However, when `output` is a deleted object, `output.getMTime()` returns undefined and therefore all numerical checks yield false.

### Expected behavior

Even when the output dataset is deleted, filters should still run if their inputs are updated.

### Environment
- **vtk.js**: 18.5.1 and lower

floryst:
### Context

- Fixes #1996 
- Adds support for cancelling debounced calls.

### Changes

- Fixes `macro shouldUpdate()` to check for output dataset deletion status
- Adds support for cancelling debounced calls via `debouncedFunction.cancel()`

### Testing

- [x] Add test for shouldUpdate fix
- [x] verify typescript defs work for vtkDebouncedFunction

### DIFF
--- a/Sources/Testing/testAlgorithm.js
+++ b/Sources/Testing/testAlgorithm.js
@@ -63,3 +63,50 @@ test('Macro methods algo tests', (t) => {
 
   t.end();
 });
+
+test('Macro shouldUpdate returns true if output is deleted', (t) => {
+  const algo = {
+    publicAPI: {},
+    model: {},
+  };
+  const input1 = {
+    publicAPI: {},
+    model: {},
+  };
+
+  const input2 = {
+    publicAPI: {},
+    model: {},
+  };
+
+  macro.obj(algo.publicAPI, algo.model);
+  macro.algo(algo.publicAPI, algo.model, 1, 1);
+
+  macro.obj(input1.publicAPI, input1.model);
+  macro.obj(input2.publicAPI, input2.model);
+
+  // trivial producer
+  algo.publicAPI.requestData = (inData, outData) => {
+    outData[0] = inData[0];
+  };
+
+  algo.publicAPI.setInputData(input1.publicAPI, 0);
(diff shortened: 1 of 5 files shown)


floryst:
@agirault @WillCMcC I will make a react-vtk-js issue to cross-ref the debounce cancel functionality.

Sources/macro.js
```diff
@@ -705,6 +705,11 @@ export function algo(publicAPI, model, numberOfInputs, numberOfOutputs) {
       if (!model.output[count]) {
         return true;
       }
+
+      if (model.output[count].isDeleted()) {
+        return true;
+      }
+
       const mt = model.output[count].getMTime();
       if (mt < localMTime) {
```
agirault, line 714: @floryst ah, we're comparing `MTime` of `output` and of the actual algo, not of the `input`, so that makes sense, since I assume the modified flag will be changed when the input is changed on the algo. I believe that covers the concern I had.

Sources/Testing/testAlgorithm.js
```diff
@@ -63,3 +63,50 @@ test('Macro methods algo tests', (t) => {
 
   t.end();
 });
+
+test('Macro shouldUpdate returns true if output is deleted', (t) => {
```
agirault, line 67: image

PR #1998, opened 2021-07-12
fix(macro): Apply macro.js shim at pkg build time

### LINKED ISSUE #1990: Can't use @kitware/vtk.js with Jest
### High-level description

@kitware/vtk.js seems to be built incorrectly, test frameworks such as Jest are unable to parse the package and throw an error.

I think the library should be compiled with ES5 as target (at least, for the `module` entry point), this way better interoperability will be provided.

### Steps to reproduce

1. Download the sample project [d8lhv.zip](https://github.com/Kitware/vtk-js/files/6784310/d8lhv.zip), or setup a Create React App and import any `@kitware/vtk.js` module from the package;
2. Add `"node_modules/(?!(@kitware)/)"` to the `transformIgnorePatterns` in package.json > jest
2. Create a test file and import `App.js`
2. Run `yarn test`

### Detailed behavior

Jest fails with the following error:

```
  FAIL  src/App.test.js
  ● Test suite failed to run

    /Users/federicozivolo/Downloads/d8lhv/node_modules/@kitware/vtk.js/Common/Core/Math.js: Cannot use import statement outside a module

      at interopRequire (node_modules/babel-plugin-macros/dist/index.js:86:15)
      at applyMacros (node_modules/babel-plugin-macros/dist/index.js:208:17)
      at ImportDeclaration (node_modules/babel-plugin-macros/dist/index.js:114:28)

```

### Expected behavior

Jest should be able to import the library without erroring.

### Environment
- **vtk.js**: 18.4.
- **OS**: macOS 11.4
- **Browser**: Node 12.20.0

floryst:
babel-plugin-macros interferes with our macro.js file. As it is not
ideal to rename it, nor is it feasible to convince babel-plugin-macros
to change its "ownership" of any files ending in "macro.js", this is an
attempt at shimming macro.js to effectively sneak past
babel-plugin-macros while still maintaining compatibility with existing
macro.js consumers.

### Context

- Fixes #1990 
- [VTK discourse on the same issue](https://discourse.vtk.org/t/how-to-tune-babel-loader-to-properly-import-a-module-named-macro-js-when-using-cra/5808/4)

### Changes

- Add a package build-time step to apply a macro.js shim.

### Results

- CRA, Jest, and other users of babel-plugin-macro should no longer have cryptic errors when trying to use vtk.js.

### Testing

- [ ] verify Typescript experience is still the same
- [ ] verify both ESM and UMD packages operate correctly

### DIFF
--- a/Utilities/ci/build-npm-package.js
+++ b/Utilities/ci/build-npm-package.js
@@ -12,6 +12,18 @@ function copyDirSync(src, dst) {
   return fs.copySync(src, path.join(dst, dirname));
 }
 
+function applyBabelPluginMacroWorkaround(srcRoot, pkgName) {
+  fs.moveSync(path.join(srcRoot, 'macro.js'), path.join(srcRoot, 'macros.js'));
+
+  const shim = fs.readFileSync(path.join('Utilities', 'ci', 'macro-shim.js'), {
+    encoding: 'utf8',
+    flag: 'r',
+  });
+  const preparedShim = shim.replace('{{ packageName }}', pkgName);
+
+  fs.writeFileSync(path.join(srcRoot, 'macro.js'), preparedShim);
+}
+
 function prepareESM() {
   const pkgdir = path.join(rootdir, 'esm');
   fs.ensureDirSync(pkgdir);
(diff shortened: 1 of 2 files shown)


floryst:
This has been converted to a draft, as there is a blocking issue:

`import * as macro from '@kitware/vtk.js/macro'` fails inside babel-plugin-macros, because it does not handle ImportNamespaceSpecifiers. (See: https://github.com/kentcdodds/babel-plugin-macros/issues/111). As of now, the only resolution is to change all instances of `import * as macro from ...` to `import macro from ...`, which I think means we do not get tree shaking.

floryst:
Closing in favor of #2001

PR #1999, opened 2021-07-12
fix(prop): support nested visibility, pickable and dragable props

### LINKED ISSUE #1988: Widget unpickability should be enforced on widget representation
If you make a widget not pickable:
`widget.setPickable(false);`
or 
`widgetInView.setPickable(false);`

It does not propagate the pickability onto its representations.
Therefore its components (think of handles) are pickable but not the widget.

Worst, setting `setPickable()` on the widget representation has no effect. It does not propagate pickability to actors. Widget representation picking is actually handled by the visibility of their actors being turned off when picking happens (see [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Representations/WidgetRepresentation/index.js#L143)).
Therefore there are 2 mechanisms to handle pickability in vtk.js: visibility in the context of picking and pickable property.

I think for a representation (Context or Handle) to be pickable it needs to be pickable AND the widget be pickable.

Therefore, I would suggest to do the following:
 - have the scenegraph be a tree instead of a list. 
 - If pickable is false on a prop, prop children should not be pickable even if their pickable value is true.

finetjul:
Prop can refer to a parent. Their final pickable property is ANDed with the ones of its ancestors

BREAKING CHANGE: updateHandleVisibility lost the widgetVisibility parameter. It is no longer useful
thanks to the nested props.

fix #1988

### Context
Should now be working...

### Changes

### Testing
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


floryst:
Looks promising from a first pass. I would like to see tests to make sure `computePickable` and friends work correctly.

finetjul:
@floryst all good to you ?

jourdain:
Seems reasonable to me

PR #2001, opened 2021-07-13
Rename macro.js to macros.js

floryst:
### Context

- As per the discussion in #1990, this patch renames macro.js to macros.js, and keeps a macro.js around for backwards compatibility with downstream users.

### Changes

- Fix all usages of `macro.js` to use `macros.js`

### Results

Downstream consumers of vtk.js should no longer have issues with babel-plugin-macros eating our macro.js.

### Testing

  - **vtk.js**: v18.6.0


floryst:
FYI @FezVrasta @WillCMcC

agirault:
@jourdain

jourdain:
Sorry for the late review. But since it involve fixing all the macro(->s) import. Should we merge macro into `Sources/vtk` and make sure we only use named import/export for the pieces we want from vtk/macro?

Or do we have some cycle dependency import issue? 

That way we have little risk of conflict down the road. 

Since it is a breaking change, I'm wondering if we are better off moving away from macro altogether. 

I'm not saying we should, but I wanted us to think about that as well... What do you think?

floryst:
Merging into `Sources/vtk` is an option, but not necessary to get around the issue. This is not a breaking change from a user perspective, because I generate `macro.js` from `macros.js` at package time. It's only a "breaking change" from someone working on vtk.js, and I don't think that should be considered a breaking change (since the purpose of such designations is to notify the user that the API has changed).

Existing (external) users of `macro.js` will not be impacted by this change. The users who are impacted by the babel-plugin-macros situation will now be able to use `macros.js` instead. `macro.js` is kept around for backwards compatibility.

What do you think?

jourdain:
I see, it make sense. All good with me. Thanks for keeping it backward compatible.

jourdain (approved):

PR #2003, opened 2021-07-17
fix(Coordinate): Incorrect coordinate calculation for multiple view p…

yarous224:
### Context
The `normalizedViewportToViewport` is not calculated according to the size of the Viewport, but is calculated using the size of the Display, which leads to incorrect coordinates when I use `getComputedViewportValue`.

### Testing
I tested it on a single ViewPort and multiple ViewPorts and it works fine.


finetjul:
Thanks @yarous224 for your contribution. I can't comment on the content of your fix (@martinken ), but would you mind creating a unit test so that we make sure we don't break your fix in the future ?
Thanks.

martinken:
LGTM +1

yarous224:
@finetjul I added a unit test for multi renderers.

PR #2007, opened 2021-07-22
fix(STLReader): fix progressCallback

v-kiniv:
Assign progressCallback property to the model object, so it will be passed to dataAccessHelper upon
calling setUrl method.

### Context
Without this fix, when you use method [setUrl](https://kitware.github.io/vtk-js/api/IO_Geometry_STLReader.html#setUrl) and pass your function as [progressCallback option](https://github.com/Kitware/vtk-js/blob/master/Sources/IO/Geometry/STLReader/index.d.ts#L7) it will never be called.

### Changes
The reason why it's never called is that it's not passed to `dataAccessHelper` as [compression option do](https://github.com/Kitware/vtk-js/blob/master/Sources/IO/Geometry/STLReader/index.js#L128). I updated the code to assign `progressCallback` property to the model object the same way as `compression`.

### Testing

  - **vtk.js**: 18.7.2 
  - **OS**: MacOS 11.4 
  - **Browser**: Chrome 91.0.4472.114 

p.s. I found that progressCallback can be passed through the reader constructor, so this fix is not mandatory. However according to the documentation you should be able to pass progressCallback when using setURL method, so you may still want to merge it.

### DIFF
--- a/Sources/IO/Geometry/STLReader/index.js
+++ b/Sources/IO/Geometry/STLReader/index.js
@@ -126,6 +126,7 @@ function vtkSTLReader(publicAPI, model) {
     model.baseURL = path.join('/');
 
     model.compression = option.compression;
+    model.progressCallback = option.progressCallback;
 
     // Fetch metadata
     return publicAPI.loadData({
--- a/Sources/IO/Geometry/STLReader/index.js
+++ b/Sources/IO/Geometry/STLReader/index.js
@@ -103,8 +103,14 @@ function vtkSTLReader(publicAPI, model) {
 
   // Internal method to fetch Array
   function fetchData(url, option = {}) {
-    const compression = option.compression || model.compression;
-    const progressCallback = option.progressCallback || model.progressCallback;
+    const compression =
+      option.compression !== 'undefined'
+        ? option.compression
+        : model.compression;
+    const progressCallback =
+      option.progressCallback !== 'undefined'
+        ? option.progressCallback
+        : model.progressCallback;
 
     if (option.binary) {
       return model.dataAccessHelper.fetchBinary(url, {
--- a/Sources/IO/Geometry/STLReader/index.js
+++ b/Sources/IO/Geometry/STLReader/index.js
@@ -104,11 +104,9 @@ function vtkSTLReader(publicAPI, model) {
   // Internal method to fetch Array
   function fetchData(url, option = {}) {
     const compression =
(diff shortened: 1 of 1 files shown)


finetjul:
Hi @v-kiniv,

Thanks for your fix. 
To be consistent, shouldn't you also remove line 133 ?

However, I think the mechanism should be tweaked so that:
 - use `option.progressCallback` when specified (not undefined)
 - use `model.progressCallback` otherwise

Same thing for `compression`. What do you think ?

v-kiniv:
I should've, but I wanted to make as less changes as possible.
I agree on conditional use of both callback and compression, check the second commit. I also changed line 131, so that the `option` object passed directly, is it ok?

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -103,7 +103,9 @@ function vtkSTLReader(publicAPI, model) {
 
   // Internal method to fetch Array
   function fetchData(url, option = {}) {
-    const { compression, progressCallback } = model;
+    const compression = option.compression || model.compression;
+    const progressCallback = option.progressCallback || model.progressCallback;
```
finetjul, line 107: I would make the difference between `option.progressCallback === undefined` and `option.progressCallback === null`.
In the latter case, it means that we do not want progress callback, in the former case, it means we want the model progressCallback.
Hence:
```
const compression = option.compression !== undefined ? option.compression : model.compression;
const progressCallback = option.progressCallback !== undefined ? option.progressCallback : model.progressCallback;
```

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -103,8 +103,14 @@ function vtkSTLReader(publicAPI, model) {
 
   // Internal method to fetch Array
   function fetchData(url, option = {}) {
-    const compression = option.compression || model.compression;
-    const progressCallback = option.progressCallback || model.progressCallback;
+    const compression =
+      option.compression !== 'undefined'
```
finetjul, line 107: why `'undefined'` and not `undefined`

v-kiniv: Sorry, I usually compare using `typeof` and I automatically added quotes to your snippet.

finetjul:
Thanks @v-kiniv

v-kiniv:
No problem, thanks for the library

PR #2009, opened 2021-07-23
Fix: ImageCroppingWidget interactions

### LINKED ISSUE #2008: ImageCroppingWidget interaction broken when origin is not [0,0,0]
When dragging an handle, it seems that only the X movement of the mouse has an effect, while dragging in the Y direction has no effect. This behaviour changes with different images. I've found that setting image origin to [0,0,0] fix the issue.

### Steps to reproduce
A simple way to reproduce the issue is using the ImageCroppingWidget example on vtk.js examples page. Just force image origin to be not [0,0,0] (ie image.setSpacing(200,200,200) )and interact with handles on the cube widget faces.

### Environment
- **vtk.js**: 18.7.2
- **Browser**: Chrome

ronzim:
Fix #2008 by removing manipulator normal transformation.


floryst:
Hi, thanks for the PR! As a housekeeping thing, please follow [semantic-release](https://github.com/semantic-release/semantic-release) commit messages.

Removing the transformVec3 is not sufficient, as I remember the cropping planes being defined in image space. That still is the buggy line of code, so we need a different fix. The `indexToWorldT` transform consists of the rotation, scaling, and translation components, when all we need is the rotation component when transforming the axis vectors. Using `gl-matrix`, I think this should do the trick:

```
// in helper.js
import { quat, mat4, vec3 } from 'gl-matrix';
function rotateVec3(vec, transform) { // transform is a mat4
  const out = vec3.create();
  const q = quat.create();
  mat4.getRotation(q, transform);
  vec3.transformQuat(out, vec, q);
  return out;
}

// in behavior.js
manipulator.setNormal(rotateVec3(axis, indexToWorldT))
```

PR #2014, opened 2021-07-26
fix(interactorstyleimage): fix vtkInteractorStyleImage.setCurrentImag…

finetjul:
…eNumber()

### Context
setCurrentImageNumber was not working when props other than vtkImageSlice were in the renderer.

### Changes
The fix filters out props that are not vtkImageSlice.
-> not needed

### Testing

  - **vtk.js**: master
  - **OS**: Windows 10
  - **Browser**: Chrome


Sources/Interaction/Style/InteractorStyleImage/index.js
```diff
@@ -267,7 +257,7 @@ const DEFAULT_VALUES = {
   windowLevelCurrentPosition: [0, 0],
   lastSlicePosition: 0,
   windowLevelInitial: [1.0, 0.5],
-  currentImageProperty: 0,
+  // currentImageProperty: null,
```
agirault, line 260: why commented & null?

Sources/Interaction/Style/InteractorStyleImage/index.js
```diff
@@ -223,29 +223,19 @@ function vtkInteractorStyleImage(publicAPI, model) {
     model.currentImageNumber = i;
 
     function propMatch(j, prop, targetIndex) {
-      if (
-        prop.isA('vtkImageSlice') &&
```
agirault, line 227: was the bug due to that specific line, now moved to l231?

Sources/Interaction/Style/InteractorStyleImage/index.js
jourdain: When you don't initialize a default value it is common practice to let it visible (so you know it is there) but there is no need to set its value to null as by default it will be undefined.

Sources/Interaction/Style/InteractorStyleImage/index.js
finetjul: That's right. 
Relying on indexes works only if the number of props is stable. There was an issue when adding multiple non-vtkImageSlice props(e.g. widgets). Those were messing up with the index, hence targetIndex could not be found.
Solution is to filter out all non vtkImageSlice first.

PR #2021, opened 2021-08-02
fix(shapewidget): fix ShapeWidgets

finetjul:
A recent change Widget behavior no longer share the widget factory model.
- Make vtkCircleContextRepresentation and vtkRectangleContextRepresentation inherit from Context representation rather than Handle representation
- Fix default values to support multiple widget instances
- `setPixelScale` has been removed. It should be replaced by point handle `scale1` mixin and `scaleInPixels`
- `useHandles` has been removed. It should be replaced by `setHandleVisibility`.
- `resetAfterPointPlacement` is now false by default.
- ShapeWidget example has been improved to demo widget configurability:
-- Rectangle widget does not show point handles
-- Ellipse widget hides point handles when not the active widget
-- Circle widget has a large point1 handle
-- Rectangle widget resets after point placement

### Changes

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
```diff
@@ -27,6 +27,13 @@ function makeBoundsFromPoints(point1, point2) {
 export default function widgetBehavior(publicAPI, model) {
   model.classHierarchy.push('vtkShapeWidgetProp');
 
+  model.keysDown = {};
+  model.modifierBehavior = model.factory.getReferenceByName('modifierBehavior');
+  model.resetAfterPointPlacement = model.factory.getReferenceByName(
+    'resetAfterPointPlacement'
+  );
+  macro.setGet(publicAPI, model, ['resetAfterPointPlacement']);
```
finetjul, line 35: @floryst I'm not very happy with this part. Do you have something better to suggest ?

To me, those properties are global (hence initializable in `vtkShapeWidget.newInstance`) but are used only by the behavior (i.e. view widget).

Maybe we should extend `methodsToLink` to propagate properties betwen Widget and ViewWidget in addition to the existing property propagation between ViewWidget and Representation.

WDYT ?

Sources/Widgets/Widgets3D/EllipseWidget/state.js
```diff
@@ -19,7 +19,7 @@ export default function generateState() {
       name: 'point2Handle',
       initialValues: {
         scale1: 10,
-        origin: [0, 0, 0],
+        origin: [undefined, undefined, undefined],
```
floryst, line 22: What's the purpose of setting these to undefined?

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
floryst: Do you mean `resetAfterPointPlacement` is global in the sense that `widget.setResetAfterPointPlacement(true)` should affect all view widgets? If so, line 32 break that assumption because `model.resetAfterPointPlacement` does not refer to the property from the factory. In that case, I would opt to be explicit, and say `publicAPI.setResetAfterPointPlacement = model.factory.setResetAfterPointPlacement`, and the same for the getter.

If you are only interested in passing initialization, then what you have should be fine for now. With regard to the `methodsToLink` extension idea, I'm not too much a fan of that either. I would much rather be explicit in initialization. 

Here's an idea: have ShapeWidget override `getWidgetForView`, like so:

```js
const superClass = { ...publicAPI };
publicAPI.getWidgetForView = (props) => {
  return superClass.getWidgetForView({
    ...props,
    initialValues: {
      modifierBehavior: model.modifierBehavior,
      resetAfterPointPlacement: model.resetAfterPointPlacement,
      ...props.initialValues,
    },
  });
};
```

You'll still need to have `macro.setGet(...)` in the widgetBehavior, but that's a consequence of how restricted our factory creation method (getWidgetForView) is.

Sources/Widgets/Widgets3D/EllipseWidget/state.js
finetjul: it is a way to have the point2 handle not initialized hence not visible (but with visible property set to true).
For some reason, I can't set it to `null`, `undefined` or `[]`.

floryst: I find this approach opaque as to its purpose. Is it not sufficient to toggle the visible property as needed?

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
finetjul: Thanks for your answer and ideas. 
> Do you mean resetAfterPointPlacement is global in the sense that widget.setResetAfterPointPlacement(true) should affect all view widgets?

Yes, this is what I want.
I like your solution `publicAPI.setResetAfterPointPlacement = model.factory.setResetAfterPointPlacement` because the property does not have to be stored twice.

Sources/Widgets/Widgets3D/EllipseWidget/state.js
finetjul: handles used to not visible by default and I got feedback from developers confused with widget initialization. For example, someone wanted to initialize the spline widget programatically, he was doing :
```
points.forEach((point) => {
      widgetFactory.getWidgetState().addHandle().setOrigin(...point);
    });
```
where he should have done
```
points.forEach((point) => {
      const handle = widgetFactory.getWidgetState().addHandle();
      handle.setOrigin(...point);
      handle.setVisible(true);
    });
```

When we create a prop, we expect it to be visible by default.

Moreover, when dealing with handles in behavior, if we keep `[0, 0, 0]` as default value, it is hard to know if the handle has yet been initialized or not.
WDYT ?

floryst: Ah, fair point. I think having a null/[] origin can effectively mean "do not show". That also means representations must be changed to support this, e.g. SphereHandleRepresentation should see that origin is null or empty and not show that state.

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
```diff
@@ -294,6 +256,35 @@ export default function widgetBehavior(publicAPI, model) {
...
+    // // const screenPoint2 = model.apiSpecificRenderWindow.worldToDisplay(...point2, model.renderer);
+    // const screenPoint2 = publicAPI.computeWorldToDisplay(
+    //   model.renderer,
+    //   ...point2
+    // );
+    // const screenBounds = vtkMath.computeBoundsFromPoints(
+    //   screenPoint1,
+    //   screenPoint2,
+    //   []
+    // );
```
floryst, line 275: Is the commented code still necessary here?

Sources/Widgets/Widgets3D/ShapeWidget/example/index.js
```diff
@@ -210,88 +206,64 @@ reader
...
+    //         VerticalTextPosition.INSIDE_TOP,
+    //         width,
+    //         height
+    //       )
+    //     );
+    //     labelRep.setLabelText(text);
+
+    //     labelRep.setTextAlign(TextAlign.RIGHT);
+    //   }
+    // );
```
floryst, line 266: ditto with the commented code comment above

floryst:
@finetjul are you thinking of waiting for #2026 first?

finetjul:
> @finetjul are you thinking of waiting for #2026 first?

Yes, I would like to fix it first (hopefully tomorrow)... I also have a big commit coming up as well in this PR, i.e. getting rid of the old style widgets. 

Are you waiting for me for merging something else ?

floryst:
I just wanted to check on the status of this PR, since it wasn't evident to me whether it was in a mergeable state or not.

floryst:
setArray changes look good to me.

finetjul:
Replaced by #2054 and #2115

PR #2022, opened 2021-08-03
feat(ScalarBarActor): make layout function settable / gettable

### LINKED ISSUE #2020: Customize automated layout of ScalarBarActor
### Need

Customize automated layout of ScalarBarActor

The automated layout function works well but I'm running into this situation:

### Feature

Automated ScalarBarActor layout function is here:
https://github.com/Kitware/vtk-js/blob/b617e02a7d09c5044f559b28e0bb4c67dad3b639/Sources/Rendering/Core/ScalarBarActor/index.js#L52

I would like to pass a custom function that would be called instead of this default layout function.

```javascript
const scalarBarActor = vtkScalarBarActor.newInstance({
  automated: true,
  autoCompute: () => {
    // layout the legend here
    ...
  }
```

An alternative proposal would be to parameterize parts of this automated layout function, but allowing a user to override the function entirely would be an extremely flexible approach.

I would be willing to submit a PR to add this small improvement.

NGimbal:
### Context

fix #2020

Allow user to get / set autoLayout function, previously called computeAndApplyAutomatedSettings

### Changes

ScalarBarActor doesn't seem to have documentation or TypeScript definitions yet.
https://kitware.github.io/vtk-js/api/Rendering_Core_ScalarBarActor.html

### Results

I've successfully been able to customize the autoLayout function! 

Not sure the best way to develop test changes to vtk.js module:
https://discourse.vtk.org/t/npm-link-for-vtk-js/6283

### Testing

  - latest master: 
  - macOS 12.2.1: 
  - Firefox 

_This contribution is funded by [Pollination](https://www.pollination.cloud/)._

### DIFF
--- a/Sources/Rendering/Core/ScalarBarActor/index.js
+++ b/Sources/Rendering/Core/ScalarBarActor/index.js
@@ -34,22 +34,23 @@ function applyTextStyle(ctx, style) {
   ctx.font = `${style.fontStyle} ${style.fontSize}px ${style.fontFamily}`;
 }
 
-function vtkScalarBarActor(publicAPI, model) {
-  // Set our className
-  model.classHierarchy.push('vtkScalarBarActor');
+// ----------------------------------------------------------------------------
+// Default autoLayout function
+// ----------------------------------------------------------------------------
 
-  // compute good values to use based on window size etc
-  // a bunch of heuristics here with hand tuned constants
-  // These values worked for me but really this method
-  // could be redically changed. The basic gist is
-  // 1) compute a resonable font size
-  // 2) render the text atlas using those font sizes
-  // 3) pick horizontal or vertical bsed on window size
-  // 4) based on the size of the title and tick labels rendered
-  //    compute the box size and position such that
-  //    the text will all fit nicely and the bar will be a resonable size
-  // 5) compute the bar segments based on the above settings
-  publicAPI.computeAndApplyAutomatedSettings = () => {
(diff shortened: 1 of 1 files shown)


jourdain:
LGTM, what do you think @martinken ?

martinken:
Yup LGTM +1

jourdain:
Can you rename your commit like `feat(ScalarBarActor): make layout function settable / gettable` and `export defaultAutoLayout` so someone can reset it or better add a method that will `resetAutoLayoutToDefault()`. Then you should be all-set.

jourdain:
Can you force push the changes so only 1 commit exist? Otherwise both entry will show up in the changelog. 

For making tests you can look at [this test](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/Glyph3DMapper/test/testGlyph3DMapper.js) while using a fake baseline. Then save the generated one when testing from the output markdown/html. You will have to register your test [here](https://github.com/Kitware/vtk-js/blob/master/Sources/tests.js) and for the debug phase just comment out all the other ones.

HTH

NGimbal:
@jourdain will do.

Regarding creating tests, are you wanting that to be part of the same commit?

jourdain:
test could happen later via another PR or commit.

jourdain:
@floryst I was going to merge that one, but since you are doing some work on vtk.js you may want to bundle it with another of your PR. Therefore I will let you merge when you see fit. Thx

PR #2025, opened 2021-08-03
fix(vtkobjectmanager.js): make vtkObjectManager.update() return a pro…

AntoineRuffion:
…mise

### Context
 Allow vtkObjectManager.update() to return a promise. Following (https://gitlab.kitware.com/vtk/vtk/-/merge_requests/8235) on VTK, needed to trigger update after receiving all data from a polydata.

### Changes
 vtkObjectManager.update() now returns a promise. Underlying functions have also been modified to return promises.

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 


Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -1,4 +1,4 @@
-import * as macro from 'vtk.js/Sources/macros';
+import * as macro from 'vtk.js/Sources/macro';
```
jourdain, line 1: This should not be changed

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -23,8 +23,6 @@ import vtkImageSlice from 'vtk.js/Sources/Rendering/Core/ImageSlice';
 import vtkImageMapper from 'vtk.js/Sources/Rendering/Core/ImageMapper';
 import vtkImageProperty from 'vtk.js/Sources/Rendering/Core/ImageProperty';
 import vtkPiecewiseFunction from 'vtk.js/Sources/Common/DataModel/PiecewiseFunction';
-import vtkCubeAxesActor from 'vtk.js/Sources/Rendering/Core/CubeAxesActor';
-import vtkScalarBarActor from 'vtk.js/Sources/Rendering/Core/ScalarBarActor';
```
jourdain, line 27: Removing code from master is bad.

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -116,9 +114,9 @@ function update(type, instance, props, context) {
   }
   const handler = TYPE_HANDLERS[type];
   if (handler && handler.update) {
-    handler.update(instance, props, context);
+    return handler.update(instance, props, context);
   } else {
-    console.log('no updater for', type);
+    return console.log('no updater for', type);
```
jourdain, line 119: This is wrong.

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -243,6 +241,7 @@ function genericUpdater(instance, state, context) {
     });
   }
 
+  let res = Promise.resolve();
```
jourdain, line 244: Maybe use `Promise.resolve(true);` instead

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -580,14 +581,6 @@ const DEFAULT_MAPPING = {
...
   },
-  vtkCubeAxesActor: {
-    build: vtkCubeAxesActor.newInstance,
-    update: genericUpdater,
-  },
-  vtkScalarBarActor: {
-    build: vtkScalarBarActor.newInstance,
-    update: genericUpdater,
-  },
 };
```
jourdain, line 584: removing code from master is bad.

AntoineRuffion: My bad, this was not my intention.

jourdain:
Update commit message to `fix(ObjectManager): update() now return a promise`

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -116,9 +116,9 @@ function update(type, instance, props, context) {
   }
   const handler = TYPE_HANDLERS[type];
   if (handler && handler.update) {
-    handler.update(instance, props, context);
+    return handler.update(instance, props, context);
   } else {
-    console.log('no updater for', type);
+    return (() => console.log('no updater for', type));
```
jourdain, line 121: console.log does not return a promise. 

Should be

```
console.log('no updater for', type);
return Promise.reject(false);
```

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -275,6 +276,7 @@ function genericUpdater(instance, state, context) {
       .finally(context.end); // -> end(arraysToBind)
   }
   context.end(); // -> end(generic-updater)
+  return res;
```
jourdain, line 279: Should you simply `return Promise.resolve(true)` since you are not doing anything with `res`?

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
jourdain: Or better 
```
return Promise.reject(new Error(`No updater for ${type}`));
```

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
AntoineRuffion: Removed the use of res. We should now only return `Promise.resolve(true)`.

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -116,9 +116,9 @@ function update(type, instance, props, context) {
   }
   const handler = TYPE_HANDLERS[type];
   if (handler && handler.update) {
-    handler.update(instance, props, context);
+    return handler.update(instance, props, context);
   } else {
-    console.log('no updater for', type);
+    return Promise.reject(new Error('No updater for ${type}'));
```
jourdain, line 121: To have `${type}` properly templated you need to use the back tick rather than the standard '.

jourdain:
Looks reasonable now. But does that work for you?

AntoineRuffion:
It does, with this I'm able to trigger rendering after a successful update of an object.
Thank you for the comments, and sorry for the amateurish merge request.

jourdain:
LGTM 

@floryst feel free to merge it with the other PR to bundle

daker:
@AntoineRuffion some tests are failing:

```js
Error:   119:5   error  Function 'update' expected no return value  consistent-return
Error:   120:10  error  Unnecessary 'else' after 'return'           no-else-return
Error:   121:5   error  Function 'update' expected no return value  consistent-return
```

jourdain:
Those are just linter feedback which indeed needs to be fixed. Thanks for extracting them from the CI @daker!

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -275,6 +274,7 @@ function genericUpdater(instance, state, context) {
       .finally(context.end); // -> end(arraysToBind)
   }
   context.end(); // -> end(generic-updater)
+  return Promise.resolve(true);
```
finetjul, line 277: didn't you wanted to wait for `bindArrays()` line 265 to be called before resolving the promise ?

jourdain: I would think so but when I suggested it, nothing changed so I wasn't sure. At least if the API is consistent that could be fixed afterward.

floryst: I'm confused by this as well. If `state.arrays` is truthy, then the immediate line of execution becomes the following:
1. L262: `context.start()`
2. L263: Create a promise that waits for all other promises
3. L276: `context.end()` (this runs before `bindArrays`)

My question: does the context need to be started for `bindArrays` to work? Because in this scenario, the context is closed well before the `.then` condition on L264 is executed.

One solution to this is to simply say `return Promise.all(promises)` on L263 and do a `return true` after L265.

jourdain: Assuming I'm thinking of the right thing:
start => +1 
end => -1
The in-between work will also call internally some start/end. So the context is not really closed.
 
basically the code should be like:
```
ctx.start()
for array in arrays:
   ctx.start()
   [...]
   ctx.end()
ctx.end()
```

floryst:
@jourdain is this still good to go? The context.start/end comments still are technically unresolved.

jourdain:
I guess that PR has 2 part. 
1. returning promise
2. resolving the promise correctly

I think (1) is ok but (2) is wrong. Since we were not expecting a promise before, having (2) be wrong is not a real issue. It will be once people start using (1). 

So it is up to you if you want to merge it now (mainly for having 1) and hope that (2) will be in focus once they realize their change was not complete.

AntoineRuffion:
Sorry for the lack of updates on that one.

I'll resolve the two comments ASAP (hopefully tomorrow) so that this can be merged.

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -261,20 +261,24 @@ function genericUpdater(instance, state, context) {
...
-        console.error(
-          'Error in array handling for state',
-          JSON.stringify(state),
-          error
-        );
-      })
-      .finally(context.end); // -> end(arraysToBind)
+    dependencies.push(
+      ...promises
+        .then(() => {
```
finetjul, line 266: I'm surprised you can do `...promises.then()`

You can do: 
 1) bind arrays as soon as they have been fetched, but this is a new behavior though, 
 2) or bind them in the `Promises.all(dependencies).then()`. But it looses "generic" meaning of "dependencies.
 3) `dependencies.push(promises); dependencies.push(Promise.all(promises)...)` and then `return Promise.all(dependencies)`

I would go for Solution 3, but I would be curious to know what @jourdain would prefer

jourdain: This syntax is wrong indeed. 

Please read my initial comments, I gave you the exact code to write that would be correct. 

```
dependencies.push(...promises);
[...]
return Promise.all(dependencies);
```

jourdain: dependencies should only contains promises, no need to nest some Promise.all in them. I would rather keep it flat rather than having a useless hierarchy.

finetjul: With what you suggest Sebastien, you take the risk to have the returned promise being resolved BEFORE the arrays are bound. I guess it is fine but that means the application can't ask for a re-render when the returned promise is resolved (i.e. arrays may not yet be bound).

jourdain: Then just add a promise in your dependencies list that will only resolve once the bind happen...

finetjul: then it should be what Antoine did with his last [push](https://github.com/Kitware/vtk-js/pull/2025/commits/fbf4bc10ada68e2210a6617d37e3af600d818c63). But I guess he can remove the intermediate promises.

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
```diff
@@ -260,21 +260,26 @@ function genericUpdater(instance, state, context) {
         })
         .finally(context.end); // -> end(arrays)
     });
+    dependencies.push(promises);
```
jourdain, line 263: This is wrong, you do not want to have `promises = [[promise, ...]]` but you want to have `promises = [promise, ...]`

jourdain: But since you push your `promise.all().then()` later on, you don't need to add anything here.

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
jourdain: Kind of, yes. With some small error above.

jourdain (changes requested):

Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js
jourdain: In summary Antoine just need to remove `dependencies.push(promises);` as this is the only remaining thing that is wrong.

finetjul:
LGTM

jourdain (approved):

PR #2030, opened 2021-08-04
fix(widgetrepresentation): fix getPixelWorldHeightAtCoord() scaling

### LINKED ISSUE #2029: [regression] worldToDisplay returns doubled coordinates
### High-level description

Starting from @kitware/vtk.ks@19.0.4 the `worldToDisplay` method returns wrong values (they are double their supposed value).

With 19.0.3 everything is okay.

### Steps to reproduce

```js
const viewCtx = useContext(Contexts.ViewContext);
const renderer = viewCtx?.renderer;
const view = renderer?.getRenderWindow()?.getViews()[0];

console.log(view?.worldToDisplay(point[0], point[1], point[2], renderer));
```

`openglRenderWindow.worldToDisplay(point[0], point[1], point[2], renderer)` has the same issue.

### Detailed behavior

The values are correct if I divide them by 2, I think it may relate to `getPixelWorldHeightAtCoord` changes made in 19.0.4

### Expected behavior

The returned values should be correct.

### Environment
- **vtk.js**: 19.0.4
- **OS**: macOS
- **Browser**: Chrome 91

finetjul:
getPixelWorldHeightAtCoord() should not apply the window.devicePixelRatio scaling. 
ParallelScale should be multiplied by 2 to get the window height.

fix #2029

### Changes

### Testing

  - **vtk.js**: master
  - **OS**: Windows 10
  - **Browser**: Chrome


finetjul:
@martinken do you confirm the [math](https://github.com/Kitware/vtk-js/pull/2030/commits/5216bca8462a7e28cb4a5730809a8d318233ddc4) ?

martinken:
Yes to the 2.0*parallel scale. The ortho matrix is setup to be 2*parallelScale ala the +- 1.0 in the code below.
```
 if (model.parallelProjection) {
      // set up a rectangular parallelipiped
      const width = model.parallelScale * aspect;
      const height = model.parallelScale;

      const xmin = (model.windowCenter[0] - 1.0) * width;
      const xmax = (model.windowCenter[0] + 1.0) * width;
      const ymin = (model.windowCenter[1] - 1.0) * height;
      const ymax = (model.windowCenter[1] + 1.0) * height;
```

PR #2039, opened 2021-08-11
fix(Rendering): support half float texturs

martinken:
When float textures are not available fall back to half float
if they are available.


swederik:
Hi @martinken,

This is very interesting. It looks like you've implemented half float support for 2D textures. It appears that gl.HALF_FLOAT should work for WebGL2 with 3D textures as well (https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/texImage3D). Have you already tried that? If not, @sedghi and I will try to do that and send a PR.

swederik:
Awesome! We actually tried it shortly after making this comment and it
seems to work fine, so we should be able to make a PR soon after some more
testing.

PR #2042, opened 2021-08-13
feat(View): Replace useFocalPointAsCenterOfRotation with more generic centerOfRotation setting

tamaynard:
### Context
This replaces the "useFocalPointAsCenterOfRotation" setting with a more generic "centerOfRotation" setting. To keep the previous behavior, simply pass the camera focal point to the new setting.

### Testing
Call setCenterOfRotation on the rotate manipulator with a non-zero vector parameter.

### DIFF
--- a/Sources/Interaction/Manipulators/MouseCameraTrackballRotateManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseCameraTrackballRotateManipulator/index.js
@@ -35,7 +35,7 @@ function vtkMouseCameraTrackballRotateManipulator(publicAPI, model) {
     mat4.identity(trans);
 
     const { center, rotationFactor } = model;
-    const useCenterOfRotation = vec3.length(model.centerOfRotation) !== 0.f;
+    const useCenterOfRotation = vec3.length(model.centerOfRotation) !== 0.0;
 
     if (useCenterOfRotation) {
       center[0] = model.centerOfRotation[0];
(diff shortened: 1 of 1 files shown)


tamaynard:
I messed up commit messages when fixing formatting issues. I'll resubmit this PR with the correct formatting in one commit.

Sources/Interaction/Manipulators/MouseCameraTrackballRotateManipulator/index.js
```diff
@@ -36,11 +36,12 @@ function vtkMouseCameraTrackballRotateManipulator(publicAPI, model) {
     mat4.identity(trans);
 
     const { center, rotationFactor } = model;
+    const useCenterOfRotation = vec3.length(model.centerOfRotation) !== 0.0;
```
finetjul, line 39: You may want to use instead `vtkMath.areEquals(model.centerOfRotation, [0, 0, 0])`

finetjul:
It is not convenient for the developer to no longer be able to simply rotate around the focal point. (i.e. the developer now has to observe the camera and update the manipulator.
It would be great to keep such feature (to rotate around focal point).

I would therefore keep `model.useFocalPointAsCenterOfRotation` in addition to `model.centerOfRotation`.
Alternatively we could say that if model.centerOfRotation is `null` (newly supported), then it rotates around camera focalPoint. I'm a bit afraid that it would become a hidden feature. I would prefer to stay explicit.

finetjul:
btw, you did not have to close the PR, you could have simply force pushed your fixed branch onto the PR

PR #2043, opened 2021-08-13
feat(View): Replace useFocalPointAsCenterOfRotation with more generic…

tamaynard:
… centerOfRotation setting

### Context
This replaces the "useFocalPointAsCenterOfRotation" setting with a more generic "centerOfRotation" setting. To keep the previous behavior, simply pass the camera focal point to the new setting.

### Testing
Call setCenterOfRotation on the rotate manipulator with a non-zero vector parameter.

### DIFF
--- a/Sources/Interaction/Manipulators/MouseCameraTrackballRotateManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseCameraTrackballRotateManipulator/index.js
@@ -36,7 +36,7 @@ function vtkMouseCameraTrackballRotateManipulator(publicAPI, model) {
     mat4.identity(trans);
 
     const { center, rotationFactor } = model;
-    const useCenterOfRotation = vec3.length(model.centerOfRotation) !== 0.0;
+    const useCenterOfRotation = vtkMath.areEquals(model.centerOfRotation, [0, 0, 0]);
 
     if (useCenterOfRotation) {
       center[0] = model.centerOfRotation[0];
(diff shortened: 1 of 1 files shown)


finetjul:
It is not convenient for the developer to no longer be able to simply rotate around the focal point. (i.e. the developer now has to observe the camera and update the manipulator.
It would be great to keep such feature (to rotate around focal point).

I would therefore keep model.useFocalPointAsCenterOfRotation in addition to model.centerOfRotation.
Alternatively we could say that if model.centerOfRotation is null (newly supported), then it rotates around camera focalPoint. I'm a bit afraid that it would become a hidden feature. I would prefer to stay explicit.

tamaynard:
My thinking was that instead of a specific one-off use-case (rotating about the focal point), this could support any arbitrary center of rotation, including the focal point if so desired. I agree though, that we shouldn't break any existing downstream code that may be using this feature already. I was the original contributor of the useFocalPointAsCenterOfRotation feature; in replacing this I had assumed I was the only user, and if not that other users could easily migrate their code to utilize this new approach. @jourdain you approved the original feature, do you have any thoughts on this alternative approach, or if we should try to support both as suggested by @finetjul ?

jourdain:
We already support [any arbitrary center of rotation](https://github.com/Kitware/vtk-js/blob/master/Sources/Interaction/Manipulators/CompositeCameraManipulator/index.js#L42) from the core design. The useFocalPointAsCenterOfRotation was added to support the focalPoint UseCase. I'm not sure I understand the need of that PR?

tamaynard:
Yes I see it basically duplicates the existing functionality provided by the parent object. It might not have been clear to me at the time what it was doing as the term 'center' is a bit vague. I do recall some issue with setting a different center of rotation along with the world up vector option. I'll test it out in combination with that option, and submit a different PR if necessary. In the meantime I'll close this for now. Thanks again for the feedback.

PR #2046, opened 2021-08-27
feat(Rendering): Add support for halfFloat 3D textures

sedghi:
Recently half float precision for 2D textures was [added](https://github.com/sedghi/vtk-js/commit/29f0cb94b57485ccbe008d0a44ed16a6004a5d89), this PR will try to add it for the 3D textures too.  

Data used is a full body CT from [TCIA](https://wiki.cancerimagingarchive.net/display/Public/Head-Neck+Cetuximab) (download .vti from [here](https://www.dropbox.com/sh/4pq08njcfmos2nf/AACFoqprysqp5migPtLS9rUua?dl=0)). 

run `npm run example -- VolumeMapperBlendModes` with the data

### Before
https://user-images.githubusercontent.com/7490180/131192577-9433a02b-084e-41de-b866-0324261749c7.mp4

Chrome task manager before 
image

### After

https://user-images.githubusercontent.com/7490180/131192631-bae03a2a-79e6-4b68-b948-601772f5186f.mp4

Chrome task manager after 

image

### PR and Code Checklist

### Changes

### Testing

  - **vtk.js**: latest master
  - **OS**: macOS 10.15.7
  - **Browser**: Chrome  92.0.4515.159 

CC'ing @swederik

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -1215,6 +1215,8 @@ function vtkOpenGLTexture(publicAPI, model) {
     // compute min and max values
     const { offset: computedOffset, scale: computedScale } =
       computeScaleOffsets(numComps, numPixelsIn, data);
+    model.volumeInfo.dataComputedScale = computedScale;
+    model.volumeInfo.dataComputedOffset = computedOffset;
 
     const useHalfFloat = checkUseHalfFloat(
       dataType,
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -474,7 +474,7 @@ function vtkOpenGLTexture(publicAPI, model) {
   };
 
   //----------------------------------------------------------------------------
-  publicAPI.getDefaultDataType = (vtkScalarType) => {
+  publicAPI.getDefaultDataType = (vtkScalarType, useHalfFloatType = false) => {
     // DON'T DEAL with VTK_CHAR as this is platform dependent.
     if (model.openGLRenderWindow.getWebgl2()) {
       switch (vtkScalarType) {
@@ -482,10 +482,10 @@ function vtkOpenGLTexture(publicAPI, model) {
         //   return model.context.BYTE;
         case VtkDataTypes.UNSIGNED_CHAR:
           return model.context.UNSIGNED_BYTE;
-        // case VtkDataTypes.SHORT:
-        //   return model.context.SHORT;
-        // case VtkDataTypes.UNSIGNED_SHORT:
-        //   return model.context.UNSIGNED_SHORT;
+        case useHalfFloatType && VtkDataTypes.SHORT:
+          return model.context.HALF_FLOAT;
+        case useHalfFloatType && VtkDataTypes.UNSIGNED_SHORT:
+          return model.context.HALF_FLOAT;
         // case VtkDataTypes.INT:
         //   return model.context.INT;
         // case VtkDataTypes.UNSIGNED_INT:
@@ -535,8 +535,11 @@ function vtkOpenGLTexture(publicAPI, model) {
   };
 
   //----------------------------------------------------------------------------
-  publicAPI.getOpenGLDataType = (vtkScalarType) => {
-    model.openGLDataType = publicAPI.getDefaultDataType(vtkScalarType);
+  publicAPI.getOpenGLDataType = (vtkScalarType, useHalfFloatType = false) => {
+    model.openGLDataType = publicAPI.getDefaultDataType(
+      vtkScalarType,
+      useHalfFloatType
+    );
     return model.openGLDataType;
   };
 
@@ -612,15 +615,20 @@ function vtkOpenGLTexture(publicAPI, model) {
   };
 
   //----------------------------------------------------------------------------
-  function updateArrayDataType(dataType, data) {
+  function updateArrayDataType(dataType, data, depth = false) {
     const pixData = [];
+
+    let pixCount = model.width * model.height * model.components;
+    if (depth) {
+      pixCount *= model.depth;
+    }
+
     // if the opengl data type is float
     // then the data array must be float
     if (
       dataType !== VtkDataTypes.FLOAT &&
       model.openGLDataType === model.context.FLOAT
     ) {
-      const pixCount = model.width * model.height * model.components;
       for (let idx = 0; idx < data.length; idx++) {
         const newArray = new Float32Array(pixCount);
         for (let i = 0; i < pixCount; i++) {
@@ -636,7 +644,6 @@ function vtkOpenGLTexture(publicAPI, model) {
       dataType !== VtkDataTypes.UNSIGNED_CHAR &&
       model.openGLDataType === model.context.UNSIGNED_BYTE
     ) {
-      const pixCount = model.width * model.height * model.components;
       for (let idx = 0; idx < data.length; idx++) {
         const newArray = new Uint8Array(pixCount);
         for (let i = 0; i < pixCount; i++) {
@@ -648,9 +655,12 @@ function vtkOpenGLTexture(publicAPI, model) {
 
     // if the opengl data type is half float
     // then the data array must be u16
-    const halfFloat = model.context.getExtension('OES_texture_half_float');
-    if (halfFloat && model.openGLDataType === halfFloat.HALF_FLOAT_OES) {
-      const pixCount = model.width * model.height * model.components;
+    const halfFloatExt = model.context.getExtension('OES_texture_half_float');
+    const halfFloat = model.openGLRenderWindow.getWebgl2()
+      ? model.openGLDataType === model.context.HALF_FLOAT
+      : halfFloatExt && model.openGLDataType === halfFloatExt.HALF_FLOAT_OES;
+
+    if (halfFloat) {
       for (let idx = 0; idx < data.length; idx++) {
         const newArray = new Uint16Array(pixCount);
         for (let i = 0; i < pixCount; i++) {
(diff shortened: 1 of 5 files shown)


Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -613,15 +613,18 @@ function vtkOpenGLTexture(publicAPI, model) {
...
     // then the data array must be float
     if (
       dataType !== VtkDataTypes.FLOAT &&
       model.openGLDataType === model.context.FLOAT
     ) {
-      const pixCount = model.width * model.height * model.components;
+      let pixCount = model.width * model.height * model.components;
+      if (depth) {
+        pixCount *= model.depth;
+      }
```
martinken, line 627: We have three copies of the pixCount calc (my fault I suspect) Can we just pull that out and have one copy near the top of the method?

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -649,9 +655,20 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+    const halfFloatExt = model.context.getExtension('OES_texture_half_float');
+    const halfFloat = model.openGLRenderWindow.getWebgl2()
+      ? model.openGLDataType === model.context.HALF_FLOAT
+      : halfFloatExt && model.openGLDataType === halfFloatExt.HALF_FLOAT_OES;
+
+    if (
+      halfFloat &&
+      (dataType === VtkDataTypes.UNSIGNED_SHORT ||
+        dataType === VtkDataTypes.SHORT)
+    ) {
```
martinken, line 667: I think this will miss the case on OpenGL1 where we have FLOAT data but are stuck rendering it as half float. I think you can just check on if (halfFloat) here as if the opengl data type is half float then we know we need to be doing a conversion. Can you double check my logic here in case I'm missing something?

martinken:
This is awesome, just had two comments on the code to review then I think it is good to go.

sedghi:
> This is awesome, just had two comments on the code to review then I think it is good to go.

Thanks @martinken for reviewing this. I will apply your comments today. 

I have one question though, this PR clearly reduces the GPU memory consumption (task manager screenshots); I was wondering about the CPU memory and how it is not getting that dramatic decrease? I checked the arrays feeding to `texImage3D` and they are not Float32 anymore as expected, but is there any other place in vtk that need to be changed for the CPU memory to also take benefit of this?

Sources/Rendering/OpenGL/Texture/index.js
sedghi: I changed them

Sources/Rendering/OpenGL/Texture/index.js
sedghi: Yes, you are right, it will miss it. I also think `if (halfFLoat)` covers it all

martinken:
I have no clue on memory issues in JS. It looks like you are saving about 200 MB on both CPU and GPU which seems reasonable to me and definitely a win.

martinken:
+1

martinken:
If you want me to merge just let me know. LGTM

sedghi:
@martinken Thanks again for your review, yes please merge it

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -483,10 +483,10 @@ function vtkOpenGLTexture(publicAPI, model) {
...
         case VtkDataTypes.UNSIGNED_CHAR:
           return model.context.UNSIGNED_BYTE;
-        // case VtkDataTypes.SHORT:
-        //   return model.context.SHORT;
-        // case VtkDataTypes.UNSIGNED_SHORT:
-        //   return model.context.UNSIGNED_SHORT;
+        case VtkDataTypes.SHORT:
+          return model.context.HALF_FLOAT;
+        case VtkDataTypes.UNSIGNED_SHORT:
+          return model.context.HALF_FLOAT;
```
martinken, line 489: OK I spoke too soon. We need to think about this section right here. The issue is that float can fully represent short and unsigned short while half float cannot. 

Short has 15 bits of precision versus about 11 in half float. Unsigned short has 16 bits versus half floats 11. So while we can stuff short and short and unsigned short into half float in webgl2 we are losing precision. So we are saving memory at the cost of precision. So it is a trade off. Any thoughts on this?

sedghi: Thanks for pointing this out. I dug deeper into the half float standard, and according to [Wikipedia](https://en.wikipedia.org/wiki/Half-precision_floating-point_format) section "Precision limitations on integer values" :

- Integers between 0 and 2048 can be exactly represented (and also between −2048 and 0)
- Integers between 2048 and 4096 round to a multiple of 2 (even number)
- Integers between 4096 and 8192 round to a multiple of 4
- Integers between 8192 and 16384 round to a multiple of 8
- Integers between 16384 and 32768 round to a multiple of 16
- Integers between 32768 and 65519 round to a multiple of 32

Since at the time we convert the buffer, we should know the `min` and `max`, one approach can be to convert by default if the `min` and `max` are in the range that can be exactly represented (-2048 to 2048), and be Optional (by flag, default false) if otherwise?

sedghi: Actually there is another extension [EXT_texture_norm16](https://developer.mozilla.org/en-US/docs/Web/API/EXT_texture_norm16), but it is totally unclear to me if it makes the texture filterable. Any comment on this extension to enable non-Float32 3D textures that are filterable? The detail of this extension is [here](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_norm16.txt), check Table 8.13

martinken: Per that table it would be filterable (TF == Texture Filterable) so it would work. However I'm not sure what systems have that extension. I checked on Chrome/Firefox on windows and neither reported it

I like the idea of both checking the data range and having a setting. The setting could be on the volume mapper and named preferSizeOverAccuracy or something like that.

sedghi: Sounds good, I'll add the `preferSizeOverAccuracy` for this PR.

Regarding the `EXT_texture_norm16`, I have it on my Chrome on Mac. I think it is GPU dependant, I checked mine on [WebGL report website](https://webglreport.com/?v=2)
Currently I'm working on a [demo](https://jsfiddle.net/sedghi/hxLegk5y/45/) to utilize `EXT_texture_norm16`, but I haven't had any luck so far, which is why I moved to half_float. However, I've begun a conversation with one of the developers of this extension, and I'll keep you updated, and hopefully create another PR later when the problem is resolved with `EXT_texture_norm16`.

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1044,17 +1051,86 @@ function vtkOpenGLTexture(publicAPI, model) {
     return true;
   };
 
+  function computeScaleOffsets(numComps, numPixelsIn, data) {
```
sedghi, line 1054: This function is moved earlier only

sedghi:
@martinken I have edited the PR to include the `preferSizeOverAccuracy` flag by checking the min and max of the data and a fallback to `Float32`.
Everything seems to work fine, except the `additiveIntensity` blend mode. Is it supposed to work for halfFloat too?

#### Update on `EXT_texture_norm16` extension
My demo is now working fine using `EXT_texture_norm16` here (https://jsfiddle.net/sedghi/j5vnb7Lc/2/)

On MacOS, however, there is a problem in Chromium that prevents correct rendering. I reported it and the bug is confirmed, and seems to be resolved soon (https://bugs.chromium.org/p/chromium/issues/detail?id=1246379 ).

martinken:
@sedghi Let me know when you want me to take another review of it. The basic gist here all looks good, I will take another careful look at it once you think it is ready for it.

sedghi:
@martinken This PR is ready for your review. Please go ahead. 
I'm opening another PR for the `EXT_texture_norm16` right now.

Sources/Rendering/OpenGL/Texture/index.js
martinken: I believe the current code will cause all 2D textures created on WebGL2 to use half float event if their range exceeds -2048 to 2048. So I think we need to pass in an argument about using half float for short/ushort into getDefaultDataType maybe? Even if we just default to always false for 2D textures.

sedghi: @martinken Thanks for catching this, I fixed the logic. Do you have any comment for the `additiveBlend` in halfFloat which is not rendering the same thing as float?

martinken:
LGTM, @sedghi can you rebase on master and squash the commits? Then we should be good to merge.

martinken:
I'll take a look at the additiveIntensity

martinken:
additive looks OK to me, I ran the volumemapperblendmodes example. Is there something you are seeing that doesn't look correct?

martinken:
Oddly enough we are counting on webgl2 supporting linear float textures and yet looking at the spec I don't think it is required. SO we are just getting lucky I guess.

sedghi:
@martinken I did rebase and squash. Are you running the `VolumemapperBlendModes` with the `mapper.setPreferSizeOverAccuracy(true)`? The `headsq` has maximum of 3926 so the flag is needed.

The problem I can see for the additiveBlend is that it renders empty compared to without the `setPreferSizeOverAccuracy` flag

martinken:
Ahh OK let me give that a try

martinken:
I just happened to be in that code yesterday on the WebGPU side so the timing was fortunate :-)

martinken:
Please give my change a look/test and see if it works for you/makes sense.

sedghi:
@martinken Thank you for correcting the additive blend and providing such a thorough explanation of the root cause of the issue; I really appreciate it.
It worked well when I tried it. I also rebased on top of the master branch.

Sources/Rendering/OpenGL/Texture/index.js
sedghi: Since is no native support for 16 bit texture in webGL as far as I know, the options are 

1) use float32 (memory inefficient)

2) use norm16Texture (not widely available)

- https://github.com/Kitware/vtk-js/pull/2058

3) use half float (at the cost of inaccuracies in rendering, but almost widely available)

In WebGPU as I understand 16 bit texture is natively supported so vtk.js should not do that in webGPU. Otherwise, I can't think of any other solution.

sedghi: I guess you can put it under a flag to do half float when memory is the end goal, but i thought I added that here 

fix(imageMapper): not exact half float should be under a flag #2786

bruyeret: Int16 and Uint16 textures are supported by WebGL so I didn't understand why all this was needed
But I found out that even if they are supported but they are not filterable, which means that we can only use nearest neighbor interpolation

Do you think that using software interpolation would be a good idea?
It would cost 8 unfiltered texture sampling and 7 call to `mix` but we don't have to do expensive operations to convert the volume, or loose memory

sedghi: They are not filterable correct. I mean I love to see software filtering on 16bits, but do you think doing 8 unfiltered texture sampling and 7 call to mix would be cheap? It might be, but if the fps is like 15 after I think we should let the user decide with some flags. But I love your idea

bruyeret: I made [this small shader](https://www.shadertoy.com/view/X3SXRw) in ShaderToy to benchmark trilinear sampling in the fragment shader
It looks like trilinear interpolation in software is around 8.5 times slower than in hardware, but I don't know how many samples are made per pixel in the volume mapper
I suppose that it depends on the options, for example if the shadows are enabled

PR #2047, opened 2021-08-31
fix(View): Fix center of rotation when used with world up vector setting

tamaynard:
Fixes an error when the world up vector setting is used with a non-origin center of rotation as mentioned [here](https://github.com/Kitware/vtk-js/pull/2043#issuecomment-898999801)


jourdain:
Why remove the `useFocalPointAsCenterOfRotation` option?

tamaynard:
It's not removed. The logic to handle that case actually applies to any case using the world up vector option and a non-origin center of rotation, so there's no need to check for that case specifically. The code for the `useFocalPointAsCenterOfRotation` option still overrides the preset center of rotation as before. This just fixes the case where `useFocalPointAsCenterOfRotation` is false and the center of rotation is not the origin.

Issue #2051, opened 2021-09-06
ImageMarchingCubes changes polydata position


gaston-perret:
### High-level description

Hello. I'm having trouble using the filter ImageMarchingCubes and I might be dealing with a bug.

I'm using a VTI file as input for the filter ImageMarchingCubes and getting a polydata as intended.
The problem is that the polydata does not keep the XYZ position of the points that the imagedata had.

### Detailed behavior

We think this happens because the vti file Extent does not have the first components coordinates starting at 0.

After we correct the extent 0, 2 and 4 elements and we adjust the origin we get the correct position for the polydata. But this generates other problems.

I add an image showing the position difference:
![image](https://user-images.githubusercontent.com/66233297/132249771-990faae4-6b34-4bb8-9064-6fdb1131d459.png)

I will also attach the vti file we use as input for the IMC filter.
[mask2.zip](https://github.com/Kitware/vtk-js/files/7117341/mask2.zip)

To correct the position we change the extent and origin as follows:

`const mask = vtkImageData.newInstance()
      mask.shallowCopy(newValue.vti)

      const extent = mask.getExtent()
      const origin = mask.getOrigin()
      const spacing = mask.getSpacing()

      const newExtent = []
      const newOrigin = []

      newExtent[0] = 0
      newExtent[1] = extent[1] - extent[0]

      newExtent[2] = 0
      newExtent[3] = extent[3] - extent[2]

      newExtent[4] = 0
      newExtent[5] = extent[5] - extent[4]

      newOrigin[0] = origin[0] + spacing[0] * extent[0]
      newOrigin[1] = origin[1] + spacing[1] * extent[2]
      newOrigin[2] = origin[2] + spacing[2] * extent[4]

      mask.setExtent(newExtent)
      mask.setOrigin(newOrigin)`

### Expected behavior

Polydata output result of applying IMC filter to imagedata should respect the points XYZ coordinates
### Environment
- **vtk.js**:  19.0.4
- **OS**:  Windows 10 / Ubuntu
- **Browser**:  Chrome / Firefox

jourdain:
Is your ImageData oriented? How did you generate the polydata? Was it with vtk.js or VTK?

gaston-perret:
> 
> 
> Is your ImageData oriented? How did you generate the polydata? Was it with vtk.js or VTK?

I'm sorry but I don't know how to orient an ImageData. The polydata is the output result of the ImageMarchingCube filter and we are using vtk js.
[...]

PR #2052, opened 2021-09-09
fix(Filters): fix marching cubes position/extent error

martinken:
The marching cubes filter was not handling images
with an extent that started with anything other than 0,0,0.

This MR fixes that issue https://github.com/Kitware/vtk-js/issues/2051


agirault:
@gaston-perret

PR #2058, opened 2021-09-16
feat(Rendering): Add support for 16bit normalized 3D texture formats

sedghi:
### Context
This PR adds support for rendering 16bit normalized textures using the `EXT_texture_norm16` WebGl extension. 

~~Because this a new extension, I began by creating a small demo here (https://jsfiddle.net/sedghi/j5vnb7Lc/2/) to showcase demonstrate the functionality. This demo seem to work fine (rendering a gradient) on 5 windows machines I tested but fails on Mac. There seems to be a bug in Chromium that prevents correct rendering. I reported it and the bug is confirmed, and seems to be resolved soon (https://bugs.chromium.org/p/chromium/issues/detail?id=1246379). To run Chrome on Mac to test this PR you need to run `/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --user-data-dir=/tmp/s1 --use-cmd-decoder=passthrough~~

### Changes
The default rendering behaviour remains unchanged; however, a new flag, `useExperimentalNorm16Texture`, is introduced to allow 16bit texture rendering if the extension is available.

### Data
The data I use is a 600 Slice CT to clearly show the benefit of 16 bit textures, you can reference it in the reader URL by  `https://s3.amazonaws.com/vtk.js-polyline/600CT.vti`

### Results
Reduced load time (from 5.5 seconds to 4.1 seconds in the VolumeBlends example), as well as memory footprint, without compromising accuracy.

Data: Whole body CT `Int16Array`

Case 1: `Float32` Rendering

image

image

Case 2: `Int16` rendering with `EXT_texture_norm16`

image

image

### Remaining bugs before merge

The Radon blend mode does not work at all, I reported it here. I don't think it is related to this PR. Basically as I saw the code the `VolumeMapperBlendModes` is using `setPreferSizeOverAccuracy(true)` and when I set that to `false` the Radon transform doesn't work

image

I guess you guys just followed the example and never tried the default scenario which is `false` for `preferSizeOverAccuracy`?

CC: @swederik 

### Testing

  - **vtk.js**: latest master
  - **OS**:  MacOS 10.15.7 
  - **Browser**:  Chrome 107.0.5304.121


floryst:
Following up on this PR again.
- Ping @martinken on the rendering aspect
- There has been some changes going into both the volume and image mapper (including some pending PRs), so this will need to be rebased.

sedghi:
@floryst @martinken Hi guys, can we please revisit this and the possibility of it to be merged?

sedghi:
@floryst I just rebased, thanks

floryst:
Please check the github annotations. There are a few variable names that aren't defined (it looks like they should be coming from the model) and some style lints.

Have you also tested this with 2D textures? I think this would be useful for 2D as well.

I can't say much to the rendering differences, other than imprecision with half float conversion.

sedghi:
Sorry for the dumb question but what is `github annotations` ? 

Haven't tested it with 2D textures, but theoretically they are not as much of a memory problem, right?

floryst:
Ah, you can just scroll through the [files tab](https://github.com/Kitware/vtk-js/pull/2058/files#diff-18c5d83bebc14325d84e520000b95e06dadffa088fe27eef138af64744171b89R1236) and see there are linter errors.

As for 2D textures, it would still be useful, as it will avoid the half-float conversion. While not as large as 3D textures, I see no reason to not support them in the 2D scenario as well, as there is a perf gain to be had there anyhow.

floryst:
Just following up on this to see if there's any progress in addressing the PR checks. I think we're okay with merging in the experimental 3D support, and creating an issue for the 2D case.

sedghi:
Sure, I can hopefully look into it this week. Thanks

sedghi:
@floryst I fixed the PR and the description, there are bugs left that I discussed in the description (one of them is not related to this PR, it is for Radon Transform)

Thanks!

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1226,18 +1231,34 @@ function vtkOpenGLTexture(publicAPI, model) {
...
 
-    const useHalfFloat = checkUseHalfFloat(
-      dataType,
-      computedOffset,
-      computedScale,
-      preferSizeOverAccuracy
-    );
+    // Check whether 16bit texture extension is available
+    model.useNorm16 =
+      useExperimentalNorm16Texture &&
```
floryst, line 1236: Can we drop `useExperimentalNorm16Texture` and just use the extension presence as the toggle condition? It may be useful to only cache this runtime `useNorm16` flag once as well.

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -525,8 +525,14 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
 
-  publicAPI.getDefaultTextureInternalFormat = (vtktype, numComps, useFloat) => {
+  publicAPI.getDefaultTextureInternalFormat = (
+    vtktype,
+    numComps,
+    useFloat = false,
+    useNorm16 = false
+  ) => {
     if (model.webgl2) {
+      const ext = model.context.getExtension('EXT_texture_norm16');
```
sankhesh, line 535: Recommend just use 16bit textures if the extension is available i.e. `if (ext)`

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -540,6 +546,31 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
             default:
               return model.context.RGBA8;
           }
+        case useNorm16 && VtkDataTypes.UNSIGNED_SHORT:
+          switch (numComps) {
+            case 1:
+              return ext.R16_EXT;
```
sankhesh, line 552: potential segfault if `ext == null` i.e. when the extension is not available

sankhesh (changes requested):

sedghi:
@floryst @sankhesh I think I applied your comments. Any idea about the MIP inconsistensies?

sankhesh:
> Any idea about the MIP inconsistencies?

Nope, not sure. Are the scale and offset getting sent right?

sedghi:
@floryst @sankhesh I fixed the MIP inconsistency it was scaling issue

Sources/Rendering/OpenGL/RenderWindow/index.d.ts
```diff
@@ -270,7 +270,7 @@ export interface vtkOpenGLRenderWindow extends vtkOpenGLRenderWindowBase {
 	 * @param {Number} numComps 
 	 * @param {Boolean} useFloat 
 	 */
-	getDefaultTextureInternalFormat(vtktype: VtkDataTypes, numComps: number, useFloat: boolean): void;
+	getDefaultTextureInternalFormat(vtktype: VtkDataTypes, numComps: number, oglNorm16Ext?:unknown): void;
```
floryst, line 273: Minor nit: space `:unknown` to `: unknown` (space). We don't have the linter turned on yet for d.ts files (and we should).

floryst (commented):
LGTM overall!

Sources/Rendering/OpenGL/RenderWindow/index.d.ts
sedghi: done

sankhesh (approved):

PR #2068, opened 2021-09-23
docs: build doc fix

ebrahimebrahim:
Update build docs to mention running `npm install` to install dependencies.

We might want to hold off on merging this until `npm run dev` and `npm run build` both work as build commands. They don't work for me at the moment.

I added a paragraph describing what building means in this case, since that was confusing for me as someone new to this type of development:

> When you use vtk.js as a dependency in your project (e.g. [like this](https://kitware.github.io/vtk-js/docs/vtk_vanilla.html)) you are pulling a package that was put together by the continuous integration. When developing vtk.js, you will have to put that package together yourself, and we call that "building."

Please let me know if I made any mistakes or could explain something more clearly.

Finally, there is a commit updating the install command on [this landing page](https://kitware.github.io/vtk-js/index.html).

### DIFF
--- a/Documentation/content/docs/develop_build.md
+++ b/Documentation/content/docs/develop_build.md
@@ -7,9 +7,11 @@ The library can be built with webpack automatically. Webpack can either gather a
 
 ## Building vtk.js
 
-In order to build the library you can run `npm run dev` for quick development usage or `npm run build` for production usage.
+Install dependencies by running `npm install`.
 
-Either of these commands will generate a `dist/vtk.js` file that can then be used as an external script.
+In order to build the library you can run `npm run dev:esm` for quick development usage or `npm run build:esm` for production usage.
+
+These commands will generate an ES module in `dist/esm`.
 
 ## Building the website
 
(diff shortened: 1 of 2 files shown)


Documentation/content/docs/develop_build.md
```diff
@@ -7,9 +7,11 @@ The library can be built with webpack automatically. Webpack can either gather a
 
 ## Building vtk.js
 
+Install dependencies by running `npm install`.
+
 In order to build the library you can run `npm run dev` for quick development usage or `npm run build` for production usage.
 
-Either of these commands will generate a `dist/vtk.js` file that can then be used as an external script.
+These commands will generate an ES module in `dist/esm`.
```
jourdain, line 14: @floryst is that true ? Or do we have a post-script for generating the ESM?

jourdain:
Looks reasonable.

Documentation/content/docs/develop_build.md
floryst: These currently aren't correct, but my plan is to make it as such. I think we need to have some changes to the build, namely integrating the post-script into the build step. The post-script simply makes it a proper package, among other fix-ups.

@ebrahimebrahim: for now, I would suggest the following:
- Change `npm run dev` to `npm run dev:esm` and `npm run build` to `npm run build:esm`. Those two will generate the ES module.

floryst:
@ebrahimebrahim Changes overall look good! I've left a comment for some minor changes.

Documentation/content/docs/develop_build.md
ebrahimebrahim: Great, I just made that change. Currently `npm run build:esm` works but there is no script for `dev:esm`.

PR #2069, opened 2021-09-23
feat(WebGPU): support more data types for volume rendering

martinken:
Use half float to support types other than uint8

Fix a matrix computation that impacts imageData that
has a direction in its transformations.

### DIFF
--- a/Sources/Common/Core/HalfFloat/index.d.ts
+++ b/Sources/Common/Core/HalfFloat/index.d.ts
@@ -1,0 +1,24 @@
+/**
+ * Helper class to convert to/from half float
+ */
+
+/**
+ * Convert a number to half float representation
+ * @param {number} input
+ * @return encoded half float number (16 bits)
+ */
+export function toHalf(input: number): number;
+
+/**
+ * Convert a half float representation to a number
+ * @param {number} input
+ * @return decoded half float number
+ */
+export function fromHalf(input: number): number;
+
+export declare const HalfFloat: {
+	toHalf: typeof toHalf;
+	fromHalf: typeof fromHalf;
+}
+
+export default HalfFloat;
--- a/Sources/Common/Core/HalfFloat/index.js
+++ b/Sources/Common/Core/HalfFloat/index.js
@@ -1,0 +1,68 @@
+/* eslint-disable no-bitwise */
+
+const floatView = new Float32Array(1);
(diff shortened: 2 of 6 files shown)


martinken:
@jourdain PTAL mainly at the typescript stuff

Sources/Common/Core/HalfFloat/index.d.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Helper class to convert to/from half float
+ */
+
+/**
+ * Convert a Base64 string to an ArrayBuffer.
```
jourdain, line 6: Is that the real description?

Sources/Common/Core/HalfFloat/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @param {number} input
+ * @return encoded half float number (16 bits)
+ */
+export default function toHalf(input: number): number;
+
+/**
+ * Convert a Base64 string to an ArrayBuffer.
+ * @param {number} input
```
jourdain, line 14: Since in/out are number, I don't see where the base64 string is coming from.

Sources/Common/Core/HalfFloat/index.d.ts
martinken: See that is why I have you review it :-) But really I was worried about the types I chose to use.

jourdain: number is fine, but now, I just noticed that you export both function as default. 

That is wrong. You should export your function as-is (no default keyword).

Then add at the bottom 

```
export default {
  toHalf,
  fromHalf,
};
```

jourdain: By doing that also in your JS you should export your function as well. That would allow someone to import them individually to match the TS description.

martinken: or maybe
```
export declare const HalfFloat: {
	toHalf: typeof toHalf;
	fromHalf: typeof fromHalf;
}

export default HalfFloat;
```

martinken: Does that look OK?

jourdain: yes it is.

PR #2073, opened 2021-09-24
feat(ImageMapper): When XYZ slicing, display intermediary slices

boucaud:
### PR and Code Checklist

### Context

This fixes https://discourse.vtk.org/t/vtkimageslice-xyz-slicing-and-image-position/6673

And discussion at the 9/23 dev meeting.
### Changes

Change getSliceAtPosition to not round slice values.

Change the vtkOpenGLImageMapper to round slice values before accessing scalars.

### Results

Slice positions are no longer rounded when slicing with X/Y/Z.
### Testing

  - **vtk.js**: 
  - **OS**: Linux
  - **Browser**: chromium 93.0.4577.82


boucaud:
@agirault @finetjul  WDYT ?

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -46,24 +48,24 @@ function vtkImageMapper(publicAPI, model) {
...
       case SlicingMode.J:
         slice = vtkMath.clampValue(ijk[1], ex[2], ex[3]);
-        slice = Math.round(slice);
         break;
       case SlicingMode.K:
         slice = vtkMath.clampValue(ijk[2], ex[4], ex[5]);
-        slice = Math.round(slice);
         break;
       default:
         return 0;
```
finetjul, line 63: return `null` (or `vtkErrorMacro`) instead of 0 which is a valid slice?

finetjul:
LGTM

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -46,24 +48,24 @@ function vtkImageMapper(publicAPI, model) {
...
     }
+
+    if (ijkMode === model.slicingMode) {
+      slice = Math.round(slice);
+    }
```
agirault, line 68: Why redo the rounding here if the goal is to do continuous index?

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -22,19 +22,21 @@ function vtkImageMapper(publicAPI, model) {
...
           break;
-        case SlicingMode.Y:
+        case SlicingMode.J:
           pos3 = [(bds[1] + bds[0]) / 2, pos, (bds[5] + bds[4]) / 2];
           break;
-        case SlicingMode.Z:
+        case SlicingMode.K:
           pos3 = [(bds[1] + bds[0]) / 2, (bds[3] + bds[2]) / 2, pos];
           break;
         default:
```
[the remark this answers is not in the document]
agirault: Example: you have no guarantee that your K axis will be along the Z axis, which is needed to compute `pos3` properly here.

Sources/Rendering/Core/ImageMapper/test/testSlicingModePosition.js
```diff
@@ -0,0 +1,42 @@
...
+
+import vtkImageMapper from 'vtk.js/Sources/Rendering/Core/ImageMapper';
+import { SlicingMode } from 'vtk.js/Sources/Rendering/Core/ImageMapper/Constants';
+import vtkImageSlice from 'vtk.js/Sources/Rendering/Core/ImageSlice';
+import vtkRTAnalyticSource from 'vtk.js/Sources/Filters/Sources/RTAnalyticSource';
+
+test('Test slice position differences between XYZ and IJK modes', (t) => {
+  const source = vtkRTAnalyticSource.newInstance({
+    wholeExtent: [0, 10, 0, 10, 0, 10],
+  });
```
agirault, line 11: Test with a spacing != 1 and an orientation that's not the identity if you want to test differences between XYZ and IJK.

Sources/Rendering/Core/ImageMapper/index.d.ts
```diff
@@ -34,10 +34,11 @@ interface IImageMapperInitialValues {
...
-	/**
-	 *
-	 * @param {Number} pos
-	 */
+  /**
+   * Get the slice index from pos.
+   * When slicing with XYZ, returns intermediary slice positions instead of IJK indices.
+   * @param {number[] | number} [pos] World point or XYZ slice value
+   */
 	getSliceAtPosition(pos: number): number;
```
agirault, line 43: `getSliceAtPosition` is meant to convert a world position to a slice index. The current slicing mode is only used when passing a single number, so it goes from 1d to 3d based on the current axis. This method should not be used to pass coordinates in index space.

**A couple of suggestions:**
1. Improve current doc to illustrate `getSliceAtPosition` 
2. Consider renaming? (`getSliceIndexAtWorldPosition`)
3. Consider a separate method to get the continuous index so we don't break too many assumptions in apps that depend on `getSliceAtPosition` ? Similar to ITK (ex: `getContinuousSliceIndexAtWorldPosition` ... ugh). No strong feelings on this atm.

Sources/Rendering/Core/ImageMapper/index.js
agirault: As discussed:
- this rounding only occurred if `slicingMode` was I, J or K
- however, this method should really only be called when in X, Y or Z (it does world to index given one single XYZ component, outputting one single IJK component). yes, the name of the method is terrible in hindsight
- either way, we'd want IJK to be able to be continuous also, and just fix the mapper to round the index before accessing the scalar data, so no rounding needed here anyway.

Sources/Rendering/Core/ImageMapper/index.js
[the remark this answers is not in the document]
agirault: As discussed: reverting those changes. Ideally, `getSliceAtPosition` should ensure the slicing mode is X, Y or Z (won't work for IJK). A better name might also be `getContinuousSliceIndexAtWorldPosition`.

Sources/Rendering/Core/ImageMapper/test/testSlicingModePosition.js
```diff
@@ -0,0 +1,42 @@
...
+  mapper.setSlicingMode(SlicingMode.K);
+
+  t.equal(3, mapper.getSliceAtPosition(3));
+  t.equal(3, mapper.getSliceAtPosition(3.1));
+
+  t.equal(5, mapper.getSliceAtPosition([0, 0, 5]));
+  t.equal(5, mapper.getSliceAtPosition([0, 0, 5.1]));
+
+  t.equal(10, mapper.getSliceAtPosition(15));
+  t.equal(10, mapper.getSliceAtPosition(15.1));
```
agirault, line 39: `getSliceAtPosition` should only be used with XYZ slicing mode.

agirault:
@jadh4v could you review?

jadh4v:
LGTM

finetjul:
@agirault can you please remove the "changes requested" ?

agirault (approved):
My bad. Thanks for the heads up @finetjul, and for the changes @boucaud !

boucaud:
@floryst I am having issues figuring out why CI is failing for this new test. The images provided as output seem identical, and the test passes when I test locally (both with and without https://github.com/Kitware/vtk-js/pull/2098).

Do you have an idea what could be causing this ?

floryst:
Not 100% sure yet, but it stems from using `run-tests-in-docker.sh`. The output looks like this:

image

I wonder if it's an issue with the opengl container that runs the tests.

boucaud:
> Not 100% sure yet, but it stems from using `run-tests-in-docker.sh`. The output looks like this:
> 
> image
> 
> I wonder if it's an issue with the opengl container that runs the tests.

I fixed the issue by explicitly setting the ColorTransferFunction to do what I needed instead of relying on window/level.

I'm still not sure why the previous version would produce different results in different testing environments.

boucaud:
@finetjul I believe this can be merged now

floryst:
Thanks, that resolved the test. I may need to look into why the original test was failing at a later point.

PR #2081, opened 2021-09-29
feat(WebGPU): add in min/max/average/additive volume blend modes

martinken:
Add support for those 4 blend modes into the volume Rendering
for WebGPU. Also implement ipScalarRange filtering but leave
it commented for now as there and some WebGL changes on another
MR that needs to land first.

### DIFF
--- a/Sources/Rendering/Core/VolumeMapper/Constants.js
+++ b/Sources/Rendering/Core/VolumeMapper/Constants.js
@@ -6,6 +6,13 @@ export const BlendMode = {
   ADDITIVE_INTENSITY_BLEND: 4,
 };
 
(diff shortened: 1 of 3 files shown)


martinken:
@jourdain does that CI failure mean something to you? I'm not seeing anything other than a timeout which I'm not sure is related to this MR.

jourdain:
The CI does not find an image (docker) that Matt created. Since other tests pass, I think this is fine. We'll have to see if that image is gone for good or not. And address the core issue. (@thewtex @floryst )

martinken:
I merged this, but yes, something is off on the CI

PR #2084, opened 2021-09-30
fix(XMLReader): use xml library over builtins

floryst:
### PR and Code Checklist

### Context
Should allow usage of XMLReaders in web workers

### Changes

- [x] remove usage of ActiveXObject and DOMParser

### Results
- [x] webworker usage of xml readers should work

### DIFF
--- a/Sources/IO/Core/BinaryHelper/index.js
+++ b/Sources/IO/Core/BinaryHelper/index.js
@@ -5,7 +5,7 @@
  * expect proper Unicode or any other encoding.
  */
 function arrayBufferToString(arrayBuffer) {
-  if ('TextDecoder' in window) {
+  if (typeof 'TextDecoder' !== 'undefined') {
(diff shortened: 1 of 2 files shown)


finetjul:
If not built-in, does it mean it will be slower ? if so, should it be optional ?

floryst:
Right now the goal of this PR is to examine the usage of readers in webworkers. DOMParser is not accessible in the webworker scope, and ActiveXObject is IE specific (I don't even know if that proprietary extension made it into Edge).

> If not built-in, does it mean it will be slower ? if so, should it be optional ?

Slower, possibly. Though I don't imagine it being significantly slower for typical usage.

floryst:
I've tested XMLPolyDataReader inside a webworker, and it appears to be working just fine after these changes.

jourdain (approved):
LGTM

finetjul:
LGTM minus the warning

PR #2088, opened 2021-10-04
fix(WebGL): zvalue hardware selection of translucent actors

martinken:
The zbuffer values were not being computed on translucent
actors in WebGL. So when zbuffer values were needed to compute
world coordinates on translucent actors they were incorrect.

Fixes issue https://github.com/Kitware/react-vtk-js/issues/37


martinken:
@jourdain PTAL

jourdain:
So would the z value be correct for actors with opacity<1?

martinken:
With this MR, yes. Translucent actors were getting selected, just without a correct zbuffer.

Issue #1553, opened 2020-08-11
Inconsistent pixel location from imagedata vs polydata


a3VonG:
Hi!
I was playing around with vtk.js and bumped upon a problem, I have tried to create a toy example where the problem becomes clear. I generated a checkerboard in imageData and want to draw a contour around a pixel using polyData. The imageData is initialized with the default coordinate system etc so the mapping between the voxel and world coordinates should be one-on-one. A square (polyData) with edge length 1 is generated starting from the pixel. You can find the code [here](https://gist.github.com/a3VonG/ff19e70a8661070713cb817168bf4138).

However, the results that I'm seeing are not in line with what I would expect: The lines around the pixel do not correspond with the pixels. 
![image](https://user-images.githubusercontent.com/32267348/89905459-80ebf900-dbea-11ea-8b71-ef7024e77f56.png)

This issue persists across grid sizes and dimensions. In general it seems like n-1 polyData squares fit in n pixels. 

I could work around it for this specific example but the issue seems more fundamental as the one-on-one mapping is not correct. Is this intended behavior? If so what is the proper way to get my desired behavior? 

Thanks in advance!

jourdain:
What you see is what you are suppose to get. If you wanted the 2 to be aligned, you will create a grid of dimensions 11x11x1 and put your scalars on the cells or adjust your spacing.

Right now, you are telling the grid to have 10 points along (x, y) which mean the minimum coordinate is (0, 0) and the max coordinate is (9, 9). And that is exactly what we see here.

a3VonG:
Thanks for the answer I understand what you are saying. However, this makes the use of the indexToWorld function on imageData a bit strange no?  

The imageData grid has 10x10x1 voxels and it has a function indexToWorld. It makes sense to me that if I were to ask the world location of a voxel and I draw this location that this world coordinate is either consistently centered inside of this voxel or in a consistent corner (e.g. bottom left) of this voxel.  But this is not the case. 
To clarify, in the following image, I have drawn a little square in each world coordinate belonging to the voxel. 

![image](https://user-images.githubusercontent.com/32267348/90031666-c62d2b00-dcbd-11ea-95c3-1e455fce24b0.png)

What would the best way be to for example always draw in the bottom left corner of the pixel? Is there another function than indexToWorld? 

Thank you so much!

jourdain:
[...]

PR #2089, opened 2021-10-05
fix(rendering): Correction to domain boundary for images and volumes

jadh4v:
(#1936).

Previously, in a vtk representation of an image/volume, pixels or voxels
are considered to exist on the points of aregular grid. This results in
an error of half-pixel/voxel length in the bounds of the rendered object
(plane or box that acts as a domain of the image/volume). While openGL
texture mapping automatically adds the half-voxel offset, we  have
corrected the size of the rendered domain to correctly represent the
bounds of the image data.

Co-authored-by: Alireza <ar.sedghi@gmail.com> @sedghi 

### PR and Code Checklist

### Changes

### Testing

  - **vtk.js**:  master
  - **OS**:  Windows 10
  - **Browser**:  Chrome 99.0.4844.51

### 2D Image Comparison (generated using example created by @a3VonG in #1553)
 Before-Nearest | After-Nearest
:---------------:|:----------------:
|
Before-Linear | After-Linear
|

#### Before
<video src="https://user-images.githubusercontent.com/22624785/136265883-1fcc2fab-4e8e-4467-bf8a-a270e726da06.mp4"></video>

#### After
<video src="https://user-images.githubusercontent.com/22624785/136267042-ea4ab8a4-8090-486d-be82-85013128157c.mp4"></video>


Sources/Common/DataModel/ImageData/index.js
```diff
@@ -233,6 +234,21 @@ function vtkImageData(publicAPI, model) {
     return bounds;
   };
 
+  // Get array extents as float / spatial coordinates
+  // without scaling for voxel/pixel spacing.
```
finetjul, line 238: documentation should go in .d.ts file

Sources/Common/DataModel/ImageData/index.js
```diff
@@ -233,6 +234,21 @@ function vtkImageData(publicAPI, model) {
...
+    const ext = model.extent;
+    const spatialExt = [
+      ext[0] - 0.5,
+      ext[1] + 0.5,
+      ext[2] - 0.5,
+      ext[3] + 0.5,
+      ext[4] - 0.5,
+      ext[5] + 0.5,
+    ];
+    return spatialExt;
```
finetjul, line 249: could have been `return vtkBoundingBox.inflate(...model.extent, 0.5)` (inflate() would need to return something though...

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -906,7 +906,7 @@ vec2 computeRayDistances(vec3 rayDir, vec3 tdims)
 {
   vec2 dists = vec2(100.0*camFar, -1.0);
 
-  vec3 vSize = vSpacing*(tdims - 1.0);
+  vec3 vSize = vSpacing*(tdims);
```
finetjul, line 909: you should probably get rid of the parentheses

Sources/Common/DataModel/ImageData/index.js
```diff
@@ -233,6 +234,21 @@ function vtkImageData(publicAPI, model) {
     return bounds;
   };
 
+  // Get array extents as float / spatial coordinates
+  // without scaling for voxel/pixel spacing.
+  publicAPI.getSpatialExtent = () => {
```
agirault, line 239: @floryst @martinken how do you like the `getSpatialExtent` name? Does this make sense here?

agirault:
@jadh4v: we'll get to some tests at some point, but before that, could you add some screenshots or videos to compare before and after? Something similar to what @sedghi did in his issue [here](https://github.com/Kitware/vtk-js/issues/1936#issuecomment-857318425) and his PR [here](https://github.com/Kitware/vtk-js/pull/1993#issue-940143144)?

@martinken @sedghi you might be interested in reviewing this WIP.

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
agirault:
```suggestion
  vec3 vSize = vSpacing*tdims;
```

martinken:
Wait, did we decide that a pixel is located in the middle?

e.g. an image ten pixels of spacing 1mm with origin 0mm, 0mm would span which of these?

a) 0mm to 10mm

b) -0.5mm to 9.5mm

The d.ts file has out of date comments FWIW and needs some updates (bounds doesn't mention direction vectors etc). Not this MR's fault but became obvious when I looked at it.

I have more comments but they really depend on knowing what we decided or the origin of a pixel. I thought someone knew what dicom used and we were going with that.

sedghi:
> Wait, did we decide that a pixel is located in the middle?

@martinken 
Pixel location in DICOM is defined as the center of the pixel (voxel). https://dicom.innolitics.com/ciods/rt-dose/image-plane/00200032

jadh4v:
The space of the return type is really some for of regularized/normalized coordinate system which doesn't have origin, spacing or orientation information but is also a continuous space. Maybe getGridBounds() sounds better?

While the pixel is located in the middle for dicom conventions, is this also true of scientific data that may use the same data-structure? A continuous vector field will really have data located on grid points. I think, as previously mentioned, this is really the difference between cell-data vs point-data (independent of interpolation scheme).

pieper:
It's good that this is being discussed and documented.  In Slicer (and medical imaging in general, as in the DICOM case cited above) we definitely want the origin in the center of the voxel/pixel to describe physical measurements in space.  In index space (i.e. IJK) we assume 0,0 is a corner.

[This coordinate system page](https://www.slicer.org/wiki/Coordinate_systems) may help, and it would be great if there were a similar page for vtk-js.

Sources/Common/DataModel/ImageData/index.js
jadh4v: `getGridBounds` is another suggestion that might work.

jadh4v:
**TL;DR:** Current PR fixes the rendering pipeline to accurately draw the image / volume data.
Fixing the processing filters and pipeline to handle the "flat" half-voxel boundary seems to
be complicated with little benefits.
We will need two flags.
Flag-1: switches the meaning of origin.
Flag-2: switches between representation (1) and (2) since the rendering (bounds) will be different.
Tcoords need to be adjusted in representation (2) to remove the CLAMP_TO_EDGE effect.

## I see 3 different representations. 1a and 1b could be represented by a simple flag / switch.
### (1) Cell-centered image:
	Voxels represented as cell-data.
	Domain boundary inflated by 0.5 voxel beyond previous grid representation.
	CLAMP_TO_EDGE when using texture-on-plane.

	(1a) vtkImageData as a medical image
		origin: center of voxel [0,0,0]

	(1b) vtkImageData as a non-medical image/volume 
		origin: corner of voxel [0,0,0]

### (2) Node-centered image:
	voxels represented as point-data.
	origin: center of voxel[0,0,0] but voxel is a grid-point.
	Domain boundary follows grid boundary.
	Tcoords need to be adjusted by half-voxel to remove the CLAMP_TO_EDGE effect.

One question is whether the grid/domain of the dataset is always expected to cover the bounds of the field-data. Current PR/fix only handles the issue in the rendering pipeline. If you consider down-stream filters such as say a contour filter or a stream line tracer, it may process the cells / geometry of the representation, assuming that this covers the entire field domain. Which therefore requires (1a) to be cell-centered so that the boundary voxels, and consequently the data boundary, can be processed correctly. However, cell-data is considered piecewise constant(?), whereas filters often use linear interpolation, which would require construction of a local "cell" that has data at its corner points rather than at its center. The boundary voxels could be broken down into smaller cells that use trilinear interpolation, but this will make the entire dataset non-regular grid.

Nevertheless, the half-voxel extended boundary of the image in representation (1a and 1b) is "flat" and often will not be amicable to processing algorithms that expect non-degenerate conditions. Which leads me to think that currently, we can ignore the half-voxel boundary of the images for processing filters, while the current PR addresses the accurate rendering of the bounds. This means we leave representation 1a and 1b as point-data for now.

I think: **The only true difference between rep (1) and (2) is the 0.5 voxel padding at the boundary. A straight-foward short term solution is to handle it in the volume and image mappers for rendering, but ignore it in the processing filters.**

martinken:
I would say update marching cubes as needed. It will be pretty quick. If we have others it would be nice to fix those as well to use the new standard.

jadh4v:
Sounds good. But wouldn't we require another class of uniform grid for simulation data which doesn't pad the border by 0.5 voxels? The bounds of these would be different from vtkImageData.

martinken:
vtkStructuredGrid

Sources/Common/DataModel/ImageData/index.js
jadh4v: done

Sources/Common/DataModel/ImageData/index.js
jadh4v: I believe we can stick with `getSpatialExtent` for now. Can be easily updated in the future.

Sources/Rendering/OpenGL/VolumeMapper/test/testVolumeMapperBounds.js
```diff
@@ -0,0 +1,91 @@
...
+import vtkRenderer from 'vtk.js/Sources/Rendering/Core/Renderer';
+import vtkVolume from 'vtk.js/Sources/Rendering/Core/Volume';
+import vtkVolumeMapper from 'vtk.js/Sources/Rendering/Core/VolumeMapper';
+import vtkColorTransferFunction from 'vtk.js/Sources/Rendering/Core/ColorTransferFunction';
+import vtkRenderWindow from 'vtk.js/Sources/Rendering/Core/RenderWindow';
+import vtkRenderWindowInteractor from 'vtk.js/Sources/Rendering/Core/RenderWindowInteractor';
+
+import baseline1 from './testVolumeMapperBounds.png';
+
+test.onlyIfWebGL.only('Test Volume Mapper Bounds', (t) => {
```
jadh4v, line 16: @sedghi please remove the `.only`.

jadh4v: Scratch that, since its the last commit, I'll take care of it.

sedghi: ooops, sorry about that, thanks for taking care of it

sedghi:
Hey @jadh4v Is there any remaining work need to be done that I can help and make this PR be ready sooner? Thanks for all your efforts so far

Sources/Common/DataModel/ImageData/index.d.ts
```diff
@@ -68,6 +68,12 @@ export interface vtkImageData extends vtkDataSet {
 	 */
 	extentToBounds(ex: Extent): Bounds;
 
+	/**
+	 * Get array extents as floats without the scaling for voxel/pixel spacing.
+	 * @internal
+	 */
+	getSpatialExtent(): number[];
```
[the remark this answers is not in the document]
jadh4v: using `Bounds` would not be accurate either, so I am fine with `Extents`. Will do the change.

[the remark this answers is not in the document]
daker: `Bounds` and `Extent` types are basicaly the same, we just limit the number of element inside the array, also Typescript doesn't have any float ot integer types, they are all numbers.

Maybe we need to define Integer or Float types?

https://github.com/Kitware/vtk-js/blob/master/Sources/types.d.ts#L15-L16

Sources/Common/DataModel/ImageData/index.js
jadh4v: Made appropriate changes.

Sources/Rendering/Core/CellPicker/test/testCellPicker.js
```diff
@@ -64,9 +64,9 @@ test('Test vtkCellPicker image mapper', (t) => {
   const positions = picker.getPickedPositions();
   t.equal(positions.length, 1);
   const xyz = positions[0];
-  t.equal(xyz[0], 64.49344091067883);
-  t.equal(xyz[1], 75.65264519589407);
-  t.equal(xyz[2], 12);
+  t.equal(xyz[0], 64.33654020304944);
+  t.equal(xyz[1], 75.54505613923392);
+  t.equal(xyz[2], 12.000000000000004);
```
jadh4v, line 69: While we expect the `resetCamera` along with added padding to the bounds to affect the x and y values, I think the z value should remain `12`. The difference is clearly just a numerical truncation error. Maybe rather than simply updating the value, I should use `floatCompare` with some tolerance?

jadh4v: changed to vtkMath.areEquals test for a more stable comparison of floats.

Sources/Common/DataModel/ImageData/index.d.ts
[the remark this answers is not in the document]
jadh4v: changed to `Extent` as suggested.

jadh4v:
@floryst @sedghi FYI: PR is ready.

floryst:
@martinken please also take a look at this PR.

floryst:
Overall, LGTM. Two questions:
- @martinken these changes will need to be applied to the WebGPU code, correct?
- As mentioned before, [ImageData.indexToWorldBounds](https://github.com/Kitware/vtk-js/blob/master/Sources/Common/DataModel/ImageData/index.js#L315) (and its inverse) may need updating. Is that still relevant?

sedghi:
thanks @floryst. I'm not sure about the indexToWorldBounds and if changes is required, maybe @jadh4v knows

floryst:
> Depends on how its being used. I don't see any usage of `ImageData.indexToWorldBounds` in vtk-js or react-vtk-js. Its output isn't necessarily wrong now, but its interpretation might be different. Anyone using this function needs to decide whether they need a padding or not.

Can we update the docs for the indexToWorldBounds and worldToIndexBounds functions to indicate the appropriate interpretation?

jadh4v:
Ok, I will push an update. Will have to re-base with newer changes.

Sources/Common/DataModel/ImageData/index.d.ts
```diff
@@ -69,18 +69,31 @@ export interface vtkImageData extends vtkDataSet {
...
+	 *   (i_min - 0.5)*Spacing[0] + Origin[0], (i_max + 0.5)*Spacing[0] + Origin[0],
+	 *   (j_min - 0.5)*Spacing[1] + Origin[1], (j_max + 0.5)*Spacing[1] + Origin[1],
+	 *   (k_min - 0.5)*Spacing[2] + Origin[2], (k_max + 0.5)*Spacing[2] + Origin[2]
 	 * ];
 	 *  ```
+	 * Note that this includes a padding of 0.5 voxel spacing in each principal direction,
+	 * as well as any orientation of the image.
 	 * You can't directly set the bounds. First you need to decide how many
 	 * pixels across your image will be (i.e. what the extent should be), and
 	 * then you must find the origin and spacing that will produce the bounds
```
finetjul, line 99: in the below example (extent = [0,9,0,9,0,9]), it might be good to also give the expected bounds `[-0.5,9.5, -0.5,9.5,-0.5,9.5]`)

jadh4v: done

jadh4v:
> Let's update the docs for the indexToWorldBounds and worldToIndexBounds functions to indicate the appropriate interpretation. I think it will be helpful, in case someone does actually use it and gets confused.

Updated the documentation

martinken:
Yes, WebGPU will need to be changed but we can do that in a different MR

martinken:
WRT worldToIndex etc someone should look at them to make sure they are correct based on the new understanding of volumes, that's all.

sedghi:
@martinken Since `worldToIndex` is actually not integer indices, there will be no problem. However, getting the correct ijk index can be done with `round` which we discussed here https://github.com/Kitware/vtk-js/issues/2295#issuecomment-1064258481

floryst:
Everything here looks resolved. Thanks for all of the hard work!

PR #2093, opened 2021-10-06
feat(labelOutline): Add outlineOpacity to labelOutline

sedghi:
### PR and Code Checklist

### Context
This PR adds outline opacity for the labelOutline property of the volumeMapper

### Changes
outlineOpacity with default value of 1.0 is added to the volumeMapper for labelOutline usage

### Results
without `outlineOpacity`

with `outlineOpacity = 0.9` 

However, I observed a non-linear relationship between opacity levels. For example, any opacity below 0.8 is nearly transparent, and the difference between 0.9 and 0.99 is considerable.

I think the behaviour is as a result of [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl#L708) when alpha is subtracted by a `pow` function (which makes it non-linear). 

`tColor.a = 1.0 - pow(1.0 - tColor.a, jitter);`

Any thoughts on this?

### Testing
  - **vtk.js**: latest master
  - **OS**: MacOS
  - **Browser**:  Chrome 94.0.4606.71


Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -465,7 +466,7 @@ vec4 getColorForValue(vec4 tValue, vec3 posIS, vec3 tstep)
 
     // If I am on the border, I am displayed at full opacity
```
finetjul, line 467: technically this comment is not 100% correct

finetjul: Sorry, I meant that this comment should be updated because it used to be true "full opacity", but with your change, it may not be "full opacity", it depends on `outlineOpacity`

floryst:
What's the status on this PR? @martinken @sedghi

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
sedghi: @martinken Thanks for your explanation. So based on my understanding there is no way to have the outlineThickness (in the above images) act linearly to the opacity. Correct me please if I'm wrong.

sedghi:
@floryst @martinken Hey guys, can we revisit this for merge? thanks!

floryst:
LGTM. @finetjul if you want to do a last check

finetjul:
LGTM

finetjul (approved):

sedghi:
@floryst @finetjul You Github actions seem to be running twice?

floryst:
That's my bad. I ended up pushing a master->beta merge. It doesn't cause any actual issues, apart from email spam.

sedghi:
@floryst Oh no worries, just wanted to let you know

PR #2098, opened 2021-10-06
chore(testUtils): set pixelmatch threshold to 0

floryst:
### PR and Code Checklist

### Context

Usage of pixelmatch threshold is incorrect. The threshold parameter for compareImages is meant to be a percent threshold, while the pixelmatch threshold is a color difference metric threshold.

pixelmatch uses a YIQ difference metric for color comparison with a
threshold, while resemblejs does component-level delta thresholding.

### Changes
- [x] Update pixelmatch usage to use a threshold of 0

### Results
Image tests should now be reporting correct deltas.
### Testing

  - **vtk.js**: 19.7.2

### DIFF
--- a/Sources/Testing/testUtils.js
+++ b/Sources/Testing/testUtils.js
@@ -52,7 +52,7 @@ async function compareImages(
       {
         alpha: 0.5,
         includeAA: false,
-        threshold,
+        threshold: 0,
       }
     );
     const percentage = (mismatched / (width * height)) * 100;


floryst:
Note: resemblejs and pixelmatch still output differing values for the number of mismatched pixels. Some of it may be due to the difference in color delta computation between the two libraries. In any case, we should now be not reporting 0 delta for incorrect images.

PR #2107, opened 2021-10-07
docs: fix vtk textbook image

ebrahimebrahim:
The book image on this page is of the wrong book:
https://kitware.github.io/vtk-js/api/

It should be an image of the VTK textbook, not the old user's guide.


floryst:
LGTM, thanks for catching that.

PR #2113, opened 2021-10-12
Mpr volume mapper

DrewLazzeriKitware:
### PR and Code Checklist

### Changes

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 


Sources/Rendering/Core/SliceHelper/index.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  publicAPI.addMapper = (mapper) => {
+    mapper.addClippingPlane(model.clipPlane1);
+    mapper.addClippingPlane(model.clipPlane2);
+  };
+
+  publicAPI.setThickness = (thickness) => {
+    superClass.setThickness(thickness);
+    update();
```
jourdain, line 36: it looks like you want to call update onModified()?

agirault:
>  In particular, setUseLabelOutline() seems to have trouble with the way I've approached clipping, which you can see here:

I agree with @floryst that we don't need an example with two volumes, though the gif shows that there is a mismatch between work coordinates of those two datasets, so we should probably investigate the source. We should be able to render two aligned images with different orientations using two of those mappers with clipping.

DrewLazzeriKitware:
I've removed the VolumeOutline example, and captured the issue here: https://github.com/Kitware/vtk-js/issues/2169

The failed CI was a timeout. Is that likely just a transient issue unrelated to this change? 

I am adding types to this, then should be ready to merge pending CI and review.

DrewLazzeriKitware:
Ok types are added and CI passed this time, please review.

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
```diff
@@ -0,0 +1,82 @@
+import macro from 'vtk.js/Sources/macros';
+import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
+import vtkPlane from 'vtk.js/Sources/Common/DataModel/Plane';
+
+function vtkSliceHelper(publicAPI, model) {
+  model.classHierarchy.push('vtkSliceHelper');
+
+  model.clipPlane1 = vtkPlane.newInstance();
```
finetjul, line 8: if you do not expose it with set/get, why do you need to store it into `model` ? Does it need to be serialized ?

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
```diff
@@ -0,0 +1,82 @@
...
+    model.clipPlane2.modified();
+  }
+
+  const subscription = publicAPI.onModified(update);
+  publicAPI.delete = () => {
+    superClass.delete();
+    subscription.unsubscribe();
+  };
+
+  publicAPI.setMapper = (mapper) => {
```
finetjul, line 48: you might want to check that mapper does not already have those clipping planes...

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
```diff
@@ -0,0 +1,82 @@
...
+    mapper.addClippingPlane(model.clipPlane2);
+  };
+}
+
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {
+  thickness: 3,
```
finetjul, line 59: a thickness of 0 (or epsilon) seems to be a better default.

finetjul:
The commit should be flagged as BREAKING_CHANGE because it removes existing files/classes

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
DrewLazzeriKitware: Agreed

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
DrewLazzeriKitware: I store it in case it is subclassed, but I'll prefix the names with '_' so they don't get serialized.

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
DrewLazzeriKitware: Agreed

Sources/Rendering/Core/AbstractMapper/index.js
```diff
@@ -13,6 +13,9 @@ function vtkAbstractMapper(publicAPI, model) {
     if (plane.getClassName() !== 'vtkPlane') {
       return;
     }
```
DrewLazzeriKitware, line 15: @finetjul I'm checking to not add a clipping plane twice here now.

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
```diff
@@ -0,0 +1,95 @@
...
+  publicAPI.registerClipPlanesToMapper = (mapper) => {
+    if (!mapper.isA('vtkVolumeMapper')) {
+      return;
+    }
+    model._clipPlane1Index = mapper.getNumberOfClippingPlanes;
+    mapper.addClippingPlane(model._clipPlane1);
+    model._clipPlane2Index = mapper.getNumberOfClippingPlanes;
+    mapper.addClippingPlane(model._clipPlane2);
+  };
+
```
DrewLazzeriKitware, line 57: @jourdain Could you check this unregister approach?

Sources/Rendering/Core/AbstractMapper/index.js
DrewLazzeriKitware: (this was changed)

Sources/Rendering/Core/VolumeMapper/index.js
```diff
@@ -98,10 +99,6 @@ export function extend(publicAPI, model, initialValues = {}) {
 
   vtkAbstractMapper.extend(publicAPI, model, initialValues);
 
-  // Build VTK API
```
DrewLazzeriKitware, line 101: These are called already in vtkAbstractMapper, so I am removing them here.

DrewLazzeriKitware:
This is updated and ready for review again.

Sources/Rendering/Core/AbstractMapper/index.js
```diff
@@ -24,9 +29,10 @@ function vtkAbstractMapper(publicAPI, model) {
 
   publicAPI.removeClippingPlane = (i) => {
     if (i < 0 || i >= 6) {
```
finetjul, line 31: shouldn't you compare with `model.clippingPlanes.length` instead of 6 ?

Sources/Rendering/Core/AbstractMapper/index.js
```diff
@@ -5,15 +5,20 @@ import macro from 'vtk.js/Sources/macros';
...
   };
 
   publicAPI.addClippingPlane = (plane) => {
     if (plane.getClassName() !== 'vtkPlane') {
-      return;
+      return false;
+    }
+    if (!model.clippingPlanes.includes(plane)) {
+      model.clippingPlanes.push(plane);
+      return true;
```
finetjul, line 19: you should also call  `modified()`.

Sources/Rendering/Core/AbstractMapper/index.js
```diff
@@ -5,15 +5,20 @@ import macro from 'vtk.js/Sources/macros';
...
 function vtkAbstractMapper(publicAPI, model) {
+  model.classHierarchy.push('vtkAbstractMapper');
   publicAPI.update = () => {
     publicAPI.getInputData();
   };
 
   publicAPI.addClippingPlane = (plane) => {
     if (plane.getClassName() !== 'vtkPlane') {
-      return;
+      return false;
```
finetjul, line 15: Your check prevents a possible `vtkPlane` subclass. You could check `plane.isA('vtkPlane')`.
But I guess, you should let typescript enforce the input parameter type instead of checking it yourself.

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
```diff
@@ -0,0 +1,105 @@
...
+    let changeDetected = mapper.addClippingPlane(model._clipPlane1);
+    changeDetected =
+      mapper.addClippingPlane(model._clipPlane2) || changeDetected;
+
+    return changeDetected;
+  };
+
+  publicAPI.unregisterClipPlanesFromMapper = (mapper) => {
+    if (!mapper || !mapper.isA('vtkAbstractMapper')) {
```
finetjul, line 61: I wouldn't bother checking the mapper type, let typescript do it instead

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
```diff
@@ -0,0 +1,105 @@
...
+      return false;
+    }
+
+    const planes = mapper.getClippingPlanes();
+    let changeDetected = mapper.removeClippingPlane(
+      planes.indexOf(model._clipPlane1)
+    );
+    changeDetected =
+      mapper.removeClippingPlane(planes.indexOf(model._clipPlane2)) ||
+      changeDetected;
```
finetjul, line 71:
```
const modified = mapper.removeClippingPlane(planes.indexOf(model._clipPlane1) || mapper.removeClippingPlane(planes.indexOf(model._clipPlane2);
```

but maybe you should change the signature of removeClippingPlane to take a vtkPlane as input parameter.
I do not see a use case where  `removeClippingPlane` would take an integer.

Sources/Rendering/Core/AbstractMapper/index.js
jourdain: That was from the original code base. But I think you are right we should make sure we are in array bounds.

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
jourdain: I don't agree, I don't use typescript and I still want that protection.

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
jourdain: probably. It may have been easier to implement with index.

Sources/Rendering/Core/AbstractMapper/index.js
jourdain: Good point but that was from before.

Sources/Rendering/Core/AbstractMapper/index.js
jourdain: good catch

Sources/Rendering/Core/AbstractMapper/index.js
```diff
@@ -24,9 +29,10 @@ function vtkAbstractMapper(publicAPI, model) {
 
   publicAPI.removeClippingPlane = (i) => {
     if (i < 0 || i >= 6) {
-      return;
+      return false;
     }
     model.clippingPlanes.splice(i, 1);
```
finetjul, line 34: call `modified()` as well

Sources/Rendering/Core/AbstractMapper/index.js
DrewLazzeriKitware: `setClippingPlanes` below also has a constraint only allowing six clipping planes. I will change this so we check array bounds but also still cap at six, unless we know why that old constraint no longer should apply.

Sources/Rendering/Core/AbstractMapper/index.js
DrewLazzeriKitware: Agreed

Sources/Rendering/Core/AbstractMapper/index.js
DrewLazzeriKitware: :+1:

DrewLazzeriKitware:
I've left the class checks in AbstractMapper but made them more general. I've also changed the removeClippingPlane method to use instances rather than indexes since this is already a breaking change. 

I'm against changing AbstractMapper more in this PR unless it's related to the original issue of MPR though.

DrewLazzeriKitware:
I've split the AbstractMapper changes into this smaller PR: https://github.com/Kitware/vtk-js/pull/2186

I'll revisit this once that gets merged.

DrewLazzeriKitware:
Since  https://github.com/Kitware/vtk-js/pull/2186 was merged, I've rebased this on master and it's ready for merge.

Sources/Rendering/Core/VolumeMapper/SliceHelper.js
jourdain: @DrewLazzeriKitware with your change on master are we not using object directly rather than index?

Sources/Rendering/Core/VolumeMapper/index.d.ts
```diff
@@ -160,12 +160,55 @@ export function extend(publicAPI: object, model: object, initialValues?: IVolume
...
+
+  /**
+   * Set the origin of the MPR
+   */
+   setOrigin(origin: Vector3): void;
+
+  /**
+   * Set the orientation of the MPR
+   */
+   setNormal(origin: Vector3): void;
```
finetjul, line 201: does not return void but a boolean (true when normal is modified) (same for all setters)

parameter name is normal and not origin.

Sources/Rendering/Core/VolumeMapper/index.d.ts
```diff
@@ -160,12 +160,55 @@ export function extend(publicAPI: object, model: object, initialValues?: IVolume
  */
 export function newInstance(initialValues?: IVolumeMapperInitialValues): vtkVolumeMapper;
 
+/**
+ * Helper class to perform MPR.
+ */
+
+export interface vtkSliceHelper {
```
daker, line 167: This needs to extends vtkObject, also have its own `newInstance` and `extend` like in VolumeMapper TS defs, maybe something like this :

```ts
export declare const vtkSliceHelper: {
	newInstance: typeof newInstanceSliceHelper,
	extend: typeof extendSliceHelper,
};
```

daker (changes requested):

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -31,6 +33,7 @@ import controlPanel from './controlPanel.html';
 
 // Force the loading of HttpDataAccessHelper to support gzip decompression
 import 'vtk.js/Sources/IO/Core/DataAccessHelper/HttpDataAccessHelper';
+import { BlendMode } from '../../../../Rendering/Core/VolumeMapper/Constants';
```
daker, line 36: i think this should be with 'vtk.js/Sources/Rendering/Core/VolumeMapper/Constants'

PR #2114, opened 2021-10-13
fix(objReader): Fix point indexation

luciemac:
### PR and Code Checklist

### Context

OBJ file can associate a point to different texture coordinates and/or normals.
To define a vtkPolyData, those points are duplicated.
To avoid some indexation issue (ie pointIds in the OBJ file does not match the polydata pointIds), those points should be duplicated only once all the OBJ vertices are added.

### Changes

Add access to the mappingPoints object with key = original point id and value = list of duplicated points ids.

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 


floryst:
Is this ready? Seems like this is mergeable.

finetjul:
Replaced by https://github.com/Kitware/vtk-js/pull/2119

PR #2116, opened 2021-10-13
fix(Interaction): Remove getters/setters on zoomScale

tamaynard:
### Changes
 It appears the zoomScale is overwritten on mouseDown event. I'm removing the setter/getter on this variable since it had no effect and I was able to achieve the desired behaviour (adjusting the speed of the zoom) using an alternate method that does not require any modification to the core library.


floryst:
For future searches, could you briefly describe your "alternate method"?

This revert looks good to me, given that `zoomScale` is more internal than external here.

PR #2118, opened 2021-10-14
Improve esm and ci

floryst:
### PR and Code Checklist

### Context

We really ought to not be bundling external dependencies if we don't need to. For reference: https://github.com/Kitware/react-vtk-js/issues/40

### Changes
- [ ] do not conditionally depend on TextDecoder status
- [ ] do not bundle external packages in ES build
### Results
- faster ES builds that depend on installed deps rather than bundled deps
- fix TextDecoder constant condition warning on CI 
### Testing

This does not affect vtk.js tests.

To test, run `npm run build:esm` and use npm link to install the package to test.


FezVrasta:
That's great, thanks!

rollup.config.js
```diff
@@ -95,39 +96,8 @@ export default {
       return name.replace(/^Sources[/\\]/, '');
     },
   },
-  external: [/@babel\/runtime/],
+  external: Object.keys(packageJSON.dependencies).map((name) => new RegExp(`^${name}`)),
```
FezVrasta, line 99: There's a [`rollup-plugin-auto-external`](https://www.npmjs.com/package/rollup-plugin-auto-external) that should cover more cases (such as transitive dependencies)

FezVrasta:
@jourdain I don't have much context about it, but it's better to rely on dependencies so that if the consumer app already uses one of them the final bundle size is not affected by the duplicate. In this case it's a very small one so it may not matter, but it's a good thing to keep in mind.

jourdain:
We started with a dependency that broke under our feet after years of usage. At that point we've integrated a working solution within vtk.js on which we can rely on while being tiny.

jourdain:
Yes but that happen when we "manually" updated the version to keep up with our dependencies.

rollup.config.js
floryst: I'll check it out. Thanks for the suggestion!

PR #2140, opened 2021-10-25
fix(widgets): fix paintWidget ignore keys alt, control and shift

luaceseb:
### PR and Code Checklist

### Context

The paintWidget does not follow the guidelines of ignoring the alt, control and shift keys like the other widgets. Example DistanceWidget
### Changes

- Add function ignoreKey that checks if the alt, control or shift keys are pressed

### Results

- The painting action does not work if any of the alt, control or shift keys are pressed when painting begins
### Testing

  - **vtk.js**: v20.2.1 
  - **OS**: Windows 10 
  - **Browser**: Chrome 94.0.4606.81

### DIFF
--- a/Sources/Rendering/Core/CubeAxesActor/index.js
+++ b/Sources/Rendering/Core/CubeAxesActor/index.js
@@ -682,11 +682,12 @@ function vtkCubeAxesActor(publicAPI, model) {
     publicAPI.update();
   });
 
-  publicAPI.setVisibility = macro.chain(
+  const setVisibility = macro.chain(
     publicAPI.setVisibility,
     model.pixelActor.setVisibility,
     model.tmActor.setVisibility
   );
+  publicAPI.setVisibility = (...args) => setVisibility(...args).some(Boolean);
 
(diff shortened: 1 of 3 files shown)


floryst:
Re-visiting this PR:
- @finetjul does the shape/spline widget support runtime config of ignoring actions when modifier keys are pressed?
- @luaceseb please rebase onto master and fix the listed conflicts. If @finetjul agrees that this is good moving forward, then we will merge.

finetjul:
@floryst ShapeWidget handles [`alt`, `shift` or `ctrl` modifiers](https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Widgets3D/ShapeWidget/index.js#L17-L25). We therefore can't rely on `ignoreKeys`.
Either `ignoreKeys` must be updated whenever `modiferBehavior` is modified (there is synchronization challenge here) or `handleLeftButtonPress` should return `macro.VOID` when the current modifier is not listed in `modifierBehavior` (seems safer)

floryst:
Thanks for the update. @luaceseb please look to using modifier behaviors to implement your ignoreKey functionality.

PR #2146, opened 2021-10-28
fix(renderwindow): fix inconsistent depth mask value

finetjul:
When a scene contained only translucent actors, depth mask was never turned off before the translucent pass. This is because the renderer was setting it to true without notifying the render window.

### PR and Code Checklist

### Context

Translucent actors without opaque actor:
image

Translucent actors with opaque actors:
image

### Changes

### Results
Got the send image even without the opaque actors.
The issue with faces ordering is still there though... (different problem)

### Testing
  - **vtk.js**: master
  - **OS**: Windows
  - **Browser**: Chrome

### DIFF
--- a/Sources/Rendering/OpenGL/Actor/index.js
+++ b/Sources/Rendering/OpenGL/Actor/index.js
@@ -132,7 +132,7 @@ function vtkOpenGLActor(publicAPI, model) {
 
   publicAPI.opaquePass = (prepass, renderPass) => {
     if (prepass) {
-      model.openGLRenderWindow.enableDepthMask();
+      model.context.depthMask(true);
       publicAPI.activateTextures();
     } else if (model.activeTextures) {
       for (let index = 0; index < model.activeTextures.length; index++) {
@@ -144,7 +144,7 @@ function vtkOpenGLActor(publicAPI, model) {
   // Renders myself
   publicAPI.translucentPass = (prepass, renderPass) => {
     if (prepass) {
-      model.openGLRenderWindow.disableDepthMask();
+      model.context.depthMask(false);
       publicAPI.activateTextures();
     } else if (model.activeTextures) {
       for (let index = 0; index < model.activeTextures.length; index++) {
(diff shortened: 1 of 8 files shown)


finetjul:
> @jourdain I'm afraid that error will show up again and again, i.e.direct access to `context.depthMask(true)` instead of going through `openGLRenderWindow.enableDepthMask()`.
> 
> Can't we remove the `depthMask()` method from the `WebGLRenderingContext` object (or override it) ?

I'm thinking of doing something like that in `publicAPI.get3DContext()` (which shouldn't it public ?) instead of the current `publicAPI.enableDepthMask()`/`publicAPI.disableDepthMask()`.
```
publicAPI.get3DContext = (...) =>{
   ...
    const depthMaskFunction = result.depthMask;
    let depthMaskEnabled = result.getParameter(result.DEPTH_WRITEMASK);
    result.depthMask = (enable) => {
      if (enable !== depthMaskEnabled) {
        depthMaskFunction.call(result, enable);
        depthMaskEnabled = enable;
      }
    return result;
```
@jourdain, do you think it would be too hackish ?

jourdain:
I'm just partially following but since we wrap the canvas to get our glContext, I would be fine getting an "advanced" glContext with more methods.

But looking at your snippet of code, it does not seem fully correct unless `result.depthMask` is already a function that does not exactly behave the way you want. 

I would rather not conflict with existing method. Also, since with the glContext we are tapping into browser specific (think native) I would be worried that some browser may not allow you to edit/modify the object/existing-method they hand you (but I can be wrong). I would rather create a method `safeDepthMask()` instead.

martinken:
I get there are two goodish options here, 

1) we just always call context.depthMask(...) and remove the cached mathods

2) we try to isolate the gl context so that people cannot bypass the cached methods

Option 1 seems easier and the only downside is maybe a trivially small performance cost so I would go that route FWIW. I mean we are a long long way from caching the full opengl state on vtk.js.

finetjul:
Fair enough, I removed caching of depth mask.

martinken:
+1

Issue #2134, opened 2021-10-21
Allow arbitrary number of clipping planes in PolyDataMapper


ebrahimebrahim:
Currently only up to 6 clipping planes can be added to [PolyDataMapper](https://github.com/Kitware/vtk-js/tree/master/Sources/Rendering/OpenGL/PolyDataMapper). This is because the number has to be [hard-coded](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/PolyDataMapper/index.js#L878) into the fragment shader.

We can make that number arbitrary by rebuilding the shader every time it changes. For example, the volume mapper's clipping plane loop count [is set based on the number of clipping planes](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/VolumeMapper/index.js#L316).

This issue should be assigned to me.

daker:
@ebrahimebrahim it should be released, right?

ebrahimebrahim:
> @ebrahimebrahim it should be released, right?

That's right, closing, thank you

PR #2152, opened 2021-11-04
allow more than 6 clipping planes on a polydata

ebrahimebrahim:
This addresses #2134 

Look, there are eight clipping planes on this sphere:

wow8planes

And there are 30 on this one:

wow30planes

Any more than that and we run into an error during the shader linking step.
It seems that there's a limit to the amount of `varying` variables that can be used ([related discussion](https://community.khronos.org/t/link-error-could-not-pack-varying/2495)).
This actual limit might depend on hardware.

- I included in this PR the example that I made to test my code. Do we want to keep that?
- I removed a seemingly redundant/inconsistent method in `Rendering/OpenGL/ShaderProgram`.
- I noticed  that a `6` seems to be hardcoded [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/VolumeMapper/index.js#L305-L306) in `VolumeMapper`. This is a bug, right? Should I open an issue, or just test and fix it in this PR?
- It looks like `ImageMapper` [also has a limit](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/ImageMapper/index.js#L333) of 6 clipping planes. It might be just as easy of a fix as this one. Do we care about this? Should I open an issue? Just do it in this PR?

Depending on answers to above bullet points, this may be ready to be merged.

@martinken  Can you have a look please?


martinken:
Can you do a quick test for me? Mod your example to render with no clipping planes and then add some and rerender. I want to make sure the shader is getting rebuilt when the number of clipping planes is changing.

ebrahimebrahim:
> Can you do a quick test for me? Mod your example to render with no clipping planes and then add some and rerender. I want to make sure the shader is getting rebuilt when the number of clipping planes is changing.

Hmm you are right, any clipping planes I add after the first render don't seem to work.

martinken:
I think if you fix AbstractMapper to call modified() in AddClipingPlane that might fix the issue as that will mark the mapper as modified.

ebrahimebrahim:
> I think if you fix AbstractMapper to call modified() in AddClipingPlane that might fix the issue as that will mark the mapper as modified.

Yup that worked, thank you! The shader now rebuilds each time as I add clipping planes one by one starting from zero.

ebrahimebrahim:
Some remaining items to address before merging:

- [ ] add a test of mesh clipping, including adding planes
- [ ] keep example but prepend directory with underscore to prevent it from being published to webpage
- [ ] Should I make `VolumeMapper` and `ImageMapper` changes and include them in this PR? If yes, do that.

ebrahimebrahim:
Added a test to cover two things:
- Many clipping planes (> 6, which was a previous limit)
- Rendering with no clipping planes, then adding clipping planes, then rendering again with them included. This used to be broken, because the shaders weren't getting rebuilt when clipping planes get introduced.

This should be ready to be reviewed and merged.

ebrahimebrahim:
@martinken This should be ready for review

martinken:
Only question is am I seeing correctly that the example is named _Mesh... ?  Is the underscore supposed to be there?

ebrahimebrahim:
> Only question is am I seeing correctly that the example is named _Mesh... ? Is the underscore supposed to be there?

Yes it was suggested to add the underscore to prevent the example from being published

ebrahimebrahim:
Oops I just noticed I kept `test.only` in my test. Don't merge yet!

martinken (approved):
+1 pending CI

PR #2158, opened 2021-11-10
feat(vtkImageData.computeHistogram): added number of pixels involved

jadh4v:
Added number of pixels involved in the computations to the return array of `computeHistogram()`.

Added the number of pixels involved in the calculation of the various values as part of the return
object, to fulfill an application side requirement. It also makes sense to have this information returned
since the function takes a mask/stencil style function as input.

Corrected the equation for variance. Returned values for variance and standard deviation were
previously wrong.

Added test cases for `vtkImageData.computeHistogram`.

### PR and Code Checklist

### Changes

### Testing

  - **vtk.js**:  master
  - **OS**:  Windows 10 pro, version: 21H1 build: 19043.1288
  - **Browser**:  Chrome 95.0.4638.69

Attn: @agirault @floryst @jourdain


agirault:
Suggestion (not required): would ideally want those two changes (variance & count) to be different commits.

Are there tests for the histogram? 
* If not, could you add one?
* If yes, could you update it to show the variance was wrong, and is now right?

Ty

Sources/Common/DataModel/ImageData/index.js
```diff
@@ -410,7 +410,7 @@ function vtkImageData(publicAPI, model) {
     }
 
     const average = inum > 0 ? isum / inum : 0;
-    const variance = sumOfSquares - average * average;
+    const variance = Math.abs(sumOfSquares / inum - average * average);
```
agirault, line 413: @dzenanz do you mind reviewing that this is accurate? ty

dzenanz: Taking absolute value is not needed, because sum of squares is guaranteed to be larger than square of average. But I believe the formula is now correct.

jadh4v:
Done!
1. Created two separate commits for the bug fix and adding count to return.
2. Added test cases. There were no test cases for vtkImageData. Added a new file at Common/DataModel/ImageData/test/.

Sources/Common/DataModel/ImageData/index.js
dzenanz: I later realized numerical considerations were probably the reason for that.

floryst: If we're checking the `inum > 0` case for the average calculation, best to do it here as well.

jadh4v: Added check for `inum`, and a test case to cover that possibility.

floryst:
LGTM

agirault (approved):
Ty 🚀

Issue #2104, opened 2021-10-07
Miscellaneous rendering tests fail


floryst:
### High-level description

A series of tests fail with pixelmatch threshold of 0.

These differences are more difficult to spot (for me), so I've just included them here. These are potentially slight color differences that are fine. The solution may be to bump up the threshold.

### Steps to reproduce

1. set testUtils.js pixelmatch threshold to 0
2. run npm test

### Detailed behavior

I've included a sampling of the failing tests.

![image](https://user-images.githubusercontent.com/1812167/136430125-7f1ac136-8394-45a4-b014-be5327ddd84b.png)
![image](https://user-images.githubusercontent.com/1812167/136430386-71d4f96a-23a2-4d87-bf18-63710218981a.png)
![image](https://user-images.githubusercontent.com/1812167/136430234-6cb70e04-96cb-4f8f-b2c8-867c60968e5f.png)
![image](https://user-images.githubusercontent.com/1812167/136430256-42cc41db-53e1-443d-8161-afb331df378a.png)
![image](https://user-images.githubusercontent.com/1812167/136430276-eecac545-e4b4-4c8a-8364-57c1389bdd62.png)
![image](https://user-images.githubusercontent.com/1812167/136430310-b2db7cb4-ea1e-49b1-b313-ceb52df5baec.png)
![image](https://user-images.githubusercontent.com/1812167/136430332-49f31482-56b5-4fc0-b8ae-b9928f45c6c7.png)
![image](https://user-images.githubusercontent.com/1812167/136430443-29111324-9843-48ef-9a0d-eadb3ff732ab.png)

### Environment
- **vtk.js**: v19.7.2

Issue #2103, opened 2021-10-07
Test fail: testScaleInPixelsWithParallel(): scaleInPixels=true, parallelProjection=true


floryst:
### High-level description

`Widgets/Core/WidgetManager/test/testWidgetManager.js` test fails with pixelmatch threshold of 0.

### Steps to reproduce

1. set testUtils.js pixelmatch threshold to 0
2. run npm test

### Detailed behavior
![image](https://user-images.githubusercontent.com/1812167/136429073-ce88ebd6-1276-4dea-a347-101539cb70c4.png)

### Environment
- **vtk.js**: v19.7.2

floryst:
FYI @finetjul, though I can also look at this if needed.

Issue #2102, opened 2021-10-07
Average Intensity Projection Volume Rendering test fails


floryst:
### High-level description

`Rendering/OpenGL/VolumeMapper/test/testAverageIntensityProjection` test fails with pixelmatch threshold of 0.

### Steps to reproduce

1. set testUtils.js pixelmatch threshold to 0
2. run npm test

### Detailed behavior

![image](https://user-images.githubusercontent.com/1812167/136428364-8a347771-3492-4dd8-b766-aeb0aed446ed.png)

### Environment
- **vtk.js**: v19.7.2

floryst:
@swederik is the baseline for this incorrect? or is the rendering incorrect? Or something else?

PR #2161, opened 2021-11-12
Update test framework

floryst:
### PR and Code Checklist

### Context
- #2104

### Changes

This results in quite some big changes, so do let me know if some choices and decisions do not make sense to you. Going commit-by-commit will make review much more straightforward.

I ended up dropping `karma-tap` because there's a race condition issue where some of the first tests to run do not get picked up by `karma-tap`, resulting in some tests just not being reported.

I'm keeping the percentage-based mismatch tolerance/threshold measure, since tracking down acceptable raw pixel count thresholds for each test is too much work.

- [x] overhauled test framework to report HTML test results
- [x] nicer image test output
- [x] globbed tests rather than Sources/tests.js
- [x] bad image tests have been skipped (see #2102, #2103, and #2104)

### Results
- `Sources/tests.js` is no longer something that needs to be maintained. Tests are picked up by glob.
- broken image tests are skipped (for now)
- Here is a sample of what the generated Test-Report.html looks like:

image

### Testing
  - **vtk.js**: latest (v21)


Sources/Widgets/Core/WidgetManager/test/testWidgetManager.js
```diff
@@ -128,9 +128,10 @@ test.onlyIfWebGL('Test getPixelWorldHeightAtCoord', (t) => {
   }
 
   function testScaleInPixelsWithParallel() {
-    t.comment(
+    t.skip(
```
jourdain, line 131: is the goal here is to print in output. Does .skip() do it?

jourdain (approved):
LGTM

finetjul:
LGTM

Sources/Widgets/Core/WidgetManager/test/testWidgetManager.js
finetjul: I think @floryst skipped it because the tests is currently failing.

martinken:
Let me quickly try this on windows and on WebGPU

martinken (approved):
Works on windows and WebGPU

WebGPU fails 6 tests, some look like real failures and some like threshold issues but that seems reasonable for now. I'll clean up WebGPU on a different MR eventually.

LGTM +1

finetjul:
@floryst it might be worth updating [CONTRIBUTING.rst](https://github.com/Kitware/vtk-js/blob/master/CONTRIBUTING.md). It contains old test framework advice.

floryst:
Good point, I'll make a note of that.

PR #2164, opened 2021-11-16
docs(breaking_changes): document that Node 10 support has been droppe…

finetjul:
…d in 19.3.0

### PR and Code Checklist

### Context
It was not documented that node 10 is no longer supported.

### Changes

### Testing
  - **vtk.js**: master -->
  - **OS**: Windows 10
  - **Browser**: Chrome


daker:
@floryst don't you think that adding a `.nvmrc` file would be a good idea for that?

floryst:
I like that idea. vtk.js targets the LTS versions (maintenance + active), so we can specify that for folks who use nvm. (Assuming they remember to run `nvm use`.)

@finetjul if you also think it's a good idea, then it'll be good to run `echo 'lts/*' > .nvmrc` and add that to this PR.

finetjul:
@floryst I have no idea what nvmrc is for but I trust you. I added the file like requested

daker:
@finetjul nvm is a version manager for node.js, it allows you to use multiple node.js installation on a same system, and .nvmrc is file that will tell nvm what node version the project is using.

On the other side, i am not sure why your .nvmrc is a binary file, it should be a text file.

https://github.com/nvm-sh/nvm#nvmrc

finetjul:
@daker gotcha, thanks for the explanation

file was detected as binary because it was UTF-16. I converted it to UTF-8

floryst:
LGTM

PR #2165, opened 2021-11-17
feat(macros.js): add a toJSON method to vtkObject publicAPI

Julesdevops:
The toJSON method simply returns the result of publicAPI.getState()
It allow user to call JSON.stringify directly on a vtkObject instead of doing JSON.stringify(vtkObject.getState())

### PR and Code Checklist

### Context
To serialize an object to JSON you have to use JSON.stringify(vtkObj.getState())

### Changes
Now we can use JSON.stringify(vtkObj) and we will get the same result.

### Testing
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Filters/General/AppendPolyData/test/testAppendPolyData.js
+++ b/Sources/Filters/General/AppendPolyData/test/testAppendPolyData.js
@@ -66,9 +66,8 @@ test('Test vtkAppendPolyData execution', (t) => {
 });
 
 test('Test addInputData edge case', (t) => {
-  const gc = testUtils.createGarbageCollector(t);
-  const appender = gc.registerResource(vtkAppendPolyData.newInstance());
-  const input = gc.registerResource(vtkPolyData.newInstance());
+  const appender = vtkAppendPolyData.newInstance();
+  const input = vtkPolyData.newInstance();
 
   appender.addInputData(input);
   const output = appender.getOutputData();
@@ -76,7 +75,7 @@ test('Test addInputData edge case', (t) => {
   t.ok(input === output, 'Single add input matches output');
   t.ok(appender.getNumberOfInputPorts() === 1, 'Expect 1 port after 1 add');
 
-  const input2 = gc.registerResource(vtkPolyData.newInstance());
+  const input2 = vtkPolyData.newInstance();
   appender.addInputData(input2);
   const output2 = appender.getOutputData();
 
--- a/Sources/macros.js
+++ b/Sources/macros.js
@@ -195,7 +195,7 @@ function enumToString(e, value) {
 }
 
 function getStateArrayMapFunc(item) {
-  if (item.isA) {
+  if (item && item.isA) {
     return item.getState();
   }
   return item;
@@ -331,6 +331,9 @@ export function obj(publicAPI = {}, model = {}) {
 
   // Add serialization support
   publicAPI.getState = () => {
+    if (model.deleted) {
+      return null;
+    }
     const jsonArchive = { ...model, vtkClass: publicAPI.getClassName() };
 
     // Convert every vtkObject to its serializable form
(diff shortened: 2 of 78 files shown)


Sources/macros.js
```diff
@@ -391,6 +391,12 @@ export function obj(publicAPI = {}, model = {}) {
     publicAPI.modified();
   };
 
+  // This function will get called when one invoke JSON.stringify(vtkObject)
+  // JSON.stringify will only stringify the return value of this function
+  publicAPI.toJSON = function vtkObjToJSON() {
```
jourdain, line 396: why not write `publicAPI.toJSON = publicAPI.getState`

finetjul: if `getState()` get overriden in subclass

jourdain: Make sense...

floryst:
Tests are hanging for some reason. Can you isolate the test and verify that it runs locally?

Julesdevops:
> Tests are hanging for some reason. Can you isolate the test and verify that it runs locally?

Tests are running fine locally ; Do you want me to create a whole new test/test file for this ?

finetjul:
Failing tests seem to be due to WidgetManagerTest

finetjul:
@jourdain WDYT about the solution to fix infinitely recursive `getState()` ? I'm not a big fan (especially for `_parentProp` in `vtkProp`), but if that's fine with you, I'm good with it. Another option would be to use "unique references" to already serialized objects to avoid infinite recursions...

Sources/Filters/General/AppendPolyData/test/testAppendPolyData.js
```diff
@@ -66,17 +66,16 @@ test('Test vtkAppendPolyData execution', (t) => {
 });
 
 test('Test addInputData edge case', (t) => {
-  const gc = testUtils.createGarbageCollector(t);
-  const appender = gc.registerResource(vtkAppendPolyData.newInstance());
```
jourdain, line 70: Why do you remove the GC in that test?

finetjul: It is not an asynchronous test, hence there is no need for GC. Did I miss something ?

finetjul:
Thanks for your review.

Ok for `macro.moveToPrivate()`. We could rename it to `macro.private` (to be more consistant with `macro.get` and `macro.set` ?)

We could also handle in `moveToPrivate()` the "_" prefix renaming of set/get accessors ? Or is it fine to leave the renaming to `capitalize()`? My motivation to handle the set/get renaming in `moveToPrivate()` is that the usefulness of `moveToPrivate()` is not very obvious because it will work fine without it for most of the times. I'm afraid developers will forget it. They won't forget it if they need accessors.

Btw, `capitalize()` works fine (i.e. `_interactor` -> `Interactor` and `interactor` -> `Interactor`), but `uncapitalize()` is more limited (i.e. `Interactor` -> `interactor`, it will never do `Interactor` -> `_interactor`).

jourdain:
I was limiting the use of `moveToPrivate` to only rename key in dict if available. 

Not all private var should have set/get. But extending set/get macro to better handle them make sense.

Sources/Filters/General/AppendPolyData/test/testAppendPolyData.js
finetjul: Are you saying ALL tests must use testUtils GarbageCollector ?

jourdain: Yes otherwise, the canvas stay around (as nobody delete them) and we run out of available one.  The gc in test help to clear things between tests.

jourdain: I mean the rendering ones. Having leaks in test with objects is not a big deal, but gl canvas are limited to 16 or something like that.

PR #2171, opened 2021-11-22
feat(XR): Update to WebXR functionality in OpenGL RenderWindow

tbirdso:
### PR and Code Checklist

### Context

Replaces WebVR with WebXR methods for running an immersive VR session from an OpenGL RenderWindow.

[WebXR](https://immersiveweb.dev/) exposes methods for both VR and AR sessions across a diverse set of platforms, including standalone headsets, fixed displays, and mobile devices. These changes use WebXR to run a VR session in a dedicated headset or mobile viewer. `Applications/SkyboxViewer` and `Geometry/VR` examples are updated for API changes.

Planned future additions (outside the scope of this PR):
- WebXR polyfill
- XR gamepad API
- AR

### Changes

- [x] `RenderWindow` now manages VR sessions with `startXR`, `enterXR`, `stopXR`, and `xrRender`
- [x] `Applications/SkyboxViewer` and `Geometry/VR` examples are updated to query availability/launch immersive sessions with VR
- [x] Gamepad support for VR examples is *deprecated* and will be reintroduced with the WebXR Gamepads API in the near future.

### Results

- `Applications/SkyboxViewer` and `Geometry/VR` examples query VR availability and launch/halt VR sessions correctly in WebXR-compatible browsers
- VR examples compatible with tethered headsets (such as HTC Vive Pro) or mobile headsets with dedicated processor (Quest)

### Testing

  - **vtk.js**: v21.1.2
  - **OS**: Windows 10
- [x] VR examples run properly in WebXR-supported environment
  - [WebXR emulator](https://github.com/MozillaReality/WebXR-emulator-extension)
  - Oculus Browser
  - Chrome (via Oculus WebXR backend)

xr-simulator

### DIFF
--- a/Examples/Applications/SkyboxViewer/index.js
+++ b/Examples/Applications/SkyboxViewer/index.js
@@ -6,10 +6,7 @@ import 'vtk.js/Sources/Rendering/Profiles/Geometry';
 import HttpDataAccessHelper from 'vtk.js/Sources/IO/Core/DataAccessHelper/HttpDataAccessHelper';
 import macro from 'vtk.js/Sources/macros';
 import vtkDeviceOrientationToCamera from 'vtk.js/Sources/Interaction/Misc/DeviceOrientationToCamera';
-import vtkForwardPass from 'vtk.js/Sources/Rendering/OpenGL/ForwardPass';
 import vtkFullScreenRenderWindow from 'vtk.js/Sources/Rendering/Misc/FullScreenRenderWindow';
-import vtkRadialDistortionPass from 'vtk.js/Sources/Rendering/OpenGL/RadialDistortionPass';
-import vtkRenderer from 'vtk.js/Sources/Rendering/Core/Renderer';
 import vtkSkybox from 'vtk.js/Sources/Rendering/Core/Skybox';
 import vtkSkyboxReader from 'vtk.js/Sources/IO/Misc/SkyboxReader';
 import vtkURLExtract from 'vtk.js/Sources/Common/Core/URLExtract';
@@ -41,14 +38,9 @@ let autoInit = true;
 const cameraFocalPoint = userParams.direction || [0, 0, -1];
 const cameraViewUp = userParams.up || [0, 1, 0];
 const cameraViewAngle = userParams.viewAngle || 100;
-const enableXR = !!userParams.xr;
 const eyeSpacing = userParams.eye || 0.0;
 const grid = userParams.debug || false;
 const autoIncrementTimer = userParams.timer || 0;
-const disableTouchNext = userParams.disableTouch || false;
-const distk1 = userParams.k1 || 0.2;
-const distk2 = userParams.k2 || 0.0;
-const cameraCenterY = userParams.centerY || 0.0;
 
 const body = document.querySelector('body');
 let fullScreenMetod = null;
@@ -151,10 +143,8 @@ function createVisualization(container, mapReader) {
   const mainRenderer = fullScreenRenderer.getRenderer();
   const interactor = fullScreenRenderer.getInteractor();
   const actor = vtkSkybox.newInstance();
-  let camera = mainRenderer.getActiveCamera();
-  let leftRenderer = null;
-  let rightRenderer = null;
-  let updateCameraCallBack = mainRenderer.resetCameraClippingRange;
+  const camera = mainRenderer.getActiveCamera();
+  const updateCameraCallBack = mainRenderer.resetCameraClippingRange;
 
   // Connect viz pipeline
   actor.addTexture(mapReader.getOutputData());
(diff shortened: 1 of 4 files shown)


tbirdso:
Current changes allow scene to be appropriately rendered onscreen before sending to XR and then in a headset once the button is pressed. However, `stopXR` method does not seem to properly reset the scene for viewing on a fixed display. After returning from XR, trying to interact with the renderwindow via click and drag gives two error messages repeatedly when `traverseAllPasses` is called:
```
WebGL: INVALID_FRAMEBUFFER_OPERATION: clear: Cannot render to a XRWebGLLayer framebuffer outside of an XRSession animation frame callback.
WebGL: INVALID_FRAMEBUFFER_OPERATION: drawArrays: Cannot render to a XRWebGLLayer framebuffer outside of an XRSession animation frame callback.
```

It seems that the XR animation frame is cancelled as expected, so the question is why the XRWebGLLayer is still used outside of expected XR methods.

This error is not observed with an XR emulator or when the page is rendered natively on a mobile headset (tested on Quest), in which cases the user can interact with the scene as expected after returning from VR. This seems to be an issue with trying to select which display to render to, perhaps as a result of making an existing GL layer XR compatible.

@floryst @martinken Any thoughts or suggestions here? I do not see this issue when running the [WebXR VR sample](https://immersive-web.github.io/webxr-samples/immersive-vr-session.html), but that example also never tries to render to the main canvas in the browser.

tbirdso:
Also pinging @jadh4v as an intro to WebXR investigation

tbirdso:
Above issue is resolved, I was not unbinding the XR context framebuffer in `stopXR`, thanks @floryst for the find. With this fix the scene can be interacted with again via mouse events after exiting VR.

Some remaining issues to note:
- Attempting to interact with the canvas while the XR session is underway still yields `INVALID_FRAMEBUFFER_OPERATION`, I assume due to trying to draw to the canvas while it is not available. Previous WebVR code handles this by checking whether the display is external or not, but so far it doesn't look like WebXR exposes this value. Will continue to investigate.
- Entering and exiting VR repeatedly with SteamVR as the OpenXR backend can cause SteamVR to crash. Not seeing this issue with Oculus as the OpenXR backend.
- Cone appears at different heights/positions for SteamVR and Oculus backends.

Examples/Applications/SkyboxViewer/index.js
```diff
@@ -189,7 +189,8 @@ function createVisualization(container, mapReader) {
     updateSkybox(allPositions[nextIdx]);
   }
 
-  if (enableVR && vtkDeviceOrientationToCamera.isDeviceOrientationSupported()) {
+  if (enableXR && navigator.xr.isSessionSupported('immersive-vr')) {
+    console.log(vtkDeviceOrientationToCamera.isDeviceOrientationSupported());
```
martinken, line 193: Yank probably

Examples/Geometry/VR/index.js
```diff
@@ -36,8 +36,7 @@ const renderWindow = fullScreenRenderer.getRenderWindow();
 // this
 // ----------------------------------------------------------------------------
 
-const coneSource = vtkConeSource.newInstance({ height: 100.0, radius: 50.0 });
-// const coneSource = vtkConeSource.newInstance({ height: 1.0, radius: 0.5 });
+const coneSource = vtkConeSource.newInstance({ height: 1.0, radius: 0.5 });
```
martinken, line 39: It is good to at least test using a cone with a crazy large/small size and positioned far from the origin to make sure everything works properly. I think your changes still use the same basic path as the old code so it should be fine. But testing unusual sizes and locations help find a lot of issues.

Examples/Geometry/VR/index.js
```diff
@@ -67,7 +66,7 @@ mapper.setInputConnection(filter.getOutputPort());
 
 const actor = vtkActor.newInstance();
 actor.setMapper(mapper);
-actor.setPosition(20.0, 0.0, 0.0);
+actor.setPosition(-1.5, 0.0, 0.0);
```
martinken, line 69: Same comment as above. The reset code should work so that it doesn't matter where the code is or what size it is, it should default to a nice view and interaction in VR.

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -1,16 +1,20 @@
...
 import vtkRenderPass from 'vtk.js/Sources/Rendering/SceneGraph/RenderPass';
-import vtkShaderCache from 'vtk.js/Sources/Rendering/OpenGL/ShaderCache';
 import vtkRenderWindowViewNode from 'vtk.js/Sources/Rendering/SceneGraph/RenderWindowViewNode';
-import vtkOpenGLTextureUnitManager from 'vtk.js/Sources/Rendering/OpenGL/TextureUnitManager';
-import vtkOpenGLHardwareSelector from 'vtk.js/Sources/Rendering/OpenGL/HardwareSelector';
-import { VtkDataTypes } from 'vtk.js/Sources/Common/Core/DataArray/Constants';
+
+// import WebXRPolyfill from 'webxr-polyfill';
+
+// const polyfill = new WebXRPolyfill();
```
martinken, line 14: Leftovers?

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -1,16 +1,20 @@
...
 
 const { vtkDebugMacro, vtkErrorMacro } = macro;
-const IS_CHROME = navigator.userAgent.indexOf('Chrome') !== -1;
+// const IS_CHROME = navigator.userAgent.indexOf('Chrome') !== -1;
```
martinken, line 17: Leftover?

Examples/Geometry/VR/index.js
tbirdso: That makes sense. The goal of changing to a smaller height/radius is so that we can test translation is working, much easier to see that we're moving around a small cone in the foreground than a very large cone in the distance.

Examples/Geometry/VR/index.js
tbirdso: Need to explore view resetting with the [WebXR gamepads module](https://github.com/immersive-web/webxr-gamepads-module), at the moment we're purely using the view we get from the XR session.

Sources/Rendering/OpenGL/RenderWindow/index.js
tbirdso: Yes, will remove here. @floryst and I discussed the WebXR polyfill briefly today, will subsequently explore adding the polyfill as an external dependency.

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -275,111 +278,143 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
-
-          model.vrSceneFrame = model.vrDisplay.requestAnimationFrame(
-            publicAPI.vrRender
-          );
-          // If Broswer is chrome we need to request animation again to canvas update
-          if (IS_CHROME) {
-            model.vrSceneFrame = model.vrDisplay.requestAnimationFrame(
-              publicAPI.vrRender
-            );
-          }
```
tbirdso, line 318: @martinken Based on testing it does not seem like we need to request animation twice for Chrome, is this consideration present elsewhere? Maybe a stopgap for a Chrome issue at the time of WebVR investigation?

Sources/Rendering/OpenGL/RenderWindow/index.js
tbirdso: Probably.. is there Chrome-specific behavior we need to consider for vtk.js to work properly? See comment below.

Examples/Applications/SkyboxViewer/index.js
```diff
@@ -189,7 +189,8 @@ function createVisualization(container, mapReader) {
     updateSkybox(allPositions[nextIdx]);
   }
 
-  if (enableVR && vtkDeviceOrientationToCamera.isDeviceOrientationSupported()) {
+  if (enableXR && navigator.xr.isSessionSupported('immersive-vr')) {
```
tbirdso, line 192: @floryst @martinken I'd like to propose that we remove the `enableVR`/`enableXR` flag entirely from SkyboxViewer.
- Subverts typical VR design by immediately trying to render a VR scene rather than presenting the user with a button to opt-in;
- Different design pattern using two different renderers, as opposed to two passes over the same renderer slightly translated;
- Relies on `vtkRadialDistortionPass` which doesn't seem to be used anywhere else and isn't necessary for at least the majority of VR viewers (potentially Cardboard?)

If preferred we can ignore the flag for now and then in a subsequent change move this functionality from the `SkyboxViewer` example into a dedicated example demonstrating how vtk.js *could* be used for a dedicated Cardboard viewing experience.

Examples/Geometry/VR/index.js
martinken: The reset camera method should (if all is working) look at the visible geometry and setup the physical scale and translation so that the geometry is near the viewer and at a reasonable scale. e.g. if you place a cone 1km away, ResetCamera should detect that and adjust the camera properties (physical translation) so that the viewer is near the cone. It should look like it is right next to you. Same idea with scale. Hope that makes sense.

tbirdso: `resetCamera` positions the camera correctly prior to entering the VR session, but once we start the session we set the camera to the position of each eye (two passes) in each frame in the local reference space we get from WebXR representing the physical confines of the viewer's VR setup. 

For 3DOF (rotation only) we only have one vantage point so the cone looks the same when it's 1km away and large vs 1m away and small. To demonstrate 6DOF it is best to be able to walk around the cone, which requires that the cone is close to the viewer within the local reference space.

We can experiment with moving the local reference space and introducing controller interactions with the WebXR gamepad API in future updates.

martinken: We have to modify the matrices from WebXR to do it correctly. 

With the old example the cone could be 1km away and super small and it would still work and look reasonable to the viewer. It isn't that the Cone is large and far away so it looks the same. It is that we take the matrices from the VR system and modify them by a world to physical matrix prior to using them. The world to physical matrix (stored in vtkjs as a few members on Camera) consists of a rotation, translation and uniform scale. It allows any size cone, placed in any location, to look reasonable when viewed in VR by default. And you can walk around it. 

If the VR example isn't showing up properly with the old Cone values, where you can walk around the cone, then we have something we need to fix.

To see what I mean try running the current example https://kitware.github.io/vtk-js/examples/VR/index.html and see how even through the cone is not at the origin you can walk around it.  If you change the values it still works.  I just set it to 2000 meters away with a height of 0.01 and it still looks good and I can walk around it. Because resetCamera is adjusting the physicaltoworld based on the currently loaded data. So if the Cone in the VR example only looks reasonable with some settings then we probably need to fix something.

martinken: The reason we care about this is that most datasets are not located at the origin with a size of about 1.0. They tend to be all over the place and with sizes that can range all over the place. So we built in a mechanism where the VR code through resetCamera will come up with a reasonable view for any of these datasets by default.

tbirdso: Thanks for the thorough explanation @martinken. I see what you mean after looking back at the WebVR example and it certainly seems reasonable to center arbitrary datasets for the VR view. It also seems possible that a user would want to fix the scale of a dataset for a walkthrough, but that can be revisited later if we want.

I've pushed a fixup commit that should resolve the issue, was previously calling `computeViewParametersFromViewMatrix` which bypasses the world-to-physical transformation as opposed to `computeViewParametersFromPhysicalMatrix` which correctly applies the transformation to the view matrix we receive from WebXR for each eye. The "large" cone seems reasonably scaled when testing in a headset and simulator. Please feel free to check on your side as well.

Examples/Geometry/VR/index.js
tbirdso: Resolved above, was skipping the world-to-physical transformation with `computeViewParametersFromViewMatrix`. Gamepad not required for the VR example in this first pass.

Sources/Rendering/OpenGL/RenderWindow/index.js
tbirdso: Removed.

Examples/Applications/SkyboxViewer/index.js
floryst: I agree that VR should have a send action rather than dropping directly into it. My guess is that it was like this for some early demo.

Examples/Geometry/VR/index.js
jadh4v: > For 3DOF (rotation only) we only have one vantage point so the cone looks the same ("right next to you") when it's 1km away and large vs 1m away and small.

I am not sure this is true. Wouldn't the convergence of the eyes matter? The user should be able to sense the difference between these two scenes.

martinken: Yes exactly @tbirdso  Perfect. On the C++ side we do provide VR GUI elements that allow the user to select an exact scale or to arbitrarily scale the world using the controllers. But agree that is all future MR stuff .

Examples/Applications/SkyboxViewer/index.js
tbirdso: Removed in fixup commit

tbirdso:
Second fixup commit contains minor bug fixes and cleanups:
- Fixed bug where `RenderWindowInteractor` tried to 1) request animation frame on mouse click + drag for rotation and 2) force render on mouse release from rotation. Mouse events have no bearing on the scene while in VR.
- Better handling for potential error where the user takes off the headset and the `XRSession` ends before `Return from VR` is clicked. For now we assume that the `XRSession` should continue until the user manually indicates that it should end. We could try to handle an end event to do the same, but risk mishandling situations such as where two users pass a headset between each other and the session "ends" in between.
- Reset the camera once the `XRSession` is manually ended to regain a good view of the scene.

Will plan to squash fixup commits before merge.

tbirdso:
@floryst @martinken this is ready for re-review when you have time

martinken:
+1

tbirdso:
Squashed fixup commits.

floryst:
LGTM. Thanks!

PR #2173, opened 2021-11-23
docs(ts): Update Typescript definitions

daker:
### PR and Code Checklist

### Context
Continue effort to add Typescript definitions & documentation

### Changes
This PR : 
* Exposes custom types directly
* Deprecates the `Point` type in favor of `Vector3` 
* Use RGBColor types instead of Color when necessary
* Fixes vtkAlgorithm type definition
* Fixes vtkLight type definition

@jourdain can you please check if the `getOutputData` & `setInputData` methods parameters types are correct?

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

Capture d’écran du 2021-11-23 00-39-21
Capture d’écran du 2021-11-23 00-40-28


floryst:
Thanks @daker! One question, since I do not remember the details: are the API docs auto-generated from the ts definition files where applicable?

daker:
@floryst yes, the api.md is generated from d.ts using [generate-api-docs.js](https://github.com/Kitware/vtk-js/blob/master/Documentation/generate-api-docs.js), which is called at build time.
https://github.com/Kitware/vtk-js/blob/master/.github/workflows/publish.yml#L39-L41

We still have api.md for the one that don't have d.ts yet and we also need to update [API documentation section](https://kitware.github.io/vtk-js/docs/develop_class.html#API-documentation) to recommand using d.ts instead.

floryst:
Thanks again!

PR #2174, opened 2021-11-23
fix(imagecropfilter): fix image crop filter instantiation

finetjul:
Since v21.0.0, default array values can be null but can't be empty array.

### PR and Code Checklist

### Context
https://discourse.vtk.org/t/unable-to-implement-imagecrop-filter/7254

### Changes

### Results
Fix ImageCropFilter example 

### Testing
  - **vtk.js**: master
  - **OS**: Windows 10
  - **Browser**: Chrome

### DIFF
--- a/Sources/Filters/General/ImageCropFilter/index.js
+++ b/Sources/Filters/General/ImageCropFilter/index.js
@@ -41,7 +41,7 @@ function vtkImageCropFilter(publicAPI, model) {
 
     const extent = input.getExtent();
     const cropped =
-      model.croppingPlanes.length === 6
+      model.croppingPlanes && model.croppingPlanes.length === 6
         ? extent.map((e, i) => {
             if (i % 2 === 0) {
               // min plane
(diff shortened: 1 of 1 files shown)


Sources/Filters/General/ImageCropFilter/index.js
```diff
@@ -130,7 +130,7 @@ function vtkImageCropFilter(publicAPI, model) {
   };
 
   publicAPI.isResetAvailable = () => {
-    if (model.croppingPlanes.length === 0) {
+    if (model.croppingPlanes == null || model.croppingPlanes.length === 0) {
```
jourdain, line 133: should you use `!model.croppingPlanes` to handle `undefined` and `null`?

jourdain: oups, you have `==` not `===`. I guess you are good there. ;-)

jourdain (approved):

PR #2178, opened 2021-11-25
fix(vtktubefilter): correcting the discrepancy in the number of point...

Nikhil216:
…s and normals of vtkTubeFilter

The polyData returned by the vtkTubeFilter has different number of points and normals when the
capping is on. The normals data is allocated more number of points depending on the numberOfSides
attribute of vtkTubeFilter.

### PR and Code Checklist

### Context

I encountered this issue while appending multiple tubeFilter outputs. The vtkTubeFilter has different number of points and normals when the capping is on. The normals in pointData has extra points which are all set to (0, 0, 0). I believe this line is to blame:
https://github.com/Kitware/vtk-js/blob/6e82cca81afe646206a1e12baed95b9aab329f73/Sources/Filters/General/TubeFilter/index.js#L666
The previous line isn't necessary as the points for capping are already accounted in `numNewPts` on this line:
https://github.com/Kitware/vtk-js/blob/6e82cca81afe646206a1e12baed95b9aab329f73/Sources/Filters/General/TubeFilter/index.js#L642

Here is a codepen to verify this claim.
https://codepen.io/nikhil216/pen/VwbWQME

### Changes

### Results

The vtkTubeFilter returns output where number of normals and points are equal.

### Testing

  - **vtk.js**: 21.1.2
  - **OS**: Windows 10

### DIFF
--- a/Sources/Filters/General/TubeFilter/index.js
+++ b/Sources/Filters/General/TubeFilter/index.js
@@ -661,10 +661,7 @@ function vtkTubeFilter(publicAPI, model) {
       size: numNewPts * 3,
       numberOfComponents: 3,
     });
-    let numNormals = 3 * numNewPts;
-    if (model.capping) {
-      numNormals = 3 * (numNewPts + 2 * model.numberOfSides);
-    }
+    const numNormals = 3 * numNewPts;
     const newNormalsData = new Float32Array(numNormals);
     const newNormals = vtkDataArray.newInstance({
       numberOfComponents: 3,
--- a/Sources/Filters/General/TubeFilter/test/testTubeFilter.js
+++ b/Sources/Filters/General/TubeFilter/test/testTubeFilter.js
@@ -112,7 +112,7 @@ test('Test vtkTubeFilter execution', (t) => {
     'Make sure the output number of points is correct with capping.'
   );
   t.ok(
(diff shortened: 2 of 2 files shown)


finetjul:
LGTM

jourdain:
Points should indeed be duplicated to properly handle the normals of the cap. Otherwise you will get a weird shading in case you share the connection points. I'm not saying there was no bug in the previous code base, but that is the reason of points duplication.

jourdain:
@finetjul

Nikhil216:
> Otherwise you will get a weird shading in case you share the connection points.
I tried to create this weird shading that you mentioned.
https://codepen.io/nikhil216/pen/VwbWQME
Is this the weird shading that you are talking about?

jourdain:
I could not see it. But the idea is that if the cap is sharing the same normal than the side of the tube, you should have some shading issue on the cap.

finetjul:
My understanding is that the points are already duplicated at extremities:
https://github.com/Kitware/vtk-js/pull/2178/files#diff-e4be6954e335829a60585792113143cfbda7429991bdcdd7f8b4d2ce3975e079R641-R643

What @Nikhil216 removed was another useless duplication.

jourdain:
Then we are all good. I was just mentioning that we need point duplication in case of capping.

PR #2183, opened 2021-11-30
docs(XR): Dynamically load XR polyfill in vtk.js VR examples

tbirdso:
### PR and Code Checklist

### Context

We are working on updating vtk.js for XR experiences with the WebXR API, however not all browsers support WebXR yet. The 
 WebXR Polyfill ([https://github.com/immersive-web/webxr-polyfill](https://github.com/immersive-web/webxr-polyfill)) allows backwards compatibility with browsers supporting WebVR or the Google Cardboard API.

The WebXR polyfill is distributed under an [Apache license](https://github.com/immersive-web/webxr-polyfill/blob/main/LICENSE). Rather than adding an extra dependency to vtk.js, we dynamically load the WebXR polyfill via CDN in XR-related examples on browsers where WebXR is not natively supported.

### Changes

- [x] Conditionally load WebXR polyfill from CDN in `Geometry/VR` and `Applications/SkyboxViewer` examples

### Results

- WebXR examples ([https://kitware.github.io/vtk-js/examples/VR.html](https://kitware.github.io/vtk-js/examples/VR.html), [https://kitware.github.io/vtk-js/examples/SkyboxViewer/SkyboxViewer.html?fileURL=https://data.kitware.com/api/v1/file/5ae8a89c8d777f0685796bae/download](https://kitware.github.io/vtk-js/examples/SkyboxViewer/SkyboxViewer.html?fileURL=https://data.kitware.com/api/v1/file/5ae8a89c8d777f0685796bae/download)) now work on desktop browsers implementing WebVR (i.e. Firefox) or mobile browsers with the Cardboard VR API.

### Testing

  - **vtk.js**: v21.3.1
  - **OS**: Windows 10
- [x] In `Geometry/VR` or `Applications/SkyboxViewer`, `navigator.xr` is now defined when trying to start the XR session. (Session init will still fail if an appropriate VR device is not available)


tbirdso:
@floryst We previously discussed adding the WebXR polyfill as an optional external dependency. Is adding the package to `package.json` and then calling with `import WebXRPolyfill from 'webxr-polyfill';`
- valid distribution under the Apache license, and
- a reasonable added dependency for vtk.js?

My impression is that adding this with `npm install --save` adds the package as a required dependency for vtk.js, which may not be reasonable as the vast majority of vtk.js functionality does not depend on WebXR. Is there a better way to condition this dependency on WebXR features being used?

floryst:
After some thought, do we need to support the polyfill in vtk.js? If we add it as a peer dependency, then we would need conditional imports, and while I suppose dynamic imports could fit the bill the user will still need to install the webxr polyfill themselves. At that point, I would add documentation indicating that, if you need the WebXR polyfill for your own projects, install it and perform the requisite import. That way we simply use the `navigator.xr` API without having to care whether it's a polyfill or not.

tbirdso:
@floryst I suppose that's reasonable. We still want our examples to be runnable on browsers that do not explicitly support WebXR yet. Can we rely on a CDN for individual examples that need to support XR? Per the polyfill repo:
```html
  <script src='https://cdn.jsdelivr.net/npm/webxr-polyfill@latest/build/webxr-polyfill.js'></script>
```

Where should relevant documentation be added?

floryst:
We can have a separate VR/AR/XR page in our docs that specifies how to use the WebXR polyfill. We can link to the WebXR polyfill project (https://github.com/immersive-web/webxr-polyfill) and instruct users to either use the CDN link or npm package approach.

tbirdso:
@floryst Agreed, it would be useful to add a discussion on backwards compatibility with the WebXR polyfill alongside the WebXR emulator in #2201.

In light of discussion I've changed this PR to dynamically load the WebXR polyfill in relevant vtk.js examples (`Geometry/VR` and `Applications/SkyboxViewer`) rather than statically loading as a vtk.js dependency. This adds a small amount of overhead for XR-based examples but better mirrors the developer scenario where vtk.js and the WebXR polyfill are consumed independently.

Examples/Applications/SkyboxViewer/index.js
```diff
@@ -16,8 +16,22 @@ import vtkURLExtract from 'vtk.js/Sources/Common/Core/URLExtract';
...
 
+// Dynamically load WebXR polyfill from CDN for WebVR and Cardboard API backwards compatibility
+if (navigator.xr === undefined) {
+  vtkResourceLoader
+    .loadScript(
+      'https://cdn.jsdelivr.net/npm/webxr-polyfill@latest/build/webxr-polyfill.js'
+    )
+    .then(() => {
+      // eslint-disable-next-line no-unused-vars, no-undef
+      const polyfill = new WebXRPolyfill();
```
floryst, line 31: I don't think we need to bind to a name. I'm assuming that the WebXRPolyfill constructor binds to `navigator.xr`, so we can drop the `polyfill` variable and the eslint exception.

tbirdso: That's fine. We still need to construct the polyfill with `new WebXRPolyfill();`, so the eslint exception `no-unused-vars` is replaced with `no-new`.

floryst:
@tbirdso I think this didn't get merged because I was waiting on docs, but turns out we had already merged docs mentioning using the polyfill. I'll merge it, since there wasn't any reason not to.

tbirdso:
Thanks @floryst !

PR #2184, opened 2021-12-01
feat(XR): Enable translation with WebXR gamepad API

tbirdso:
### PR and Code Checklist

### Context

In moving from WebVR to WebXR we must also integrate changes from the updated [WebXR gamepad API](https://www.w3.org/TR/webxr-gamepads-module-1/). VR controllers are exposed solely under the [`xr-standard` extension to the Gamepad API](https://www.w3.org/TR/gamepad/#dom-gamepadmappingtype) which remaps buttons/axes and exposes controller transforms as XRPose objects.

### Changes

- [x] `updateXRGamepads` is updated to reflect WebXR gamepads where controllers are identified by handedness rather than ID or index (see [https://www.w3.org/TR/webxr-gamepads-module-1/#gamepad-differences](https://www.w3.org/TR/webxr-gamepads-module-1/#gamepad-differences)).
- [x] VR locomotion in `InteractorStyleTrackballCamera` is revised from variable-speed translation based on the touchpad direction to fixed-speed translation in the direction of the right-handed controller. The WebXR Gamepad API requires that VR gamepads report a trigger button, while the touchpad/thumbstick is not guaranteed. Locomotion can be triggered with either a trigger press or (if available) a touchpad press.

### Results

- VR locomotion is re-introduced to `OpenGL/RenderWindow` and the `Geometry/VR` example so that the user may "fly" around their dataset. Locomotion is technically present in the `Applications/SkyboxViewer` example as well, but is not visible to the user as that example consist of just a skybox.

### Testing

  - **vtk.js**: v21.1.4
  - **OS**: Windows 10

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -12,11 +12,13 @@ const { vtkWarningMacro, vtkErrorMacro, normalizeWheel, vtkOnceErrorMacro } =
 // ----------------------------------------------------------------------------
 
 const deviceInputMap = {
-  'OpenVR Gamepad': [
-    Input.TrackPad,
+  'xr-standard': [
     Input.Trigger,
     Input.Grip,
-    Input.ApplicationMenu,
+    Input.TrackPad,
+    Input.Thumbstick,
+    Input.A,
+    Input.B,
   ],
 };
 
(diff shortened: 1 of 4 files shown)


martinken (approved):
LGTM +1

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -12,11 +12,13 @@ const { vtkWarningMacro, vtkErrorMacro, normalizeWheel, vtkOnceErrorMacro } =
 // ----------------------------------------------------------------------------
 
 const deviceInputMap = {
-  'OpenVR Gamepad': [
-    Input.TrackPad,
+  'xr-standard': [
```
floryst, line 15: Where does this get used?

tbirdso: In [line 451](https://github.com/Kitware/vtk-js/pull/2184/files#diff-6f8b2d910dfd95c1cc6d62e23311ede827ceae8efc6ad2b705fff3c15a01f3b3R451):
```js
input:
                deviceInputMap[gp.mapping] && deviceInputMap[gp.mapping][b]
                  ? deviceInputMap[gp.mapping][b]
                  : Input.Trigger,
```

Essentially we need to query what gamepad standard we are using so that we can translate from an ordered list of button inputs into meaningful enums about which buttons were pressed. I believe the idea of `deviceInputMap` is that we could extend this for other standards in the future. The [Gamepad API](https://www.w3.org/TR/gamepad/#dom-gamepadmappingtype) outlines the difference between the "standard" and "xr-standard" gamepad layouts.

tbirdso: @floryst Did this resolve your question?

floryst: Yep, thanks!

floryst:
LGTM

PR #2193, opened 2021-12-09
Update beta branch

floryst:
Beta branch still contains f1a04e2ad84b476f0b4ac98d2bfd2f972a25ad45 changes, but we are not yet merging it because we still are working on unifying the image data origin concept between both the ImageMapper and VolumeMapper. (As it stands, there is still an inconsistency if an image and a volume of the same image are shown together.)


floryst:
Update: CI is complaining about badly formatted commit messages. I will need to update the starting commit for analysis.

floryst:
Closing in favor of a separate PR.

Issue #2162, opened 2021-11-12
itk-wasm backwards compatibility with ITKHelper


weeeina:
### High-level description
With the itk.js --> itk-wasm migration, I believe I have discovered 2 minor changes that need to be made to keep ITKHelper compatible and able to read ITK file formats

### Steps to reproduce
Using the `vtkITKImageReader` to read an ITK image type, such as...
```
const niftiReader = vtkITKImageReader.newInstance();
    niftiReader.setFileName('image.nii');

    niftiReader
      .parseAsArrayBuffer(res)
      .then(() => {
        const source = niftiReader.getOutputData();
        return source;
      })
      .then((source) => {
        ...
```
You run into errors in ITKHelper as is saying that you cannot index into `undefined` (caused by itkImage.direction.data not existing) and then imageData always returning as null (caused by the switch statement always returning the default case, null).

### Detailed behavior
https://github.com/Kitware/vtk-js/blob/e8c8aec95323513e4b3baa754bdb27bf998b576e/Sources/Common/DataModel/ITKHelper/index.js#L51
Needs to be changed to:
```
      direction[col + idx * 3] = itkImage.direction[idx + col * itkImage.imageType.dimension];
```
because `itkImage.direction` is an array, not an object with key `data`

https://github.com/Kitware/vtk-js/blob/e8c8aec95323513e4b3baa754bdb27bf998b576e/Sources/Common/DataModel/ITKHelper/index.js#L73
Needs to be changed to:
```
  switch (ITKPixelTypes[itkImage.imageType.pixelType]) {
```
Otherwise, `itkImage.imageType.pixelType` is a string, e.g. `'Scalar'` while the switch statement looks for keys in `ITKPixelTypes`, which are all mapped to numbers. So something like `'Scalar'` should map to `1` and then the switch statement will work as expected. As is, this will always hit the default case as is and return null. 

### Expected behavior
Making these two changes reads .nii files as expected.

### Environment
- **vtk.js**: 17.17.2
- **itk-wasm**: ^1.0.0-a.8
- **OS**:  iOS 11.3.1
- **Browser**: Chrome 95.0.4638.69

Also using craco to put all of this into a react app. The craco.config I'm using is as follows:
```
const path = require('path');
const CopyWebpackPlugin = require('copy-webpack-plugin');

module.exports = () => ({
  overrideWebpackConfig: ({ webpackConfig, context: { env, paths } }) => {
    //console.log(webpackConfig)
[...]

PR #2196, opened 2021-12-13
ITKHelper itk-wasm compatibility fixes

weeeina:
### PR and Code Checklist

### Context
See description in issue: https://github.com/Kitware/vtk-js/issues/2162

### Changes

- remove itkImage.direction.data since it is an array instead of object
- add ITKPixelTypes lookup so that switch statement is not always default case

### Results

ITKHelper now successfully loads .nii files when used in following way:
```
const niftiReader = vtkITKImageReader.newInstance();
    niftiReader.setFileName('image.nii');

    niftiReader
      .parseAsArrayBuffer(res)
      .then(() => {
        const source = niftiReader.getOutputData();
        return source;
      })
      .then((source) => {
        ...
 ```

### Testing

- vtk.js: 17.17.2
- itk-wasm: ^1.0.0-a.8
- OS: iOS 11.3.1
- Browser: Chrome 95.0.4638.69

### DIFF
--- a/Sources/Common/DataModel/ITKHelper/index.js
+++ b/Sources/Common/DataModel/ITKHelper/index.js
@@ -4,8 +4,8 @@ import vtkDataArray from 'vtk.js/Sources/Common/Core/DataArray';
 
 const { vtkErrorMacro } = macro;
 
-// see itk.js/PixelTypes.js
-const ITKPixelTypes = {
+// see itk.js PixelTypes.js
+const ITKJSPixelTypes = {
   Unknown: 0,
   Scalar: 1,
   RGB: 2,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by thewtex — CHANGES_REQUESTED** [review]
@weeeina thank you for the contribution! :pray: 

A few improvements will result keep the code working that has not yet migrated and result in the desired behavior:

Near the top of the function, we should detect whether the `itkImage` is an `itk.js` `Image` or an `itk-wasm` `Image`.

This can look something like:

```
// Is this an itk.js Image or an itk-wasm Image?
const bool isITKWasm = itkImage.direction.data === undefined
```

then,

```
if (isITKWasm) {
  itkImage.direction[idx + col * itkImage.imageType.dimension];
} else {
   itkImage.direction[idx + col * itkImage.imageType.dimension];
}
```

To address the pixel type detection, rename `const ITKPixelTypes = {` to `const ITKJSPixelTypes = {`, and also define `const ITKWASMPixelTypes = {`. These were changed to descriptive values to provide a more explicit description when serialized. See:

https://github.com/InsightSoftwareConsortium/itk-wasm/blob/master/src/core/PixelTypes.ts

Then, something like this may work (needs to be tested):

```
const ITKPixelTypes = isITKWasm ? ITKWASMPixelTypes : ITKJSPixelTypes
```

weeeina:
@thewtex Thanks for the tips! I have made the changes to detect whether we are dealing with and itk or itk-wasm image in the convertVtkToItkImage function. However, there's also the convertItkToVtkImage function, so I've kept that as the ITKJSPixelTypes unless there is a way we would like to give the user a choice and/or move to the WASM version on that as well. Can you confirm whether it's okay that it goes itk.js OR itk-wasm --> vtk but vtk --> itk.js _only_?

thewtex:
@weeeina thanks for the updates!

We will also have to update vtk -> itk-wasm in addition to itk-wasm -> vtk, either in this PR or another.

weeeina:
> @weeeina thanks for the updates!
> 
> We will also have to update vtk -> itk-wasm in addition to itk-wasm -> vtk, either in this PR or another.

Thanks @thewtex! I'll leave the vtk --> itk-wasm to another PR, as I'm not entirely clear on how you would be able to differentiate based on what version of itk is being used by the user... 

I noticed that I messed up my first commit message and so the commitlint was not happy with that. Other than that, is there anything I need to do to get this ready for merge?

thewtex:
@weeeina thanks for the update!

I rebased on `master` to address the commit message issues.

Looks good to me!

thewtex (approved):

mhalle:
I think this patch isn't quite right yet. If using WASM, you want to do   
`switch (ITKPixelTypes[itkImage.imageType.pixelType]) {   // aliased to ITKWASMPixelTypes`
if using JS, you want to do:
`switch (itkImage.imageType.pixelType) {   // an integer `

PR #2200, opened 2021-12-16
fix(macro): enforce the creation of a copy when array is passed to se…

finetjul:
…t accessor

When calling obj.setMyProp(myArray), a copy of the array was not made
(a reference of the same array was kept). Changing myArray was therefore
also modifying the obj model.myProp.

Note that it breaks reslice cursor when camera viewUp was set outside
of the widget (e.g. with ctrl+left mouse)

### PR and Code Checklist

### Context
Found the issue with the reslice cursor widget example.
Ctrl click is rolling the view up of the camera
Then moving the plane line from another view was "NOT resetting" the viewUp.
It should not reset the viewUp but the code was not coded as such.

### Changes

### Results
The reslice cursor widget is now broken, the viewUp is being reset.

### Testing
  - **vtk.js**: master
  - **OS**:  Windows 10
  - **Browser**: Chrome


finetjul:
@LucasGandel

jourdain (approved):
Seems reasonable

floryst (approved):
LGTM

PR #2202, opened 2021-12-17
feat(XR): Add base WebXR AR support to OpenGL RenderWindow

tbirdso:
### PR and Code Checklist

### Context

First step in adding AR support as part of our WebXR investigation.

### Changes

- Updates `OpenGL/vtkRenderWindow` to support `immersive-ar` sessions via the [WebXR AR module](https://www.w3.org/TR/webxr-ar-module-1/)
- Forks VR cone example for a basic AR example

### Results

- AR is functional in WebXR emulator and in mobile device testing

cone-ar-emulator

### Testing

  - **vtk.js**: v21.3.0
  - **OS**: Windows 10
- [x] AR example functions on mobile and in WebXR emulator
  - See [WebXR emulator extension repo](https://github.com/MozillaReality/WebXR-emulator-extension) for install instructions
  - For testing on mobile, run the example with `npm test example -- AR` and visit the displayed LAN address in your mobile browser. May require adjusting firewall rules on your host machine.

### DIFF
--- a/Examples/Applications/GeometryViewer/index.js
+++ b/Examples/Applications/GeometryViewer/index.js
@@ -33,6 +33,7 @@ import icon from '../../../Documentation/content/icon/favicon-96x96.png';
 
 let autoInit = true;
 let background = [0, 0, 0];
+let fullScreenRenderWindow;
 let renderWindow;
 let renderer;
 let scalarBarActor;
@@ -115,13 +116,13 @@ function emptyContainer(container) {
 // ----------------------------------------------------------------------------
 
 function createViewer(container) {
-  const fullScreenRenderer = vtkFullScreenRenderWindow.newInstance({
+  fullScreenRenderWindow = vtkFullScreenRenderWindow.newInstance({
     background,
     rootContainer: container,
     containerStyle: { height: '100%', width: '100%', position: 'absolute' },
   });
-  renderer = fullScreenRenderer.getRenderer();
-  renderWindow = fullScreenRenderer.getRenderWindow();
+  renderer = fullScreenRenderWindow.getRenderer();
+  renderWindow = fullScreenRenderWindow.getRenderWindow();
   renderWindow.getInteractor().setDesiredUpdateRate(15);
 
   container.appendChild(rootControllerContainer);
@@ -139,7 +140,7 @@ function createViewer(container) {
     }
     fpsMonitor.setRenderWindow(renderWindow);
     fpsMonitor.setContainer(container);
-    fullScreenRenderer.setResizeCallback(fpsMonitor.update);
+    fullScreenRenderWindow.setResizeCallback(fpsMonitor.update);
   }
 }
 
@@ -194,6 +195,10 @@ function createPipeline(fileName, fileContents) {
   labelSelector.setAttribute('class', selectorClass);
   labelSelector.innerHTML = fileName;
 
+  const immersionSelector = document.createElement('button');
+  immersionSelector.setAttribute('class', selectorClass);
+  immersionSelector.innerHTML = 'Start AR';
+
   const controlContainer = document.createElement('div');
   controlContainer.setAttribute('class', style.control);
   controlContainer.appendChild(labelSelector);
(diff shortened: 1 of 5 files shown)


Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -311,6 +315,21 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
+        // Default setting for AR is slightly smaller / more distant
+        scale *= 3;
+        offset[2] *= -5;
+        renderer.getActiveCamera().setPhysicalScale(scale);
+        renderer.getActiveCamera().setPhysicalTranslation(offset);
+      } else {
+        // Move VR scene a bit forward so that we don't start in the middle of objects
+        offset[2] *= -3;
+        renderer.getActiveCamera().setPhysicalTranslation(offset);
+      }
```
tbirdso, line 331: These settings in `enterXR` adjust the world-to-physical matrix we originally get from `renderer.resetCamera()` a bit so that the scene is a reasonable size for VR vs AR. The values here are arbitrary and based on testing for what makes the cone scene "comfortable", I'd welcome suggestions on how we can refine this.

tbirdso: After more testing, these adjustments don't completely solve the positioning issue. Half the time the offset will move the scene in the wrong direction so that the scene is behind the viewer. Scaling/positioning the scene for VR or AR experiences is still a consideration that should be pursued, but perhaps is best left out of these initial changes.

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -362,6 +368,12 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
 
     if (xrPose) {
       const gl = publicAPI.get3DContext();
+
+      if (model.xrSessionIsAR) {
+        gl.canvas.width = model.oldCanvasSize[0];
```
floryst, line 373: Are we guaranteed that `model.oldCanvasSize` is not undefined here? I'm asking because the previous usage of `model.oldCanvasSize` checks against the case where `oldCanvasSize` has not yet been set on the model.

tbirdso: Good call, I don't think we can guarantee that `model.oldCanvasSize` is defined. In practice it succeeds for the AR case but there's no harm in adding the extra check just in case.

tbirdso: Fixed.

martinken:
Only thing I would suggest is adding basic (just empty or similar)  startXR methods (or some query for the feature) to RenderWindowViewNode which is the superclass for API specific RenderWindows. Otherwise with WebGPU I think it will cause console errors yes?

tbirdso:
@martinken I haven't specifically tested for WebGPU behavior, are there resources I could use to look into this?

In any case, isn't it reasonable to throw a console error if we try to enter XR on an unsupported backend?

martinken:
If we only call them when someone is asking to go into VR then sure (which looks like the case here) , though the console message will be that the method doesn't exist as opposed to something nicer. Might be nice to have a query method for if the API supports it so we could hide the toggle. e.g. if on an API (e.g. WebGPU) that doesn't support XR, don;t show the toggle.  So even if the browser supports WebXR, if the backend doesn;t support it we don;t show the GUI toggle. It could be something like RenderWindowViewNode.getSupportsFeature('WebXR') and defaults to false for everything. Then in WebGL we return true for that.

Examples/Applications/GeometryViewer/index.js
```diff
@@ -202,6 +207,14 @@ function createPipeline(fileName, fileContents) {
   controlContainer.appendChild(colorBySelector);
   controlContainer.appendChild(componentSelector);
   controlContainer.appendChild(opacitySelector);
+
+  if (
+    navigator.xr !== undefined &&
+    navigator.xr.isSessionSupported('immersive-ar') &&
+    fullScreenRenderWindow.getApiSpecificRenderWindow().getXrSupported()
```
tbirdso, line 214: `Applications/GeometryViewer` will only include a button for AR if 1) WebXR is available, 2) the device supports AR, and 3) the rendering backend supports AR (aka WebGL and not WebGPU)

Examples/Geometry/AR/index.js
```diff
@@ -0,0 +1,107 @@
...
+renderer.resetCamera();
+renderWindow.render();
+
+// -----------------------------------------------------------
+// UI control handling
+// -----------------------------------------------------------
+
+fullScreenRenderer.addController(controlPanel);
+const arbutton = document.querySelector('.arbutton');
+arbutton.disabled = !fullScreenRenderer
```
tbirdso, line 81: The `Geometry/AR` example will always include a button but disable it to show that AR is not supported if necessary

martinken:
Love it!

martinken (approved):

PR #2203, opened 2021-12-20
feat(XR): Adjust world-to-physical factors for XR modalities

tbirdso:
### PR and Code Checklist

### Context

There are different factors that we want to consider when transforming data for viewing in AR or VR. In AR we are probably using a small window such as a phone screen and want to display data in a smaller physical region of space, such as in front of us on a table. In VR we are using an HMD and are more likely to want to walk around the data representation at room scale. Each requires that the default camera parameters after resetting the scene with `renderer.resetCamera` be adjusted based on the current modality.

~Draft PR requires further in-person testing to determine how bounded XR sessions where the HMD is not at the physical origin to start will be handled. Current testing with physical origin defined by device origin at session initialization (real+emulated AR, emulated VR) show good scene positioning.~

After some in-person testing with different WebXR reference spaces I've concluded that for now it may be best to avoid overcomplicating with different reference spaces. Instead I've exposed the default rescale/translation settings as parameters that a hypothetical application could override for individual uses. As we continue to expand our WebXR support we can delve further into advantages of introducing the concept of user height and bounded spaces for user comfort.

### Changes

- Add a call within `enterXR` to explicitly update camera world-to-physical factors based on the immersive modality.
- Expose default values for world-to-physical adjustments so that applications can reset based on individual use cases.

### Results

- AR objects are displayed at 1/4 size and by default 0.5m in front of the camera.
- VR objects are displayed by default at 0.7m in front of the camera

### Testing

  - **vtk.js**: master
  - **OS**: Windows


Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -320,6 +326,37 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
+    let rescaleFactor = 0.0;
+
+    if (model.xrSessionIsAR) {
+      // Scale down + translate forward in AR
+      rescaleFactor = 0.25;
+      distToTranslateZ = -0.5;
+    } else {
+      // Translate without scaling in VR
+      rescaleFactor = 1.0;
+      distToTranslateZ = -1.0;
```
tbirdso, line 342: Based on testing, initializing the scene at 1m in front of the user is ok for seated VR experiences but potentially out of bounds for a standing VR walk-around experience. Need to experiment more with different XR reference spaces before settling on init factors.

martinken (approved):
Seems reasonable to me +1

floryst:
One question: are your rescale and translation constants empirically determined from the VR/AR emulators?

Overall, LGTM

tbirdso:
@floryst Good question, I settled on these from testing in-person with VR/AR hardware rather than the emulators, though the results are also visibly better in emulation

floryst:
Sounds good. We can make adjustments as real-world use-cases come to light.

floryst (approved):

tbirdso:
To clarify, my thought is that calling applications can call `resetXRScene` with application-specific parameters as needed, so ideally we won't need to adjust the default parameters directly. Either way, can easily revisit as needed in the future.

PR #2207, opened 2022-01-05
fix(SynchronizableRenderWindow): prevent invalid object modification

jourdain:
### PR and Code Checklist

### Context

SynchronizedRenderWindow was rebuilding several objects even if their content were remaining the same.
This fix that issue which improve greatly the performances when using local rendering with remote vtk view object.

### Changes

No API change

### Results

See major perf boost when dealing with large geometries

### Testing

Manual testing using trame app in `examples/VTK/PerfTests/scene-export-time.py`.

### DIFF
--- a/Sources/Rendering/Core/ColorTransferFunction/index.js
+++ b/Sources/Rendering/Core/ColorTransferFunction/index.js
@@ -214,22 +214,29 @@ function vtkColorTransferFunction(publicAPI, model) {
   // Set nodes directly
   publicAPI.setNodes = (nodes) => {
     if (model.nodes !== nodes) {
+      const before = JSON.stringify(model.nodes);
       model.nodes = nodes;
-      publicAPI.sortAndUpdateRange();
+      const after = JSON.stringify(model.nodes);
+      return publicAPI.sortAndUpdateRange() || before !== after;
     }
+    return false;
   };
 
   //----------------------------------------------------------------------------
   // Sort the vector in increasing order, then fill in
   // the Range
   publicAPI.sortAndUpdateRange = () => {
+    const before = JSON.stringify(model.nodes);
     model.nodes.sort((a, b) => a.x - b.x);
+    const after = JSON.stringify(model.nodes);
 
     const modifiedInvoked = publicAPI.updateRange();
     // If range is updated, Modified() has been called, don't call it again.
-    if (!modifiedInvoked) {
+    if (!modifiedInvoked && before !== after) {
       publicAPI.modified();
+      return true;
     }
+    return modifiedInvoked;
   };
 
   //----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -214,22 +214,29 @@ function vtkColorTransferFunction(publicAPI, model) {
   // Set nodes directly
   publicAPI.setNodes = (nodes) => {
     if (model.nodes !== nodes) {
+      const before = JSON.stringify(model.nodes);
       model.nodes = nodes;
-      publicAPI.sortAndUpdateRange();
+      const after = JSON.stringify(model.nodes);
+      return publicAPI.sortAndUpdateRange() || before !== after;
```
finetjul, line 220: shouldn't `after` be set AFTER sortAndUpdateRange() is called ?

what if calling `setNodes([0, 2, 1])` with `model.nodes = [0, 1, 2]` ? `modified()` should NOT be called and false be returned. This is not the case here.

Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -214,22 +214,29 @@ function vtkColorTransferFunction(publicAPI, model) {
...
     }
+    return false;
   };
 
   //----------------------------------------------------------------------------
   // Sort the vector in increasing order, then fill in
   // the Range
```
finetjul, line 227: you might want to update the documentation to explain what is returned

Sources/Rendering/Core/ColorTransferFunction/index.js
jourdain: I see the idea, but the result would be the same as the `sortAndUpdateRange()` would have then triggered a modified(). 

This mostly fixed the intended use case when nodes provided have the same content/layout as the one stored in the instance. We could push to support unsorted nodes and skip the modified, but so far I don't see the use case.

Sources/Rendering/Core/ColorTransferFunction/index.js
jourdain: done (did not create a release for it) but one will be available soon with the speed of vtk.js releases. ;-)

Sources/Rendering/Core/ColorTransferFunction/index.js
finetjul: I see your point. It will be a hard to find bug in case such unlikely scenario happens.
But I do not have a quick and elegant solution (I'm actually concerned that `stringify` gets costly in case of large CTF)  to suggest so I guess it's fine to leave it as is.

jourdain: modified() is not that evil. But once you have large meshes, you see the performance hit when GL structures get rebuilt. After it is just a matter of figuring which object mtime trigger the rebuild. Once you have a use case highlighting the issue, it is easy to optimize it or suggest changes on the app side to properly get the intended behavior. That issue never arose before, until I was dealing with large meshes with the SyncView.

PR #2208, opened 2022-01-05
feat(rendering): support wide lines with vtkMapper and vtkMapper2D

sankhesh:
When the line width attribute on vtkProperty2D is more than 1, the 2D mapper makes multiple
instanced draw calls and the vertex shader offsets the vertex position based on the instance index.

### PR and Code Checklist

- [ x ] [semantic-release](https://github.com/semantic-release/semantic-release) commit messages
- [ x ] Run `npm run reformat` to have correctly formatted code

### Results

Screenshot from 2022-01-04 23-26-02
Screenshot from 2022-01-04 23-25-31

### DIFF
--- a/Sources/Rendering/OpenGL/Helper/index.js
+++ b/Sources/Rendering/OpenGL/Helper/index.js
@@ -3,6 +3,19 @@ import vtkCellArrayBufferObject from 'vtk.js/Sources/Rendering/OpenGL/CellArrayB
 import vtkShaderProgram from 'vtk.js/Sources/Rendering/OpenGL/ShaderProgram';
 import vtkVertexArrayObject from 'vtk.js/Sources/Rendering/OpenGL/VertexArrayObject';
 
+import { Representation } from 'vtk.js/Sources/Rendering/Core/Property/Constants';
+
+export const primTypes = {
+  Start: 0,
+  Points: 0,
+  Lines: 1,
+  Tris: 2,
+  TriStrips: 3,
+  TrisEdges: 4,
+  TriStripsEdges: 5,
+  End: 6,
+};
+
 // ----------------------------------------------------------------------------
 // vtkOpenGLHelper methods
 // ----------------------------------------------------------------------------
@@ -12,7 +25,8 @@ function vtkOpenGLHelper(publicAPI, model) {
   model.classHierarchy.push('vtkOpenGLHelper');
 
   publicAPI.setOpenGLRenderWindow = (win) => {
-    model.program.setContext(win.getContext());
+    model.context = win.getContext();
+    model.program.setContext(model.context);
     model.VAO.setOpenGLRenderWindow(win);
     model.CABO.setOpenGLRenderWindow(win);
   };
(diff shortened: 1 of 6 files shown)


Sources/Rendering/OpenGL/PolyDataMapper2D/index.js
```diff
@@ -514,6 +520,29 @@ function vtkOpenGLPolyDataMapper2D(publicAPI, model) {
...
+        [
+          'VTK::PositionVC::Impl',
+          ' if (gl_InstanceID != 0)',
+          ' {',
+          '   vec4 tmpPos = gl_Position;',
+          '   vec3 tmpPos2 = tmpPos.xyz / tmpPos.z;',
+          '   tmpPos2.x = tmpPos2.x + 1.4*(float(gl_InstanceID % 2) - 0.5)/viewportSize[0];',
+          '   tmpPos2.y = tmpPos2.y + 1.4*(float(gl_InstanceID / 2) - 0.5)/viewportSize[1];',
+          '   gl_Position = vec4(tmpPos2.xyz * tmpPos.w, tmpPos.w);',
+          ' }',
```
martinken, line 541: So this is my semi working code which is fine for emphasizing line widths of 2 or so. But we probably should suck it up and do the right thing here. So what does this do. Basically we are trying to do a sort of dilation by the line width. So we shift the line in X and then Y and then both. But for larger line widths it isn't correct as gl_InstanceID % 2 is always either 0 or 1. So it can only shift at most 1.4/2.0 pixels in X.  (1.4 from the code, the 2.0 is because device coordinates are -1.0 to 1.0 for a range of 2.0)  The code for the Y shift is better but the -0.5 only centers it for 2 pixels, larger widths will result in the line being wider on one side than the other. Not a huge issue but still fixable so let's fix it. To fix it we need something like the following
```
float offset = floor(gl_InstanceID / 2) * lineWidthStepSize - halfLineWidth;
tmpPos2.x = tmpPos2.x + 2.0*offset*float(gl_InstanceID % 2)/viewportSize[0];
tmpPos2.y = tmpPos2.y + 2.0*offset*float((gl_InstanceID + 1) %2)/viewportSize[1];
```

I forget if gl_InstanceID starts at 0 or 1 so these may be off by one. The idea is to create an X centered about the original line vertex of the right width. The number of instances should be a ceil(lineWidth)*2 as even a 1.5 width line will require 4 invocations, a 2.1 width line would require 6 etc. The halfLineWidth should probably be float lineWidth/2.0 and the stepSize should be float lineWidth/ceil(lineWidth)  e.g. for a width of 2.1 we have 6 invocations each shifting the pixel by 2.1/3 pixels on X or Y. Some of this I tested out when I was messing around earlier with this, but there are probably some issues in it. Like I worry that moving by exactly one pixel could cause some holes due to rasterization issues etc. So there may need to be a 0.99 in there somewhere to move slightly less than a full pixel etc. I can expand more on it later this afternoon if you want but that is the basic gist of what I think we need to do.

Sources/Rendering/OpenGL/PolyDataMapper2D/index.js
```diff
@@ -514,6 +520,29 @@ function vtkOpenGLPolyDataMapper2D(publicAPI, model) {
     const GSSource = shaders.Geometry;
     const FSSource = shaders.Fragment;
 
+    if (publicAPI.haveWideLines(ren, actor)) {
+      VSSource = vtkShaderProgram.substitute(
+        VSSource,
+        '//VTK::PositionVC::Dec',
+        ['VTK::PositionVC::Dec', 'uniform vec2 viewportSize;']
+      ).result;
+      VSSource = vtkShaderProgram.substitute(
```
finetjul, line 529: it should probably be in a helper to be reused by PolyDataMapper as well

Sources/Rendering/OpenGL/PolyDataMapper2D/index.js
sankhesh: @martinken This logic helps with creating a `+` centered about the vertex but does not account for the line normal.

## Expected

Screenshot from 2022-01-10 09-06-39

## Result

Screenshot from 2022-01-10 09-07-00

martinken: There must be a bug in the code to create the + as a + should look fine in that example. I'll take a look.

martinken: A couple problems with that code. First instanceID starts at zero even when using instancing. Second you are always moving along the +X -Y diagonal as your sign is the same for X and Y. It needs to be different (not just by -). More like the code I originally posted. e.g. even values offset X with a 0 Y offset, odd values offset Y with a 0 X offset. You can use the even odd of the instance ID to determine if you are offsetting X or Y.

Sources/Rendering/OpenGL/PolyDataMapper2D/index.js
```diff
@@ -514,6 +520,42 @@ function vtkOpenGLPolyDataMapper2D(publicAPI, model) {
...
+        VSSource,
+        '//VTK::PositionVC::Impl',
+        [
+          '//VTK::PositionVC::Impl',
+          ' if (halfLineWidth > 0.0)',
+          '   {',
+          '   float inst = float(gl_InstanceID);',
+          '   float offset = 0.5 * inst * lineWidthStepSize - halfLineWidth;',
+          '   vec4 tmpPos = gl_Position;',
```
martinken, line 549: offset should probably be
```
float offset = float(gl_InstanceID / 2) * lineWidthStepSize - halfLineWidth;
```

aylward:
Can you add a test for this new functionality?   Ken suggested that perhaps the example can be made into a test.

sankhesh:
> Can you add a test for this new functionality? Ken suggested that perhaps the example can be made into a test.

Definitely 👍

Sources/Rendering/OpenGL/Helper/index.js
```diff
@@ -22,13 +36,204 @@ function vtkOpenGLHelper(publicAPI, model) {
...
+          0,
+          model.CABO.getElementCount(),
+          2 * Math.ceil(actor.getProperty().getLineWidth())
+        );
+      } else {
+        gl.lineWidth(actor.getProperty().getLineWidth());
+        publicAPI.updateShaders(ren, actor, oglMapper);
+        gl.drawArrays(mode, 0, model.CABO.getElementCount());
+        // reset the line width
+        gl.lineWidth(1);
```
finetjul, line 61: maybe linewidth was not 1 before

Sources/Rendering/OpenGL/Helper/index.js
```diff
@@ -22,13 +36,204 @@ function vtkOpenGLHelper(publicAPI, model) {
...
+          return false;
+        }
+      }
+      return true;
+    }
+    return false;
+  };
+
+  publicAPI.getNeedToRebuildShaders = (ren, actor, oglMapper) => {
+    // has something chaged that would require us to recreate the shader?
```
finetjul, line 105: chaged -> changed

Sources/Rendering/OpenGL/Helper/index.js
finetjul: fair enough

martinken (approved):
LGTM

PR #2212, opened 2022-01-06
refactor(WebGL): remove support for WebGL1

martinken:
Now that safari supports webgl2 move to that as the webg1 code
and conditionals added a lot of complexity and limited what
we could do.

Now that WebGL2 is fairly universal this change should not
break anything other than iOS devices still running iOS 14
or earlier.

There is a second step (not in this MR) which is to update vtk.js
to assume that we have depth textures. This will remove the need
for the opaque depth pass code and some other sections of code
but to keep this MR reasonably small it is being left out.

- [ ] Up-to-date market analysis (https://mixpanel.com/trends/#report/ios_15)

### DIFF
--- a/Sources/Rendering/OpenGL/Framebuffer/index.js
+++ b/Sources/Rendering/OpenGL/Framebuffer/index.js
@@ -31,7 +31,7 @@ function vtkFramebuffer(publicAPI, model) {
   };
 
   publicAPI.saveCurrentBuffers = (modeIn) => {
-    // noop on webgl 1
+    // noop on webgl
   };
 
   publicAPI.restorePreviousBindingsAndBuffers = (modeIn) => {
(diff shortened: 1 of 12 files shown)


martinken:
Going to let this sit for a bit as I'm not sure how much iOS 14 is still out there in the wild right now.

floryst:
Here are Mixpanel's iOS 14 and 15 usage stats: https://mixpanel.com/trends/#report/ios_15

jourdain:
We should still ask the community once we are getting closer to the time where we want to merge that. As I'm worried that some of the usage of vtk.js would be in hospitals on iOS devices... On the other hand, they could still use older version of vtk.js to keep things working...

daker:
iOS 17 is released and iOS 14 market share is less than 0.05% https://telemetrydeck.com/blog/ios-market-share-13-23/

daker:
Three.js will remove support for WebGL1 in r163 which is due to March or April 2024

daker:
I think iOS 14 & 15 have almost disappeared from the market

floryst:
> I think iOS 14 & 15 have almost disappeared from the market

We will revisit this when someone gets the bandwidth!

daker:
To close, replaced with #3523

PR #2221, opened 2022-01-12
feat(XR): Handle WebXR XRInputSourceEvents

tbirdso:
### PR and Code Checklist

### Context

The [WebXR API](https://developer.mozilla.org/en-US/docs/Web/API/XRInputSourceEvent) exposes a limited number of user input events, which we expose to the user here through `RenderWindowInteractor` alongside other typical browser events.

Note that the granularity offered through XR events does not match that of the XR gamepad API, through which we can query continuous values such as touchpad or joystick position through our [`updateXRGamepads`](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/RenderWindowInteractor/index.js#L420) method. Developers should take both approaches into account.

### Changes

- [x] WebXR events are exposed through `RenderWindowInteractor`.

### Results

Exposes six new events ("select"/start/end, "squeeze"/start/end) that developers can choose to handle.

### Testing

  - **vtk.js**: `master`
  - **OS**: Windows
  - **Browser**: Chrome 97.0.4692.71 (Windows 10)

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -59,6 +59,12 @@ const handledEvents = [
   'StartInteraction',
   'Interaction',
   'EndInteraction',
+  'StartXrSelect',
+  'XrSelect',
+  'EndXrSelect',
+  'StartXrSqueeze',
+  'XrSqueeze',
+  'EndXrSqueeze',
 ];
 
 function preventDefault(event) {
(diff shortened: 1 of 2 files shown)


tbirdso:
Note that XR events seem at the moment to be the best way to handle mobile AR events, which will usually be simpler (i.e. touching the screen). VR events may be more complicated due to different controller inputs/configurations.

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -59,6 +59,12 @@ const handledEvents = [
   'StartInteraction',
   'Interaction',
   'EndInteraction',
+  'StartXrSelect',
+  'XrSelect',
+  'EndXrSelect',
+  'StartXrSqueeze',
+  'XrSqueeze',
+  'EndXrSqueeze',
```
floryst, line 67: I think capitalizing the 'R' in 'XR' for consistency is a good idea here.

floryst (commented):
Looks good so far. Just a small comment on capitalization.

tbirdso:
Note that I do not expect to be able to revisit this PR. Will leave open in case others would like to take up this development, otherwise project maintainers are welcome to close it.

Sources/Rendering/Core/RenderWindowInteractor/index.js
bourdaisj: still relevant?

bourdaisj:
this PR is probably useful so we should just merge it someday imo

PR #2222, opened 2022-01-12
fix(XR): Resolve mobile AR issue with absent gripspace

tbirdso:
### PR and Code Checklist

### Context

Fixes an issue where an XRInputSource is not guaranteed to have a `gripSpace` property. In the case of the [AR example](https://kitware.github.io/vtk-js/examples/AR.html) tapping the screen creates an input source with a `targetRaySpace` but not a `gripSpace` because there is no independently tracked game controller.

### Changes

- [x] Handle absent `gripSpace` to prevent exceptions.

### Results

Previous behavior: tapping the mobile device screen in the [AR example](https://kitware.github.io/vtk-js/examples/AR.html) causes an exception to be thrown and halts rendering.

New behavior: tapping the screen does not throw an exception or affect rendering.

### Testing

  - **vtk.js**: `master`
  - **OS**: Windows
  - **Browser**: Chrome 97.0.4692.71 (Windows 10)


floryst (approved):
LGTM

PR #2226, opened 2022-01-13
fix(rendering): take inverseShiftScaleMatrix into account when comput…

### LINKED ISSUE #2225: PolyDataMapper coordinate shift and scale not applied to clipping planes
### High-level description

Due to webgl vertex buffer precision limitations, polydata vertex coordinates are currently centered and scaled in CellArrayBufferObject when some size condition is met. However, clipping planes coordinates are computed without accounting for this eventual shift and scale, therefore leading to incorrect clipping.

### Steps to reproduce

* Create a large polydata source (e.g. sphere with radius of 30)
* Add a clipping plane to its mapper
* Notice that its clipping is incorrect

### Detailed behavior

* Live example of the issue, on a sphere of radius 30, with a plane supposed to clip z < 0.25: https://codepen.io/Lillypucien/pen/gOGZwwM
* Live reference example, on a smaller sphere (radius 5), on which there is no coordinate shift, and therefore no bug: https://codepen.io/Lillypucien/pen/RwLEGVr
* Screenshot of the issue with the clipping plane drawn, on a sphere of radius 30:
bigSphereClipped
* Reference screenshot on a smaller sphere (radius 5):
smallSphereClipped

### Expected behavior

When large polydata coordinates are scaled and shifted, clipping planes should also account for this scale and shift in their clipping computations.

### Environment
- **vtk.js**: 22.0.0
- **OS**: Windows 10
- **Browser**: Chrome 97.0.4692.71

Lillypucien:
…ing clipping plane equations

fix #2225

### PR and Code Checklist

### Context

fix #2225

vtkPolyDataMapper did not take into account coordinates shift and scale when computing clipping plane equations.

### Changes

Rendering/OpenGL/PolyDataMapper now checks if shift and scale was applied to coordinates, and applies the inverseShiftAndScaleMatrix to the plane coordinates if necessary.

### Results

Clipping planes now clip correctly large polydata (e.g. sphere with radius 30).

See live examples and screenshots in issue #2225.

### Testing

  - **vtk.js**: 22.0.0
  - **OS**: Windows 10
  - **Browser**: Chrome 97.0.4692.71


floryst:
@martinken can you review?

floryst:
FYI the ImageMapper tests are failing here. Running `npm test` will generate a `Utilities/TestResults` folder containing an HTML file with more test info.

Lillypucien:
If I understand correctly the test results, it seems that ImageMapper also doesn't account for coordinates' shift/scale, and that the reference image may reflect an unwanted behaviour.


image

floryst:
> Considering that both the image slice and polydata are positioned at (0.3, 0, 0), with a 30° rotation around z, and that the clipping plane is positioned at (0, 0, 0) with a normal of (0.707, 0.707, 0), it shouldn't clip anything, unlike what is seen in the reference image.

I would have assumed that the polydata gets clipped in this case, since the clipping plane intersects both the polydata and the image. I might be missing where the polydata and image is spatially located with respect to the clipping plane, though.

@thewtex can you comment on this issue? It does appear that the image mapper is having clip issues, as per the last two images above.

Lillypucien:
> I would have assumed that the polydata gets clipped in this case, since the clipping plane intersects both the polydata and the image. I might be missing where the polydata and image is spatially located with respect to the clipping plane, though.

If my math is correct, the image's corners are at (0, 0, 0), (173, 100, 0), (-100, 173, 0), (73, 273, 0), all translated by a negligible (0.3, 0, 0). So a clipping plane positioned at (0, 0, 0) with a (1, 1, 0) normal doesn't intersect nor clip the image ? And neither does it clip the slightly smaller polydata.

floryst:
@Lillypucien thanks for your detailed explanation. Would it be possible for you to examine and fix the same issue for the ImageMapper as well? If not, let us know and I'll bring @martinken in to help.

Lillypucien:
I don't think I'm familiar enough with the ImageMapper to do the same, it's probably best to let @martinken look at it

thewtex:
Indeed, the same issue was present with the ImageMapper.

I am pushing a fix along with new test baselines covering the cases @Lillypucien mentioned.

testImageCroppingPlanesNoRotation

testImageCroppingPlanesRotation

testImageCroppingPlanesRotationClipPlane

thewtex:
This only addresses the WebGL case. Issue #2293 created to track WebGPU.

thewtex:
@Lillypucien thanks for the contribution! :pray:

PR #2229, opened 2022-01-14
fix(RenderWindowInteractor): fix undefined error in publicAPI.getFirs…

Enside:
### Context
UI Framework: Vue 3.x
Related Example: ResliceCursorWidget
Related File: RenderWindowInteractor
Description: Using compositon-api and calling genericRenderWindow.delete in unBeforeMount life cycle callback, there will be an error with `read property getRenderersByReference of undefined` in getFirstRenderer or `read property getRenderers of undefined` in findPokedRenderer, then the whole render broke down. It seems that vtkRenderWindowInteractor still doing something when render is deleting.

### Changes
Add some code for null value judgment.

### Results
Property is not read on null value and render will not break down. But Macro logs an error 'Can not forward events without a current renderer on the interactor' without influence of use.

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -701,7 +701,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
   };
 
   publicAPI.getFirstRenderer = () =>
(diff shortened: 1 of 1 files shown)


jourdain:
LGTM

PR #2230, opened 2022-01-14
refactor(macros): optimize code structure of shouldUpdate

Enside:
Just optimize code structure of shouldUpdate.

### DIFF
--- a/Sources/macros.js
+++ b/Sources/macros.js
@@ -74,6 +74,7 @@ export function vtkWarningMacro(...args) {
 }
 
 const ERROR_ONCE_MAP = {};
+
 export function vtkOnceErrorMacro(str) {
   if (!ERROR_ONCE_MAP[str]) {
     loggerFunctions.error(str);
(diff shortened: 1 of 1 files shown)


jourdain:
LGTM

Sources/macros.js
```diff
@@ -763,22 +755,13 @@ export function algo(publicAPI, model, numberOfInputs, numberOfOutputs) {
     count = model.numberOfInputs;
     while (count--) {
       if (
-        model.inputConnection[count] &&
-        model.inputConnection[count].filter.shouldUpdate()
+        model.inputConnection[count]?.filter.shouldUpdate() ||
```
finetjul, line 758: @floryst @jourdain are we OK with optional chaining in VTK.JS ?

jourdain: I thought we are based on the latest update of the toolchain. Am I correct @floryst ?

floryst: Yep, we handle optional chaining (and null coalescing, for that matter).

finetjul: Good to know, thanks !

PR #2232, opened 2022-01-18
feat(scalarbaractor): add option to hide extra colors

boucaud:
### PR and Code Checklist

### Context

We would like to have the ability to not show Nan and above colors.
### Changes

Add three flags `drawNanAnnotation`, `drawAboveRangeSwatch` and `drawBelowRangeSwatch` to toggle the display of extra colors

### Results

Users can now toggle whether to show extra color bars.
### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


boucaud:
@finetjul WDYT ?

Sources/Rendering/Core/ScalarBarActor/index.d.ts
```diff
@@ -212,8 +218,14 @@ export interface vtkScalarBarActor extends vtkActor {
 	setScalarsToColors(scalarsToColors: vtkScalarsToColors): boolean;
 
 	/**
-	 * 
-	 * @param tickLabelPixelOffset 
+	 * Control whether to show extra colors
```
finetjul, line 221: please explain what extra colors are (nan, above, below...)

finetjul:
Why not reuse the same variable names in C++ (e.g. `DrawNanAnnotation`) ?

boucaud:
> Why not reuse the same variable names in C++ (e.g. `DrawNanAnnotation`) ?

I did not check whether these flags were present in the C++ implementation. I have adapted these changes to be closer to the C++ interface.

Sources/Rendering/Core/ScalarBarActor/index.js
```diff
@@ -682,12 +685,27 @@ function vtkScalarBarActor(publicAPI, model) {
...
+    }
+    if (
+      model.drawAboveRangeSwatch &&
+      typeof model.scalarsToColors.getAboveRangeColor === 'function'
+    ) {
+      numberOfExtraColors += 1;
+    }
 
-    const numPts = 4 + (haveExtraColors ? 12 : 0);
+    const numPts = 4 * (1 + numberOfExtraColors);
```
finetjul, line 708: odd, it used to be `4 * 12` and now it is `4 * 4`

Sources/Rendering/Core/ScalarBarActor/index.js
```diff
@@ -682,12 +685,27 @@ function vtkScalarBarActor(publicAPI, model) {
...
+    }
+    if (
+      model.drawBelowRangeSwatch &&
+      typeof model.scalarsToColors.getBelowRangeColor === 'function'
+    ) {
+      numberOfExtraColors += 1;
+    }
+    if (
+      model.drawAboveRangeSwatch &&
+      typeof model.scalarsToColors.getAboveRangeColor === 'function'
```
finetjul, line 703: shouldn't it be `haveAbove` instead ?

Sources/Rendering/Core/ScalarBarActor/index.js
boucaud: It was `4+12` before, now it's `4+4*numberOfExtras`

finetjul: my bad, did not read correctly

Sources/Rendering/Core/ScalarBarActor/index.js
```diff
@@ -682,12 +685,29 @@ function vtkScalarBarActor(publicAPI, model) {
...
+    if (
+      model.drawNanAnnotation &&
+      typeof model.scalarsToColors.getNanColor === 'function'
+    ) {
+      numberOfExtraColors += 1;
+    }
+    if (
+      model.drawBelowRangeSwatch &&
+      typeof model.scalarsToColors.getBelowRangeColor === 'function' &&
+      model.scalarsToColors.getUseBelowRangeColor()
```
finetjul, line 698: strictly speaking, it is not because getBelowRangeColor is a function that get**Use**BelowRangeColor is a function.

I'm not sure about the difference between those 2 functions...

jourdain: `scalarsToColors` seems to be a vtk instance and you are tapping directly to methods from its publicAPI. To some extent, I'm not sure checking for type is needed as if present, the publicAPI "~only" have methods.

jourdain:
Seems reasonable. Do we have an image test?

boucaud:
@finetjul @jourdain I've added two image tests and fixed the publicAPI checks.

Sources/Rendering/Core/ScalarBarActor/index.js
```diff
@@ -382,20 +382,20 @@ function vtkScalarBarActor(publicAPI, model) {
...
+      model.scalarsToColors.getUseBelowRangeColor &&
       model.scalarsToColors.getUseBelowRangeColor()
     ) {
       pushSeg('Below', [-0.1, -0.1, -0.1, -0.1]);
     }
 
     const haveAbove =
-      typeof model.scalarsToColors.getUseAboveRangeColor === 'function' &&
+      model.scalarsToColors.getUseAboveRangeColor &&
       model.scalarsToColors.getUseAboveRangeColor();
```
finetjul, line 399: you can use optional chaining operator (.?)

Sources/Rendering/Core/ScalarBarActor/test/testScalarBarNoExtras.js
```diff
@@ -0,0 +1,70 @@
...
+import testUtils from 'vtk.js/Sources/Testing/testUtils';
+
+import 'vtk.js/Sources/Rendering/Misc/RenderingAPIs';
+import vtkRenderer from 'vtk.js/Sources/Rendering/Core/Renderer';
+import vtkRenderWindow from 'vtk.js/Sources/Rendering/Core/RenderWindow';
+
+import vtkScalarBarActor from 'vtk.js/Sources/Rendering/Core/ScalarBarActor';
+import vtkColorTransferFunction from 'vtk.js/Sources/Rendering/Core/ColorTransferFunction';
+
+import baselineNoExtras from './testScalarBarNoExtras.png';
```
finetjul, line 11: without extra, the scalarbar does not go as high as the case with extras.

boucaud: Is that an issue ? It was already the case before so I'm guessing it's intended

boucaud:
@finetjul I think this can be merged now

PR #2233, opened 2022-01-18
doc(AnnotatedCubeActor): add missing type annotations

yasushi-saito:
### PR and Code Checklist

### Context

`IO/Core/ImageStream/example` doesn't compile with typescript without this change.
This PR only changes the typescript definitions, not the actual js code.

### Changes

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

### DIFF
--- a/Sources/Rendering/Core/AnnotatedCubeActor/index.d.ts
+++ b/Sources/Rendering/Core/AnnotatedCubeActor/index.d.ts
@@ -10,15 +10,16 @@ export interface IStyle {
 	fontSizeScale?: (res: number) => number;
 	edgeThickness?: number;
 	edgeColor?: string;
+	resolution?: number;
 }
 
-export interface IFaceProperty {
+export interface IFaceProperty extends IStyle {
 	text?: string;
 	faceRotation?: number;
 }
 
 /**
- * 
+ *
  */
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/AnnotatedCubeActor/index.d.ts
```diff
@@ -10,15 +10,16 @@ export interface IStyle {
 	fontSizeScale?: (res: number) => number;
 	edgeThickness?: number;
 	edgeColor?: string;
+	resolution: number;
```
daker, line 13: can you make it optional with `?`

daker (changes requested):

Sources/Rendering/Core/AnnotatedCubeActor/index.d.ts
yasushi-saito: done.

daker (approved):
LGTM

finetjul:
Please make it a "doc" type of commit instead of "fix".

yasushi-saito:
ok, done.

finetjul:
Thanks for your contribution !

PR #2234, opened 2022-01-19
WIP: fix(vtkResliceCursor): Correctly bounds plane when reslice close to image bounds limits

laurennlam:
When setting bounds on cube source, we want that the output data has the same bounds as input ones.
So we need to use 64bit numbers instead of 32bit in order to keep precision.

@finetjul 

After discussing with @finetjul , we thaught that it could be a good point to also "upgrade" source filter from float32 to float64 in order to avoid issue with precision. @martinken @jourdain WDYT ?

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
@@ -18,6 +18,7 @@ const EPSILON = 0.00001;
  * @param {Array} origin
  * @param {Array} p1
  * @param {Array} p2
+ * @return {Boolean} false if no bounds have been found, else true
  */
 export function boundPlane(bounds, origin, p1, p2) {
   const v1 = [];
@@ -32,12 +33,16 @@ export function boundPlane(bounds, origin, p1, p2) {
   vtkMath.cross(v1, v2, n);
   vtkMath.normalize(n);
 
+  // Inflate bounds in order to avoid precision error when cutting cubesource
+  const inflatedBounds = [...bounds];
+  vtkBoundingBox.inflate(inflatedBounds, EPSILON);
+
   const plane = vtkPlane.newInstance();
   plane.setOrigin(...origin);
   plane.setNormal(...n);
 
   const cubeSource = vtkCubeSource.newInstance();
-  cubeSource.setBounds(bounds);
+  cubeSource.setBounds(inflatedBounds);
 
   const cutter = vtkCutter.newInstance();
   cutter.setCutFunction(plane);
@@ -45,7 +50,7 @@ export function boundPlane(bounds, origin, p1, p2) {
 
   const cutBounds = cutter.getOutputData();
   if (cutBounds.getNumberOfPoints() === 0) {
-    return;
+    return false;
   }
   const localBounds = STATIC.computeLocalBounds(
     cutBounds.getPoints(),
@@ -61,6 +66,7 @@ export function boundPlane(bounds, origin, p1, p2) {
     p2[i] =
       localBounds[0] * v1[i] + localBounds[3] * v2[i] + localBounds[4] * n[i];
   }
+  return true;
 }
 // Project point (inPoint) to the bounds of the image according to a plane
 // defined by two vectors (v1, v2)
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testBoundPlane.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testBoundPlane.js
@@ -58,11 +58,26 @@ test('Test boundPlane no intersection', (t) => {
 
   const bounds = [0, 1, 0, 1, 1, 2];
 
-  boundPlane(bounds, origin, point1, point2);
+  t.ok(boundPlane(bounds, origin, point1, point2) === -1);
 
   t.ok(areEquals(origin, [0, 0, 0]));
   t.ok(areEquals(point1, [2, 0, 0]));
   t.ok(areEquals(point2, [0, 2, 0]));
+
+  t.end();
+});
+
+test('Test boundPlane with points close to bounds limits', (t) => {
+  const origin = [-146.45421827279782, -101.95882636683, -64.91712337082552];
+  const p1 = [147.22047922902922, -101.95882636683, -64.91712337082552];
+  const p2 = [-146.45421827279782, -101.95882636683, 194.58407789818668];
+
+  const bounds = [
+    -117.53152848, 137.57958263142996, -101.95882637, 52.04117363,
+    -63.819600831429966, 191.29151028,
+  ];
+
+  t.ok(boundPlane(bounds, origin, p1, p2));
 
   t.end();
 });


Sources/Filters/Sources/CubeSource/test/testCube.js
```diff
@@ -85,3 +85,13 @@ test.onlyIfWebGL('Test vtkCubeSource Rendering', (t) => {
   });
   renderWindow.render();
 });
+
+test('Test vtkCubeSource : Same output bounds as inputs', (t) => {
```
finetjul, line 89: no space before ":"

Sources/Filters/Sources/CubeSource/test/testCube.js
```diff
@@ -85,3 +85,13 @@ test.onlyIfWebGL('Test vtkCubeSource Rendering', (t) => {
...
+  const bounds = [
+    -10.2368459, 10.3651298, -10.2569841, 10.8746985, -10.5987456, 10.6549872,
+  ];
+  const cube = vtkCubeSource.newInstance();
+  cube.setBounds(bounds);
+  t.deepEqual(cube.getOutputData().getBounds(), bounds);
```
finetjul, line 95: is the test failing if pointType is 'Float32Array' ?

laurennlam:
> @laurennlam did you check the code base if there are no other places like this to fix ?

I didn't. I will !

Sources/Filters/Sources/CubeSource/test/testCube.js
laurennlam: Not tested yet because I have issue with these data. I have an error of 10e-9. I suppose that it's correct, isn't it ? That's why it's WIP.

finetjul: error of 10e-9 in the case of `Float64Array` or `Float32Array` ?

laurennlam: After deep test, here is the status:
- In float64, I have errors of 10e-14
- In float32, I have errors between 10e-6 and 10e-7

Sources/Filters/Sources/LineSource/index.js
```diff
@@ -22,9 +22,9 @@ function vtkLineSource(publicAPI, model) {
     // Check input
     const pointDataType = dataset
       ? dataset.getPoints().getDataType()
-      : 'Float32Array';
+      : 'Float64Array';
     const pd = vtkPolyData.newInstance();
-    const v21 = new Float32Array(3);
+    const v21 = new Float64Array(3);
```
finetjul, line 27: should simply be `const v21 = []`

Sources/Filters/Sources/LineSource/index.js
```diff
@@ -22,9 +22,9 @@ function vtkLineSource(publicAPI, model) {
     // Check input
     const pointDataType = dataset
       ? dataset.getPoints().getDataType()
-      : 'Float32Array';
+      : 'Float64Array';
```
finetjul, line 25: model.pointType

Sources/Filters/Sources/CubeSource/test/testCube.js
finetjul: If you still have errors in float64, it means that this fix will not be sufficient for our reslice cursor widget problem.
We need to inflate the bounds anyway.
So maybe this entire fix is not mandatory (for our problem) and we can leave everything in float32.

WDYT @martinken @jourdain  ? Do we want to default to 64bit points (and optionally normals+tc) or is it useless because it will be converted back to 32bits for WebGL ?

martinken: IMO all source/widget points should be float64 unless they have a huge number of points.  Normals and tcoords should be float32 as their ranges are much better behaved in normal usage. 

WebGL and WebGPU backends will convert to float32 but it will use shift scale to best render the float64 data. For geophysical data float32 points are almost always a problem unless some sort of local reference frame is used.

martinken: FWIW in VTK I have had to  convert a number of sources/widgets to float64 to prevent significant positional jitter. It is very noticeable. Moving to float64 with shiftscale makes it render correctly.

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -32,6 +33,10 @@ export function boundPlane(bounds, origin, p1, p2) {
   vtkMath.cross(v1, v2, n);
   vtkMath.normalize(n);
 
+  // Inflate bounds in order to avoid precision error when cutting cubesource
+  const inflatedBounds = [...bounds];
+  vtkBoundingBox.inflate(inflatedBounds, EPSILON);
```
finetjul, line 38: why not `scale` like we discussed ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -32,6 +33,10 @@ export function boundPlane(bounds, origin, p1, p2) {
   vtkMath.cross(v1, v2, n);
   vtkMath.normalize(n);
 
+  // Inflate bounds in order to avoid precision error when cutting cubesource
+  const inflatedBounds = [...bounds];
```
finetjul, line 37: aren't you supposed to use it somewhere ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
```diff
@@ -18,6 +18,7 @@ const EPSILON = 0.00001;
  * @param {Array} origin
  * @param {Array} p1
  * @param {Array} p2
+ * @return {Number} -1 if no bounds have been found, else 1
```
finetjul, line 21: why not a boolean ?

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
laurennlam: It doesn't do what we want. The output after scaling is too close to the original bounds.

Sources/Filters/Sources/CubeSource/test/testCube.js
finetjul: Gotcha ! Thanks for your input.
@laurennlam we can therefore keep that commit 'Float32Array' to 'Float64Array'

Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
laurennlam: Well... that breaks other test on bound plane. As we modify the bounds, then the expected output data (origin, p1, p2) are not correct. This fix is not the correct solution I think...

finetjul: Indeed, it should be an addition and not a multiplication...
The following should work though...
```
const eps = [...n];
vtkMath.multiplyScalar(eps, EPSILON);
vtkBoundingBox.addBounds(inflatedBounds, bounds[0] + eps[0], bounds[1] + eps[0], ...);
vtkBoundingBox.addBounds(inflatedBounds, bounds[0] - eps[0], bounds[1] - eps[0], ...);
```

laurennlam: It's not correct to inflate on all bounds limits: cf [Test boundPlane oriented](https://github.com/Kitware/vtk-js/blob/6aaa03adbfe8a8bc0985a3db03906272ec3c3df5/Sources/Widgets/Widgets3D/ResliceCursorWidget/test/testBoundPlane.js#L38)

finetjul: for this test, `eps` = `[0,0,EPSILON]`, therefore it should not be an issue.

laurennlam: For this test, `n != [0, 0, 1]`. We are in an oblique case

Sources/Filters/Sources/CubeSource/test/testCube.js
laurennlam: #2237

laurennlam:
This MR is replaced by #2238 and #2237

PR #2237, opened 2022-01-20
perf(sources filters): use 64bits numbers for points coordinates

laurennlam:
### Context
Use Float64 instead of Float32 to define points coordinate of all source filters. It will avoid precision error.

@finetjul @martinken

### DIFF
--- a/Sources/Filters/Sources/CubeSource/index.js
+++ b/Sources/Filters/Sources/CubeSource/index.js
@@ -192,7 +192,7 @@ function vtkCubeSource(publicAPI, model) {
       .apply(points);
 
     // Define quads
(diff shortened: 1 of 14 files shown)


Sources/Filters/Sources/CubeSource/index.js
```diff
@@ -192,7 +192,7 @@ function vtkCubeSource(publicAPI, model) {
       .apply(points);
 
     // Define quads
-    const polys = macro.newTypedArray(model.pointType, numberOfPolys * 5);
+    const polys = macro.newTypedArray('Uint16Array', numberOfPolys * 5);
```
finetjul, line 195: no need to use the macro when the type of the array is not dynamic, simply use new `Uint16Array`

PR #2238, opened 2022-01-20
Fix reslice cursor precision

laurennlam:
@finetjul 
Finally, it won't work with 10e-6 but ok with 10e-7

Need to be reviewed after #2237

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers.js
@@ -9,7 +9,7 @@ import vtkMatrixBuilder from 'vtk.js/Sources/Common/Core/MatrixBuilder';
 
 import { ViewTypes } from 'vtk.js/Sources/Widgets/Core/WidgetManager/Constants';
 
-const EPSILON = 0.00001;
+const EPSILON = 10e-7;
 
 /**
  * Fit the plane defined by origin, p1, p2 onto the bounds.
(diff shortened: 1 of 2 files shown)


Sources/Filters/Sources/PlaneSource/index.js
```diff
@@ -28,10 +28,10 @@ function vtkPlaneSource(publicAPI, model) {
     // Check input
     const pointDataType = dataset
       ? dataset.getPoints().getDataType()
-      : 'Float32Array';
+      : model.pointType;
     const pd = vtkPolyData.newInstance();
-    const v10 = new Float32Array(3);
-    const v20 = new Float32Array(3);
+    const v10 = new Float64Array(3);
+    const v20 = new Float64Array(3);
```
finetjul, line 34:
```
const v10 = [];
const v20 = [];
```

Sources/Filters/Sources/CubeSource/index.js
```diff
@@ -273,7 +273,7 @@ const DEFAULT_VALUES = {
   zLength: 1.0,
```
finetjul, line 273: line 195 should be fixed, polys should not be stored with floating number, please use Uint16Array instead (can be a new commit in the same PR)

finetjul:
Can you please explain in your commit message that the cube source output bounds are different than the input bounds.Hence cutting the cube with a plane exactly on a side of the input bounds may generate empty cut.

PR #2240, opened 2022-01-20
perf(WebGL): rework WebGL volumeMapper interactive rendering

martinken:
Update the renderwindowinteractor to set the recentFrameTime
from the new approach which uses at least a one second
timeframe to compute rendering rates.

Update the WebGL VolumeMapper to use a callback for when the
animation rate changes on the renderwindowinteractor
and simplify a bit how the use of the smaller viewport
is handled.

Adjust the volumeviewer example to be 30 fps because 15fps
feels slow.

### DIFF
--- a/Sources/Rendering/OpenGL/VolumeMapper/index.js
+++ b/Sources/Rendering/OpenGL/VolumeMapper/index.js
@@ -947,107 +947,84 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
     }
   };
 
+  // unsubscribe from our listeners
+  publicAPI.delete = macro.chain(() => {
+    if (model._animationRateSubscription) {
+      model._animationRateSubscription.unsubscribe();
+      model._animationRateSubscription = null;
+    }
+  }, publicAPI.delete);
+
   publicAPI.getRenderTargetSize = () => {
-    if (model.lastXYF > 1.43) {
-      const sz = model.framebuffer.getSize();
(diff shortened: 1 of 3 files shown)


Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -947,107 +947,84 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
-    // if (2.0 * diag / (newSize[0] + newSize[1]) < 0.25 * model.renderable.getSampleDistance()) {
-    //   model.renderable.setSampleDistance(0.25 * model.renderable.getSampleDistance());
-    // }
+    // use/create/resize framebuffer if needed
+    if (model._useSmallViewport) {
+      const size = model.openGLRenderWindow.getFramebufferSize();
 
-    // create/resize framebuffer if needed
-    if (xyf > 1.43) {
+      // adjust viewportSize to always be at most the dest fo size
```
floryst, line 1012: I assume this is meant to be "dest fb size"?

floryst (approved):
LGTM

Sources/Rendering/OpenGL/VolumeMapper/index.js
martinken: fo = framebuffer object

martinken: But yes, I use framebuffer above when it should technically be framebuffer texture

PR #2241, opened 2022-01-22
docs(ts): allow vtk classes discovery

daker:
### PR and Code Checklist

### Context
This changes will allow the discovery of the available vtk classes, types and interfaces

This changes only affects the ESM build.

### Changes

### Results
image

image

### Testing
  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

_This contribution is funded by CSD._

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -20,6 +20,7 @@ import copy from 'rollup-plugin-copy';
 import pkg from './package.json';
 
 import { rewriteFilenames } from './Utilities/rollup/plugin-rewrite-filenames';
+import { generateDtsReferences } from './Utilities/rollup/plugin-generate-references';
 
 const absolutifyImports = require('./Utilities/build/absolutify-imports.js');
 
@@ -169,7 +170,10 @@ export default {
           dest: outputDir,
           rename(name, ext, fullPath) {
             const filename = `${name}.${ext}`;
-            if (filename === 'index.d.ts') {
+            if (
+              filename === 'index.d.ts' &&
+              path.dirname(fullPath) !== 'Sources'
+            ) {
               const moduleName = path.basename(path.dirname(fullPath));
               return `../${moduleName}.d.ts`;
             }
@@ -193,6 +197,9 @@ export default {
         },
       ],
     }),
+    generateDtsReferences({
+      dest: outputDir,
+    }),
     copy({
       flatten: true,
       targets: [
@@ -204,6 +211,7 @@ export default {
         { src: 'Utilities/build/macro-shim.js', dest: outputDir, rename: 'macro.js' },
         { src: '*.txt', dest: outputDir },
         { src: '*.md', dest: outputDir },
+        { src: 'tsconfig.json', dest: outputDir },
         { src: '.npmignore', dest: outputDir },
         { src: 'LICENSE', dest: outputDir },
         {
@@ -214,6 +222,7 @@ export default {
             pkg.name = '@kitware/vtk.js';
             pkg.main = './index.js';
             pkg.module = './index.js';
+            pkg.types = './index.d.ts';
             return JSON.stringify(pkg, null, 2);
           },
         },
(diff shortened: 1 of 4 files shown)


daker:
@floryst can you take a look at this when you have time?

jourdain:
I'll let @floryst merge as I'm worry of breaking the esm build with the edit of the rollup transform logic... ;-)

floryst:
Overall, LGTM. I am curious about whether it's worthwhile to autogenerate the `index.d.ts` files prior to the copy+transform step. That is, for every `index.d.ts` in a terminating (VTK class) folder, append a corresponding `<reference path=...>` to the `index.d.ts` in the above folder (creating as necessary). Then do the same for those generated `index.d.ts` files, up until you hit the root with a `vtk.d.ts` file.

What do you think?

daker:
@floryst sure i'll take a look at it.

daker:
@floryst what do you think the best way to do it? a rollup plugin? with which hook ? buildEnd ?

floryst:
Maybe try a plugin hooked on the [`generateBundle`](https://rollupjs.org/guide/en/#generatebundle). You can use `this.emitFile` in the plugin to generate these d.ts files. I'm a bit fuzzy as to whether this matches the desired output that you have, but it would, for each leaf d.ts, write a `<reference path=...` line in the parent folder's d.ts file (creating it when necessary). These second level files then do the same, building the d.ts reference files bottom-up. Does that seem like an approach compatible with your typescript additions?

daker:
> Maybe try a plugin hooked on the [`generateBundle`](https://rollupjs.org/guide/en/#generatebundle). You can use `this.emitFile` in the plugin to generate these d.ts files. I'm a bit fuzzy as to whether this matches the desired output that you have, but it would, for each leaf d.ts, write a `<reference path=...` line in the parent folder's d.ts file (creating it when necessary). These second level files then do the same, building the d.ts reference files bottom-up. Does that seem like an approach compatible with your typescript additions?

That's what i was trying to do, the issue is that plugin are triggered 3 times, if you throw a console.log string inside the plugin it will appear 3 times on the console, any idea why?

floryst:
If you push your WIP, I can help take a look. I'm not familiar with why that would run 3 times.

daker:
@floryst can you check what i propose? basically instead of creating a separate index.d.ts for each folder, i use the main index.d.ts to reference all the .d.ts files which also works.

floryst:
Oh nice! I like this approach, and it's cleaner than my idea. I'll try it out, but it looks good from my end so far.

daker:
@floryst i made one last change, you can test now

Utilities/rollup/plugin-generate-refrences.js
```diff
@@ -0,0 +1,35 @@
...
+      const files = Object.keys(bundle);
+
+      for (let i = 0; i < files.length; i++) {
+        const file = files[i];
+        const filename = file.replace('.js', '.d.ts');
+
+        const dts = path.join(pluginOptions['dest'], filename);
+
+        if (fs.existsSync(dts) && filename !== rootDts) {
+          dtsReferences.push(`/// <reference path="./${filename}" />`);
```
floryst, line 24: I would add `filename.replace(/\\/g, '/')` here, since Windows will use backslashes for the path. This is minor, since we build and publish from a Unix system.

rollup.config.js
```diff
@@ -20,6 +20,7 @@ import copy from 'rollup-plugin-copy';
 import pkg from './package.json';
 
 import { rewriteFilenames } from './Utilities/rollup/plugin-rewrite-filenames';
+import { generateDtsReferences } from './Utilities/rollup/plugin-generate-refrences';
```
floryst, line 23: Minor typo: `refrences` to `references` (same with the actual filename).

floryst (commented):
Looks good! Just two nits, nothing major. Built and tested on Windows. It didn't work without replacing the path separators, but that's because I built on Windows.

daker:
@floryst done, if you want to take a look again.

floryst:
LGTM! Thanks!

PR #2242, opened 2022-01-24
fix(Picking): fix issue with picking and widgets

martinken:
When using the hardware selector and widgets that are
in a different renderer (such as the orientation widget)
there was an issue with picking where the widget's renderer
was not being used.

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -7,7 +7,7 @@ import vtkSVGRepresentation from 'vtk.js/Sources/Widgets/SVG/SVGRepresentation';
 import { diff } from './vdom';
 
 const { ViewTypes, RenderingTypes, CaptureOn } = Constants;
-const { vtkErrorMacro } = macro;
+const { vtkErrorMacro, vtkWarningMacro } = macro;
 const { createSvgElement, createSvgDomElement } = vtkSVGRepresentation;
 
 let viewIdCount = 1;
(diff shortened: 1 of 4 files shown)


floryst (approved):
LGTM.

Would it be helpful to deprecate `renderWindow.getSelector()` in favor of `createSelector()` to avoid potential sharing issues (like [this post](https://discourse.vtk.org/t/interactiveorientationwidget-loses-interactivity-once-hardwareselector-is-used/7702)) in the future?

martinken:
> 

Let me think about this a bit more. Really we only need one selector and the renderer gets passed to it when capturing buffers. I think I just never updated the WidgetManager to use the new async API. Let me give that a shot.

martinken:
OK I think this is an approach that will work with the async hardware selection. I'm not 100% sure of all the gotchas as async is always a bit tricky, but local testing on both backends seems to work.

jourdain:
Might be worth checking the API call made to vtk.js for picking in [react-vtk-js](https://github.com/Kitware/react-vtk-js/blob/master/src/core/View.js) and [vue-vtk-js](https://github.com/Kitware/vue-vtk-js/blob/master/src/core/VtkView/script.js)

martinken:
Do you know if either of those systems use the WidgetManager? The two links just use the selector which should be fine as the deprecation is in the WidgetManager.

jourdain:
No widget, just plain hardware selector. Great to know... Sorry, I'm only half following. ;-)

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -293,11 +301,10 @@ function vtkWidgetManager(publicAPI, model) {
       subscriptions.pop().unsubscribe();
     }
 
-    model.selector = model.apiSpecificRenderWindow.getSelector();
+    model.selector = model.apiSpecificRenderWindow.createSelector();
```
floryst, line 304: Do we still need to create a new selector, or can we use the existing one?

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -440,7 +451,50 @@ function vtkWidgetManager(publicAPI, model) {
...
+        await captureBuffers(x, y, x, y);
+      }
+
+      model.selections = model._capturedBuffers.generateSelection(x, y, x, y);
+    }
+    return publicAPI.getSelectedData();
+  };
+
   publicAPI.updateSelectionFromXY = (x, y) => {
+    console.warn(
```
floryst, line 495: Maybe use `vtkWarningMacro` here.

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -459,20 +513,16 @@ function vtkWidgetManager(publicAPI, model) {
...
-
-      if (pickingAvailable) {
-        model.selections = model.selector.generateSelection(x, y, x, y);
+        captureBuffers(x, y, x, y);
       }
     }
   };
 
   publicAPI.updateSelectionFromMouseEvent = (event) => {
+    console.warn(
```
floryst, line 523: Ditto with `vtkWarningMacro`

floryst (commented):
The changes look good to me overall.

PR #2243, opened 2022-01-24
docs(DeviceOrientationToCamera): fix skybox refs

floryst:
### PR and Code Checklist

### Context

- Fixes the DeviceOrientationToCamera example ([bug report](https://discourse.vtk.org/t/deviceorientationtocamera-example-page-not-working/7700))

### Changes

### Results

When running on the desktop, the scene shown is as follows: 
image

### Testing


floryst:
@martinken is the initial skybox view supposed to be upside-down? Seems strange to me.

martinken:
I think you flipped the Y images. py should be top and ny should be bottom

floryst:
That's what I originally thought too, but with py=top and ny=bottom, I get this:

image


With py=bottom and ny=top, I get the following:

image

Maybe the original py/ny to bottom/top renaming commit got it mixed up. (43ed0fb80398f2f8a8c11d63b26ef95e7abefa19)

martinken:
Well we could look at the images to see if they are correctly labeled. Another thing to consider though is that we have "flip Y" logic in vtk.js as some image formats use Y down versus Y up. So it could be that these images are all flipped (up to down) and once correctly flipped they bottom/top settings would be correct.

jourdain:
on desktop the default orientation (angles, ...) could be upside down too...
Those angles are supposed to come from a device that know its orientation (phone).

floryst:
I tested this on my phone and in the devtools. On desktop the orientation is upside-down, but on mobile the orientation is correct.

The last bit to resolve is the labeling of the bottom and top images. It might be that the true bottom is mis-named as top, and vice-versa. I'm not sure how to look at the images directly, so I can't verify this.

martinken:
To look at any type of vtk file you can use paraview. Binaries are up for most platforms.

martinken:
Odds are https://github.com/f3d-app/f3dwould work as well for viewing, but I know ParaView will work.

floryst:
They seem to be correct. Top is the sky, and Bottom is the water. The sequence I have also renders correctly, so maybe there was a naming issue prior.

floryst:
Last check: are we okay with the renaming? The renders are correct at least.

jourdain:
I am if we don't break any other examples using the old names?

floryst:
The skybox test is the only other reference to the mountain dataset, and that passes.

PR #2244, opened 2022-01-25
fix(SynchronizableRenderWindow): add type annotations

yasushi-saito:
Add typescript definitions for SynchronizableRenderWindow.

Minor fix & tightening of type definitions for RenderWindow & ImageStream


jourdain:
LGTM

daker:
@jourdain is `viewId` a string or number? if it's a string then this should be fixed, right?

https://github.com/Kitware/vtk-js/blob/758989ee8d01fbe61a27b14973d1e83f4ec7f572/Sources/IO/Core/ImageStream/DefaultProtocol.d.ts#L6-L7

jourdain:
It is a string that is computed from a number. "1"

yasushi-saito:
Ok, changed all `viewId: number` to `viewId: string`.

jourdain:
@floryst I will let you merge it when you want to bundle it with other vtk.js changes.

PR #2247, opened 2022-01-27
fix(splinewidget): fix crash when moveHandle is not initialized

tomsuchel:
Hovering a spline handle while the moveHandle was not initialized (origin=null) was causing errors.
Fix this by calculating the distance between moveHandle and handle only if their positions are set.

### PR and Code Checklist

### Changes

### Testing

  - **vtk.js**: 22.2.1
  - **OS**: Ubuntu 20.04 LTS


finetjul:
LGTM

PR #2249, opened 2022-01-28
chore(patch-package): only patch on rollup build

floryst:
### PR and Code Checklist

### Context

When building vtk.js as a dependency on CI, patch-package returns an error because we patch a dev dependency, which is not installed.

@finetjul 

### Changes

- We only have patches relevant to rollup plugins, so only patch on rollup build.
### Results

using vtk.js as a dependency on CI should not fail.

### Testing


finetjul:
LGTM

PR #2259, opened 2022-02-04
Add support for defining a full url in array.refs in index.json 

Julesdevops:
### PR and Code Checklist

### Context
The URL to fetch a dataset is currently built using a base url and the basepath and id  located in data.ref.
(See https://github.com/Kitware/vtk-js/blob/master/Data/volume/headsq.vti/index.json#L31)
Instead of building the url, it can be useful to directly provide the url where to fetch the array. A use case is for signed URLs.

### Changes
Add a check for array.ref.url in HttpDataAccessHelper.fetchArray, and use it if it exists, else use the default process.

### Testing
  - **vtk.js**: master
  - **OS**: Ubuntu 20.04
  - **Browser**: Chrome 98.0.4758.80


finetjul:
LGTM.
Your commit title could be a bit more "imperative and present tense", e.g. "feat(httpdataaccesshelper): Add index.json full url support in array.refs"

jourdain (approved):

PR #2261, opened 2022-02-07
fix(ScalarBarActor): Fix missing text updates

martinken:
This MR fixes the missing text updates for the
scalar bar actor on some browsers. It also separates
the View dependent data and functions from the non-view
dependent through a new helper class. One helper should be
created per view. Unfortunately this results in a huge
code change and one breaking API change. When users provided
their own function to compute values, those functions will
need to be reworked to deal with the new helper class.
The default autolayout function has been modified and
can be used as an example of how to update your own function.


finetjul:
@boucaud

jourdain (approved):
Seems reasonable.

PR #2262, opened 2022-02-07
fix(shapewidget): absorb mouseMove event while moving shapewidget handles

tomsuchel:
Mouse move events are no longer propagated to other widgets when dragging ShapeWidget handles

### PR and Code Checklist

### Changes

### Testing
  - **vtk.js**: 22.4.0
  - **OS**: Ubuntu 20.04 LTS

### DIFF
--- a/Documentation/content/docs/develop_widget.md
+++ b/Documentation/content/docs/develop_widget.md
@@ -199,6 +199,10 @@ The widget behavior has also access to the renderer, openGLRenderWindow and inte
 
 ##### Focus
 
+"Focus" as a widget concept means that a particular widget should be the only one interactable. This use-case arises when there is a primary widget.
+
+An example is the paint widget needs focus, since when it's active it shouldn't be possible to move rulers, crosshairs, etc. The widget manager should only allow the paint state to be active, and not activate any other state.
+
 When a widget is given the focus, the widget behavior is notified through the `grabFocus()` method. This is usually the place to setup complex interaction states such as initializing the widget behavior's internal state (distinct from the widget state), starting animations (see Animations) or setting up the active state (see Active State).
 
 The `loseFocus()` method is called when the widget manager removes the focus from the widget. It can also be called by the widget behavior itself if necessary. For example, a widget might decide to lose the focus after the escape key was pressed.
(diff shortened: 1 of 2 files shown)


Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
```diff
@@ -498,7 +498,7 @@ export default function widgetBehavior(publicAPI, model) {
       publicAPI.invokeInteractionEvent();
     }
 
-    return model.hasFocus ? macro.EVENT_ABORT : macro.VOID;
+    return model.hasFocus || model.isDragging ? macro.EVENT_ABORT : macro.VOID;
```
finetjul, line 501: @floryst Shouldn't the widget get "focus" when its handle is being dragged ? Apparently widgets get focus only at creation-time. Is it on purpose ? When should they have "focus" ? Thanks !

floryst: "Focus" as a widget concept means that a particular widget should be the only one interactable. This use-case arises when there is a primary widget.

An example is the paint widget needs focus, since when it's active you shouldn't be able to move rulers, crosshairs, etc. The widget manager (should) only allow the paint state to be active, and not activate any other state.

When dragging, that particular widget state is considered active, but the widget is not focused. Conversely, focus usually implies some state is active (like in the paint widget case).

finetjul: Thanks for your explanation ! it makes sense.

@tomsuchel please update the documentation with Forrest's words.

Documentation/content/docs/develop_widget.md
```diff
@@ -203,6 +203,10 @@ When a widget is given the focus, the widget behavior is notified through the `g
 
 The `loseFocus()` method is called when the widget manager removes the focus from the widget. It can also be called by the widget behavior itself if necessary. For example, a widget might decide to lose the focus after the escape key was pressed.
 
+"Focus" as a widget concept means that a particular widget should be the only one interactable. This use-case arises when there is a primary widget.
+
+An example is the paint widget needs focus, since when it's active it shouldn't be possible to move rulers, crosshairs, etc. The widget manager should only allow the paint state to be active, and not activate any other state.
+
```
finetjul, line 209: to move at the beginning of the "Focus" section

finetjul:
LGTM

floryst:
@tomsuchel Thanks for adding the doc change regarding focus!

PR #2263, opened 2022-02-07
Fix(pick): null actor object when returning selection

jadh4v:
### PR and Code Checklist

### Context

In `vtkOpenglHardwareSelector`, functions `getPixelInformation()` and `getPixelInformationWithData()` return a valid flag when actor id read from the pixel buffer is positive. However, the actor id may not correspond to a valid actor object stored
in the model.props array. 

### Changes
An additional check is added for validating that the actor id corresponds to a valid array index. Else, a null selection is returned. This fix is related to a corresponding fix in react-vtk-js [PR 47](https://github.com/Kitware/react-vtk-js/pull/47) for picking the closest object from the mouse pointer position.

- [ ] ~~Documentation and TypeScript definitions were updated to match those changes~~ NA

### Results
Actors that are not explicitly registered with the selector through a call to [`vtkOpenGLHardwareSelector.renderProp(prop)`](https://github.com/Kitware/vtk-js/blob/f7ee7bfd7bf3bb568d44a0d722ad732cb54a9ea4/Sources/Rendering/OpenGL/HardwareSelector/index.js#L506) will not participate in hardware picking, since the pointer to those actors will not be available to populate [`info.prop`](https://github.com/Kitware/vtk-js/blob/f7ee7bfd7bf3bb568d44a0d722ad732cb54a9ea4/Sources/Rendering/OpenGL/HardwareSelector/index.js#L597)

### Testing

- [ ] ~~This change adds or fixes unit tests~~ No
  - **vtk.js**: 
  - **OS**: Windows 10 pro 21H2 19044.1466
  - **Browser**: Chrome 97.0.4692.99 (Windows 10)

### DIFF
--- a/Sources/Rendering/OpenGL/HardwareSelector/index.js
+++ b/Sources/Rendering/OpenGL/HardwareSelector/index.js
@@ -65,7 +65,7 @@ function getPixelInformationWithData(
       buffdata.pixBuffer[PassTypes.ACTOR_PASS],
       buffdata.area
     );
-    if (actorid <= 0) {
+    if (actorid <= 0 || actorid - idOffset >= buffdata.props.length) {
       // the pixel did not hit any actor.
       return null;
     }
@@ -585,7 +585,7 @@ function vtkOpenGLHardwareSelector(publicAPI, model) {
         model.pixBuffer[PassTypes.ACTOR_PASS],
         model.area
       );
-      if (actorid <= 0) {
+      if (actorid <= 0 || actorid - idOffset >= model.props.length) {
         // the pixel did not hit any actor.
         return null;
       }


jadh4v:
@floryst Please take a look. I am not sure if my approach is correct. Another way to fix this could be to change the if condition here from:
```javascript
if(info && info.valid)
```
to
```javascript
if(info && info.valid && info.prop)
```
throughout the file.
The main issue is that the function `getPixelInformation` should continue searching for nearest object if a valid `actor` reference is not found in `model.props`.

floryst:
@jourdain @martinken

I would assume that `info.valid` implies that `info.prop` is a vtkProp that exists in the scene. If so, then your guard checks should be sufficient. I find it interesting that such a condition where `info.valid` is true, yet `info.prop` is null/undefined. In what scenario does this occur?

jourdain:
I think it is related to the fact that the picking successfully happened (no rendering issue or else) but yet nothing under the pointer was found (no actor).

floryst:
In that case, would it be better to go with @jadh4v's second option, with the change to `if(info && info.valid && info.prop)`?

jourdain:
I would have @martinken weight in as I was just making an assumption and I'm not fully focused on this issue.

jourdain:
I would have @martinken weight in as I was just making an assumption and I'm not fully focused on this issue.

jadh4v:
maybe @sankhesh you can also please take a look / chime in on this if you have time.

sankhesh:
IMHO, it should not be considered a valid pick if the actorID is invalid. That would mean an issue with rendering/picking. 
The change looks good as is to me. @martinken do you see a case that we might be missing here where the pick is valid without a valid actorID?

martinken:
all valid picks should have a valid actorid, I think code/callers rely on that.

jadh4v:
I think the reason some picks don't return valid actors is because not all mappers are actually registering for pickable integer values by calling `model.openGLRenderer.getSelector().renderProp(actor)`. Only `PolyDataMapper` and `PolyDataMapper2D` have this call, but `ImageMapper` doesn't.

jadh4v:
I believe we agree that the current change is a reasonable guard against empty / null picks? Can we get this merged, since its a very small change and is blocking another of my tasks.

floryst:
If `actorID - offset` is out of range of the prop list, you might as well effectively return null. I'm good to move forward with this change.

PR #2268, opened 2022-02-08
fix(rendering): take fieldDataTupleId into account when mapping colors

### LINKED ISSUE #2266: fieldDataTupleId not taken into account to chose the index of vector data by which a model should be colorized
Hello, I found what I think is a bug, as I tried to colorize a geometry using different components of a vector data.

### High-level description
When loading a PolyData, and wanting to colorize by a vector field data array,  any call to `setFieldDataTupleId` is inoperant, and the colorization is always done by the vector's first element.

### Steps to reproduce
Steps to reproduce the behavior.:
  * Load a polydata containing a cell data of type vector.
  * Create a mapper on this data
    ```js
      const mapper = vtkMapper.newInstance( { ... });
      mapper.setArrayAccessMode(1);
      mapper.setScalarModeToUseCellFieldData();
      mapper.setColorByArrayName("myDataArrayName");
      mapper.setColorModeToMapScalars();
    
      mapper.setFieldDataTupleId(1); // this call is inoperant
    ```

### Detailed behavior
Colorization is always done by the vector's first component. The call to `setFieldDataTupleId` is ignored

### Expected behavior
Colorization is expected to be done by the vector's component passed to `setFieldDataTupleId()`.

### Environment
- **vtk.js**: 22.4.2

### Analysis
I think in `Mapper/index.js`, line 164, `mapScalars` is passed wrongly a `-1` value, while it should be passed `publicAPI.getFieldDataTupleId()`.
I tested a version with this change and it indeed fixes the issue for me. It's visible [here](https://github.com/eino/vtk-js/commit/cb2b65b5ec36092385ecd4a0d16a3795472a93dc); if it looks good I can open a PR.

eino:
Call `getFieldDataTupleId()` instead of a static `-1` when computing `colorMapColors`,
allowing `setFieldDataTupleId` to set the index of the data tuple by which the model should be colorized.

### PR and Code Checklist

### Context

- fix #2266

### Changes
In both `Mapper` and `Mapper2D`, when mapping scalars to colors, pass `publicAPI.getFieldDataTupleId()` rather than a static `-1` as third argument to `lut.mapScalars()`.
As `fieldDataTupleId` defaults to `-1`, base behavior is not changed, while a call to `setFieldDataTupleId` is now taken into account.
- [ ] Documentation and TypeScript definitions were updated to match those changes --> No need, rather the behavior is aligned with current description of  `setFieldDataTupleId`.

### Results
- Before: call of `setFieldDataTupleId` was inoperant, the vector values at first index was always used
- After: values at expected index are used to color the geometry

### Testing
  - **vtk.js**: 22.4.2
  - **OS**: Ubuntu 20
  - **Browser**: Any (Firefox, Chrome)

### DIFF
--- a/Sources/Rendering/Core/Mapper/index.js
+++ b/Sources/Rendering/Core/Mapper/index.js
@@ -161,7 +161,11 @@ function vtkMapper(publicAPI, model) {
       if (lut) {
         // Ensure that the lookup table is built
         lut.build();
-        model.colorMapColors = lut.mapScalars(scalars, model.colorMode, -1);
+        model.colorMapColors = lut.mapScalars(
+          scalars,
+          model.colorMode,
+          model.fieldDataTupleId
+        );
       }
     }
     model.colorBuildString = `${publicAPI.getMTime()}${scalars.getMTime()}${alpha}`;
--- a/Sources/Rendering/Core/Mapper2D/index.js
+++ b/Sources/Rendering/Core/Mapper2D/index.js
@@ -133,7 +133,11 @@ function vtkMapper2D(publicAPI, model) {
     if (lut) {
       // Ensure that the lookup table is built
       lut.build();
-      model.colorMapColors = lut.mapScalars(scalars, model.colorMode, -1);
+      model.colorMapColors = lut.mapScalars(
+        scalars,
+        model.colorMode,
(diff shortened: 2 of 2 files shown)


jourdain:
LGTM

PR #2271, opened 2022-02-09
perf: improve computeRange performance

FezVrasta:
### PR and Code Checklist

### Context

The existing `createRangeHelper` and `computeRange` functions trigger the garbage collector a lot and slow down the whole operation.

### Changes

A new `fastComputeRange` function has been added, that provides a 3x speed improvement over the existing function. While not triggering a single GC action.

~I would like to completely remove the slow `createRangeHelper` but I need to better understand how it's used first. Some help would be apreciated.~

### Results

**Before**
CleanShot 2022-02-09 at 5 49 14@2x
CleanShot 2022-02-09 at 6 01 45@2x

**After**
CleanShot 2022-02-09 at 5 43 55@2x

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

_This contribution is funded by [Sirona Medical](https://sironamedical.com)._

### DIFF
--- a/Sources/Common/Core/DataArray/index.js
+++ b/Sources/Common/Core/DataArray/index.js
@@ -9,6 +9,34 @@ const TUPLE_HOLDER = [];
 // Global methods
 // ----------------------------------------------------------------------------
 
+// Original source from https://www.npmjs.com/package/compute-range
+// Modified to accept type arrays
+function fastComputeRange(arr) {
+  const len = arr.length;
+  let min;
+  let max;
+  let x;
+  let i;
+
+  if (len === 0) {
+    return null;
+  }
+  min = arr[0];
+  max = min;
+  for (i = 1; i < len; i++) {
+    x = arr[i];
+    if (x < min) {
+      min = x;
+    } else if (x > max) {
+      max = x;
+    }
+  }
+  return { min, max };
+}
+
+/**
+ * @deprecated please use fastComputeRange instead
+ */
 function createRangeHelper() {
   let min = Number.MAX_VALUE;
   let max = -Number.MAX_VALUE;
@@ -36,29 +64,21 @@ function createRangeHelper() {
 }
 
 function computeRange(values, component = 0, numberOfComponents = 1) {
-  const helper = createRangeHelper();
-  const size = values.length;
-  let value = 0;
-
   if (component < 0 && numberOfComponents > 1) {
     // Compute magnitude
-    for (let i = 0; i < size; i += numberOfComponents) {
-      value = 0;
-      for (let j = 0; j < numberOfComponents; j++) {
-        value += values[i + j] * values[i + j];
+    const size = values.length;
+    const numberOfValues = size / numberOfComponents;
+    const data = new Float64Array(numberOfValues);
+    for (let i = 0, j = 0; i < numberOfValues; ++i) {
+      for (let nextJ = j + numberOfComponents; j < nextJ; ++j) {
+        data[i] += values[j] * values[j];
       }
-      value **= 0.5;
-      helper.add(value);
+      data[i] **= 0.5;
     }
-    return helper.getRange();
+    return fastComputeRange(data);
   }
 
-  const offset = component < 0 ? 0 : component;
-  for (let i = offset; i < size; i += numberOfComponents) {
-    helper.add(values[i]);
-  }
-
-  return helper.getRange();
+  return fastComputeRange(values);
 }
 
 function ensureRangeSize(rangeArray, size = 0) {
@@ -94,6 +114,7 @@ function getMaxNorm(normArray) {
 export const STATIC = {
   computeRange,
   createRangeHelper,
+  fastComputeRange,
   getDataType,
   getMaxNorm,
 };
--- a/Sources/Common/Core/DataArray/index.js
+++ b/Sources/Common/Core/DataArray/index.js
@@ -11,7 +11,7 @@ const TUPLE_HOLDER = [];
 
 // Original source from https://www.npmjs.com/package/compute-range
 // Modified to accept type arrays
-function fastComputeRange(arr) {
+function fastComputeRange(arr, offset, numberOfComponents) {
   const len = arr.length;
   let min;
   let max;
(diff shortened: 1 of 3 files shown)


jourdain:
LGTM

Sources/Common/Core/DataArray/index.js
```diff
@@ -9,6 +9,31 @@ const TUPLE_HOLDER = [];
...
+  let max;
+  let x;
+  let i;
+
+  if (len === 0) {
+    return null;
+  }
+  min = arr[0];
+  max = min;
+  for (i = 1; i < len; i++) {
```
jadh4v, line 26: I think you will have to pass `offset` and `numberOfComponents` as arguments and do `i=offset` and `i += numberOfComponents` to properly handle multi-channel data.

FezVrasta: I sent a commit that _should_ address multi-channel data too, may you take a look please?

sankhesh (approved):
LGTM

Sources/Common/Core/DataArray/index.js
```diff
@@ -36,29 +64,24 @@ function createRangeHelper() {
...
 function computeRange(values, component = 0, numberOfComponents = 1) {
-  const helper = createRangeHelper();
-  const size = values.length;
-  let value = 0;
-
   if (component < 0 && numberOfComponents > 1) {
     // Compute magnitude
+    const size = values.length;
+    let value = 0;
+    const data = [];
```
finetjul, line 71: you could allocate the final size of the array instead of "push"

FezVrasta: I don't quite get how this part is working. Care to send a patch?

finetjul:
```
const data = [];
data.length = size / numberOfComponents;
...
  data[i/numberOfComponents] = value;
```
(you may have to have `i` ranging from 0 to size / numberofComponents to reduce the number of operations).
e.g.
```
const numberOfValues = size / numberOfComponents)
const data = new Float64Array(numberOfValues);
for (let i = 0, j = 0; i < numberOfValues; ++i) {
  for (let nextJ = j + numberOfComponents; j < nextJ; ++j) {
     data[i] += values[j]*values[j];
   }
   data[i] *= 0.5;
}
```

FezVrasta: Couldn't this lead to non-integer numbers as `numberOfValues`?

finetjul: it should never be the case.

FezVrasta: I pushed the change.

finetjul: Is it any faster ? 
if not, then it may not worth it.

FezVrasta: Your version is 25% faster

Sources/Common/Core/DataArray/index.js
```diff
@@ -9,6 +9,34 @@ const TUPLE_HOLDER = [];
...
+  let max;
+  let x;
+  let i;
+
+  if (len === 0) {
+    return null;
+  }
+  min = arr[0];
+  max = min;
+  for (i = 1; i < len; i++) {
```
jadh4v, line 26:
```suggestion
  for (i = offset; i < len; i += numOfComponents) {
```

Sources/Common/Core/DataArray/index.js
```diff
@@ -9,6 +9,34 @@ const TUPLE_HOLDER = [];
 // Global methods
 // ----------------------------------------------------------------------------
 
+// Original source from https://www.npmjs.com/package/compute-range
+// Modified to accept type arrays
+function fastComputeRange(arr) {
```
jadh4v, line 14:
```suggestion
function fastComputeRange(arr, offset, numOfComponents) {
```

Sources/Common/Core/DataArray/index.js
```diff
@@ -36,29 +64,21 @@ function createRangeHelper() {
...
-      helper.add(value);
+      data[i] **= 0.5;
     }
-    return helper.getRange();
-  }
-
-  const offset = component < 0 ? 0 : component;
-  for (let i = offset; i < size; i += numberOfComponents) {
-    helper.add(values[i]);
+    return fastComputeRange(data);
```
jadh4v, line 78:
```suggestion
    return fastComputeRange(data, 0, 1);
```

Sources/Common/Core/DataArray/index.js
```diff
@@ -36,29 +64,21 @@ function createRangeHelper() {
...
   }
 
-  return helper.getRange();
+  return fastComputeRange(values);
```
jadh4v, line 81:
```suggestion
  return fastComputeRange(values, component < 0 ? 0 : component, numberOfComponents);
```

sankhesh (approved):

FezVrasta:
Any ETA to get this released?

finetjul:
@FezVrasta would you mind writing a unit test (and squashing your commits to keep only 2: one from yours and one from @jadh4v ?

@jourdain how about removing `createRangeHelper()`, or can it be flagged as obsolete ?

FezVrasta:
@finetjul don't you already have some unit tests for publicAPI.getRange? I'd defer the unit test creation to someone else since I don't completely understand how the most advanced modes of this utility work.

jourdain:
If `createRangeHelper()` is not needed anymore, we can surely remove it, unless it was public and used outside of this context.

FezVrasta:
`createRangeHelper` is exposed to `STATIC` so I suppose someone may be using it in userland?

FezVrasta:
@jourdain @jadh4v I experimented with an alternative approach and I noticed that `Array.prototype.sort` is approx 20% faster than the algorithm this PR introduced, probably the fact the built-in sorth algorithm is ran as native code gives it an additional perf boost.

This can only work when `numberOfComponents` is set to `1` unfortunately, but I think it may be worth it implementing this change?

Here's the benchmarks:
https://jsbench.me/kokzo24f6t/1

jourdain:
What about an array with NaN

PR #2276, opened 2022-02-12
perf: replace blueimp-md5 with spark-md5

FezVrasta:
### PR and Code Checklist

### Context

Gains a 30% performance boost. I tried a few implementations and this one was the fastest I found.

I wonder if we need to md5 this for a specific reason though?

### Changes
Replaces blueimp-md5 with spark-md5, it's slightly larger but 30% faster.

### Results
30% performance boost on benchmarks

CleanShot 2022-02-12 at 11 07 20@2x

CleanShot 2022-02-12 at 11 12 19@2x

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 


jourdain:
How big is the size change? I'm just wondering

FezVrasta:
I have been benchmarking the vtk.js initialization and I'm addressing the low hanging fruits I identified, this was one of the largest ones.

Size wise, blueimp-md5 is 1.7kb gzipped, while spark-md5 is 2.66kb. Considering the already large size of vtk.js I don't think it makes any difference, but the performance boost is welcome.

I'd like to stress more on the need of this md5 utility though, why do you need to md5 that string? Couldn't we leave it as is?

FezVrasta:
If size is an issue I can extract just the md5 string function from spark-md5, this way the total size will be 1.39kb instead.

jourdain:
Size is not an issue. I was just wondering. Thanks for sharing and pushing this topic forward.

martinken:
LGTM +1

FezVrasta:
Nobody is able to explain the md5 requirement?

floryst:
As far as I can tell, the md5 is solely used to determine if a shader's code has changed, and to perform necessary ops if so. This is mostly relevant in OpenGL/SurfaceLIC code. Perf-wise, raw string comparisons might be faster overall if the shader program text isn't super long. It may be useful to see that comparison.

jourdain:
md5 is how we index things within VTK/C++ for the GL generated code. Maybe the plain string approach might work better in JS but I guess we never tried, so it is hard to tell which one is best here.

If I'm not mistaken, that md5 pass only needs to happen when mapper mtime are changing which should happen at a slow rate. So that overhead should not be an issue. @martinken please correct me if I'm wrong.

Also, we could create a facade for the MD5 lib which we could toggle on/off for using md5 or plain string depending on the use cases.

martinken:
For WebGL (not WebGPU) we cache shaders based on the md5 of their source code. The md5 should only be computed when something has changed that can cause the shader code to change, not every frame typically. A straight string comparison can be used if we are storing the full shader code in the cache. I think for JS we may be storing the full shader code so that would be fine.

floryst:
I'm good to merge this for now. We can examine the direct string comparison approach at a later point, which may yield better results.

PR #2277, opened 2022-02-14
fix(ArrowHandleRepresentation): do not zero out Z

floryst:
### PR and Code Checklist

### Context
- LineWidget example fails to have oriented handles when using a non-default plane manipulator normal.

### Changes
- [ ] remove code that zeros out the Z component of an orientation vector

### Results
- oriented handles work properly in the LineWidget example


floryst:
@finetjul just checking in with you: this looked like a debugging statement of some sort that got left in. Removing it allows orientation to work regardless of the plane manipulator's orientation.

finetjul:
From what I remember, this code is not a bug nor a debugging statement, it was there to enforce the arrow to always "face" the camera.

floryst:
Oh, I see now. Seems like there is a separate issue causing the handle orientations to not properly line up. Here is a sample where the plane normal is along Y and the camera is looking along Y:

image

floryst:
I'm moving this to an issue.

PR #2281, opened 2022-02-15
perf: further improve fastComputeRange performance (~20% boost in some cases)

FezVrasta:
### PR and Code Checklist

### Context

`Array.prototype.sort` can be used to speed up the range computation when `numberOfComponents` equals `1`.

Benchmarks: [jsbench.me/kokzo24f6t/1](https://jsbench.me/kokzo24f6t/1)

### Changes

### Testing

  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 


FezVrasta:
Oh, nevermind, I realized the default `.sort` function sorts alphabetically, which is not okay for our use case, if a custom sorting function is provided it gets slower though :-(

PR #2296, opened 2022-03-01
fix(volumeMapper): halfVoxel extension of the volumeMapper

sedghi:
### Background
This PR is the same as https://github.com/Kitware/vtk-js/pull/1993 which was merged in `beta`. @floryst asked me to open another one for master. In addition to the base branch change, it: 
- Adds a test for comparison to the baseline image to make sure breaking volume boundary changes are picked up
- Updates one of the volumeMapper tests which were showing 1.71% diff 

The test data I have used is synthetically generated by me and can be found [here](https://sedghi.github.io/dicom-test/src/data/synthetic_dicom_11Slices_1mm.vti). it is a useful small volume that has helped me debug rendering. I was wondering if we can host it along side other datasets that vtk.js use (e.g., headsq.vti). 

Thanks 

PS: broken baseline was


Sources/Rendering/OpenGL/VolumeMapper/test/testVolumeMapperBounds.js
```diff
@@ -0,0 +1,91 @@
...
+  const reader = vtkHttpDataSetReader.newInstance({ fetchGzip: true });
+  // create color and opacity transfer functions
+  const ctfun = vtkColorTransferFunction.newInstance();
+  ctfun.addRGBPoint(0, 0, 0, 0);
+  ctfun.addRGBPoint(1, 1, 1, 1);
+  // const ofun = vtkPiecewiseFunction.newInstance();
+  // ofun.addPoint(0.0, 0.0);
+  // ofun.addPoint(255.0, 1.0);
+  actor.getProperty().setRGBTransferFunction(0, ctfun);
+  // actor.getProperty().setScalarOpacity(0, ofun);
```
floryst, line 49: Should the commented code be uncommented, or removed?

Sources/Rendering/OpenGL/VolumeMapper/test/testVolumeMapperBounds.js
```diff
@@ -0,0 +1,91 @@
...
+  // Interactor
+  const interactor = vtkRenderWindowInteractor.newInstance();
+  interactor.setStillUpdateRate(0.01);
+  interactor.setView(glwindow);
+  interactor.initialize();
+  interactor.bindEvents(renderWindowContainer);
+
+  reader
+    .setUrl(
+      'https://sedghi.github.io/dicom-test/src/data/synthetic_dicom_11Slices_1mm.vti'
```
floryst, line 68: If the dataset is small enough, we may commit it to the vtk.js Data folder. That way, this test won't rely on external sources.

floryst: @jourdain

floryst:
Thanks for pushing the PR! This one will go into master, and I'll drop the beta branch PR.

@jadh4v can you see if these changes are compatible with what you're working on in #2089? Also, FYI @martinken

Sources/Rendering/OpenGL/VolumeMapper/test/testVolumeMapperBounds.js
jourdain: We could put it in https://github.com/Kitware/vtk-js-datasets too. Still external but less than another repo that we don't control.

Sources/Rendering/OpenGL/VolumeMapper/test/testVolumeMapperBounds.js
sedghi: removed

Sources/Rendering/OpenGL/VolumeMapper/test/testVolumeMapperBounds.js
sedghi: Thank you both. I created a PR here https://github.com/Kitware/vtk-js-datasets/pull/12

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -951,15 +953,16 @@ vec2 computeRayDistances(vec3 rayDir, vec3 tdims)
 void computeIndexSpaceValues(out vec3 pos, out vec3 endPos, out float sampleDistanceIS, vec3 rayDir, vec2 dists)
 {
   // compute starting and ending values in volume space
-  pos = vertexVCVSOutput + dists.x*rayDir;
+  vec3 halfVoxel = vOriginPlusHalfVoxelVC - vOriginVC;
+  pos = vertexVCVSOutput + dists.x*rayDir + halfVoxel;
   pos = pos - vOriginVC;
```
martinken, line 958: The code here looks odd. The ray position start and end points (in View Coords) are being offset by half a voxel. That seems confusing as the ray positions should be unchanged, it should be the mapping of the ViewCoords to volume TextureCoords that should be updated to me. Otherwise intersections with cropping planes etc get a bit off as the ray is no longer the original ray, but this offset ray. Does that make sense?

jadh4v: Instead of correcting for the padding directly inside the shader, the spatial extents can be correctly set to pad by 0.5 voxels, like here:
https://github.com/Kitware/vtk-js/pull/2089/files#diff-653a0c06ccb1ec727a15802ff1d90b9cb6599f82d58610b2522dcfe483ab7cf1R168

This way, there would be minimal changes to the shader, like here:
https://github.com/Kitware/vtk-js/pull/2089/files#diff-4755e93326d6496e13ee38881a618a5accbe127aa1c4de56cdac22555b5fdc4cR909

sedghi: @jadh4v Thanks. As I see your PR has both image and volume mapper fixes but it is in Draft, can I ask what is the plan for your PR to get merged in master? We are in a rush to have this bug fixed for volumeMapper in vtk, but was wondering if yours is ready? if not I can go ahead adding the `getSpatialExtent` approach instead of modifying shader code.

jadh4v: @sedghi We kinda got stuck into a discussion regarding how to handle medical vs simulation data (which doesn't require the padding), and convention regarding the definition of the `Origin` point given the padding.

But my PR already addresses the core issue of proper padding for both Image and Volume mappers. So I think we can quickly merge it. How about this, I will rebase and re-push the updated branch and share it with you. You can test it locally if it works for your use-case, and also push the unit test that you have created to the branch.

sedghi: @jadh4v Yes, that would be great! thanks for your help

jadh4v: @sedghi  Great! I have updated my branch and added you as a collaborator. Please go ahead with further modifications: https://github.com/jadh4v/vtk-js/tree/image-volume-half-voxel

jadh4v: There are some tests failing, but I can take care of it after you push your changes.

sedghi: @jadh4v Thanks again for your help, I tested your fixes, and it made sense and also works for our usecase. I added two commits in your fork:
1. Added my test for the volumeMapperBounds, but there are some tests failing as you mentioned, but the new added test is passing. 
2. Fixed the halfVoxel for the labelmapOutline

sedghi:
@floryst @jourdain @jadh4v @swederik 
Closing this PR in favor of https://github.com/Kitware/vtk-js/pull/2089 as we have tested the solution and works for us too, and also has minimal changes in the shader as @martinken requested.

PR #2298, opened 2022-03-02
fix(LabelmapOutline): Take more than one renderer into account

sedghi:
### Context

`LabelmapOutline` does not render properly if there are more than one renderer in the renderWindow. To see the problem you can run the example included (but revert the shader changes). Since the example has been specifically written for demonstrating the problem and might not be useful for others, maybe we remove it after the review? 

**Before:** 

**After:** 

### Results
By taking into account the correct size and offset we are able to render it properly. 

### Changes
Two new uniforms for `vtkVolumeFS.glsl` which passes the offsetX and offsetY to be used in the volumeOutline logic in the shader.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: latest master
  - **OS**: MacOS 10.15.7
  - **Browser**: Chrome 98.0.4758.109

Tagging @floryst for review :)
CC'ing @swederik


finetjul:
Your example should be turned into a unit test

sedghi:
Thanks @finetjul for your comment. 
I went ahead and replaced the example with a test, but there is a problem which is the renderWindow is the whole page (in manyRenderers [example](https://github.com/Kitware/vtk-js/blob/master/Examples/Rendering/ManyRenderers/index.js#L131) too)  and the image that is captured from it becomes window size dependent which is not something we like for tests, I guess.

I thought if I manually `setSize` on the `glWindow` and the style width of the container (like I do [here](https://github.com/Kitware/vtk-js/pull/2298/commits/99985748d07ed75e81784eb3f414afc883444213#diff-7a69900e652fad24a125240187b678f6ced8a219ed56b5f049ed3aa784f90c05R54-R55)) would work but it just shrinks everything which I don't understand why. Any insight on this problem?

floryst:
I'm not 100% sure why that's happening, but calling `setSize` should work. Maybe try that without setting the rootContainer's width, since the effective canvas drawing space (set from `setSize`) is independent of the canvas element's actual width and height.

floryst:
The changes seem reasonable thus far. We may also need to look at this PR's applicability to the WebGPU code. @martinken

sedghi:
Thank you @floryst, yeah I didn't need the canvas style resize.
Now my test works with `.only` and compares the rendered image to baseline successfully; however, letting it run with other tests (without `.only`) will result in a failed test. 

Looking at the tests report, I see that actually it doesn't render anything when it doesn't have single run with `only`. Have you faced this issue before?

floryst:
Odd, I've not seen that happen before. You can modify `karma.conf.js` and edit the `files` array to explicitly include your test and maybe one other webgl test to see if only two tests cause the issue.

sedghi:
@floryst So I tried the `files` property on the karma config, I wanted to see which other tests are causing trouble, so I added them one by one and I ended up adding them all without any errors .... which was surprising to me, since outline test I wrote was passing successfully with the explicit file names. See below:

```
karma.config.js

files: [
      'Sources/Testing/setupTestEnv.js',
      // 'Sources/**/test*.js',
      'Sources/Rendering/**/test*.js',
      'Sources/Filters/**/test*.js',
      'Sources/Common/**/test*.js',
      'Sources/Testing/**/test*.js',
      'Sources/Widgets/**/test*.js',
      'Sources/IO/**/test*.js',
      'Sources/Imaging/**/test*.js',
      'Sources/Misc/**/test*.js',
      'Sources/Proxy/**/test*.js',
      'Sources/Interaction/**/test*.js',
      { pattern: 'Data/**', watched: false, served: true, included: false },
    ],
```

Is it possible the order that the tests run have effect? 


image

floryst:
That is quite odd. These tests shouldn't have ordering affect their results. If you put the `Rendering` tests last, does that impact anything?

martinken:
The basic gist of this change is fine, +1 for that. Just need to resolve the testing issue.  This change is not applicable to WebGPU yet as the label mapping isn't in coded in the volume renderer yet on WebGPU.

floryst:
Ok, that's something we need to dig into. Can you find which tests (or class of tests) that, when run before the Rendering tests, cause failure? I'm guessing either tests from Interaction or Widgets.

sedghi:
@floryst I confirm it is `Interactor` test, I found one and added release resources by `gc` to it. It seems to work for my test along with other ones too. Now I'm getting an error for `Test Composite Volume Rendering: intermixed geometry
`. 

image

floryst:
[This setTimeout](https://github.com/Kitware/vtk-js/blob/master/Sources%2FWidgets%2FCore%2FAbstractWidgetFactory%2Findex.js#L177-L181) is causing issues. For the purposes of your test, I would recommend wrapping the contents of the setTimeout with an `if (model.widgetState)` guard.

I'll be opening a separate issue to address this post-initialization hack. It seems to be on the brittle side.

sedghi:
@floryst all tests passing :)

floryst:
LGTM! Last request: can you rename your `feat(test)` commit messages to `test(LabelmapOutline)`? Commit message-wise, the latter is more in-line with what the change focuses on. Other than that, this is good to go.

PR #2302, opened 2022-03-04
Imagemapper lut range

sankhesh:
### Context

This change adds functionality to map pixel data to colors using one of two mechanisms in the image
mapper:

- Use color window/level values from the image property
- Use the range set on the lookup table in the image property

By default, the color window/level values are used.

BREAKING CHANGE: The original behavior of the image mapper was that if a lookup table is provided,
it mapped the lookup table's scalar range by default. The new behavior disables using the lookup
table scalar range by default. Instead, the window/level values are used.
### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  latest master
  - **OS**:  Linux
  - **Browser**:  Brave, Firefox


jourdain:
If that breaking change makes sense to the medical folks and @martinken, I'm cool with it.

Sources/Rendering/Core/ImageProperty/index.d.ts
```diff
@@ -173,6 +174,12 @@ export interface vtkImageProperty extends vtkObject {
 	 * @param func 
 	 */
 	setScalarOpacity(index: any, func: any): boolean;
+
+	/**
+	 *
```
finetjul, line 179: please add documentation

finetjul (commented):
LGTM, it is the same behavior as in VTK C++

martinken (approved):

Sources/Rendering/Core/ImageProperty/index.d.ts
sankhesh: Thanks, will do.

osiux:
> Nope. Not when a LUT is supplied. When a lut is supplied, the lut is range was used and there was no mechanism to window/level the range.

What is LUT? The lookup table? Does this mean that if this is used, is not possible to change window/level?

martinken:
Ahh yes I see. I keep thinking we do W/L in the fragment shader and we don't yet. LGTM

PR #2306, opened 2022-03-08
fix(Interaction): improve mouse wheel interaction

martinken:
readjust the mouse wheel timer delay to 200ms so that
we do not get unintended zooms. Ideally the mouse wheel
movements would be single fire events as opposed to
start, move, stop but there may be code relying on the
start/move/stop approach.

Add the notion of extending the animation end time.
Now when a mouse wheel event is ending we tell the
RenderWindowInteractor to continue animation for another
600ms. The idea is to stay in animation mode for
longer so that a slow non-interactive render doesn't happen
when the user is really still being interactive.


floryst:
LGTM!

floryst:
LGTM!

PR #2314, opened 2022-03-11
docs(ts): use relative instead of absolute imports

### LINKED ISSUE #2311: Type definitions are not properly resolved when using `npm link`
### High-level description

When working on the typescript definitions using `npm link` some types are unable to be resolved. Perhaps I'm doing something wrong? Or this is an unsupported was of developing types. @daker would you share how you've been developing the types?

### Steps to reproduce

```
# in vtk.js repo
npm run dev:esm

cd dist/esm/ # so we get the @kitware/vtk.js package rather than the vtk.js
npm link

# in the linked project dir
npm link @kitware/vtk.js
``` 

```typescript
import vtkOrientationMarkerWidget from "@kitware/vtk.js/Interaction/Widgets/OrientationMarkerWidget"

// type error! .onModified does not exist on vtkOrientationMarkerWidget despite it extending vtkObject
vtkOrientationMarkerWidget.newInstance().onModified()
```

This is happening because `vtkObject` cannot be resolved, if we look at the defintion we can see:
```typescript
import { vtkObject } from '@kitware/vtk.js/interfaces';
```
Which I guess doesn't work in an `npm link` environment?

However if the import was:
```typescript
import { vtkObject } from './../../interfaces';
```
Everything types correctly.

My solution for now has been to override the absolute import code in `rollup.config.js`:
https://github.com/Kitware/vtk-js/blob/ea51f4538773febf4cbc36e103b1ba677ff682f1/rollup.config.js#L184,L194

With quick hack to leave it as relative and "mostly" get it hooked up again:
```typescript
return absolutifyImports(content.toString(), (relImport) => {
  if (relImport.startsWith('../')) {
    return './' + relImport.slice(3);
  }
  return relImport;
});
```

### Environment
- **vtk.js**: master branch
- **OS**: Debian
- **Browser**: N/A

daker:
Use relative instead of absolute imports to make npm link
and regular packages work correclty.

fix #2311

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


Sources/types.d.ts
```diff
@@ -1,5 +1,26 @@
...
+ * @ignorenaming
+ */
+declare type double = number;
+
+/**
+ * Alias type for number that are integer
+ * @ignorenaming
+ */
+declare type int = number;
+
```
daker, line 23: @jadh4v i have added those new types

Sources/Rendering/OpenGL/RenderWindow/index.d.ts
```diff
@@ -241,20 +241,20 @@ export interface vtkOpenGLRenderWindow extends vtkOpenGLRenderWindowBase {
 
 	/**
 	 *
-	 * @param vtktype
-	 * @param numComps
-	 * @param useFloat
+	 * @param {VtkDataTypes} vtktype
```
floryst, line 244: Please chomp the trailing spaces.

floryst: We will need to get the linter to format d.ts files. I don't think it does presently.

daker: The trailing space is needed, it indicates the description of the param it's just we don't it everywhere
* @param {paramType} paramName paramDescription

floryst: Just to follow-up: I'm pretty sure if I let prettier run on d.ts files, it'll chomp trailing whitespace. I don't know if there is any special setting to tell prettier to avoid docstring whitespace.

floryst:
Cool, I just ran a build of your branch, and it looks good. Haven't tried an `npm link` yet.

Georift:
Thanks for fixing this @daker :grin:

PR #2316, opened 2022-03-14
feat: add support for Mouse Lock API to MouseRangeManipulator

FezVrasta:
### Context

This change allows to set a `usePointerLock` flag to the `MouseRangeManipulator` so that the mouse is hidden while the user is interacting with the view.

This is especially useful because users may move their mouse outside the view while dragging their cursor to quickly interact with the view.

The PR is based on https://github.com/Kitware/vtk-js/pull/1272

### Results

Before, the user would keep seeing their cursor as they dragged it around, if the mouse went outside the view it could cause different issues since it could interfere with other UI elements.

After, the cursor is hidden while the user holds their button down, preventing the above issues.

### Changes

I couldn't find TS definitions for this module.

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

### Funding

### DIFF
--- a/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
@@ -9,11 +9,6 @@ function vtkMouseRangeManipulator(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkMouseRangeManipulator');
 
-  const internal = {
-    interactor: null,
-    renderer: null,
-  };
-
   // Keep track of delta that is below the value
   // of one step to progressively increment it
   const incrementalDelta = new Map();
@@ -171,9 +166,8 @@ function vtkMouseRangeManipulator(publicAPI, model) {
     model.containerSize = size.map((v) => v * ratio);
 
     if (model.usePointerLock && !interactor.isPointerLocked()) {
-      Object.assign(internal, { interactor, renderer });
       interactor.requestPointerLock();
-      publicAPI.startPointerLockInteraction();
+      publicAPI.startPointerLockInteraction(interactor);
     }
   };
 
@@ -190,9 +184,7 @@ function vtkMouseRangeManipulator(publicAPI, model) {
 
   //--------------------------------------------------------------------------
 
-  publicAPI.startPointerLockInteraction = () => {
-    const { interactor } = internal;
-
+  publicAPI.startPointerLockInteraction = (interactor) => {
     // TODO: at some point, this should perhaps be done in
     // RenderWindowInteractor instead of here.
     // We need to hook into mousemove directly for two reasons:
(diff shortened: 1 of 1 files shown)


floryst:
LGTM. @jourdain we may need to look at how we can support this kind of interaction with the interactor, or if we need to at all.

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -164,6 +164,63 @@ function vtkMouseRangeManipulator(publicAPI, model) {
...
+      publicAPI.startPointerLockInteraction(interactor);
+    }
+  };
+
+  publicAPI.onButtonUp = () => {
+    publicAPI.exitPointerLock();
+  };
+
+  publicAPI.bindEvents = (container) => {
+    document.addEventListener(
```
floryst, line 179: This is already handled in the [RenderWindowInteractor bindEvents](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/RenderWindowInteractor/index.js#L243), and so it shouldn't be necessary here.

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -164,6 +164,63 @@ function vtkMouseRangeManipulator(publicAPI, model) {
...
+      subscription?.unsubscribe();
+    };
+    subscription = interactor?.onEndPointerLock(endInteraction);
+  };
+
+  publicAPI.onPointerLockMove = (e) => {
+    model.previousPosition.x += e.movementX;
+    model.previousPosition.y += e.movementY;
+
+    publicAPI.onMouseMove(null, null, model.previousPosition);
```
floryst, line 211: In thinking about this, I'm thinking that we can instead cache the interactor and renderer in onButtonDown, and used those to pass into onMouseMove here.

floryst (commented):
I did a second pass, and have two comments/changes.

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
floryst: And then clear said cache in onButtonRelease/Up (I forget what it's called).

FezVrasta: Why do we care if they are not used by onMouseMove?

floryst: In thinking generally, the caching pattern would be useful. But you are right, they aren't used in MouseRangeManipulator.onMouseMove, so keeping the nulls is fine.

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
FezVrasta: It should be good now

Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -166,11 +166,64 @@ function vtkMouseRangeManipulator(publicAPI, model) {
...
   publicAPI.onMouseMove = (interactor, renderer, position) => {
     if (!model.verticalListener && !model.horizontalListener) {
       return;
     }
+
+    // We only want to initialize the pointer lock listener
+    // after the user starts moving their mouse, this way
+    // we don't interfere with other events such as doubleClick,
+    // for this reason we don't call this from `onButtonDown`
+    if (model.usePointerLock && !interactor.isPointerLocked()) {
```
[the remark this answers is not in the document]
FezVrasta: done!

floryst:
Thanks! Looks good!

PR #2318, opened 2022-03-15
feat(vtkimageinterpolator): add linear interpolation

laurennlam:
### Context
If we use data headsq.vti in the vtkResliceCursorWidget example, we can see that when we rotate a vue, there are artifacts in the rotated view because of the interpolation. A linear interpolation instead of a nearest interpolation is needed.
image

### Results
Before
image
After
image

### Changes
Add methods in vtkImageInterpolator called when interpolation mode is LINEAR.
Allow to set the interpolation mode from vtkImageReslice.

### PR and Code Checklist


laurennlam:
@finetjul

finetjul:
LGTM

finetjul (approved):

Issue #1291, opened 2019-12-13
How to solve the triangulation problem when the models overlap？


JerryGit1:
In the process of model flipping, there is a problem as shown in the figure
![image](https://user-images.githubusercontent.com/34857855/70787169-179cb780-1dc9-11ea-9d51-2f5d0480d1e4.png)

Normal condition：
![image](https://user-images.githubusercontent.com/34857855/70786916-9fce8d00-1dc8-11ea-8b9d-9f0aff44f6cf.png)

Is there any way to solve this problem?

thank you very much.

martinken:
In the new WebGPU backend we are handling this case using an order independent transparency pass. I don;t have any plans to add that to WebGL but I'll keep this open as feature request.

PR #2319, opened 2022-03-16
feat(WebGL2): add order independent translucent rendering

martinken:
Add a basic order independent translucent rendering technique
for the WebGL backend. Requires WebGL2 support and a common
extension to work. No changes required in existing code as
this is now the default for webgl2.

Addresses https://github.com/Kitware/vtk-js/issues/1291

### DIFF
--- a/Sources/Rendering/OpenGL/Framebuffer/index.js
+++ b/Sources/Rendering/OpenGL/Framebuffer/index.js
@@ -22,6 +22,13 @@ function vtkFramebuffer(publicAPI, model) {
   };
 
   publicAPI.saveCurrentBindings = (modeIn) => {
+    if (!model.context) {
+      macro.vtkErrorMacro(
+        'you must set the OpenGLRenderWindow before calling saveCurrentBindings'
+      );
+      return;
+    }
+
     const gl = model.context;
     model.previousDrawBinding = gl.getParameter(
       model.context.FRAMEBUFFER_BINDING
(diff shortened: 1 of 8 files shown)


Sources/Rendering/OpenGL/Framebuffer/index.js
```diff
@@ -242,7 +247,8 @@ export function extend(publicAPI, model, initialValues = {}) {
   // Build VTK API
   macro.obj(publicAPI, model);
 
-  macro.setGet(publicAPI, model, ['colorTexture']);
+  model.colorTextures = [];
```
finetjul, line 250: by doing so, you prevent passing color textures in constructor.
```
function defaultValues (initialValues) {
  return {
  // openGLRenderWindow: null,
  // glFramebuffer: null,
  colorTextures: [],
  // depthTexture: null,
  previousDrawBinding: 0,
  previousReadBinding: 0,
  prev...,
  ..initialValues
  };
}
function extend(...) {
```

finetjul:
Will it work in case of 2 translucent actors ? (i.e. replaces depth peeling ?)

Sources/Rendering/OpenGL/Framebuffer/index.js
```diff
@@ -242,7 +247,8 @@ export function extend(publicAPI, model, initialValues = {}) {
...
+  macro.setGet(publicAPI, model, ['colorTextures']);
```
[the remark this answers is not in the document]
martinken: Yes, should be just getArray. Fixed.

martinken:
Yes, handles multiple props.

Sources/Rendering/OpenGL/Framebuffer/index.js
martinken: Updated to error on this case and cleaned up the naming and added some error checks.

finetjul:
LGTM

martinken:
@berkgeveci  FYI here is the translucent support for WebGL2

finetjul:
@martinken can you please share some doc/reference on the technique being used here. I have a mismatch between the result rendering and what I expect when cells are opaque (alpha=1).

martinken:
You mean a transparent object with some opaque cells I'm guessing. That will not look correct. The algorithm can be thought of as an opacity and depth weighted average of colors for all translucent pixels. The resulting total opacity is correct but the resulting color is a heuristic. This post gives the basics http://casual-effects.blogspot.com/2014/03/weighted-blended-order-independent.html though our weighting by depth may be different.

Issue #2192, opened 2021-12-09
VtkMath definition error


aShuaizzz:
### High-level description

The matrix parameters defined are inconsistent with the actual matrix parameters used in vtkMath
### Steps to reproduce

      const mP = [
        [dP[0], 0, 0],
        [dP[1], 0, 0],
        [dP[2], 0, 0],
      ];
      const mQ = [
        [dQ[0], dQ[1], dQ[2]],
        [0, 0, 0],
        [0, 0, 0],
      ];

      const H1 = [
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
      ];
      vtkMath.multiplyMatrix(mP, mQ, 3, 1, 1, 3, H1);

### Detailed behavior

![1639033540(1)](https://user-images.githubusercontent.com/45392896/145349872-8806fe0f-9548-4fea-ad4f-2c013dc8aaf4.png)
![image](https://user-images.githubusercontent.com/45392896/145349946-206d45dc-3cdd-4bc3-92c9-459a2e0869ff.png)
![image](https://user-images.githubusercontent.com/45392896/145350008-a096280c-07d0-418d-acbe-0c2da4fc95c3.png)

### Environment
- **vtk.js**: 21.2.0
- **OS**: 
- **Browser**:

aShuaizzz:
both MAT3 and MAT4 in GL-matrix are one-dimensional arrays, but the Matrix operations in vtkMath are two-dimensional arrays, so you need to distinguish them when writing the definition file

daker:
This can be easily fixed if you want to propose a PR https://github.com/Kitware/vtk-js/blob/master/Sources/Common/Core/Math/index.d.ts#L413-L421

daker:
@aShuaizzz i would recommand creating a new type `Matrix3x3` & `Matrix2x2` and replace the gl-matrix types with the new one

```typescript
declare type Matrix2x2 = [Vector2, Vector2];
declare type Matrix3x3 = [Vector3, Vector3, Vector3];
```

aShuaizzz:
@daker That's great,sometimes I need to use m* N matrices when using methods such as solveLeastSquares and would like to include those methods in the definition as well.
![企业微信截图_16391001066819](https://user-images.githubusercontent.com/45392896/145502381-50859c8c-2d11-4e70-912c-9231320d3a2e.png)

PR #2325, opened 2022-03-22
Fix are matrices equal

### LINKED ISSUE #2324: `areMatricesEqual` is not working
### High-level description
`areMatricesEqual` is not working
The function is just a copy of `areEquals` and is only meant to be used on `number[]` not `number[][]`

### Steps to reproduce

 ```js
vtk.Common.Core.vtkMath.areMatricesEqual([[1,1], [2,2]], [[1,1], [2,2]]) \\ returns false
 ```

### Expected behavior
Function should test equality between matrices and return true if all the values are equals (given an epsilon for floating points).

### Possible fix
```js
export function areMatricesEqual(a, b, eps = 1e-6) {
  if (a.length !== b.length) {
    return false;
  }
  return a.every((line,i) => areEquals(line, b[i], eps));
}
```

### Environment
- **vtk.js**: latest
- **OS**: any
- **Browser**: any

stity:
### Context
- fix #2324 

### Results
Fixes `areMatricesEqual`

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Common/Core/Math/index.d.ts
+++ b/Sources/Common/Core/Math/index.d.ts
@@ -466,7 +466,7 @@ export function areEquals(a: number[], b: number[], eps?: number): boolean;
  * @param {Number[][]} b A matrix of numbers
  * @param {Number} [eps] The tolerance value.
  */
-export function areEquals(a: number[][], b: number[][], eps?: number): boolean;
+export function areMatricesEqual(a: number[][], b: number[][], eps?: number): boolean;
 
 /**
  *
--- a/Sources/Common/Core/Math/index.d.ts
+++ b/Sources/Common/Core/Math/index.d.ts
@@ -454,11 +454,19 @@ export function quaternionToMatrix3x3(quat_4: Vector4, mat_3x3: Matrix3x3): void
 
 /**
  * Returns true if elements of both arrays are equals.
- * @param {Number[]} a An array of numbers (vector, point, matrix...)
- * @param {Number[]} b An array of numbers (vector, point, matrix...)
+ * @param {Number[]} a An array of numbers (vector, point...)
+ * @param {Number[]} b An array of numbers (vector, point...)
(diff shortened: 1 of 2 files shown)


stity:
> Why do you want to support matrices stored as numbe[][] ? It is not a matrix format supported in VTK.js, is it ?

In `Sources/types.d.ts` you have the following code:
```ts
declare type Matrix = number[][];
```
Is there another type for matrices that would be supported by the current version of `areMatricesEqual` ?

daker:
@finetjul for ex you can see this in `multiplyMatrix` https://github.com/Kitware/vtk-js/blob/83927a18db3a513ef1fcf141491606c32d784841/Sources/Common/Core/Math/index.js#L574

`a` & `b` are two-dimensional arrays
https://github.com/Kitware/vtk-js/blob/83927a18db3a513ef1fcf141491606c32d784841/Sources/Common/Core/Math/index.js#L588

also in other places in the ts def https://github.com/Kitware/vtk-js/blob/41aea103cf66c7823fa6a40d7e2f1b188d15426e/Sources/Common/Core/Math/index.d.ts

I am wrong?

jourdain:
I think you are correct. Those were ported from the c++ as close as we could. But I don't think they are used. On the JS side we prefer to use the gl-matrix helpers. It might be worth trimming down that vtkMath to only expose what we need and remove any code that is coming from C++ but that is not getting used/exercised.

daker:
We had an issue before about this https://github.com/Kitware/vtk-js/issues/2192

finetjul:
For consistency reasons, we should have only 1 way to represent matrices in VTK.js .
For performance reasons, it should be the `number[]` notation and not `number[][]`.

Who wants to convert all those methods (BREAKING_CHANGES !) to take `number[]` instead of `number[][]`?

PR #2327, opened 2022-03-23
refactor(boundingbox): return updated bounds instead of nothing

finetjul:
Also speed-up vtkBoundingBox constructor

### Context
Some vtkBoundingBox functions do not return anything.
It would be more convenient to return the updated bounds.

### Results
BEFORE:
```
const bbox = vtkBoundingBox.newInstance();
bbox.setBounds(originalBounds);
bbox.inflate(10);
return bbox.getBounds();
```
AFTER:
```
const bbox = vtkBoundingBox.newInstance();
bbox.setBounds(originalBounds);
return bbox.inflate(10);
```

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 10
  - **Browser**: Chrome 99.0.4844.82

### DIFF
--- a/Sources/Common/DataModel/BoundingBox/api.md
+++ b/Sources/Common/DataModel/BoundingBox/api.md
@@ -21,10 +21,12 @@ is less than the min point then the  min point will also be changed
 Change bounding box so it includes the point p
 Note that the bounding box may have 0 volume if its bounds
 were just initialized.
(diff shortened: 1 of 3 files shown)


jourdain:
Love it!

PR #2338, opened 2022-03-29
fix(OBBTree): send almost-zero components to zero

floryst:
### Context
~`vtkMath.jacobi` is giving some strange results when off-diagonal components are not quite zero. In the below code, if `v=0`, then we get what we'd expect: the eigenvectors are the unit basis vectors. But if `v=1e-12`, then we do not get that, as shown below the code listing.~

Upon some more testing, the values are valid. I will do some last-minute testing to verify OBB outputs.

```
const vals = [0, 0, 0];
const vecs = [
  [0, 0, 0],
  [0, 0, 0],
  [0, 0, 0],
];
const v = 1e-12;
vtkMath.jacobi(
  [
    [0.5, v, v],
    [v, 0.5, v],
    [v, v, 0.5],
  ],
  vals,
  vecs
);

console.log(vals)
console.log(vecs)

// eigenvalues: [0.500000000002, 0.499999999999, 0.49999999999899997]
// eigenvectors: 
//  [0.577348850073642, 0.7071067811853349, 0.40825029739163465]
//  [0.5773488500736421, -0.7071067811877602, 0.40825029738743396]
//  [0.5773531074111288, 2.425284842538276e-12, -0.8164945739946553]
```

In the interest of getting the `beta` branch ready, I'm opting to add the zeroing logic to OBBTree, and reserve the `vtkMath.jacobi` fix+discussion for later.

### Results
- OBBTree should(?) be more correct

### Changes
- updated tests for OBBTree

### PR and Code Checklist

### Testing

### DIFF
--- a/Sources/Common/Core/Math/index.js
+++ b/Sources/Common/Core/Math/index.js
@@ -867,7 +867,7 @@ export function jacobiN(a, n, w, v) {
     tmp = w[k];
     for (i = j + 1; i < n; i++) {
       // boundary incorrect, shifted already
-      if (w[i] >= tmp) {
+      if (w[i] >= tmp || Math.abs(w[i] - tmp) < VTK_SMALL_NUMBER) {
         // why exchange if same?
         k = i;
         tmp = w[k];
--- a/Sources/Filters/General/OBBTree/index.js
+++ b/Sources/Filters/General/OBBTree/index.js
@@ -157,6 +157,10 @@ function vtkOBBTree(publicAPI, model) {
     for (let i = 0; i < 3; i++) {
       for (let j = 0; j < 3; j++) {
         a[i][j] = a[i][j] / totMass - mean[i] * mean[j];
+        // send to zero if close to zero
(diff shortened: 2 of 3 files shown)


floryst:
I'm temporarily basing this off of master so that tests can run.

floryst:
Ok, this has subtle effects on Firefox vs Chrome again. This time in determining which eigenvector corresponds to which eigenvalue...

FYI @finetjul

floryst:
Tests pass. @finetjul the updated test values seem to make more sense to me than the previous ones. For instance, the default ArrowSource should have an axis-aligned bounding box with side lengths ~0.2, 0.2, and 1. (Tip radius is 0.1, shaft+tip is 1.) In it's default orientation (along one of the coordinate axes), the arrow source's AABB should be equivalent to its OBB.

finetjul:
Thanks @floryst , I'm fine with your fix for timing reasons (to get beta merged).
We'll also investigate over the next weeks.

PR #2339, opened 2022-03-30
Fix obbtree tests

floryst:
### Context
This is an attempt at more correctly fixing the OBB tests. I noticed that it's entirely possible for two different browsers (either on the same machine, or different machines) to output a different corner of the OBB. See the screenshots below.

Additionally, different platforms and Firefox versions exhibit a drift in computed values, so I increased the epsilon tolerance for comparisons to 0.1. The list of tested configurations is provided below.

Firefox (arrow tip resolution is 120):
image

Chrome (arrow tip resolution is also 120):
image

### Results
- OBB tests should pass regardless of browser.

### Changes
- [x] tests were updated

### PR and Code Checklist

### Testing

- Tested on:
  - github CI
  - vtk.js master on Chromium: 99.0.4844.88 (Official Build) (64-bit) for Windows 10
  - vtk.js master on Firefox: 98.0.2 (64-bit) for Windows 10
  - vtk.js master on Firefox: 91.0 (64-bit) on Ubuntu 18.04.5


floryst:
@finetjul You can use this as a starting point for your own investigations, if you'd like. I went ahead and did some since I already was looking at this yesterday.

joannacirillo:
Here are what the investigations I did led to.
The difference in computations is due to the way floating points are stored and the way the various browsers deal with the issues on computation inherently present with operations on floating points, in particular here due to imprecisions after 15 digits. The floating points are handled according to the [ECMAScript specification](https://en.wikipedia.org/wiki/Double-precision_floating-point_format).
Whatever the corner computed is,  what is important for further computations with the OBBTree is the box and not the actual corner computed. Also in the general case several bounding boxes could be correct, possibly an infinity even (if there are some invariances in rotation for example). Therefore, what is actually necessary to check is only the intersections computed, which is what is done in other tests.
So the conclusion is that actually this test is not relevant and should be removed.

joannacirillo:
Here are what the investigations I did led to.
The difference in computations is due to the way floating points are stored and the way the various browsers deal with the issues on computation inherently present with operations on floating points, in particular here due to imprecisions after 15 digits. The floating points are handled according to the [ECMAScript specification](https://en.wikipedia.org/wiki/Double-precision_floating-point_format).
Whatever the corner computed is, what is important for further computations with the OBBTree is the box and not the actual corner computed. Also in the general case several bounding boxes could be correct, possibly an infinity even (if there are some invariances in rotation for example). Therefore, what is actually necessary to check is only the intersections computed, which is what is done in other tests.
So the conclusion is that actually this test is not relevant and should be removed. The epsilon can be reduced again to 0.0001

finetjul:
LGTM

Issue #2285, opened 2022-02-23
Mobile version cannot pick points to ajust planes


ramaobalta:
### High-level description
Testing ResliceCursorWidget example at https://kitware.github.io/vtk-js/examples/ResliceCursorWidget.html on my mobile, I could not angulate the planes using the points. This feature works fine on the PC Navigator unless you change to debug mode and set mobile mode.

### Steps to reproduce
Open https://kitware.github.io/vtk-js/examples/ResliceCursorWidget.html on the phone
try to pick a point, it looks very small, but this is not the problem, I changed the size on my code, same behavior.
Drag the points. It will only apply window level, the planes will not change.

### Detailed behavior
![image](https://user-images.githubusercontent.com/100292669/155335491-afa15695-92b8-4e5e-ad27-89f365825374.png)

### Expected behavior
The plane should move on the views

### Environment
We tried this example in several phones, Motorola, Samsung, Iphones near January 20, 2022
Chrome, Samsung navigator, default android navigator
And also tried with Chrome for windows - debug mode

finetjul:
The widget does not implement `handleTouchMove`, hance mobile events (no left/right mouse button pressed) are not supported. It shouldn't be too difficult to support if you want to give it a shot.

ramaobalta:
I tried to debug the events and noticed that the touch events are implemented and transfered to mousedown/mouse move events. Is it to possible that it could be using something differente? Maybe a picker?

floryst:
Relevant: https://github.com/Kitware/vtk-js/pull/2348

PR #2348, opened 2022-04-01
Touch support with the widget manager

floryst:
### Context
- Touch events on widgets is currently not supported. Related issues: #2285

### Results
- Touch start should select widget handles.

### Changes

- [x] Added a way to measure promise execution time
- [x] Add touch start support for widget handles
- [x] marks touchstart, touchmove, and touchend as non-passive events to suppress browser warnings about preventDefault in passive event handlers.
- [x] Adds a warning if widget selection takes longer than 0.1 seconds.

#### approach

This change means widget selection no longer happens asynchronously alongside interaction. Instead, widget selection occurs first, and then re-triggers the original event, allowing interactors to handle the event. If widget selection takes significant time, interaction will suffer (which was sort of the case prior to this PR, but only wrt widget picking).

### PR and Code Checklist

### Testing

- Tested widgets:
  - PolyLineWidget
  - SphereWidget
  - ResliceCursorWidget


jourdain:
I'm not following all the details, but seems reasonable.

finetjul:
LGTM

PR #2351, opened 2022-04-04
feat(CubeAxesActor): add bounds extimate to CubeAxesActor

martinken:
Add a basic getBounds method that expands the grid bounds
by a specified factor to accomodate text annotations.

Also fix an issue with get_tmAtlas which may have been introduced
with recent macros.js changes.

### DIFF
--- a/Sources/Common/DataModel/BoundingBox/index.js
+++ b/Sources/Common/DataModel/BoundingBox/index.js
@@ -148,6 +148,28 @@ export function getCenter(bounds) {
   ];
 }
 
+export function scaleAboutCenter(bounds, sx, sy, sz) {
+  if (!isValid(bounds)) {
+    return false;
+  }
+  const center = getCenter(bounds);
+  bounds[0] -= center[0];
+  bounds[1] -= center[0];
+  bounds[2] -= center[1];
+  bounds[3] -= center[1];
+  bounds[4] -= center[2];
+  bounds[5] -= center[2];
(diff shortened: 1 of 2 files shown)


Sources/Common/DataModel/BoundingBox/index.js
```diff
@@ -148,6 +148,28 @@ export function getCenter(bounds) {
...
+  bounds[5] -= center[2];
+  scale(bounds, sx, sy, sz);
+  bounds[0] += center[0];
+  bounds[1] += center[0];
+  bounds[2] += center[1];
+  bounds[3] += center[1];
+  bounds[4] += center[2];
+  bounds[5] += center[2];
+
+  return true;
```
finetjul, line 170: I think it make more sense to return the bounds rather than a boolean

martinken: And it looks like you mostly followed Sebastien's earlier implementation when creating the function

martinken: Regardless, if we change one we should probably change both.

finetjul: Because it's already merged, I'll make a pass on it eventually.

martinken: Likely an API breaking change so best to do it in a sep MR anyhow, but there was a bug in this change so another MR

PR #2355, opened 2022-04-06
feat(manipulator): harmonize and improve widget manipulators

tomsuchel:
Harmonizes the way manipulators are used in widgets so its more consistent. 

Improves the manipulators so they can use 3 different origins: `userOrigin` (set by the user outside of the
widget code), `handleOrigin` (automatically set if the handle have a manipulator mixin) and
`widgetOrigin` (occasionally set in the widget code to give a "global" origin to the manipulator).
The manipulator normal is handled the same way, but can also use camera normal if the property
`useCameraNormal` is set to true.

### Results

Widget manipulator are now more customizable. It is possible to set a manipulator for each handle (ex: `widget.getWidgetState().getHandle().setManipulator(manipulator)`) or for the whole widget (ex: `widget.setManipulator(manipulator)`). In this case, the manipulator will be shared between each handle.

Manipulators will try to set their origin using the following order:
- userOrigin > handleOrigin > widgetOrigin

Manipulators will try to set their normal using the following order:
- userNormal > cameraNormal (*if useCameraNormal is true*) > handleNormal > widgetNormal

This allow to redefine the origin/normal of a manipulator (ex: to force a handle to move along a plane) without creating a custom widgets and override methods.

### Changes

In `LineManipulator`, `PlaneManipulator` and `TrackballManipulator`, added the following APIs:
  - `getUserNormal` / `setUserNormal`
  - `getHandleNormal` / `setHandleNormal`
  - `getWidgetNormal` / `setWidgetNormal`
  - `getUserOrigin` / `setUserOrigin`
  - `getHandleOrigin` / `setHandleOrigin`
  - `getWidgetOrigin` / `setWidgetOrigin`
  - `getUseCameraNormal` / `setUseCameraNormal`
 
`setOrigin` and `setNormal` APIs have been removed
- [x] Documentation and TypeScript definitions were updated to match those changes (todo)

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: v24.3.0
  - **OS**: Ubuntu 20.04 LTS

### DIFF
--- a/Sources/Widgets/Manipulators/AbstractManipulator/index.d.ts
+++ b/Sources/Widgets/Manipulators/AbstractManipulator/index.d.ts
@@ -1,0 +1,221 @@
+import { vtkObject } from "../../../interfaces";
+import { vtkOpenGLRenderWindow } from "../../../Rendering/OpenGL/RenderWindow"
+import { Vector3 } from "../../../types";
+
+/**
+ *
+ */
+export interface IAbstractManipulatorInitialValues {
+    userOrigin?: Vector3;
+    handleOrigin?: Vector3;
+    widgetOrigin?: Vector3;
+    userNormal?: Vector3;
+    handleNormal?: Vector3;
+    widgetNormal?: Vector3;
+}
+
+export interface vtkAbstractManipulator extends vtkObject {
+
+    /**
+     * Get the normal of the line
+     */
+    getNormal(callData: any): Vector3;
+
+    /**
+     * Get the origin of the line
+     */
+    getOrigin(callData: any): Vector3;
+
+    /**
+     * Get the value of useCameraFocalPoint
+     */
+    getUseCameraFocalPoint(): boolean;
+
+    /**
+     * Set the value of useCameraFocalPoint
+     * @param useCameraFocalPoint if true, the focal point of the camera will be used if userOrigin is not set.
+     */
+    setUseCameraFocalPoint(useCameraFocalPoint: boolean): boolean;
+
+    /**
+     * Get the value of useCameraNormal
+     */
+    getUseCameraNormal(): boolean;
+
+    /**
+     * Set the value of useCameraNormal
+     * @param useCameraNormal if true, the normal of the camera will be used if userNormal is not set.
+     */
+    setUseCameraNormal(useCameraNormal: boolean): boolean;
+
+    /**
+     * 
+     * @param callData 
+     * @param glRenderWindow 
+     */
+    handleEvent(callData: any, glRenderWindow: vtkOpenGLRenderWindow): Vector3;
+
+    /* ------------------------------------------------------------------- */
+
+    /**
+     * Set the user normal. 
+     * This normal take precedence on the handleNormal and the widgetNormal.
+     * This normal should not be set within the widget internal code.
+     * @param {Vector3} normal The normal coordinate.
+     */
+    setUserNormal(normal: Vector3): boolean;
+
+    /**
+     * Set the user normal (see setUserNormal).
+     * @param {Number} x The x coordinate.
+     * @param {Number} y The y coordinate.
+     * @param {Number} z The z coordinate.
+     */
+    setUserNormal(x: number, y: number, z: number): boolean;
+
+    /**
+     * Set the user normal (see setUserNormal).
+     * @param {Vector3} normal The normal coordinate.
+     */
+    setUserNormalFrom(normal: Vector3): boolean;
+
+    /**
+     * Set the user origin.
+     * This origin take precedence on the handleOrigin and the widgetOrigin.
+     * This origin should not be set within the widget internal code.
+     * @param {Vector3} origin The coordinate of the origin point.
+     */
+    setUserOrigin(origin: Vector3): boolean;
+
+    /**
+     * Set the user origin (see setUserOrigin).
+     * @param {Number} x The x coordinate of the origin point.
+     * @param {Number} y The y coordinate of the origin point.
+     * @param {Number} z The z coordinate of the origin point.
+     */
+    setUserOrigin(x: number, y: number, z: number): boolean;
+
+    /**
+     * Set the user origin (see setUserOrigin).
+     * @param {Vector3} origin The coordinate of the origin point.
+     */
+    setUserOriginFrom(origin: Vector3): boolean;
+
+    /* ------------------------------------------------------------------- */
+
+    /**
+     * Set the handle normal. 
+     * This normal is used after the userNormal and before the widgetNormal.
+     * This normal is automatically set by any state having a manipulatorMixin,
+     * and can be overridden in the widget code.
+     * @param {Vector3} normal The normal coordinate.
+     */
+    setHandleNormal(normal: Vector3): boolean;
+
+    /**
+     * Set the handle normal (see setHandleNormal).
+     * @param {Number} x The x coordinate.
+     * @param {Number} y The y coordinate.
+     * @param {Number} z The z coordinate.
+     */
+    setHandleNormal(x: number, y: number, z: number): boolean;
+
+    /**
+     * Set the handle normal (see setHandleNormal).
+     * @param {Vector3} normal The normal coordinate.
+     */
+    setHandleNormalFrom(normal: Vector3): boolean;
+
+    /**
+     * Set the handle origin.
+     * This origin is used after the userOrigin and before the widgetOrigin.
+     * This origin is automatically set by any state having a manipulatorMixin,
+     * and can be overridden in the widget code.
+     * @param {Vector3} origin The coordinate of the origin point.
+     */
+    setHandleOrigin(origin: Vector3): boolean;
+
+    /**
+     * Set the handle origin (see setHandleOrigin).
+     * @param {Number} x The x coordinate of the origin point.
+     * @param {Number} y The y coordinate of the origin point.
+     * @param {Number} z The z coordinate of the origin point.
+     */
+    setHandleOrigin(x: number, y: number, z: number): boolean;
+
+    /**
+     * Set the handle origin (see setHandleOrigin).
+     * @param {Vector3} origin The coordinate of the origin point.
+     */
+    setHandleOriginFrom(origin: Vector3): boolean;
+
+    /* ------------------------------------------------------------------- */
+
+    /**
+     * Set the widget normal. 
+     * This normal is used if no other normals are set. 
+     * It can be used to define a normal global to the whole widget.
+     * @param {Vector3} normal The normal coordinate.
+     */
+    setWidgetNormal(normal: Vector3): boolean;
+
+    /**
+     * Set the widget normal (see setWidgetNormal).
+     * @param {Number} x The x coordinate.
+     * @param {Number} y The y coordinate.
+     * @param {Number} z The z coordinate.
+     */
+    setWidgetNormal(x: number, y: number, z: number): boolean;
+
+    /**
+     * Set the widget normal (see setWidgetNormal).
+     * @param {Vector3} normal The normal coordinate.
+     */
+    setWidgetNormalFrom(normal: Vector3): boolean;
+
+    /**
+     * Set the widget origin.
+     * This origin is used if no other origins are set.
+     * It can be used to define an origin global to the whole widget.
+     * @param {Vector3} origin The coordinate of the origin point.
+     */
+    setWidgetOrigin(origin: Vector3): boolean;
+
+    /**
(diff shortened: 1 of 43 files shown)


Sources/Widgets/Manipulators/LineManipulator/index.js
```diff
@@ -36,17 +36,34 @@ export function projectDisplayToLine(
 // ----------------------------------------------------------------------------
 
 function vtkLineManipulator(publicAPI, model) {
-  // Set our classNae
+  // Set our className
   model.classHierarchy.push('vtkLineManipulator');
 
+  // prettier-ignore
```
finetjul, line 42: why ignore ?

Sources/Widgets/Manipulators/LineManipulator/index.js
```diff
@@ -36,17 +36,34 @@ export function projectDisplayToLine(
...
+  publicAPI.getOrigin = () => {
+    if (model.userOrigin) { return model.userOrigin; }
+    if (model.handleOrigin) { return model.handleOrigin; }
+    if (model.widgetOrigin) { return model.widgetOrigin; }
+    return [0, 0, 0]
+  };
+
+  // prettier-ignore
+  publicAPI.getNormal = (camera) => {
```
finetjul, line 51: pass in the renderer instead of the camera

Sources/Widgets/Manipulators/LineManipulator/index.js
```diff
@@ -36,17 +36,34 @@ export function projectDisplayToLine(
 // ----------------------------------------------------------------------------
 
 function vtkLineManipulator(publicAPI, model) {
-  // Set our classNae
+  // Set our className
   model.classHierarchy.push('vtkLineManipulator');
 
+  // prettier-ignore
+  publicAPI.getOrigin = () => {
```
finetjul, line 43: pass in `calldata` instead

Sources/Widgets/Manipulators/TrackballManipulator/index.js
```diff
@@ -83,15 +100,34 @@ function vtkTrackballManipulator(publicAPI, model) {
...
+    [
+      'userNormal',
+      'userOrigin',
+      'handleNormal',
+      'handleOrigin',
+      'widgetNormal',
+      'widgetOrigin',
+    ],
+    3
+  );
```
finetjul, line 130: looks like you need to create a "base" manipulator that exposes all those models/setters/getters/getOrigin()/getNormal()

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
```diff
@@ -70,17 +71,14 @@ export default function widgetBehavior(publicAPI, model) {
     if (
       model.pickable &&
       model.dragable &&
-      model.manipulator &&
       model.activeState &&
       model.activeState.getActive() &&
+      model.activeState.getManipulator &&
```
finetjul, line 76: if activeState has no manipulator, it should use the widget manipulator.

Sources/Widgets/Widgets3D/AngleWidget/index.js
```diff
@@ -69,6 +69,14 @@ function vtkAngleWidget(publicAPI, model) {
     return vtkMath.angleBetweenVectors(vec1, vec2);
   };
 
+  publicAPI.setManipulator = (manipulator) => {
+    model.manipulator = manipulator;
```
finetjul, line 73: call superclass setManipulator (to have the ModifiedEvent triggered)

Sources/Widgets/Manipulators/LineManipulator/index.js
tomsuchel: It was to compact the function but it was unnecessary after all (prettier forbids `if (cond) { return val; }` but not `if (conf) return val;` )

Sources/Widgets/Manipulators/LineManipulator/index.js
tomsuchel: I've replaced `calldata` by `camera` because we sometimes need to call `manipulator.getNormal` without having access to callData. Example in ShapeWidget/behavior.js:

```js
const computeTextPosition = (worldBounds, textPosition, worldMargin = 0) => {
    const viewPlaneOrigin = model.manipulator.getOrigin();
    const viewPlaneNormal = model.manipulator.getNormal(
      model._renderer.getActiveCamera()
    );
    const viewUp = model._renderer.getActiveCamera().getViewUp();
    // ...
```
How could I do this otherwise?

Sources/Widgets/Manipulators/LineManipulator/index.js
tomsuchel: even if it is not used in the function?

Sources/Widgets/Manipulators/LineManipulator/index.js
finetjul: In the ShapeWidget example you shared, it does not seem to be correct to get the "origin"/"normal" from the manipulator.
Text position is not really related to the manipulator but maybe more related ot the handles or the widget itself.

Sources/Widgets/Widgets3D/ImageCroppingWidget/index.js
```diff
@@ -131,31 +131,59 @@ function vtkImageCroppingWidget(publicAPI, model) {
...
+  );
 }
 
 // ----------------------------------------------------------------------------
 
-const DEFAULT_VALUES = {};
+const DEFAULT_VALUES = {
+  // cornerManipulator: null,
+  // edgeManipulator: null,
+  // faceManipulator: null
```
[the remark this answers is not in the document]
tomsuchel: just in case we need to change them? (but currently, there is indeed no need to store them)

Sources/Widgets/Manipulators/LineManipulator/index.js
tomsuchel: I could get it to work using the bounds center as the origin and the camera normal:
```js
const viewPlaneOrigin = [
      (bounds[0] + bounds[1]) * 0.5,
      (bounds[2] + bounds[3]) * 0.5,
      (bounds[4] + bounds[5]) * 0.5,
    ];
const viewPlaneNormal = model._renderer.getActiveCamera().getDirectionOfProjection();
const viewUp = model._renderer.getActiveCamera().getViewUp();
```
I think this makes sense as we always want the text to be centered in the volume (by default) and facing the user camera.

finetjul: +1

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
```diff
@@ -70,14 +71,14 @@ export default function widgetBehavior(publicAPI, model) {
...
       model.activeState &&
       model.activeState.getActive() &&
       !ignoreKey(callData)
     ) {
-      model.manipulator.setOrigin(model.activeState.getOrigin());
-      model.manipulator.setNormal(model._camera.getDirectionOfProjection());
-      const worldCoords = model.manipulator.handleEvent(
+      const manipulator = model.activeState.getManipulator
+        ? model.activeState.getManipulator()
+        : model.manipulator;
```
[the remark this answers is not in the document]
tomsuchel: I could eventually do the following but I fear it can cause confusion.
```js
const manipulator = model.activeState.getManipulator 
    ? model.activeState.getManipulator() || model.manipulator 
    : model.manipulator;
```
I can't find a case where we would want to have a handle with `getManipulator()` returning null. This behavior would probably be an error and it would be best to notify the user rather than use an alternative manipulator. What do you think?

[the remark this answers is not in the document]
finetjul: `const manipulator = model.activeState.?getManipulator() || model.manipulator`

Interestingly enough, setting a null manipulator onto a handle would be a nice way to "disable" mouse events on such handle (and not for the whole widget).

Sources/Widgets/Manipulators/AbstractManipulator/index.d.ts
```diff
@@ -0,0 +1,218 @@
...
+    widgetOrigin?: number[];
+    widgetNormal?: number[];
+}
+
+export interface vtkAbstractManipulator extends vtkObject {
+
+    /**
+     * Get the normal of the line
+     */
+    getNormal(): number[];
```
finetjul, line 20: incorrect signature

Sources/Widgets/Manipulators/AbstractManipulator/index.d.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    /**
+     * Get the normal of the line
+     */
+    getNormal(callData: any): number[];
+
+    /**
+     * Get the normal of the line
+     */
+    getNormalByReference(): number[];
```
finetjul, line 25: you did not implement this method

Sources/Widgets/Manipulators/LineManipulator/index.js
```diff
@@ -36,17 +38,15 @@ export function projectDisplayToLine(
...
 
-  // --------------------------------------------------------------------------
-
   publicAPI.handleEvent = (callData, glRenderWindow) =>
     projectDisplayToLine(
       callData.position.x,
       callData.position.y,
-      model.origin,
-      model.normal,
+      publicAPI.getOrigin(),
```
finetjul, line 48: pass `callData` as parameter

Sources/Widgets/Widgets3D/LineWidget/index.js
```diff
@@ -174,12 +184,15 @@ function vtkLineWidget(publicAPI, model) {
   });
 
   // Default manipulator
-  model.manipulator = vtkPlanePointManipulator.newInstance();
+  publicAPI.setManipulator(
+    vtkPlanePointManipulator.newInstance({ useCameraNormal: true })
+  );
```
finetjul, line 189: a manipulator could be set via initial values, hence you should not overwrite it

Sources/Widgets/Widgets3D/SphereWidget/behavior.js
```diff
@@ -58,7 +56,9 @@ export default function widgetBehavior(publicAPI, model) {
   }
 
   function currentWorldCoords(e) {
-    return model.manipulator.handleEvent(e, model._apiSpecificRenderWindow);
+    return model.activeState
+      .getManipulator()
```
finetjul, line 60: to be safe, I would use model.manipulator as fallback

finetjul (changes requested):

Sources/Widgets/Widgets3D/EllipseWidget/index.js
```diff
@@ -103,7 +105,7 @@ function defaultValues(initialValues) {
 
 export function extend(publicAPI, model, initialValues = {}) {
   vtkShapeWidget.extend(publicAPI, model, defaultValues(initialValues));
-  macro.setGet(publicAPI, model, ['manipulator', 'widgetState']);
```
finetjul, line 106: why remove the getManipulator() ?

Sources/Widgets/Manipulators/AbstractManipulator/index.d.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * Get the normal of the line
+     */
+    getNormal(callData: any): number[];
+
+    /**
+     * Get the origin of the line
+     */
+    getOrigin(callData: any): number[];
```
floryst, line 26: I would recommend using Vector3 from `interfaces.d.ts` for these arrays, since it'll always be in 3D.

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
```diff
@@ -67,17 +68,18 @@ export default function widgetBehavior(publicAPI, model) {
   // --------------------------------------------------------------------------
 
   publicAPI.handleMouseMove = (callData) => {
+    const manipulator = model.activeState?.getManipulator
+      ? model.activeState.getManipulator()
+      : model.manipulator;
```
floryst, line 73: This can be rewritten as `const manipulator = model.activeState?.getManipulator?.() ?? model.manipulator` for conciseness. Same with all of the other instances of this.

floryst (commented):
What do you think about adding a `useCameraFocalPoint` as a default origin? For instance, the PolyLineWidget does this [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Widgets3D/PolyLineWidget/behavior.js#L23-L24).

floryst (approved):
LGTM

PR #2357, opened 2022-04-07
fix(WebGL): fix translucent selection on WebGL

martinken:
The new order independent translucent pass was not
being bypassed during selection.


martinken:
Tested by @laurennlam

PR #2358, opened 2022-04-07
docs(Examples): delete mesh clip plane example

### LINKED ISSUE #2333: MeshClipPlane example is 404
### High-level description

The MeshClipPlane example is not build correctly due to the use of the underscore in the folder name.

### Steps to reproduce

https://kitware.github.io/vtk-js/examples/_MeshClipPlane.html

### Detailed behavior

MeshClipPlane example returns 404
image

### Expected behavior

MeshClipPlane page should be open and not return 404

### Environment
- **vtk.js**: 
- **OS**: 
- **Browser**: 

@ebrahimebrahim

floryst:
### Context
- Closes #2333 

@ebrahimebrahim This will remove this example, since it was meant for testing only. Let me know if you'd rather keep it, in which case we'd remove the underscore.

### Results
- Removes the mesh clip plane example, since it wasn't meant to show up in the examples anyways.

### Changes

### PR and Code Checklist

### Testing
No tests were affected.


ebrahimebrahim:
This is fine with me. The underscore was initially there because that was a way to keep the example without publishing it. That aspect seems to have stopped working, so I'm fine with just deleting it if it's not useful.

PR #2369, opened 2022-04-14
feat(hardwareselector): add hardware picking for point and cell ids

boucaud:
### Context
The WebGL hardware selector currently does not support picking point and cell ids.

### Results
This adds support for picking points and cells with the hardware selector. Currently only implemented for the polydata mapper.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/Actor/index.js
+++ b/Sources/Rendering/Core/Actor/index.js
@@ -158,6 +158,12 @@ function vtkActor(publicAPI, model) {
 
   publicAPI.getSupportsSelection = () =>
     model.mapper ? model.mapper.getSupportsSelection() : false;
+
+  publicAPI.processSelectorPixelBuffers = (selector, pixelOffsets) => {
+    if (model.mapper && model.mapper.processSelectorPixelBuffers) {
+      model.mapper.processSelectorPixelBuffers(selector, pixelOffsets);
+    }
+  };
 }
 
 // ----------------------------------------------------------------------------
--- a/Sources/Rendering/OpenGL/Helper/index.js
+++ b/Sources/Rendering/OpenGL/Helper/index.js
@@ -44,6 +44,10 @@ function vtkOpenGLHelper(publicAPI, model) {
       const mode = publicAPI.getOpenGLMode(rep);
       const wideLines = publicAPI.haveWideLines(ren, actor);
       const gl = model.context;
+      const depthMask = gl.getParameter(gl.DEPTH_WRITEMASK);
+      if (model.pointPicking) {
+        gl.depthMask(false);
+      }
       const drawingLines = mode === gl.LINES;
       if (drawingLines && wideLines) {
         publicAPI.updateShaders(ren, actor, oglMapper);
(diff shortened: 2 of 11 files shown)


boucaud:
@martinken @finetjul I would like to have your opinion on this before I go further.

finetjul:
How about an example that prints whatever information you can get from picking when you move the mouse around ?

Sources/Rendering/OpenGL/Helper/index.js
```diff
@@ -169,11 +179,35 @@ function vtkOpenGLHelper(publicAPI, model) {
...
+      model.primitiveType === primTypes.Points ||
+      actor.getProperty().getRepresentation() === Representation.POINTS
+    ) {
+      publicAPI
+        .getProgram()
+        .setUniformf('pointSize', actor.getProperty().getPointSize());
+    } else if (model.pointPicking) {
+      publicAPI
+        .getProgram()
+        .setUniformf('pointSize', publicAPI.getPointPickingPrimitiveSize());
```
finetjul, line 192: why not use actor property point size ?

finetjul:
How about an example that prints whatever information you get from the picking when moving the mouse around ?

Sources/Rendering/Core/Actor/index.js
```diff
@@ -158,6 +158,12 @@ function vtkActor(publicAPI, model) {
```
finetjul, line 158: How about an example that prints out all the information you get by picking when moving the mouse around ?

martinken:
Wow, this is a lot of progress on a tough and tricky area. A couple quick comments in terms of approach.

1) I would suggest not using gl_PrimitiveID. The reason is that some hardware systems increment gl_PrimitiveID with degenerate triangles and other hardware doesn't. In vtk c++ we actually check the mesh for degenerate primitives and build a mapping of vtk cell ids to gl primitiveIds based on that (plus other stuff). Long story short avoid using gl_PrimitiveID. I found this out the hard way when some systems started failing on cell coloring with meshes that had degenerate triangles.  Was a real pain to work around at that point.

2) can you use flat with webgl1 ? If not then wouldn't this code break webgl1? We are dropping webgl1 soon so maybe not a big deal.

3) Not sure if you did this, but the approach for point picking on the C++ side is that we first render the surface to set the zbuffer, then we switch to point mode, offset the points a bit using coincident settings and render as points. The idea is that we want to prevent points that are occluded by the surface from being picked. We don't have to do that. The question is do we want a pick that visually looks to have missed any visible points, to return a point that is occluded by the surface. On C++ we decided no. 

I can take a more detailed look but wanted to get these ideas to you first to get your feeling on them.

boucaud:
Thanks ! I'm going for option 2. This implementation already builds a map from gl_VertexID to vtkPointId in the CABO building method, so it should be fairly simple.

martinken:
Looking now.

Sources/Rendering/OpenGL/Helper/index.js
martinken: Ahh you do have the 6, 4, 2 code here.

martinken:
Code looks good, only one issue that I see. There is a failing test that is designed to make sure picking is fast. So you'll want to check that out and see what is happening there and fix the issue or change the test. I looked at the code and it doesn't look like you rebuild the CABO during selection passes which would be my big concern as that can take a long time for large datasets (rebuilding the cabo every pick). So I'm guessing it is just the cost of extra pointId / cellId conversions but take a look.

boucaud:
@martinken I have added a few relevant tests, can you take another look ?

floryst:
Last ping @martinken, since @boucaud needs this soon. Otherwise I think this good to merge.

martinken (approved):
LGTM

PR #2370, opened 2022-04-15
Fix inconsistent matrix representation

joannacirillo:
### Context

fix #2325.
In issue #2325, it was mentionned that it is preferable to have only one format for matrices. For performance reason, it should be number[] .

### Results

The matrix format used in the math module (and therefore wherever those functions are called) are in number[] format. Tests have been added for these functions.

### Changes

Changed : testMath, math, TextureMapToPlane, MeshClipPlane, OBBTree, LandmarkTransform, ImageMapper, Line, Triangle

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 24.5.2
  - **OS**: Windows 10
  - **Browser**: Chrome 100.0.4896.88

### DIFF
--- a/Sources/Common/Core/Math/index.js
+++ b/Sources/Common/Core/Math/index.js
@@ -17,22 +17,34 @@ function notImplemented(method) {
   return () => vtkErrorMacro(`vtkMath::${method} - NOT IMPLEMENTED`);
 }
 
-function vtkSwapVectors3(v1, v2) {
-  for (let i = 0; i < 3; i++) {
-    const tmp = v1[i];
-    v1[i] = v2[i];
-    v2[i] = tmp;
+// Swap rows for n by n matrix
+function swapRowsMatrix_nxn(matrix, n, row1, row2) {
+  let tmp;
+  for (let i = 0; i < n; i++) {
+    tmp = matrix[row1 * n + i];
+    matrix[row1 * n + i] = matrix[row2 * n + i];
+    matrix[row2 * n + i] = tmp;
   }
 }
 
-function createArray(size = 3) {
-  // faster than Array.from and/or while loop
-  return Array(size).fill(0);
+// Swap columns for n by n matrix
+function swapColumnsMatrix_nxn(matrix, n, column1, column2) {
+  let tmp;
+  for (let i = 0; i < n; i++) {
+    tmp = matrix[i * n + column1];
+    matrix[i * n + column1] = matrix[i * n + column2];
+    matrix[i * n + column2] = tmp;
+  }
 }
 
 // ----------------------------------------------------------------------------
 // Global methods
 // ----------------------------------------------------------------------------
+
+export function createArray(size = 3) {
+  // faster than Array.from and/or while loop
+  return Array(size).fill(0);
+}
 
 export const Pi = () => Math.PI;
 
@@ -195,11 +207,15 @@ export function dot(x, y) {
 }
 
 export function outer(x, y, out_3x3) {
-  for (let i = 0; i < 3; i++) {
-    for (let j = 0; j < 3; j++) {
-      out_3x3[i][j] = x[i] * y[j];
-    }
-  }
+  out_3x3[0] = x[0] * y[0];
+  out_3x3[1] = x[0] * y[1];
+  out_3x3[2] = x[0] * y[2];
+  out_3x3[3] = x[1] * y[0];
+  out_3x3[4] = x[1] * y[1];
+  out_3x3[5] = x[1] * y[2];
+  out_3x3[6] = x[2] * y[0];
+  out_3x3[7] = x[2] * y[1];
+  out_3x3[8] = x[2] * y[2];
 }
 
 export function cross(x, y, out) {
@@ -378,11 +394,10 @@ export function gaussianWeight(mean, variance, position) {
 }
 
 export function outer2D(x, y, out_2x2) {
-  for (let i = 0; i < 2; i++) {
-    for (let j = 0; j < 2; j++) {
-      out_2x2[i][j] = x[i] * y[j];
-    }
-  }
+  out_2x2[0] = x[0] * y[0];
+  out_2x2[1] = x[0] * y[1];
+  out_2x2[2] = x[1] * y[0];
+  out_2x2[3] = x[1] * y[1];
 }
 
 export function norm2D(x2D) {
@@ -396,6 +411,44 @@ export function normalize2D(x) {
     x[1] /= den;
   }
   return den;
+}
+
+export function rowsToMat4(row0, row1, row2, row3, mat) {
+  for (let i = 0; i < 4; i++) {
+    mat[i] = row0[i];
+    mat[4 + i] = row1[i];
+    mat[8 + i] = row2[i];
+    mat[12 + i] = row3[i];
+  }
+  return mat;
+}
+
+export function columnsToMat4(column0, column1, column2, column3, mat) {
+  for (let i = 0; i < 4; i++) {
+    mat[4 * i] = column0[i];
+    mat[4 * i + 1] = column1[i];
+    mat[4 * i + 2] = column2[i];
+    mat[4 * i + 3] = column3[i];
+  }
+  return mat;
+}
+
+export function rowsToMat3(row0, row1, row2, mat) {
+  for (let i = 0; i < 3; i++) {
+    mat[i] = row0[i];
+    mat[3 + i] = row1[i];
+    mat[6 + i] = row2[i];
+  }
+  return mat;
+}
+
+export function columnsToMat3(column0, column1, column2, mat) {
+  for (let i = 0; i < 3; i++) {
+    mat[3 * i] = column0[i];
+    mat[3 * i + 1] = column1[i];
+    mat[3 * i + 2] = column2[i];
+  }
+  return mat;
 }
 
 export function determinant2x2(...args) {
@@ -416,11 +469,11 @@ export function LUFactor3x3(mat_3x3, index_3) {
 
   // Loop over rows to get implicit scaling information
   for (let i = 0; i < 3; i++) {
-    largest = Math.abs(mat_3x3[i][0]);
-    if ((tmp = Math.abs(mat_3x3[i][1])) > largest) {
+    largest = Math.abs(mat_3x3[i * 3]);
+    if ((tmp = Math.abs(mat_3x3[i * 3 + 1])) > largest) {
       largest = tmp;
     }
-    if ((tmp = Math.abs(mat_3x3[i][2])) > largest) {
+    if ((tmp = Math.abs(mat_3x3[i * 3 + 2])) > largest) {
       largest = tmp;
     }
     scale[i] = 1 / largest;
@@ -429,41 +482,40 @@ export function LUFactor3x3(mat_3x3, index_3) {
   // Loop over all columns using Crout's method
 
   // first column
-  largest = scale[0] * Math.abs(mat_3x3[0][0]);
+  largest = scale[0] * Math.abs(mat_3x3[0]);
   maxI = 0;
-  if ((tmp = scale[1] * Math.abs(mat_3x3[1][0])) >= largest) {
+  if ((tmp = scale[1] * Math.abs(mat_3x3[3])) >= largest) {
     largest = tmp;
     maxI = 1;
   }
-  if ((tmp = scale[2] * Math.abs(mat_3x3[2][0])) >= largest) {
+  if ((tmp = scale[2] * Math.abs(mat_3x3[6])) >= largest) {
     maxI = 2;
   }
   if (maxI !== 0) {
-    vtkSwapVectors3(mat_3x3[maxI], mat_3x3[0]);
+    swapRowsMatrix_nxn(mat_3x3, 3, maxI, 0);
     scale[maxI] = scale[0];
   }
   index_3[0] = maxI;
 
-  mat_3x3[1][0] /= mat_3x3[0][0];
-  mat_3x3[2][0] /= mat_3x3[0][0];
+  mat_3x3[3] /= mat_3x3[0];
+  mat_3x3[6] /= mat_3x3[0];
 
   // second column
-  mat_3x3[1][1] -= mat_3x3[1][0] * mat_3x3[0][1];
-  mat_3x3[2][1] -= mat_3x3[2][0] * mat_3x3[0][1];
-  largest = scale[1] * Math.abs(mat_3x3[1][1]);
+  mat_3x3[4] -= mat_3x3[3] * mat_3x3[1];
+  mat_3x3[7] -= mat_3x3[6] * mat_3x3[1];
+  largest = scale[1] * Math.abs(mat_3x3[4]);
   maxI = 1;
-  if ((tmp = scale[2] * Math.abs(mat_3x3[2][1])) >= largest) {
+  if ((tmp = scale[2] * Math.abs(mat_3x3[7])) >= largest) {
     maxI = 2;
-    vtkSwapVectors3(mat_3x3[2], mat_3x3[1]);
+    swapRowsMatrix_nxn(mat_3x3, 3, 1, 2);
     scale[2] = scale[1];
   }
   index_3[1] = maxI;
-  mat_3x3[2][1] /= mat_3x3[1][1];
+  mat_3x3[7] /= mat_3x3[4];
 
   // third column
-  mat_3x3[1][2] -= mat_3x3[1][0] * mat_3x3[0][2];
-  mat_3x3[2][2] -=
-    mat_3x3[2][0] * mat_3x3[0][2] + mat_3x3[2][1] * mat_3x3[1][2];
+  mat_3x3[5] -= mat_3x3[3] * mat_3x3[2];
+  mat_3x3[8] -= mat_3x3[6] * mat_3x3[2] + mat_3x3[7] * mat_3x3[5];
   index_3[2] = 2;
 }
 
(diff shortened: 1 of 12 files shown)


joannacirillo:
@luciemac

Sources/Filters/Texture/TextureMapToPlane/index.js
```diff
@@ -87,19 +87,23 @@ function vtkTextureMapToPlane(publicAPI, model) {
...
     }
 
-    matrix[0] = v;
-    matrix[1] = c2;
-    matrix[2] = c3;
+    for (i = 0; i < 3; i++) {
+      matrix[i] = v[i];
+      matrix[i + 3] = c2[i];
+      matrix[i + 6] = c3[i];
+    }
```
finetjul, line 100: for performance reasons, I would simply do `matrix = [v[0], v[1], v[2], c2[0], ...]` instead of a for loop.

Sources/Rendering/OpenGL/PolyDataMapper/test/testMoreClippingPlanes.js
```diff
@@ -96,9 +96,15 @@ test('Test PolyDataMapper Clipping Planes 2', (t) => {
...
-    [0, 0, 1],
+    Math.cos(theta),
+    Math.sin(theta),
+    0,
+    -Math.sin(theta),
+    Math.cos(theta),
+    0,
+    0,
+    0,
+    1,
```
finetjul, line 107: please list 3 elements per line (and disable linter temporarily

Sources/Common/Core/Math/index.js
```diff
@@ -383,7 +375,7 @@ export function gaussianWeight(mean, variance, position) {
 export function outer2D(x, y, out_2x2) {
   for (let i = 0; i < 2; i++) {
     for (let j = 0; j < 2; j++) {
-      out_2x2[i][j] = x[i] * y[j];
+      out_2x2[i * 2 + j] = x[i] * y[j];
```
finetjul, line 378: replace nested for loops with hardcoded exhaustive assignment (can be done in a separate PR)

Sources/Common/Core/Math/index.js
```diff
@@ -17,14 +17,6 @@ function notImplemented(method) {
   return () => vtkErrorMacro(`vtkMath::${method} - NOT IMPLEMENTED`);
 }
 
-function vtkSwapVectors3(v1, v2) {
```
finetjul, line 20: you could keep it with the following signature: `vtkSwapMatrixColumns(matrix, column1, column2)` (or vtkSwapMatrixRows?)

Sources/Common/Core/Math/index.js
```diff
@@ -530,44 +532,29 @@ export function linearSolve3x3(mat_3x3, x_3, y_3) {
...
-      a_3x3[1][1] * b_3x3[1][i] +
-      a_3x3[1][2] * b_3x3[2][i];
-    tmp[2][i] =
-      a_3x3[2][0] * b_3x3[0][i] +
-      a_3x3[2][1] * b_3x3[1][i] +
-      a_3x3[2][2] * b_3x3[2][i];
+  const copyA = createArray(9);
+  const copyB = createArray(9);
+  for (let j = 0; j < 9; j++) {
+    copyA[j] = a_3x3[j];
```
finetjul, line 548: it is useless to `createArray` filled with 0s (more costly than `new Array(9)` ) and overwrite it just after

finetjul: use `a_3x3.slice(0, 9)`

Sources/Common/Core/Math/index.js
```diff
@@ -577,47 +564,57 @@ export function multiplyMatrix(a, b, rowA, colA, rowB, colB, out_rowXcol) {
...
   }
 
+  const copyA = createArray(9);
+  const copyB = createArray(9);
+  for (let j = 0; j < rowA * colA; j++) {
+    copyA[j] = a[j];
+  }
+  for (let j = 0; j < rowB * colB; j++) {
+    copyB[j] = b[j];
+  }
```
finetjul, line 574: why a copy is needed here ? please add comment

Sources/Common/Core/Math/index.js
```diff
@@ -577,47 +564,57 @@ export function multiplyMatrix(a, b, rowA, colA, rowB, colB, out_rowXcol) {
...
       // sum for this point
       for (let k = 0; k < colA; k++) {
-        out_rowXcol[i][j] += a[i][k] * b[k][j];
+        out_rowXcol[i * colB + j] += copyA[i * colA + k] * copyB[j + colB * k];
       }
     }
   }
 }
 
 export function transpose3x3(in_3x3, outT_3x3) {
```
finetjul, line 589: why not keep the same logic as the previous implementation (using a tmp value)?

Sources/Common/Core/Math/index.js
```diff
@@ -634,38 +631,43 @@ export function invert3x3(in_3x3, outI_3x3) {
 
   // Divide by the determinant
   const det = a1 * d1 + b1 * d2 + c1 * d3;
+  if (det === 0) {
+    vtkWarningMacro('Matrix has determinant 0');
```
finetjul, line 635: has 0 determinant

Sources/Common/Core/Math/index.js
```diff
@@ -634,38 +631,43 @@ export function invert3x3(in_3x3, outI_3x3) {
...
 }
 
+export function identity3x3(mat_3x3) {
+  for (let i = 0; i < 9; i++) {
+    mat_3x3[i] = 0.0;
+    if (i === 0 || i === 4 || i === 8) {
+      mat_3x3[i] = 1.0;
+    }
+  }
+}
```
finetjul, line 669: previous implementation was more elegant (and faster)

Sources/Common/Core/Math/index.js
```diff
@@ -757,23 +766,28 @@ export function jacobiN(a, n, w, v) {
...
-    for (iq = 0; iq < n; iq++) {
-      v[ip][iq] = 0.0;
+  for (ip = 0; ip < n * n; ip++) {
+    v[ip] = 0.0;
+    if (ip % (n + 1) === 0) {
+      v[ip] = 1.0;
     }
-    v[ip][ip] = 1.0;
+    // v[ip] = ip % (n + 1) === 0 ? 1.0 : 0.0;
   }
```
finetjul, line 788: previous implementation was more elegant.

(remove comment)

you could create an identity function that takes n as parameter

Sources/Common/Core/Math/index.js
```diff
@@ -757,23 +766,28 @@ export function jacobiN(a, n, w, v) {
   let tmp;
   const b = createArray(n);
   const z = createArray(n);
+  const a_temp = createArray(n * n);
+  for (let index = 0; index < n * n; index++) {
+    a_temp[index] = a[index];
+  }
```
finetjul, line 772: `a_temp = [...a]` or  `a.slice();`

Sources/Common/Core/Math/index.js
```diff
@@ -890,66 +904,60 @@ export function jacobiN(a, n, w, v) {
...
+        v[i * n + j] *= -1.0;
       }
     }
   }
+  // for (i = 0; i < n; i++) {
+  //   w[i] = roundNumber(w[i], 6);
+  //   for (j = 0; j < n; j++) {
+  //     v[i * n + j] = roundNumber(v[i * n + j], 6);
+  //   }
+  // }
```
finetjul, line 923: to remove

Sources/Common/Core/Math/index.js
finetjul: why not `copyA` to be consistent with the rest of your changes ?

Sources/Common/Core/Math/index.js
```diff
@@ -798,21 +812,21 @@ export function jacobiN(a, n, w, v) {
...
 
         // after 4 sweeps
         if (
           i > 3 &&
           Math.abs(w[ip]) + g === Math.abs(w[ip]) &&
           Math.abs(w[iq]) + g === Math.abs(w[iq])
         ) {
-          a[ip][iq] = 0.0;
-        } else if (Math.abs(a[ip][iq]) > tresh) {
+          a_temp[ip * n + iq] = 0.0;
```
finetjul, line 823: it no longer modifies input `a`, is it on purpose ?

Sources/Common/Core/Math/index.js
```diff
@@ -1083,40 +1104,38 @@ export function diagonalize3x3(a_3x3, w_3, v_3x3) {
...
-    identity3x3(v_3x3);
+    v_3x3[0] = 1.0;
+    v_3x3[4] = 1.0;
+    v_3x3[8] = 1.0;
+    v_3x3[1] = 0.0;
+    v_3x3[2] = 0.0;
+    v_3x3[3] = 0.0;
+    v_3x3[5] = 0.0;
+    v_3x3[6] = 0.0;
+    v_3x3[7] = 0.0;
```
finetjul, line 1120: use identity3x3

Sources/Common/Core/Math/index.js
```diff
@@ -1126,29 +1145,49 @@ export function diagonalize3x3(a_3x3, w_3, v_3x3) {
...
+        vectTmp1,
+        [v_3x3T[maxI * 3], v_3x3T[maxI * 3 + 1], v_3x3T[maxI * 3 + 2]],
+        vectTmp2
+      );
+      for (let t = 0; t < 3; t++) {
+        v_3x3T[k * 3 + t] = vectTmp1[t];
+      }
+      for (let t = 0; t < 3; t++) {
+        v_3x3T[j * 3 + t] = vectTmp2[t];
+      }
```
finetjul, line 1187: merge both loops into one loop

Sources/Common/Core/Math/index.js
```diff
@@ -1126,29 +1145,49 @@ export function diagonalize3x3(a_3x3, w_3, v_3x3) {
...
+      v_3x3T[j * 3 + 2] = 0.0;
+      v_3x3T[j * 3 + j] = 1.0;
+      const vectTmp1 = createArray(3);
+      cross(
+        [v_3x3T[maxI * 3], v_3x3T[maxI * 3 + 1], v_3x3T[maxI * 3 + 2]],
+        [v_3x3T[j * 3], v_3x3T[j * 3 + 1], v_3x3T[j * 3 + 2]],
+        vectTmp1
+      );
+      normalize(vectTmp1);
+      const vectTmp2 = createArray(3);
```
finetjul, line 1176: not needed -> `const vecTmp2 = cross (...., []);`

Sources/Common/Core/Math/index.js
```diff
@@ -1083,40 +1104,38 @@ export function diagonalize3x3(a_3x3, w_3, v_3x3) {
...
+    v_3x3[6] = 0.0;
+    v_3x3[7] = 0.0;
     return;
   }
 
   // transpose temporarily, it makes it easier to sort the eigenvectors
-  transpose3x3(v_3x3, v_3x3);
+  // copy eigenvectors vector before transpose
+  const v_3x3T = createArray(9);
+  transpose3x3(v_3x3, v_3x3T);
```
finetjul, line 1127: why create a new variable, simply use `v_3x3`, no ? like before

Sources/Common/Core/Math/test/testMath.js
```diff
@@ -1,6 +1,14 @@
 import test from 'tape-catch';
 import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
 
+const createArray = (size = 3) => {
```
finetjul, line 4: export it from vtkMath if you need it here instead of re-declaring it

finetjul (changes requested):

Sources/Common/Core/Math/index.js
joannacirillo: Yes, the modification of `a` during computation of eigen values and vectors made the test of diagonalize3x3 function tricky. The verification of `D = P^-1 A P` is done by computing the eigen vectors on one hand and computing the diagonalization on the another hand. As `a` was changed in the computation of the eigen values and vectors `D = P^-1 A P` was false even though the computation of `D` and `P` was correct. Since the modified value of `a` was not meaningfull, I thought it was better to copy the matrix inside of jacobiN before doing the computation than having to copy it whenever we want to pass it to jacobiN. It is not the same implementation as vtk C++ though.

Sources/Common/Core/Math/index.js
joannacirillo: Do you mean avoiding the copy of a and b by storing the result in a temporary Array and then copy it into out_rowXcol ?

Sources/Common/Transform/LandmarkTransform/test/testLandmarkTransform.js
```diff
@@ -0,0 +1,54 @@
...
+  transform.update();
+  transformMatrix = transform.getMatrix();
+  t.ok(
+    vtkMath.areEquals(
+      transformMatrix,
+      [
+        1, -4.9960036106613045e-17, -1.1657341758720898e-16, 0,
+        4.9960036106613045e-17, 1, -4.9960036106612817e-17, 0,
+        1.1657341758720898e-16, 4.9960036106612817e-17, 1, 0,
+        2.9999999999999996, 3, 3.000000000000001, 1,
```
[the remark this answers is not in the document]
joannacirillo: I rounded the values to 1e-6 (since it is the order used in `areEquals`)

Sources/Common/Core/Math/index.js
```diff
@@ -881,7 +881,7 @@ export function jacobiN(a, n, w, v) {
     tmp = w[k];
     for (i = j + 1; i < n; i++) {
       // boundary incorrect, shifted already
-      if (w[i] >= tmp) {
```
finetjul: such fix, if needed, must be done in a separate PR

Sources/Filters/Texture/TextureMapToPlane/index.js
```diff
@@ -87,7 +87,7 @@ function vtkTextureMapToPlane(publicAPI, model) {
     const c1 = [m[0], m[1], m[2]];
```
finetjul, line 87: line 61-63 no longer seem to be needed

Sources/Filters/Texture/TextureMapToPlane/index.js
```diff
@@ -87,7 +87,7 @@ function vtkTextureMapToPlane(publicAPI, model) {
...
     const c2 = [m[3], m[4], m[5]];
     const c3 = [m[6], m[7], m[8]];
-    const matrix = [c1, c2, c3];
+    const matrix = m;
```
finetjul: this new variable matrix does not seem needed anymore, you could use `m` instead

Sources/Common/Core/Math/index.js
```diff
@@ -17,7 +17,17 @@ function notImplemented(method) {
   return () => vtkErrorMacro(`vtkMath::${method} - NOT IMPLEMENTED`);
 }
 
-function createArray(size = 3) {
+// Swap columns for n by n matrix
+function swapRowsOrColumnsMatrix_nxn(matrix, n, col1, col2) {
```
finetjul: why name it "swapRows" if it only swap columns ? -> swapColumnsMatrix_nxn

Sources/Common/Core/Math/index.js
```diff
@@ -660,14 +661,21 @@ export function determinant3x3(mat_3x3) {
...
-      mat_3x3[i] = 1.0;
-    }
+  for (let i = 0; i < 3; i++) {
+    /* eslint-disable-next-line no-multi-assign */
+    mat_3x3[i * 3] = mat_3x3[i * 3 + 1] = mat_3x3[i * 3 + 2] = 0;
+    mat_3x3[i * 3 + i] = 1;
   }
 }
 
+export function identity(n) {
```
finetjul, line 672: pass mat as function input parameter

Sources/Common/Core/Math/index.js
```diff
@@ -766,10 +774,7 @@ export function jacobiN(a, n, w, v) {
   let tmp;
   const b = createArray(n);
   const z = createArray(n);
-  const a_temp = createArray(n * n);
-  for (let index = 0; index < n * n; index++) {
-    a_temp[index] = a[index];
-  }
+  const copyA = a.slice();
```
finetjul: use spread operator

finetjul (changes requested):

Sources/Common/Core/Math/index.js
joannacirillo: Some confusion on my side (due to the fact that both swaping columns and rows are required in different part of the code), last commit should solve this.

Sources/Common/Core/Math/index.js
joannacirillo: I think last commit solves this problem.

finetjul (changes requested):

Sources/Common/Core/Math/test/testMath.js
```diff
@@ -162,7 +162,13 @@ test('Test Identity3x3', (t) => {
...
+  // prettier-ignore
+  const m = [
+    0, 4, 0, 0,
+    2, 2, 1, 0,
+    0, 0, 9, 0,
+    0, 3, 7, 0]
+  vtkMath.identity(m, 4);
   // prettier-ignore
   t.ok(
     vtkMath.areEquals(m, [
```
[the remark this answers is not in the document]
joannacirillo: Is it required that m can be any size before being set to Identity ?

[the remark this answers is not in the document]
finetjul: no, just empty.
fyi, the following works:
```
const m = [];
m[10] = 0;
```

Sources/Filters/General/OBBTree/index.js
```diff
@@ -153,39 +153,37 @@ function vtkOBBTree(publicAPI, model) {
...
-    min[2] = v[2][2];
+    max[0] = v[0];
+    max[1] = v[3];
+    max[2] = v[6];
+    mid[0] = v[1];
+    mid[1] = v[4];
+    mid[2] = v[7];
+    min[0] = v[2];
+    min[1] = v[5];
+    min[2] = v[8];
```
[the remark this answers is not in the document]
joannacirillo: Yes. `a` represents the covariance matrix on the three coordinate axis thus it is a 3x3 matrix. The matrix `v` is also 3x3 as it contains the 3 eigen vectors of `a`.

Sources/Common/Core/Math/index.js
joannacirillo: It has been reverted in further commits. This change does not appear on the final changes anymore.

Sources/Common/Transform/LandmarkTransform/index.js
```diff
@@ -13,23 +13,6 @@ function vtkLandmarkTransform(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkLandmarkTransform');
 
-  // Convert a mat4 matrix to a Matrix 2 dimensions
-  function mat4To2DArray(mat) {
```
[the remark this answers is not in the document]
finetjul: it seems that you are actually doing such conversion in some parts of your code, don't you ? 
If so, it means there is a need and it would reduce code duplciation.

Sources/Common/Core/Math/index.js
```diff
@@ -1202,22 +1209,22 @@ export function diagonalize3x3(a_3x3, w_3, v_3x3) {
 
 export function singularValueDecomposition3x3(a_3x3, u_3x3, w_3, vT_3x3) {
   let i;
-  const B = [createArray(3), createArray(3), createArray(3)];
+  const B = createArray(9);
```
finetjul, line 1212: should not be needed to zero initialize: `const B = []` is enough (or `const B = new Array(9)`)

Sources/Common/Core/Math/index.js
```diff
@@ -1202,22 +1209,22 @@ export function diagonalize3x3(a_3x3, w_3, v_3x3) {
...
 
   // copy so that A can be used for U or VT without risk
   for (i = 0; i < 3; i++) {
-    B[0][i] = a_3x3[0][i];
-    B[1][i] = a_3x3[1][i];
-    B[2][i] = a_3x3[2][i];
+    B[i] = a_3x3[i];
+    B[3 + i] = a_3x3[3 + i];
+    B[6 + i] = a_3x3[6 + i];
   }
```
finetjul, line 1219: actually, isn't the following good enough: ? `const B = [...a_3x3]`

you might want to do a pass on all your modified code to make sure you have not let such unneeded complexity around.

Sources/Common/Core/Math/index.js
```diff
@@ -1202,22 +1209,22 @@ export function diagonalize3x3(a_3x3, w_3, v_3x3) {
...
   // temporarily flip if determinant is negative
   const d = determinant3x3(B);
   if (d < 0) {
     for (i = 0; i < 3; i++) {
-      B[0][i] = -B[0][i];
-      B[1][i] = -B[1][i];
-      B[2][i] = -B[2][i];
+      B[i] = -B[i];
+      B[3 + i] = -B[3 + i];
+      B[6 + i] = -B[6 + i];
```
finetjul, line 1227: loop on 9 instead of 3

Sources/Common/Transform/LandmarkTransform/index.js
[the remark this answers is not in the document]
joannacirillo: Actually I think it is espacially the rowsToMat3 that is usefull. I created all of them so if in further refactorings they can be used.

Sources/Common/Core/Math/index.js
joannacirillo: Change was made at that place.

PR #2371, opened 2022-04-18
fix(WidgetManager): Fixed SVG positioning incorrectly in multiple vie…

yarous224:
The current SVG Widget does not consider the case of multiple renderers.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


floryst:
LGTM. FYI the [SVG layer will be deprecated](https://kitware.github.io/vtk-js/docs/concepts_widgets.html#SVG-Layer-Deprecated), but it'll remain in vtk.js until I write a migration guide for existing users of the SVG representations, as well as the existing SVG examples. We decided it's not worth the effort to maintain the layer, especially since application developers may choose to use their own layer for 2D overlays.

PR #2373, opened 2022-04-19
docs(ts): renderWindowInteractor events and FieldData

Julesdevops:
### Context
Some TS definitions.

### Changes

### PR and Code Checklist

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
@@ -1,4 +1,4 @@
-import { vtkObject } from "../../../interfaces";
+import { vtkObject, vtkSubscription } from "../../../interfaces";
 import vtkRenderer from "../Renderer";
 
 export enum Device {
@@ -34,6 +34,20 @@ export interface IRenderWindowInteractorInitialValues {
 
 interface IPosition {
 	type: string;
+}
+
+export type InteractorEventCallback = (e: IRenderWindowInteractorEvent) => void
+
+export type InteractorEventType = "StartInteractionEvent" | "InteractionEvent" | "EndInteractionEvent"
+
+export interface IRenderWindowInteractorEvent {
+	altKey: boolean;
+	controlKey: boolean;
+	firstRenderer: vtkRenderer;
+	pokedRenderer: vtkRenderer;
+	position: { x: number; y: number; z: number };
+	shiftKey: boolean;
+	type: InteractorEventType;
 }
 
 export interface vtkRenderWindowInteractor extends vtkObject {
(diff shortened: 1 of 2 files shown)


daker (changes requested):

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
```diff
@@ -113,398 +127,398 @@ export interface vtkRenderWindowInteractor extends vtkObject {
 	/**
 	 *
 	 */
-	invokeStartAnimation(callData: object): any;
+	invokeStartAnimation(callData: IRenderWindowInteractorEvent): any;
```
daker, line 130: The [invoke${_capitalize(eventName)}](https://github.com/Kitware/vtk-js/blob/master/Sources/macros.js#L941)  calls `invoke` which returns void i guess, can we have void on them instead of any?

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
```diff
@@ -113,398 +127,398 @@ export interface vtkRenderWindowInteractor extends vtkObject {
...
-	invokeEndInteractionEvent(callData: object): any;
+	invokeEndInteractionEvent(callData: IRenderWindowInteractorEvent): any;
 
 	/**
 	 *
-	 * @param args
+	 * @param cb The callback to be called
 	 */
-	onStartAnimation(args: any): any;
+	onStartAnimation(cb: InteractorEventCallback): Readonly<vtkSubscription>;
```
daker, line 311: The [on${_capitalize(eventName)}](https://github.com/Kitware/vtk-js/blob/master/Sources/macros.js#L943) does have two args `callback, priority = 0.0 => callback:InteractorEventCallback, priority?:number`, if i am not wrong the callback needs to have an [apply](https://github.com/Kitware/vtk-js/blob/master/Sources/macros.js#L944-L947) attribute otherwise the function will exit.

@jourdain can you give us your input on this?

daker (changes requested):

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
jourdain: I confirm, `invoke` return `void`.

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
jourdain: So the [apply](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) is only to check that it is a `function`. It will be called with `this` resolving to the publicAPI of the event source where `invokeXXX` was called.

jourdain: But technically, it could also be an object with as apply function ;-).

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
Julesdevops: Thanks Sébastien

daker (approved):
LGTM

PR #2375, opened 2022-04-19
feat(vtkscalarbaractor): adds generateTicks field to vtkScalarBarActor

NGimbal:
### Context
I'd like to enable customizing ticks in vtkScalarBarActor to do things like the following:
image

Discussed approach with @martinken on [Discourse thread here](https://discourse.vtk.org/t/set-vtkscalarbaractor-ticks/8311/5).

### Results
The approach I took was to create a new `generateTicks` field that is implemented similarly to `autoLayout`.

I ended up adding `ticks` and `tickstrings` setters / getters to the `vtkScalarBarActorHelper` `publicAPI` where those values had been private to the `vtkScalarBarActorHelper` model previously. The `generateTicks` function then has to set those values. This is very similar to the current implementation of `autoLayout`, although probably could be improved.

Here are results from my client app:
Screen Shot 2022-04-19 at 2 12 15 PM

Setting a new `generateTicks` function requires the user to update the `vtkScalarBarActor` annotations. This also seems to be true for other similar functions, `setAxisLabel` for example. This is how I'm doing that:

``` javascript
      const mapper = scalarBarActor.getMapper()
      if (mapper) {
        mapper.getLookupTable().resetAnnotations()
      }
```

### Changes
- [x] Adds new `generateTicks` field to `vtkScalarBarActor`
- [x] Added test for new generateTicks initial value

### PR and Code Checklist

### Testing
Adds new test testScalaBarSetGenerateTicks.js, tests against image testScalarBarSetGenerateTicks.png.

- [x] Tested environment:
  - **vtk.js**: latest master
  - **OS**: macOS Monterey 12.3.1
  - **Browser**:  Chrome Version 100.0.4896.127 (Official Build) (x86_64)

### Funding
This contribution is funded by [Pollination](https://pollination.cloud).


martinken:
Very nice MR! I suspect the failing test is due to font differences. We need to add another valid image (see testScalarBar.js for how to have multiple baselines) To get the image that failed you can go to actions at the top of this page and then click on your failed workflow and then download the tests artifact (zip file) and open that in a browser, click on the hide/show  browser tests and the results should show up. If you get stuck just holler of check this page https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts

 The only other minor thing I see is since we are making tickstrings public let's change the capitalization to be tickStrings and getTickStrings() (tickstrings was my mistake, but let's fix it now if that's OK)

martinken:
artifacts will look like 

image

NGimbal:
@martinken Thanks - I'll add the images from the failing tests as you describe (I was wondering where the multiple baseline images came from!) then convert back to a reviewable PR.

Sources/Rendering/Core/ScalarBarActor/index.d.ts
```diff
@@ -175,6 +180,11 @@ export interface vtkScalarBarActor extends vtkActor {
 	 */
 	setAutoLayout(autoLayout: any): boolean;
 
+  /**
+   *
```
finetjul, line 184: please add more comments about the expected function behavior

Sources/Rendering/Core/ScalarBarActor/index.js
```diff
@@ -145,6 +145,23 @@ function defaultAutoLayout(publicAPI, model) {
   };
 }
 
+// ----------------------------------------------------------------------------
+// Default generateTicks function
+// ----------------------------------------------------------------------------
+function defaultGenerateTicks(publicApi, model) {
```
finetjul, line 151: please add more comments about the default behavior

Sources/Rendering/Core/ScalarBarActor/index.d.ts
```diff
@@ -29,6 +29,7 @@ export interface IStyle {
 export interface IScalarBarActorInitialValues extends IActorInitialValues {
 	automated?: boolean,
 	autoLayout?: (publicAPI: object, model: object) => void,
+  generateTicks?: (helper: any) => void,
```
finetjul, line 32: move line next to other tick member variables ? -> line 42

Sources/Rendering/Core/ScalarBarActor/index.d.ts
NGimbal: How does this look to you?

``` javascript
/**
 * Sets the function used to generate legend ticks. 
 * 
 * This function takes a vtkScalarBarActorHelper and returns true on success. 
 * To have the desired effect, the function must call: `helper.setTicks(ticks: num[])` and `helper.setTickStrings(tickStrings: string[])`.
 * 
 * After setting the generateTicks function you must regenerate the vtkScalarBarActor for your changes to take effect. 
 * One way to do that is:
 * ```
 *  const mapper = scalarBarActor.getMapper()
 *  if (mapper) {
 *    mapper.getLookupTable().resetAnnotations()
 *  }
 * ```
 * @param generateTicks 
 */
   ```

Comment (finetjul) [reply Sources/Rendering/Core/ScalarBarActor/index.d.ts:173]: This is great, thanks.

You also need to give a type to the `generateTicks` parameter

Comment (NGimbal) [reply Sources/Rendering/Core/ScalarBarActor/index.d.ts:173]: Just added a type for this parameter. I'd like to add a type for the vtkScalarBarActorHelper as well but will have to do that at a later date.

File: Sources/Rendering/Core/ScalarBarActor/index.d.ts [hunk 4]
```diff
@@ -27,21 +27,10 @@ export interface IStyle {
...
-	boxPosition?: Vector2,
-	boxSize?: Size,
-	scalarToColors?: null,
-	axisTitlePixelOffset?: number,
-	axisTextStyle?: IStyle,
-	tickLabelPixelOffset?: number,
-	tickTextStyle?: IStyle,
-	drawNanAnnotation?: boolean,
-	drawBelowRangeSwatch?: boolean,
-	drawAboveRangeSwatch?: boolean,
```
finetjul, line 44: whydo you get rid of those values ?

Sources/Rendering/Core/ScalarBarActor/index.js
```diff
@@ -898,6 +922,10 @@ function vtkScalarBarActor(publicAPI, model) {
   publicAPI.resetAutoLayoutToDefault = () => {
     model.autoLayout = defaultAutoLayout(publicAPI, model);
   };
+
+  publicAPI.resetTicks = () => {
+    model.generateTicks = defaultGenerateTicks(publicAPI, model);
```
finetjul, line 927: you might want to call setGenerateTicks() instead of setting it manually (to make sure all the modifiedTime is updated)

Sources/Rendering/Core/ScalarBarActor/index.d.ts
NGimbal: Total mistake, not sure how that happened. Thanks for catching that.

finetjul (commented):
LGTM

boucaud:
@finetjul @NGimbal what is the status on this ? This will affect a project I'm working on now and I would like to start the integration.

NGimbal:
@boucaud - I think this is ready to merge pending @martinken's review.

PR #2376, opened 2022-04-20
Fix getContainer() return type

cjh1:
### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/OpenGL/RenderWindow/index.d.ts
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.d.ts
@@ -66,7 +66,7 @@ export interface vtkOpenGLRenderWindow extends vtkOpenGLRenderWindowBase {
 	/**
 	 * Get the container element.
 	 */
-	getContainer(): void;
+	getContainer(): Nullable<HTMLElement>;
 
 	/**
 	 * Get the container size.


Sources/Rendering/OpenGL/RenderWindow/index.d.ts
```diff
@@ -66,7 +66,7 @@ export interface vtkOpenGLRenderWindow extends vtkOpenGLRenderWindowBase {
 	/**
 	 * Get the container element.
 	 */
-	getContainer(): void;
+	getContainer(): HTMLElement;
```
floryst, line 69: The container can be null.

cjh1: Fixed

daker:
Can you change the commit message to : docs(ts): fix vtkOpenGLRenderWindow defs

Sources/Rendering/OpenGL/RenderWindow/index.d.ts
daker: You can use Nullable from [types.d.ts](https://github.com/Kitware/vtk-js/blob/master/Sources/types.d.ts#L4)
`Nullable<HTMLElement>`

cjh1: Great, thanks

daker (approved):
LGTM

PR #2379, opened 2022-04-20
fix(VolumeMapper): Fixed depth test of volume was incorrect in multi-…

yarous224:
### Context
I think the viewport is an area that scissor the entire frame buffer, and did not change the coordinates of the internal objects, so framebufferSize should be used in depth testing.

### Results
I placed a tube in front of a volume to test depth issues with multiple renderers.

#### Before
image
The graphics display is incomplete, and the volume is always on top of other actors.

#### After
image
Display and order are normal.

### Changes

- [ ] VolumeMapper

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


martinken:
Wow very nice tracking this down and the images look great. +1 I'll update the branch and merge if the checks are all still good.

martinken (approved):

PR #2381, opened 2022-04-20
fix(widget): fix widget picking when substates are invisible

### LINKED ISSUE #2374: ImageCroppingWidget can't interact properly
I found some problems with the examples on the official website(https://kitware.github.io/vtk-js/examples/ImageCroppingWidget.html)

I can't drag my selected point correctly in ImageCroppingWidget

https://user-images.githubusercontent.com/19411208/164004193-9a62d0e8-e85d-4344-bd79-82ded04db3a2.mp4

.

finetjul:
getSelectedState() is calling getRepresentationStates(). When hiding a handle, the compositeID of
the visible handles change because of the invisible handle not being rendered.
WidgetRepresentation.getRepresentationStates() returns all handle substates, visible or not.

fix #2374

### Context
Hiding widget handles (i.e. setVisible(false)) disrupts the handle selection. 
When a sphere handle gets invisible, it does not get rendered.
However, an invisible handle could still get selected. 

### Results
Invisible handles no longer disrupt visible handle picking.

### Changes
Synchronize handle state between rendering (requestData) and selecting (getSelectedState)

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 10
  - **Browser**: Chrome

### DIFF
--- a/Sources/Widgets/Representations/CircleContextRepresentation/index.js
+++ b/Sources/Widgets/Representations/CircleContextRepresentation/index.js
@@ -111,19 +111,16 @@ function vtkCircleContextRepresentation(publicAPI, model) {
     model.pipelines.circle.actor.getProperty().setOpacity(opacity);
   };
 
+  const superGetRepresentationStates = publicAPI.getRepresentationStates;
+  publicAPI.getRepresentationStates = (input = model.inputData[0]) =>
+    superGetRepresentationStates(input).filter(
+      (state) => state.getOrigin?.() && state.isVisible?.()
+    );
   // --------------------------------------------------------------------------
 
   publicAPI.requestData = (inData, outData) => {
     const { points, scale, color, direction } = model.internalArrays;
-    const list = publicAPI
-      .getRepresentationStates(inData[0])
-      .filter(
-        (state) =>
-          state.getOrigin &&
-          state.getOrigin() &&
-          state.isVisible &&
-          state.isVisible()
-      );
+    const list = publicAPI.getRepresentationStates(inData[0]);
     const totalCount = list.length;
 
     if (color.getNumberOfValues() !== totalCount) {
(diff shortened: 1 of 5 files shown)


finetjul:
@floryst I'm happy to move my change into `WidgetRepresentation.getRepresentationStates()` if you think it makes more sense.

floryst:
While `origin` and `visible` are fairly fundamental properties, I hesitate to move the logic to `WidgetRepresentation.getRepresentationStates()` because it's state-specific.

I think this was introduced [here](https://github.com/Kitware/vtk-js/commit/974d8cac88799d2f23f0377f4b47f5b4c321d56a#diff-522f5e38132d70308386c1f4e0fcba1c2227476dc40231101d3baff8f1efea57L170-R180). I presume setting the scale to 0 still allocated a compositeID for that state, but the filter no longer allows for that behavior.

Sources/Widgets/Representations/CircleContextRepresentation/index.js
```diff
@@ -111,19 +111,20 @@ function vtkCircleContextRepresentation(publicAPI, model) {
...
   };
 
+  const superGetRepresentationStates = publicAPI.getRepresentationStates;
+  publicAPI.getRepresentationStates = (input = model.inputData[0]) =>
+    superGetRepresentationStates(input).filter(
+      (state) =>
+        state.getOrigin &&
+        state.getOrigin() &&
+        state.isVisible &&
+        state.isVisible()
```
floryst, line 121: Simplify to `state.getOrigin?.() && state.isVisible?.()`.

finetjul:
> While `origin` and `visible` are fairly fundamental properties, I hesitate to move the logic to `WidgetRepresentation.getRepresentationStates()` because it's state-specific.

I agree with you, that's why I left the filtering in subclasses.

> 
> I think this was introduced [here](https://github.com/Kitware/vtk-js/commit/974d8cac88799d2f23f0377f4b47f5b4c321d56a#diff-522f5e38132d70308386c1f4e0fcba1c2227476dc40231101d3baff8f1efea57L170-R180). I presume setting the scale to 0 still allocated a compositeID for that state, but the filter no longer allows for that behavior.

Indeed, this is the reason of the regression

floryst (approved):
LGTM

PR #2382, opened 2022-04-21
Include direction on XML reader and writers.

### LINKED ISSUE #2380: Add direction to `XMLImageDataReader`
### Need
When overlaying images that have different origin, if the direction is not included, visualization will not be correct. 

### Feature

Include the direction on the XML reader (probably on the writer too). If reading an XML file generated from a VTK C++ writer then the following should be enough to read:
```js
const direction = imageDataElem
      .getAttribute('Direction')
      .split(' ')
      .map((t) => Number(t));
```

ramonemiliani93:
### Context
- fix #2380
- Adds the capability of reading and writing directions on XML files.

### Results
- New `Direction` field on XML file generated.

### Changes
- `Direction` must now be included in the XML file used on the reader.

### PR and Code Checklist

### Testing
- [X] This change adds or fixes unit tests.


Sources/IO/XML/XMLImageDataWriter/index.js
```diff
@@ -23,6 +23,7 @@ function vtkXMLImageDataWriter(publicAPI, model) {
       WholeExtent: dataObject.getExtent().join(' '),
       Origin: dataObject.getOrigin().join(' '),
       Spacing: dataObject.getSpacing().join(' '),
+      Direction: dataObject.getDirection().join(' '),
```
finetjul, line 26: it seems that it will crash if no direction is set on the vtkImageData.

Sources/IO/XML/XMLImageDataWriter/test/testXMLImageDataWriter.js
```diff
@@ -0,0 +1,23 @@
...
+import vtkImageData from '../../../../Common/DataModel/ImageData';
+
+test('Test XML writer file content', (t) => {
+  const expectedSubstring =
+    '<ImageData WholeExtent="0 -1 0 -1 0 -1" Origin="10 20 30" Spacing="1 2 3" Direction="-1 0 0 0 1 0 0 0 -1">';
+
+  const imageData = vtkImageData.newInstance();
+  imageData.setOrigin([10, 20, 30]);
+  imageData.setSpacing([1, 2, 3]);
+  imageData.setDirection([-1, 0, 0, 0, 1, 0, 0, 0, -1]);
```
finetjul, line 12: add a test step when direction is not set

Sources/IO/XML/XMLImageDataReader/test/testXMLImageDataReader.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const enc = new TextEncoder();
+  const arrayBuffer = enc.encode(fileContent);
+
+  const reader = vtkXMLImageDataReader.newInstance();
+
+  reader.parseAsArrayBuffer(arrayBuffer);
+  const imageData = reader.getOutputData();
+
+  t.ok(
```
finetjul, line 34: `t.equal()` did not work for you ? 

Alternativelly, you might want to consider `vtkMath.areEquals()` instead of doing string comparison

finetjul:
Looks fine to me besides the minor comments.

Do you confirm that VTK C++ always writes "Direction" in vti files ?

ramonemiliani93:
@finetjul well I am using the Python binding and the following code:
```python
import pathlib
import vtkmodules.all as vtk

image_data = vtk.vtkImageData()

writer = vtk.vtkXMLImageDataWriter()
writer.WriteToOutputStringOn()
writer.SetInputData(image_data)
writer.Update()
s = writer.GetOutputString()

with (pathlib.Path() / "test.xml").open("w") as file:
    file.write(s)
```
Generates the following file:
```xml
<?xml version="1.0"?>
<VTKFile type="ImageData" version="0.1" byte_order="LittleEndian" header_type="UInt32" compressor="vtkZLibDataCompressor">
  <ImageData WholeExtent="0 -1 0 -1 0 -1" Origin="0 0 0" Spacing="1 1 1" Direction="1 0 0 0 1 0 0 0 1">
    <Piece Extent="0 -1 0 -1 0 -1"                                                    >
      <PointData>
      </PointData>
      <CellData>
      </CellData>
    </Piece>
  </ImageData>
  <AppendedData encoding="base64">
   _
  </AppendedData>
</VTKFile>
```
I am assuming that the `Direction` field is always included.

Sources/IO/XML/XMLImageDataReader/test/testXMLImageDataReader.js
ramonemiliani93: I'll try it out thanks!

ramonemiliani93:
Thanks @jourdain 🙌 I'll add the case for when `Direction` is not present.

ramonemiliani93:
@finetjul @jourdain I added the optional chaining to the direction that gives an undefined when `Direction` is not present, which then gets evaluated to false on the ImageData constructor and sets the direction to the default value. If there is a better way let me know.

jourdain:
LGTM @floryst feel free to merge when you want to bundle it with something else...

Sources/IO/XML/XMLImageDataWriter/index.js
finetjul: @ramonemiliani93 can you please address this issue ? it will crash if no Direction is set on the image data.

ramonemiliani93: Sure @finetjul let me fix it 👌

jourdain: in vtk.js vtkImageData will always add a direction so it should not be a problem at write time. The issue was at read time but that has been addressed.

finetjul: Oh, good point, I did not see what was going on in `vtkImageData::extent()`. My bad

PR #2383, opened 2022-04-21
fix(cellpicker): return the correct cell id

Julesdevops:
CellPicker used to return a wrong cell id if having lines and verts in the dataset. This change
makes it so we return a consistent cell id.

### Context
bad cell id value in CellPicker

### Results
We now return a consistent cell id if we have vertices and lines in the dataset.
### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: master
  - **OS**: Archlinux
  - **Browser**: Firefox


Julesdevops:
@finetjul

finetjul:
Made obsolete by #2394

PR #2388, opened 2022-04-25
fix(Manipulators): use renderer size

floryst:
### Context

Some of the interactions are "slower" for renderers that do not take up the full canvas size, since some of the interaction logic incorrectly depends on the renderwindow size rather than the renderer/viewport size.

This is relevant for when the interaction surface is limited to a particular renderer that may be smaller than the entire render window.

### Results

- certain manipulators now use viewport size rather than the entire canvas size.

### Changes

- [x] Changed certain usages of `.getSize()` to `.getViewportSize(renderer)`.

### PR and Code Checklist

### Testing

### DIFF
--- a/Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
@@ -30,7 +30,8 @@ function vtkMouseCameraUnicamManipulator(publicAPI, model) {
 
   //----------------------------------------------------------------------------
   const normalize = (position, interactor) => {
-    const [width, height] = interactor.getView().getSize();
+    const renderer = interactor.findPokedRenderer();
+    const [width, height] = interactor.getView().getViewportSize(renderer);
 
     const nx = -1.0 + (2.0 * position.x) / width;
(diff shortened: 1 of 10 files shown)


Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
```diff
@@ -62,7 +63,8 @@ function vtkMouseCameraUnicamManipulator(publicAPI, model) {
     vtkMath.normalize(atV);
     const l = vtkMath.dot(cameraToPointVec, atV);
     const viewAngle = vtkMath.radiansFromDegrees(camera.getViewAngle());
-    const [width, height] = interactor.getView().getSize();
+    const renderer = interactor.findPokedRenderer();
+    const [width, height] = interactor.getView().getSize(renderer);
```
jourdain, line 67: `getViewportSize`?

jourdain:
LGTM minus my question. ;-)

Sources/Interaction/Manipulators/MouseCameraUnicamManipulator/index.js
floryst: Yep, thanks for catching that!

PR #2389, opened 2022-04-26
feat(WebGPU): add support for index buffers

martinken:
Add support for index buffers to the webgpu backend and use them in the
CellArrayMapper by default.

This approach uses a flat interpolation approach to support celldata. In
that each primitive must start with a provoking vertex that is only used
for that cell (not primitive). The flat ibo building code handles this
and sets up maps from flatIds to source point and cell IDs. These
mappings can be used to build flat array buffers for point or cell data
to go along with the index buffer.

Note that this flat packing is typically larger than the original point
array as we have to be careful not to use the same point to provoke
multiple cells. In a fully connected triangle mesh each point is shared
by 6 triangles. The old code would explode the point array by 6 as a
brute force solution. This flat packing would theoretically achieve a 2x
size, significantly smaller that the old 6x.

Also includes a fix to getStatistics for RenderWindow

Includes a performance/memory improvement to GetNumberOfCells

### DIFF
--- a/Sources/Common/Core/CellArray/index.js
+++ b/Sources/Common/Core/CellArray/index.js
@@ -18,7 +18,12 @@ function extractCellSizes(cellArray) {
 }
 
 function getNumberOfCells(cellArray) {
-  return extractCellSizes(cellArray).length;
+  let cellId = 0;
+  for (let cellArrayIndex = 0; cellArrayIndex < cellArray.length; ) {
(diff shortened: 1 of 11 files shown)


martinken:
@jourdain can you review the CellArray and Mapper changes ?  Feel free to review the WebGPU stuff too but the CellArray and Mapper changes are more central.

jourdain:
The `Core/CellArray` and `Core/Mapper` fixes looks good.

PR #2394, opened 2022-05-03
Add edge picking support to the Cellpicker

Julesdevops:
### Context
Currently, the CellPicker can only pick triangles.

### Results
CellPicker will also try to pick edges.

### Changes
- Added a PolyLine data model class.
- Implemented evaluateLocation in vtkLine
- Changes the picking workflow ; now using `vtkPolyData.getCell` and `vtkPolyData.getCellType`

### PR and Code Checklist

### Testing
- [ ] Tested environment:
  - **vtk.js**: v19.4.0
  - **OS**: ArchLinux
  - **Browser**: Firefox 99.0.1

### DIFF
--- a/Sources/Common/DataModel/PolyLine/index.d.ts
+++ b/Sources/Common/DataModel/PolyLine/index.d.ts
@@ -1,0 +1,63 @@
+import { Vector2, Vector3 } from '../../../types';
+import vtkCell, { ICellInitialValues } from '../Cell';
+import { IIntersectWithLine } from '../Line';
+
+export interface IPolyLineInitialValues extends ICellInitialValues { }
+
+export interface vtkPolyLine extends vtkCell {
+
+	/**
+	 * Get the topological dimensional of the cell (0, 1, 2 or 3).
+	 */
+	getCellDimension(): number;
+
+	/**
+	 * @param {number} t1
+	 * @param {number} t2
+	 * @param {Vector3} p1 The first point coordinate.
+	 * @param {Vector3} p2 The second point coordinate.
+	 * @param {Number} tol The tolerance to use.
+	 * @param {Vector3} x The point which intersect the line.
+	 * @param {Vector3} pcoords The parametric coordinates.
+	 */
+	intersectWithLine(t1: number, t2: number, p1: Vector3, p2: Vector3, tol: number, x: Vector3, pcoords: Vector3): IIntersectWithLine;
+
+  /**
+   * Determine global coordinate (x[3]) from subId and parametric coordinates.
+   * Also returns interpolation weights. (The number of weights is equal to
+   * the number of points in the cell.)
+   * 
+   * @param {number} subId
+   * @param {Vector3} pcoords The parametric coordinates
+   * @param {Vector3} x The global coordinate
+   * @param {Vector2} weights The interpolation weights
+   */
+  evaluateLocation(subId: number, pcoords: Vector3, x: Vector3, weights: Vector2): void
+}
+
+/**
+ * Method used to decorate a given object (publicAPI+model) with vtkPolyLine characteristics.
+ *
+ * @param publicAPI object on which methods will be bounds (public)
+ * @param model object on which data structure will be bounds (protected)
+ * @param {IPolyLineInitialValues} [initialValues] (default: {})
+ */
+export function extend(publicAPI: object, model: object, initialValues?: IPolyLineInitialValues): void;
+
+/**
+ * Method used to create a new instance of vtkPolyLine.
+ * @param {IPolyLineInitialValues} [initialValues] for pre-setting some of its content
+ */
+export function newInstance(initialValues?: IPolyLineInitialValues): vtkPolyLine;
+
+/** 
+ * vtkPolyLine is a cell which representant a poly line.
+ * 
+ * @see vtkCell
+ */
+export declare const vtkPolyLine: {
+	newInstance: typeof newInstance,
+	extend: typeof extend;
+};
+
+export default vtkPolyLine;
--- a/Sources/Common/DataModel/PolyLine/index.js
+++ b/Sources/Common/DataModel/PolyLine/index.js
@@ -1,0 +1,102 @@
+import macro from 'vtk.js/Sources/macros';
+import vtkCell from 'vtk.js/Sources/Common/DataModel/Cell';
+import vtkLine from 'vtk.js/Sources/Common/DataModel/Line';
+
+function vtkPolyLine(publicAPI, model) {
+  model.classHierarchy.push('vtkPolyLine');
+  const superClass = { ...publicAPI };
+
+  const line = vtkLine.newInstance();
+  line.getPoints().setNumberOfPoints(2);
+
+  publicAPI.getCellDimension = () => 1;
+  publicAPI.intersectWithLine = (t1, t2, p1, p2, tol, x, pcoords) => {
+    const outObj = {
+      intersect: 0,
+      t: Number.MAX_VALUE,
+      subId: 0,
+      betweenPoints: null,
+    };
+
+    const numLines = superClass.getPoints().getNumberOfPoints() - 1;
+    let pDistMin = Number.MAX_VALUE;
+    const minXYZ = [0, 0, 0];
+    const minPCoords = [0, 0, 0];
+    for (let subId = 0; subId < numLines; subId++) {
+      const pCoords = [0, 0, 0];
+
+      line
+        .getPoints()
+        .getData()
+        .set(model.points.getData().subarray(3 * subId, 3 * (subId + 2)));
+
+      const lineIntersected = line.intersectWithLine(p1, p2, tol, x, pcoords);
+
+      if (
+        lineIntersected.intersect === 1 &&
+        lineIntersected.t <= outObj.t + tol &&
+        lineIntersected.t >= t1 &&
+        lineIntersected.t <= t2
+      ) {
+        outObj.intersect = 1;
+        const pDist = line.getParametricDistance(pCoords);
+        if (
+          pDist < pDistMin ||
+          (pDist === pDistMin && lineIntersected.t < outObj.t)
+        ) {
+          outObj.subId = subId;
+          outObj.t = lineIntersected.t;
+          pDistMin = pDist;
+          for (let k = 0; k < 3; k++) {
+            minXYZ[k] = x[k];
+            minPCoords[k] = pCoords[k];
+          }
+        }
+      }
+    }
+
+    return outObj;
+  };
+
+  publicAPI.evaluatePosition = (
+    x,
+    closestPoint,
+    subId,
+    pcoords,
+    dist2,
+    weights
+  ) => {}; // virtual
+
+  publicAPI.evaluateLocation = (subId, pcoords, x, weights) => {
+    line
+      .getPoints()
+      .getData()
+      .set(model.points.getData().subarray(3 * subId, 3 * (subId + 2)));
+
+    return line.evaluateLocation(pcoords, x, weights);
+  };
+}
+
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {};
+
+// ----------------------------------------------------------------------------
+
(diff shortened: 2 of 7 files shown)


Julesdevops:
@finetjul

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -0,0 +1,120 @@
+import macro from 'vtk.js/Sources/macros';
+import vtkCell from 'vtk.js/Sources/Common/DataModel/Cell';
+import vtkLine from 'vtk.js/Sources/Common/DataModel/Line';
+
+function vtkPolyLine(publicAPI, model) {
+  model.classHierarchy.push('vtkPolyLine');
+  const superClass = { ...publicAPI };
+
+  model.line = vtkLine.newInstance();
```
finetjul, line 9: no need to store in model, can be a private variable (`const line = vtkLine.newInstance()`).

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -0,0 +1,120 @@
...
+    const numLines = superClass.getPoints().getNumberOfPoints() - 1;
+    let pDistMin = Number.MAX_VALUE;
+    const minXYZ = [0, 0, 0];
+    const minPCoords = [0, 0, 0];
+
+    for (let subId = 0; subId < numLines; subId++) {
+      const pCoords = [0, 0, 0];
+
+      model.line.getPoints().setPoint(0, ...model.points.getPoint(subId));
+      model.line.getPoints().setPoint(1, ...model.points.getPoint(subId + 1));
```
finetjul, line 30: The following should be faster:
`model.line.getPoints().getData().set(model.points.getData().subarray(3 * subId, 3 * (subId + 2))`

Sources/Common/DataModel/PolyLine/index.d.ts
```diff
@@ -0,0 +1,55 @@
+import { Vector2, Vector3 } from '../../../types';
+import vtkCell from '../Cell';
+import { IIntersectWithLine } from '../Line';
+
+interface IPolyLineInitialValues { }
```
daker, line 5: can i have an export here and extends from `ICellInitialValues` ?

Sources/Common/DataModel/PolyLine/index.d.ts
```diff
@@ -0,0 +1,55 @@
...
+   * @param {number} t1
+   * @param {number} t2
+	 * @param {Vector3} p1 The first point coordinate.
+	 * @param {Vector3} p2 The second point coordinate.
+	 * @param {Number} tol The tolerance to use.
+	 * @param {Vector3} x The point which intersect the line.
+	 * @param {Vector3} pcoords The parametric coordinates.
+	 */
+	intersectWithLine(t1: number, t2: number, p1: Vector3, p2: Vector3, tol: number, x: Vector3, pcoords: Vector3): IIntersectWithLine;
+
```
daker, line 24: Can you reformat the file to indent it correctly?

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -0,0 +1,120 @@
...
+
+    weights[0] = 1.0 - pcoords[0];
+    weights[1] = pcoords[0];
+  };
+
+  publicAPI.getParametricDistance = (subId, pCoords) => {
+    model.line.getPoints().setPoint(subId, model.points.getPoint(subId));
+    model.line
+      .getPoints()
+      .setPoint(subId + 1, model.points.getPoint(subId + 1));
```
finetjul, line 93: same perf improvement as above

Sources/Common/DataModel/PolyLine/index.d.ts
```diff
@@ -0,0 +1,55 @@
...
+  evaluateLocation(subId: number, pcoords: Vector3, x: Vector3, weights: Vector2): void
```
daker, line 25: Can you add documentation for `evaluateLocation` & `getParametricDistance`

daker (changes requested):

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -324,7 +364,12 @@ function vtkCellPicker(publicAPI, model) {
         weights[i] = 0.0;
       }
       const point = [];
-      minCell.evaluateLocation(minPCoords, point, weights);
+
+      if (vtkCellTypes.hasSubCells(cellType)) {
```
finetjul, line 368: Instead of cellType, shouldn't it be `minCellType` ? or `minCell.getType()` ? or `getCellType(minCell)`

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -264,30 +271,56 @@ function vtkCellPicker(publicAPI, model) {
...
-      // cross all cells
-      for (let i = 0; i < cellData.length; cellId++) {
+      if (!data.getCells()) {
+        data.buildLinks();
+      }
+
+      const triangle = vtkTriangle.newInstance();
+      const line = vtkLine.newInstance();
+      const polyLine = vtkPolyLine.newInstance();
+      const numberOfCells = data.getNumberOfCells();
```
finetjul, line 281: how about creating an object ?
```
const tempCell  = {
  CellType.VTK_LINE: vtkLine.newInstance(),
  CellType.VTK_POLY_LINE: ... };
...
  const cell = tempCell[ cellType];
```

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -239,7 +241,12 @@ function vtkCellPicker(publicAPI, model) {
     let pDistMin = Number.MAX_VALUE;
     const minPCoords = [0, 0, 0];
     let minCellId = -1;
```
finetjul, line 243: let's use `let minCellId // = undefined;` to be consistent with `subId`

finetjul (changes requested):

Julesdevops:
@daker @finetjul Thanks for the feedback, I addressed your comments.
I got rid of `getParametricDistance` override in `vtkPolyLine`, because it seems that it is not needed at all.

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -1,15 +1,25 @@
...
 // ----------------------------------------------------------------------------
 // Global methods
 // ----------------------------------------------------------------------------
 
+function getCellTempMap() {
+  return {
+    [CellType.VTK_LINE]: vtkLine.newInstance(),
+    [CellType.VTK_POLY_LINE]: vtkPolyLine.newInstance(),
+    [CellType.VTK_TRIANGLE]: vtkTriangle.newInstance(),
+  };
```
finetjul, line 20: `return new Map(CellType.keys().map((cellType) => return [cellType, CellType.getClassNameFromTypeId(cellType)])`

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  publicAPI.evaluatePosition = (
+    x,
+    closestPoint,
+    subId,
+    pcoords,
+    dist2,
+    weights
+  ) => {}; // virtual
```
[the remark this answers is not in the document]
Julesdevops: I don't know. I've done like here: https://github.com/Kitware/vtk-js/blob/master/Sources/Common/DataModel/Line/index.js#L220
If it make sense to add an error message, then I'm adding it to both locations

Sources/Rendering/Core/CellPicker/index.js
Julesdevops: imo this is elegant, but we can't use it.
We don't support all cells, and we would need to dynamically import the relevant module from the class string, and I don't know how to do this properly without creating problems.
If someday we supports all the cells listed in `vtk-js/Sources/Common/DataModel/CellTypes/Constants.js` and also have a map directly with the corresponding modules we shall use it.

Sources/Common/DataModel/PolyLine/index.js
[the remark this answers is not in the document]
finetjul: Actually, both Line and PolyLine should not even overwrite that method. Its implementation already exist in vtkCell (which should have the vtkErrorMacro)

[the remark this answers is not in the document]
Julesdevops: Make sense

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -0,0 +1,98 @@
+import macro from 'vtk.js/Sources/macros';
+import vtkCell from 'vtk.js/Sources/Common/DataModel/Cell';
+import vtkLine from 'vtk.js/Sources/Common/DataModel/Line';
+
+function vtkPolyLine(publicAPI, model) {
+  model.classHierarchy.push('vtkPolyLine');
+  const superClass = { ...publicAPI };
+
+  const line = vtkLine.newInstance();
+  line.getPoints().setNumberOfPoints(2);
```
finetjul, line 10: shouldn't this be done automatically in `vtkLine::extend()`  ?

Sources/Rendering/Core/CellPicker/index.js
finetjul: fair enough

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -264,30 +276,46 @@ function vtkCellPicker(publicAPI, model) {
     if (locator) {
       // TODO when cell locator will be implemented
     } else if (data.getPolys) {
```
finetjul, line 278: this test is no longer needed

Julesdevops: Which test ? `data.getPolys` or `locator`?

finetjul: data.getPolys

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -0,0 +1,98 @@
...
+    const a2 = [];
+    model.points.getPoint(subId, a1);
+    model.points.getPoint(subId + 1, a2);
+
+    for (let i = 0; i < 3; i++) {
+      x[i] = a1[i] + pcoords[0] * (a2[i] - a1[i]);
+    }
+
+    weights[0] = 1.0 - pcoords[0];
+    weights[1] = pcoords[0];
```
finetjul, line 72: why not do:
```
line
        .getPoints()
        .getData()
        .set(model.points.getData().subarray(3 * subId, 3 * (subId + 2)));

      return line.evaluateLocation(...);
```

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -263,31 +275,47 @@ function vtkCellPicker(publicAPI, model) {
...
-        const nbPointsInCell = cellData[i++];
 
-        cellPoints.setNumberOfPoints(nbPointsInCell);
-        // Extract cell points
-        for (let j = 0; j < nbPointsInCell; j++) {
-          pointsIdList[j] = cellData[i++];
+        minCellType = data.getCellType(cellId);
+        const cell = tempCellMap[minCellType];
+
+        if (cell === undefined) {
```
finetjul, line 294: `if (cell == null)` is more flexible

Julesdevops: Sure but it will never be null, or am I missing something?
If it's null it's because one of the `newInstance`calls returned null so in this case I think it's better to have an error than silently fail.

Sources/Common/DataModel/PolyLine/index.js
Julesdevops: yes, make sense. Thanks

Sources/Rendering/Core/CellPicker/index.js
finetjul: To me it is not this function responsibility to trigger such error.
However, this function needs to support such scenario and not crash.

Julesdevops:
@finetjul all done

finetjul (approved):

daker (approved):
The TS definitions looks good

PR #2396, opened 2022-05-04
fix(WidgetRepresentation): Use device pixel ratio

laurennlam:
### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Widgets/Representations/WidgetRepresentation/index.js
+++ b/Sources/Widgets/Representations/WidgetRepresentation/index.js
@@ -221,7 +221,7 @@ function vtkWidgetRepresentation(publicAPI, model) {
     }
 
     const rHeight = rendererPixelDims[1];
-    return scale / rHeight;
+    return (window.devicePixelRatio * scale) / rHeight;
   };
 
   // Make sure setting the labels at build time works with string/array...


laurennlam:
@finetjul can you test it please ?

Sources/Widgets/Representations/WidgetRepresentation/index.js
```diff
@@ -221,7 +221,7 @@ function vtkWidgetRepresentation(publicAPI, model) {
     }
 
     const rHeight = rendererPixelDims[1];
-    return scale / rHeight;
+    return (window.devicePixelRatio * scale) / rHeight;
```
finetjul, line 224: are parenthesis needed ?

finetjul:
This does not seem right, it would be a regression from #2030 

I'll have a deeper look

Sources/Widgets/Representations/WidgetRepresentation/index.js
jourdain: Also that assume the vtkRenderWindow use `window.devicePixelRatio` to scale up its resolution.

PR #2399, opened 2022-05-05
docs(ts): fix typescript defs

daker:
### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Filters/General/ImageCropFilter/index.d.ts
+++ b/Sources/Filters/General/ImageCropFilter/index.d.ts
@@ -1,10 +1,12 @@
 import { vtkAlgorithm, vtkObject } from "../../../interfaces";
 
+type CroppingPlanes = number[];
+
 /**
  *
  */
-interface IImageCropFilterInitialValues {
-	croppingPlanes?: any;
+export interface IImageCropFilterInitialValues {
+	croppingPlanes?: CroppingPlanes;
 }
 
 type vtkImageCropFilterBase = vtkObject & vtkAlgorithm;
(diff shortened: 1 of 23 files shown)


Sources/Filters/General/ImageCropFilter/index.d.ts
```diff
@@ -1,10 +1,11 @@
 import { vtkAlgorithm, vtkObject } from "../../../interfaces";
+import { Extent } from "../../../types";
 
 /**
  *
  */
-interface IImageCropFilterInitialValues {
-	croppingPlanes?: any;
+export interface IImageCropFilterInitialValues {
+	croppingPlanes?: Extent;
```
jourdain, line 8: is it really `Extent`? I haven't checked but that seems strange...

daker: No but it's an array of 6 elements, i wanted to make a local type CroppingPlanes, should i do it?

jourdain: I think it use to be 6 vtkPlane but now it can be as many as we want...

daker: from the code it's 6 number not vtkPlane

jourdain: You right, @floryst mentioned to me that the ImageCropFilter was indeed using some ijk bounds. Therefore using extent make sense. It might be off by 1 in the true meaning of extent but that is fine for the type definition...

PR #2405, opened 2022-05-08
feat(objwriter): add support for vtkOBJWriter

### LINKED ISSUE #2404: Feature Request: OBJWriter
### Need

An OBJWriter to allow exporting meshes to OBJ format

daker:
fix #2404

### Context

Add support for vtkOBJWriter

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


Sources/IO/Misc/OBJWriter/index.js
```diff
@@ -0,0 +1,179 @@
...
+// ----------------------------------------------------------------------------
+
+function vtkOBJWriter(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkOBJWriter');
+
+  publicAPI.requestData = (inData, outData) => {
+    const input = inData[0];
+
+    if (!input || input.getClassName() !== 'vtkPolyData') {
```
jourdain, line 143: replace `input.getClassName() !== 'vtkPolyData'` by `!input.isA('vtkPolyData')` to support inheritance too.

jourdain (approved):
LGTM

Sources/IO/Misc/OBJWriter/index.js
```diff
@@ -0,0 +1,179 @@
...
+  if (tcoords) {
+    for (let i = 0; i < nbPts; i++) {
+      p = tcoords.getTuple(i, p);
+      outputData += `vt ${p[0]} ${p[1]}\n`;
+    }
+  }
+  return outputData;
+};
+
+const writeOBJ = (polyData) => {
```
finetjul, line 86: why not making it public ?

Sources/IO/Misc/OBJWriter/index.d.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * 
+ * @param {vktPolyData} polyData 
+ */
+export function writeOBJ(polyData: vtkPolyData): vtkPolyData;
+
+/**
+ * vtkOBJWriter writes wavefront obj (.obj) files in ASCII form. OBJ files
+ * contain the geometry including lines, triangles and polygons. Normals and
+ * texture coordinates on points are also written if they exist.
```
finetjul, line 45: you might want to mention that materials are not written out.

Sources/IO/Misc/OBJWriter/index.js
daker: i followed the same logic as in others Writers, it's exported in the default

finetjul: my bad :-)

Sources/IO/Misc/OBJWriter/index.d.ts
daker: ok i will re-add the mtl code again

daker: i am trying to finish this PR, by calling the getOutputData(1) i am getting an error

```js
Uncaught TypeError: model.output[count].isDeleted is not a function
    at publicAPI.shouldUpdate (macros.js:853:55)
    at Object.getOutputData (macros.js:839:19)
```

in both outputs i am returning strings

```js
const objContent = writer.getOutputData(0);
const mtlContent = writer.getOutputData(1);
```

Displaying the content of `model.output[count]`, i can see it's returned string hence the error, so i am not sure if it's a bug or not.

image

finetjul: Did you check vtkMoleculeToRepresentation that produces 2 outputs ?

daker: Yes but it does produce two polydatas that's why it works, here i am producing two strings to be returned

finetjul: I see, I think I got confused at some point. A writer is not really meant to have any output.

`getOutputData()` is supposed to create a `vtkDataSet`, in your case `getOutput()` should be enough.

Sources/IO/Misc/OBJWriter/example/index.js
```diff
@@ -0,0 +1,69 @@
...
+    a.style.left = '50%';
+    a.style.bottom = '10px';
+    document.body.appendChild(a);
+    a.style.background = 'white';
+    a.style.padding = '5px';
+  });
+
+global.writer = writer;
+global.writerReader = writerReader;
+global.fullScreenRenderer = fullScreenRenderer;
```
finetjul, line 69: can you demonstrate how to write the obj, mtl and texture ?

Sources/IO/Misc/OBJWriter/index.d.ts
daker: i don't think `getOutput()` exist in vtk.js

Sources/IO/Misc/OBJWriter/example/index.js
daker: I added `exportAsZip()` to get the obj+mtl+texture

jourdain:
I'm ok with the code, knowing that there is some limitation in the current implementation as only one texture can exist.

jourdain (approved):

finetjul (commented):
Please add the following screenshot in the example gallery
VTK js OBJWriter Example Screenshot

finetjul (approved):
LGTM

PR #2406, opened 2022-05-09
refactor(mtlreader): add a boolean to not automatically request and l…

luciemac:
…oad materials files

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/IO/Misc/MTLReader/index.js
+++ b/Sources/IO/Misc/MTLReader/index.js
@@ -43,7 +43,7 @@ function vtkMTLReader(publicAPI, model) {
         model.materials[model.currentMaterial] = {};
       }
       model.materials[model.currentMaterial][tokens[0]] = tokens.slice(1);
-      if (tokens[0] === 'map_Kd') {
+      if (model.autoLoadMaterials && tokens[0] === 'map_Kd') {
         const image = new Image();
         image.onload = () => setTimeout(imageReady, 0);
         image.src = [model.baseURL, tokens[1]].join('/');
@@ -165,6 +165,7 @@ const DEFAULT_VALUES = {
   requestCount: 0,
   materials: {},
   interpolateTextures: true,
+  autoLoadMaterials: true,
   // baseURL: null,
   // dataAccessHelper: null,
   // url: null,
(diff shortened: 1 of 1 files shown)


luciemac:
@finetjul

Sources/IO/Misc/MTLReader/index.js
```diff
@@ -165,6 +165,7 @@ const DEFAULT_VALUES = {
   requestCount: 0,
   materials: {},
   interpolateTextures: true,
+  autoLoadMaterials: true,
```
finetjul, line 168: rename to `loadMaterialImages` ?

Sources/IO/Misc/MTLReader/index.js
```diff
@@ -43,7 +43,7 @@ function vtkMTLReader(publicAPI, model) {
         model.materials[model.currentMaterial] = {};
       }
       model.materials[model.currentMaterial][tokens[0]] = tokens.slice(1);
-      if (tokens[0] === 'map_Kd') {
+      if (model.autoLoadMaterials && tokens[0] === 'map_Kd') {
```
finetjul, line 46: `setImageSrc()` seems to be implemented to overwrite the image. It's similar to what you need (you already have the image in cache and you do not want to load it again).

Maybe we could combine both usage:
e.g.
```
image.src = model.imageSourceHelper(tokens[1]);
...
model.imageSourceHelper = (source) => [model.baseURL, source].join('/');
macro.setGet(['imageSourceBuilder']);
...
mtlReader.setImageSourceBuilder((source) => {return ''});
```

@jourdain what is the purpose of `setImageSrc()` ? What's the use case ?

luciemac: I don't necessarily have it in cache.
Sometimes "map_Kd" is mentioned, but we don't want to use it because the image source file does not exist.

jourdain: the mtl file will have a name/path and you want to link it to an HTML image so you call 
`setImageSrc(mtl_name, image_url)`

You can see it being used [here](https://github.com/Kitware/vtk-js/blob/master/Examples/Applications/OBJViewer/index.js#L76)

finetjul: thanks and all my apologies for not finding it myself, I blame github online search (couldn't find it), and my VSCode search that is configured to search only the "Sources" folder. My bad

jourdain: No worries and sorry for not providing that info sooner, but still pretty busy on trame v2.

PR #2408, opened 2022-05-10
test(scaleinpixels): restore testScaleInPixelsWithParallel

finetjul:
The baseline must be updated to fix the failing test.

### Context
Test has been skipped because it was failing.
It has always been failing due to an incorrect baseline.
When contributed, the test was not failing due to the incorrect test image comparison method in place in vtk.js
When the test image comparison method has been fixed, the test started to fail, hence it has been skipped.

### Results
testScaleInPixelsWithParallel is now passing and no more skipped.

### Changes
The baseline has been updated so that the test is now passing.
The test is no longer skipped.

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 10
  - **Browser**: Chrome


floryst:
LGTM!

PR #2409, opened 2022-05-10
Refactor constructors

joannacirillo:
### Context

When creating a new object, no setters are currently called. However, if they have a side effect (which is the case if they have been created through the use of `macro.set(...)`), we have to call them so we can see those side effects. Therefore, setters need to be called when creating a new instance if they exist.
In particular, with `A` and `B` two classes, `B` being a child class of `A` and the function `setProp1` existing in both classes, only the setter from child class `B` must be called when creating a new object `B` .

### Results

The setters are called if they exist when instanciating a new object. Changes might be needed to be done in classes that customly handle properties in constructor (`extend` function) like it is the case of the `DataArray` class.

### Changes

In the function `newInstance`, `publicAPI.set(values)` is called. The setters are thus called when creating a new object but not if extending it. Classes need to be created this way:

#### Parent class constructor example:
function defaultValues(initialValues) {
  return {
    prop1: 1,
    prop3: 70,
    prop5: 0,
    ...initialValues,
  };
}

  // Build VTK API
  macro.obj(publicAPI, model);
  macro.get(publicAPI, model, ['prop1']);

  vtkClassA(publicAPI, model);
}

#### Child class constructor example:
`function defaultValues(initialValues) {
  return {
    prop1: 111,
    prop2: 2,
    prop4: 9999,
    prop5: 2000,
    ...initialValues,
  };
}

 // values passed to parent are default of the child and initialValues
  vtkClassA.extend(publicAPI, model, defaultValues(initialValues));
  vtkClassB(publicAPI, model);
}`

***

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


joannacirillo:
@jourdain @finetjul @martinken This pull request is a proposal to solve the problem mentioned two months ago about the refactoring of constructors. I propose a way to handle that issue on the three highlighted cases as a PoC.

Sources/macros.js
```diff
@@ -972,6 +972,8 @@ export function newInstance(extend, className) {
     const model = {};
     const publicAPI = {};
     extend(publicAPI, model, initialValues);
+    const val = { ...initialValues, ...model };
```
jourdain, line 975: Are the `initialValues` suppose to override the `model` ones?

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,44 +318,38 @@ function vtkDataArray(publicAPI, model) {
...
-    model.size = model.values.length;
-    model.dataType = getDataType(model.values);
-  }
+  Object.assign(model, defaultValues(model));
 
   // Object methods
   macro.obj(publicAPI, model);
   macro.set(publicAPI, model, ['name', 'numberOfComponents']);
 
+  const type = initialValues.dataType ? initialValues.dataType : model.dataType;
```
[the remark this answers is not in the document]
jourdain: I pinned the wrong line but I'm talking about the `initialValues.values` if you tend to skip passing `initialValues` to the `extend()` function.

Sources/Common/Core/Points/index.js
```diff
@@ -89,10 +89,10 @@ const DEFAULT_VALUES = {
 
 // ----------------------------------------------------------------------------
 
-export function extend(publicAPI, model, initialValues = {}) {
-  Object.assign(model, DEFAULT_VALUES, initialValues);
+export function extend(publicAPI, model) {
```
jourdain, line 92: for good practice I will keep initialValues arg. Maybe we just need something to mention that we don't use it...

Sources/Common/Core/DataArray/index.js
```diff
@@ -262,9 +262,14 @@ function vtkDataArray(publicAPI, model) {
   };
 
   publicAPI.setData = (typedArray, numberOfComponents) => {
-    model.values = typedArray;
-    model.size = typedArray.length;
-    model.dataType = getDataType(typedArray);
+    if (typedArray) {
```
finetjul, line 265: if typedArray is null, you should empty model.values, no ?

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,44 +318,38 @@ function vtkDataArray(publicAPI, model) {
...
+  const type = initialValues.dataType ? initialValues.dataType : model.dataType;
+  if (!initialValues.values) {
+    initialValues.values = macro.newTypedArray(type, model.size);
+  } else if (Array.isArray(initialValues.values)) {
+    initialValues.values = macro.newTypedArrayFrom(type, initialValues.values);
+  }
+
   // Object specific methods
   vtkDataArray(publicAPI, model);
+  publicAPI.setData(initialValues.values, initialValues.numberOfComponents);
```
finetjul, line 352: it looks like it makes sense to create a test that tests ALL the different ways to create a DataArray

Sources/Common/Core/Points/index.js
finetjul: I do not understand why we do not want to support initialValues

Sources/Common/DataModel/Cell/test/testCell.js
```diff
@@ -47,13 +47,13 @@ test('Test vtkCell initialize with pointsIds', (t) => {
 
 test('Test vtkCell deepCopy', (t) => {
   const points = vtkPoints.newInstance();
-  points.setData([0, 0, 0, 2, 0, 0, 2, 2, 0]);
+  points.setData(Float64Array.from([0, 0, 0, 2, 0, 0, 2, 2, 0]));
```
finetjul, line 50: looking at the `vtkDataArray.setData` code, it seems that JS array (not typed) are supported.

If not, you need to update the TS doc

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -94,8 +94,14 @@ function vtkFullScreenRenderWindow(publicAPI, model) {
   model.interactor.initialize();
   model.interactor.bindEvents(model.container);
 
+  const superClass = { ...publicAPI };
+
   // Expose background
-  publicAPI.setBackground = model.renderer.setBackground;
+  publicAPI.setBackground = (background) => {
```
finetjul, line 100: use `macro.chain` ?

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -188,19 +194,22 @@ function vtkFullScreenRenderWindow(publicAPI, model) {
...
-  background: [0.32, 0.34, 0.43],
-  containerStyle: null,
-  controlPanelStyle: null,
-  listenWindowResize: true,
-  resizeCallback: null,
-  controllerVisibility: true,
-};
+function defaultValues(initialValues) {
+  return {
+    ...initialValues,
```
finetjul, line 199: shouldn't it be last to overwrite default values ?

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -94,8 +94,14 @@ function vtkFullScreenRenderWindow(publicAPI, model) {
...
 
+  const superClass = { ...publicAPI };
+
   // Expose background
-  publicAPI.setBackground = model.renderer.setBackground;
+  publicAPI.setBackground = (background) => {
+    // Synchronize the setting of background from here to renderer
+    model.renderer.setBackground(background);
+    superClass.setBackground(background);
+  };
```
finetjul, line 105: `publicAPI.setBackground()` call line 147 should probably be moved here or at the end of this function

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -94,8 +94,14 @@ function vtkFullScreenRenderWindow(publicAPI, model) {
...
```
finetjul, line 104: Actually, it does not seem to make sense to store in vtkFullScreenRenderWindow the background.

Maybe you could do:
```
publicAPI.setBackground = model.renderer.setBackground
publicAPI.getBackground = model.renderer.getBackground
...
extend(...) {
...
vtkFullScreenRenderWindow(...);
// we do not want to "store" background, only forward it to renderer
initialValues.background = model.background;
delete model.background;
}
```
...

Sources/Testing/testMacro.js
```diff
@@ -35,6 +35,7 @@ const DEFAULT_VALUES = {
   myProp11: 11,
   _myProp12: [12],
   _myProp13: 13,
+  myProp14: 14,
```
finetjul, line 38: how myProp14 is different from myProp11 ?

Sources/macros.js
finetjul: Hasn't `model` been updated with `initialValues` in `extend()` ?

Sources/Common/Core/DataArray/index.js
```diff
@@ -262,9 +262,14 @@ function vtkDataArray(publicAPI, model) {
   };
 
   publicAPI.setData = (typedArray, numberOfComponents) => {
-    model.values = typedArray;
-    model.size = typedArray.length;
-    model.dataType = getDataType(typedArray);
+    if (typedArray) {
+      model.values = typedArray;
+      model.size = typedArray.length;
+      model.empty = model.size === 0;
```
luciemac, line 268: Probably not the right test to do.
If initialValues.values is null, then you define an empty typedArray of size = model.size.
publicAPI.setData is called afterwards, so model.empty will be set to false whereas it should be true.

Sources/macros.js
```diff
@@ -317,9 +317,9 @@ export function obj(publicAPI = {}, model = {}) {
       } else {
         // Set data on model directly
         if (['mtime'].indexOf(name) === -1 && !noWarning) {
-          vtkWarningMacro(
-            `Warning: Set value to model directly ${name}, ${map[name]}`
-          );
+          // vtkWarningMacro(
```
luciemac, line 320: Should not be commented

Sources/Testing/testMacro.js
joannacirillo: myProp14 is the same as myProp11 except it does not have any setter. It tests that in this case the `publicAPI.set(...)` call pushes directly to the model the protected property.

Sources/Common/Core/DataArray/index.js
joannacirillo: I updated the function to correspond to the behavior it had before (throw an error if typedArray is either null or undefined).

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,44 +318,57 @@ function vtkDataArray(publicAPI, model) {
...
+  };
+}
 
 // ----------------------------------------------------------------------------
 
 export function extend(publicAPI, model, initialValues = {}) {
-  Object.assign(model, DEFAULT_VALUES, initialValues);
-
-  if (!model.empty && !model.values && !model.size) {
+  if (initialValues.empty === false && !initialValues.values) {
```
finetjul, line 337: `vtkCellarray` `empty`  is not passed as `initialValues` here

Sources/Common/Core/Points/index.js
```diff
@@ -80,18 +80,20 @@ function vtkPoints(publicAPI, model) {
...
+    ...initialValues,
+  };
+}
 
 // ----------------------------------------------------------------------------
 
 export function extend(publicAPI, model, initialValues = {}) {
-  Object.assign(model, DEFAULT_VALUES, initialValues);
-
+  Object.assign(model, defaultValues(model));
```
finetjul, line 96: I do not understand why you pass model as parameter to defaultValues

joannacirillo: It is in order that defaultValues of child classes are passed to parent class without being redundant with the passing of initialValues and risk being incoherent then when calling ``publicAPI.set(initialValues)` in new instance. However it will cause the setters of defaultValues not being called, so I'll update that.

Sources/Common/Core/DataArray/index.js
joannacirillo: This is resolved by last update (default values are passed through initialValues)

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,44 +318,55 @@ function vtkDataArray(publicAPI, model) {
...
-    model.size = model.values.length;
-    model.dataType = getDataType(model.values);
-  }
+  Object.assign(initialValues, defaultValues(initialValues));
 
   // Object methods
   macro.obj(publicAPI, model);
   macro.set(publicAPI, model, ['name', 'numberOfComponents']);
 
+  model.dataType = initialValues.dataType;
```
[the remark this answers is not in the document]
joannacirillo: At that point `initialValues.dataType` cannot be undefined because default values have been pushed to the `initialValues` if they don't exist in `initialValues`  (`Object.assign(initialValues, defaultValues(initialValues)`)

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -94,8 +94,14 @@ function vtkFullScreenRenderWindow(publicAPI, model) {
   model.interactor.initialize();
   model.interactor.bindEvents(model.container);
 
-  // Expose background
-  publicAPI.setBackground = model.renderer.setBackground;
+  publicAPI.setBackground = macro.chain(
+    publicAPI.setBackground,
```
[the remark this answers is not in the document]
joannacirillo: The setter previously existing was supposing that one can give an array of size 3 instead of size 4 (which is the case for the setBackground of renderer) with a default value.

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -213,9 +221,10 @@ export function extend(publicAPI, model, initialValues = {}) {
     'container',
     'controlContainer',
   ]);
-
+  macro.setArray(publicAPI, model, ['background'], 3, 1.0);
```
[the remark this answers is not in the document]
joannacirillo: It allows the size of the array given for the background to be 3 instead of 4 in the renderer class

Sources/Common/Core/DataArray/index.js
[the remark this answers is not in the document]
finetjul: In such case: `vtkDataArray.newInstance({dataType: null})`

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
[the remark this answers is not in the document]
finetjul: if you want to support array of 3 and 4, then rely on the `setBackground()` of the `renderer`.

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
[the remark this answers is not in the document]
finetjul: let's drop this feature to set background with just 3 elements.
 it is confusing to have 1 API with 3 (FullScreenRenderWindow) and one API with 4 (vtkRenderer).

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,44 +318,59 @@ function vtkDataArray(publicAPI, model) {
...
+      initialValues.values
+    );
+  }
+
   // Object specific methods
   vtkDataArray(publicAPI, model);
+  // We need to call setData manually here because it is not a setter for a property
+  // in itself but calling it is still the proper way to set the datas of the model
+  publicAPI.setData(initialValues.values, initialValues.numberOfComponents);
+  delete model.empty;
```
finetjul, line 368: should not be needed

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,44 +318,59 @@ function vtkDataArray(publicAPI, model) {
...
+      model.dataType,
+      initialValues.values
+    );
+  }
+
   // Object specific methods
   vtkDataArray(publicAPI, model);
+  // We need to call setData manually here because it is not a setter for a property
+  // in itself but calling it is still the proper way to set the datas of the model
+  publicAPI.setData(initialValues.values, initialValues.numberOfComponents);
```
finetjul, line 367: do yo uconfirme that initialValues.values will NEVER be null ?

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,44 +318,59 @@ function vtkDataArray(publicAPI, model) {
...
+  };
+}
 
 // ----------------------------------------------------------------------------
 
 export function extend(publicAPI, model, initialValues = {}) {
-  Object.assign(model, DEFAULT_VALUES, initialValues);
-
-  if (!model.empty && !model.values && !model.size) {
+  if (!initialValues.empty && !initialValues.values && !initialValues.size) {
```
finetjul, line 337: can you please document "empty" in index.d.ts ?

Sources/Common/Core/DataArray/index.js
finetjul, line 367: @jourdain , to avoid this exception, wouldn't it make sense to expose a convenient `setValues()` setter that simply calls setData() ? e.g. `publicAPI.setValues = (v) => publicAPI.setData(v);`

Sources/Common/Core/DataArray/test/testDataArray.js
```diff
@@ -1,10 +1,168 @@
...
+
+  t.throws(
+    () => vtkDataArray.newInstance({ empty: false }),
+    'Create instance with empty false, no values'
+  );
+
+  t.throws(
+    () => vtkDataArray.newInstance({}),
+    'Create instance without values'
+  );
```
finetjul, line 28: should be moved line 19

Sources/Common/Core/DataArray/test/testDataArray.js
```diff
@@ -1,10 +1,168 @@
...
+
+test.only('Test vtkDataArray instance', (t) => {
   t.ok(vtkDataArray, 'Make sure the class definition exists');
-  const instance = vtkDataArray.newInstance({ size: 256 });
-  t.ok(instance);
+
+  t.throws(
+    () => vtkDataArray.newInstance({ empty: false }),
+    'Create instance with empty false, no values'
+  );
```
finetjul, line 23: add a test with `empty: true`

Sources/Common/Core/DataArray/test/testDataArray.js
```diff
@@ -1,10 +1,168 @@
...
+  t.deepEqual(
+    {
+      dataType: DefaultDataType,
+      size: 0,
+      numberOfComponents: 1,
+      values: macro.newTypedArray(DefaultDataType),
+    },
+    getDataArrayProperties(dataArray3),
+    'Empty an instance (pass [] array)'
+  );
```
finetjul, line 103: to move in a separate function (e.g. testSetData)

Sources/Common/Core/DataArray/test/testDataArray.js
```diff
@@ -1,10 +1,168 @@
...
+    },
+    getDataArrayProperties(dataArray4),
+    'Change number of components at instanciation'
+  );
+
+  const dataArray5 = vtkDataArray.newInstance({
+    values: [],
+    numberOfComponents: 1,
+  });
+  dataArray5.setData([1, 2, 3, 4, 5, 6], 2),
```
finetjul, line 127: typo

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,44 +318,55 @@ function vtkDataArray(publicAPI, model) {
...
+  };
+}
 
 // ----------------------------------------------------------------------------
 
 export function extend(publicAPI, model, initialValues = {}) {
-  Object.assign(model, DEFAULT_VALUES, initialValues);
-
-  if (!model.empty && !model.values && !model.size) {
+  if (initialValues.empty === false && !initialValues.values) {
```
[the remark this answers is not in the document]
jourdain: empty flag was intended at construction time so you can allocate the array later on rather than ahead of time. Basically to ensure the user was aware that it created an empty dataArray.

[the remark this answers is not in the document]
finetjul: looking at [the code in master](https://github.com/Kitware/vtk-js/blob/master/Sources/Common/Core/DataArray/index.js#L329-L341), it seems that an array is ALWAYS provided or created, no ?

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -188,19 +191,24 @@ function vtkFullScreenRenderWindow(publicAPI, model) {
...
+  };
+}
 
 // ----------------------------------------------------------------------------
 
 export function extend(publicAPI, model, initialValues = {}) {
-  Object.assign(model, DEFAULT_VALUES, initialValues);
+  Object.assign(initialValues, defaultValues(initialValues));
+  // we do not want to "store" background, only forward it to renderer
+  model.background = initialValues.background;
```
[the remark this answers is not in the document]
joannacirillo: not needed anymore. Not removed from `initialValues` either (because then in `newInstance` it will call `setBackground`)

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -94,8 +94,14 @@ function vtkFullScreenRenderWindow(publicAPI, model) {
...
-  // Expose background
-  publicAPI.setBackground = model.renderer.setBackground;
+  publicAPI.setBackground = macro.chain(
+    publicAPI.setBackground,
+    model.renderer.setBackground
+  );
+  publicAPI.getBackground = model.renderer.getBackground;
+
+  // Update BG color
+  publicAPI.setBackground(...model.background);
```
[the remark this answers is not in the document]
joannacirillo: It is not needed anymore (since if the `background` is given in the initialValues it wall the setters, or if it is not as it is in the defaultValues it will be pushed in the initialValues)

joannacirillo:
Changes to setData do not full work for now.

Sources/Common/Core/DataArray/index.js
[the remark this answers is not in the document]
joannacirillo: I updated the `setData` function so that if it is the case it shows an error. I made an update here too to have the Default dataType as the type of the array an not the default type set in the `macro.newTypedArray` function.

Sources/Common/Core/DataArray/index.js
joannacirillo: Added an error log and return so that `values` is not updated to null.

Sources/Common/Core/DataArray/index.js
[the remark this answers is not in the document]
jourdain: First nothing is asynchronous here. 

Then I agree that extend() should still always have the initialValues.

Finally what I'm talking about is that if you process something in `extend()` you should clear it so it does not get re-process at the newInstance/set call.

```
extend()
   setData(...)
      model.values = ...
newInstance()
   set()
      model.values = initialValues.values
```

Sources/Common/Core/DataArray/index.js
jourdain: setData setup the tuple size. So not keen to change the model name to data either.

joannacirillo: I made an update so that the property `values` is now `data` in DataArray and `Points` and `CellArray` to keep an example of child classes refactoring running. That change needed the `initialValues` to be sorted so that setData is called lastly (to be able to set properly the other properties). The update on the other classes using DataArray will be needed to be done.
Some other classes work similarly as DataArray (with a property `values` and a function `setData`), a similar change thus will be needed to be done.

jourdain: Sorry that I wasn't clear earlier, but I do-not want to have `values` renamed to `data` especially because setData set 2 things rather than setting only the values.

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -188,19 +185,25 @@ function vtkFullScreenRenderWindow(publicAPI, model) {
...
+    ...initialValues,
+  };
+}
 
 // ----------------------------------------------------------------------------
 
 export function extend(publicAPI, model, initialValues = {}) {
-  Object.assign(model, DEFAULT_VALUES, initialValues);
+  // a background property can be set in the initialValues but we do not want
+  // to "store" background, only forward it to the renderer
```
finetjul, line 204: remove comment

Sources/Common/Core/DataArray/index.js
```diff
@@ -262,9 +268,19 @@ function vtkDataArray(publicAPI, model) {
   };
 
   publicAPI.setData = (typedArray, numberOfComponents) => {
-    model.values = typedArray;
+    if (!typedArray) {
+      vtkErrorMacro(`Cannot call setData with given array : ${typedArray}`);
+      return;
+    }
```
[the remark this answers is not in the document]
joannacirillo: The previous implementation would have raised an error in that case. It was not checking the typed array but was doing `model.size = typedArray.length`, which would have raised an error. It put the `values` to null but not update the other parameters. Such implementation avoiding an error can also be implemented.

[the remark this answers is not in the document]
finetjul: let's support `setData(null)`

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,42 +329,51 @@ function vtkDataArray(publicAPI, model) {
...
+    );
+  } else if (Array.isArray(initialValues.data)) {
+    initialValues.data = macro.newTypedArrayFrom(
+      model.dataType,
+      initialValues.data
+    );
+  }
+  // Sort given values so that setData is called lastly in newInstance
+  sortProperties(initialValues);
+
```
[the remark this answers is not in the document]
joannacirillo: I think it's best to let the user put the `values` as parameter, since when the instance is created the array passed to constructor will be called that way. It seems strange to me that one pass a parameter to instanciation that is called after another way. We can support both too.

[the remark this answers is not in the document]
joannacirillo: I think it's best to let the user put the `values` as parameter, since when the instance is created the array passed to constructor will be called that way. It seems strange to me that one pass a parameter to instanciation that is called after another way. We can support both too.

Sources/Rendering/Misc/FullScreenRenderWindow/index.js
joannacirillo: Maybye it could be better to have that information in the index.d.ts file or somewhere, in order that users know that when they pass it in the initialValues it will not be in the model afterwards.

finetjul: ok to document 'background' property in index.d.ts for `newInstance()`. No need to explain it is not "stored", this is an implementation detail, it is not a useful information for the "user".

Sources/Common/Core/DataArray/index.js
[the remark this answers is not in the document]
finetjul: fair enough, so let's see if we all agree on the following:
```
if initialValues = {empty: true, size: undefined|null|0} then model.values = null
if initialValues = {empty: true, size: 10} then model.values = new TypedArray(10)
if initialValues = {empty: false} then throw error
if initialValues = {empty: false, data: aTypedArray} then model.values = aTypedArray
if initialValues = {empty: false, data: aJSArray} then model.values = TypedArray.from(aJSArray)
if setData(null) then model.values = null
if setData(aTypedArray) then model.values = aTypedArray
if setData(aJSArray) then model.values = TypedArray.from(aJSArray)
```

Sources/Common/Core/DataArray/index.js
joannacirillo: Current implementation: null can be passed so values is set to null, size to 0, numberOfComponents to 1.

Sources/Common/Core/DataArray/index.js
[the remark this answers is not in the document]
finetjul: @jourdain Are you ok with https://github.com/Kitware/vtk-js/pull/2409#discussion_r876080336 ?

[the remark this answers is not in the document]
jourdain: yes, sorry about that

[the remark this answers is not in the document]
jourdain: For `if initialValues = {empty: true, size: undefined|null|0}`

I will be fine with `model.values = null || TypedArray(0)`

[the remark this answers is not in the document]
joannacirillo: > 
> ```
> if initialValues = {empty: true, size: undefined|null|0} then model.values = null
> if initialValues = {empty: true, size: 10} then model.values = new TypedArray(10)
> if initialValues = {empty: undefined|false} then throw error
> if initialValues = {empty: undefined|false, data: aTypedArray} then model.values = aTypedArray
> if initialValues = {empty: undefined|false, data: aJSArray} then model.values = TypedArray.from(aJSArray)
> if setData(null) then model.values = null
> if setData(aTypedArray) then model.values = aTypedArray
> if setData(aJSArray) then model.values = TypedArray.from(aJSArray)
> ```

It should be implemented this way in last commit.

joannacirillo:
After some troubleshooting while applying the changes on the DataArray to classes using it, it appeared that it was necessary to handle the field `values` in `extend`. While copying the object through the call of `vtk(dataArrayToCopy)` the object we copy will always have a property called values. In order to ensure that the array can be given by only one mean in `newInstance`, I chose to call it `values` and to transpose it to `data` in the function `extend` so that `setData` can be called in the constructor.

Sources/Common/Core/DataArray/index.js
```diff
@@ -313,42 +319,57 @@ function vtkDataArray(publicAPI, model) {
...
+      );
+  } else if (Array.isArray(initialValues.values)) {
+    initialValues.values = macro.newTypedArrayFrom(
+      model.dataType,
+      initialValues.values
+    );
+  }
+
+  // We store the values given in a field called data in order to call setData
+  // lastly in the constructor
```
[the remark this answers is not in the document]
jourdain: Wait what? `setData(values, nbComponents)` is not the same thing as `values`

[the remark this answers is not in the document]
joannacirillo: As `setData` is the proper way to set `values` and `numberOfComponents`, it should be called when the DataArray is created. When calling `publicAPI.set(initialValues)` in the constructor, the `set` function looks for setters with the format `setMyPropName` and call them if they exist. Thus in order to call `setData` when creating the DataArray a property `data` should exist in the object `initialValues`. That's why @finetjul proposed to change `values` to `data` when calling `newInstance`. However, when copying an object through the call of `vtk(obj)`, the function calls the `constructor` method with the given properties of the object hence the model, which in the case of the DataArray contains a property named `values`. The `extend` function of the DataArray must therefore expect a property named values being given while still calling `setData`.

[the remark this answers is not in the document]
jourdain: Normally by the time `publicAPI.set(initialValues)` get called any "processed" data should be removed from initialValues which means you don't have any problem if you clean/remove any field that is getting processed in the `extend` function...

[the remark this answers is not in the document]
finetjul: The idea is to let `publicAPI.set(initialValues)` process the "data/values" initial parameter (by calling `setData()` on it) instead of handling it in the `extend()` function.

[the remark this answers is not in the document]
jourdain: In general when we have a 1 to 1 mapping with the data that is passed in, I would agree. But for something like that, I'm not sure. Are you aiming to do `vtkDataArray.newInstance({ data: [values, 3]})` ?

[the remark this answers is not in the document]
finetjul: After discussion, `setData()` will be called in `extend()`

Sources/Rendering/Core/Actor/index.js
```diff
@@ -185,6 +166,27 @@ function vtkActor(publicAPI, model) {
...
+    mapper: null,
+    property: null,
+    backfaceProperty: null,
+
+    forceOpaque: false,
+    forceTranslucent: false,
+
+    bounds: [1, -1, 1, -1, 1, -1],
+    // Internal objects
+    boundsMTime: macro.obj({}),
```
finetjul, line 184: to move above `mapper`

Sources/Rendering/Core/Actor2D/test/testActor2D.js
```diff
@@ -16,7 +16,7 @@ import { Representation } from 'vtk.js/Sources/Rendering/Core/Property/Constants
 
 import baseline from './testActor2D.png';
 
-test('Test Actor2D', (t) => {
+test.only('Test Actor2D', (t) => {
```
finetjul, line 19: to not commit

Sources/Rendering/Core/Camera/index.js
```diff
@@ -213,9 +213,9 @@ function vtkCamera(publicAPI, model) {
   };
 
   publicAPI.roll = (angle) => {
-    const eye = model.position;
-    const at = model.focalPoint;
-    const up = model.viewUp;
+    const eye = model.position ? model.position : defaultValues().position;
```
finetjul, line 216: when `roll` is called BEFORE model.position is initialized ?

Sources/Rendering/OpenGL/ShaderProgram/index.js
```diff
@@ -572,14 +577,9 @@ function extend(publicAPI, model, initialValues = {}) {
...
   // Instantiate internal objects
-  model.attributesLocs = {};
-  model.uniformLocs = {};
-  model.vertexShader = vtkShader.newInstance();
-  model.vertexShader.setShaderType('Vertex');
-  model.fragmentShader = vtkShader.newInstance();
-  model.fragmentShader.setShaderType('Fragment');
-  model.geometryShader = vtkShader.newInstance();
-  model.geometryShader.setShaderType('Geometry');
+  initialValues.vertexShader.setShaderType('Vertex');
```
finetjul, line 580: can't shader type be passed as initialValue ?

joannacirillo: It was not the case before, but as now the shaders could be passed in initialValues it would be logical to accept setting the shaderType.

Sources/Common/DataModel/DataSetAttributes/FieldData.js
```diff
@@ -67,7 +67,7 @@ function vtkFieldData(publicAPI, model) {
               : a,
           { array: null, index: -1 }
         )
-      : null;
+      : { array: null, index: -1 };
```
finetjul, line 70: `(model.arrays || []).reduce(...)`

Sources/Common/DataModel/DataSetAttributes/index.js
```diff
@@ -261,8 +264,8 @@ export function extend(publicAPI, model, initialValues = {}) {
     'activePedigreeIds',
   ]);
 
-  if (!model.arrays) {
-    model.arrays = {};
+  if (!initialValues.arrays) {
+    initialValues.arrays = {};
```
finetjul, line 268: should be done in `defaultValues()`

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -2045,16 +2048,16 @@ export function extend(publicAPI, model, initialValues = {}) {
...
 
-  model.primitives = [];
-  model.primTypes = primTypes;
+  initialValues.primitives = [];
+  initialValues.primTypes = primTypes;
 
-  model.tmpMat3 = mat3.identity(new Float64Array(9));
-  model.tmpMat4 = mat4.identity(new Float64Array(16));
+  initialValues.tmpMat3 = mat3.identity(new Float64Array(9));
+  initialValues.tmpMat4 = mat4.identity(new Float64Array(16));
```
finetjul, line 2055: to move in `defaultValues()`

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -2045,16 +2048,16 @@ export function extend(publicAPI, model, initialValues = {}) {
     initialValues
   );
 
-  model.primitives = [];
-  model.primTypes = primTypes;
+  initialValues.primitives = [];
+  initialValues.primTypes = primTypes;
```
finetjul, line 2052: to move in `defaultValues()` ?

Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -2045,16 +2048,16 @@ export function extend(publicAPI, model, initialValues = {}) {
...
 
-  model.tmpMat3 = mat3.identity(new Float64Array(9));
-  model.tmpMat4 = mat4.identity(new Float64Array(16));
+  initialValues.tmpMat3 = mat3.identity(new Float64Array(9));
+  initialValues.tmpMat4 = mat4.identity(new Float64Array(16));
 
   for (let i = primTypes.Start; i < primTypes.End; i++) {
```
finetjul, line 2057:
```
if (initialValues.primitives == null) {
  for (let ...) {...}
}
```

Sources/Common/DataModel/DataSetAttributes/index.js
joannacirillo: `arrays` need to be in the model before the setters of active properties are called. thus, this part has been remplaced by pushing to the model `initialValues.arrays` before the call of the setters.

MilanPlFn2:
Hello, no news for a while, but where do we stand?

PR #2410, opened 2022-05-11
chore(reslicecursorwidget): same display handle size on all devices

laurennlam:
### Context
Make an example of displaying the reslice cursor widget with the same handle size on all devices.

### Results
Reslice cursor widget handle will have the same display size.


laurennlam:
@finetjul

finetjul:
LGTM

Please change the commit  from chore() to doc()

laurennlam:
> LGTM
> 
> Please change the commit from chore() to doc()

Done !

PR #2415, opened 2022-05-12
feat(light,volumeMapper): add positional light and multiple light support

JiayiXuu:
### Context
This pull request adds new features to the light module, which include: positional lighting in world coordinate, arbitrary number and combination of lights, updated camera module to account for the transformation of camera light, two sided lighting, and corresponding surface rendering change in the shader. These features are made in preparation for the upcoming photorealistic rendering features. All changes are in sync with the recent updates on VTK.

### Results
A example dataset is used to compare renderings before and after modifications. In each image a default light is present (head light with intensity 1). The red sphere shows the location of light source. The added new features avoid modifying existing functions as much as possible, although surface lighting function is changed to be keep up with latest updates on VTK.
- There's cameraLightTransformMatrix errors for camera light update, need to investigate 

image

### Changes

- [ ] Documentation and TypeScript definitions were updated to match those changes (not yet)
- camera: implemented `getCameraLightTransformMatrix()` and `computeCameraLightTransform().`
- light: implemented `getTransformedPosition()` and `getTransformedFocalPoint()`, exposed` attenuationValues`
- openGL volumeMapper: replaced `applyLighting()` with new surface lighting methods, updated uniforms and variables, support multiple lights
- shader template: added declarations for substitution

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Windows 10
  - **Browser**: Chrome Version 101.0.4951.54 (Official Build) (64-bit)


floryst (commented):
Looks good so far! cc @martinken for looking at the light code as well.

JiayiXuu:
@martinken for review

martinken:
I'm wondering if we should move some of these larger sections of shader code into the shader template. Specifically some of the functions. They look like they could be placed into the shader template and then we call them if needed. Might make the JS code cleaner and make editing the glsl functions easier.

Sources/Rendering/Core/Camera/index.js
```diff
@@ -348,7 +348,26 @@ function vtkCamera(publicAPI, model) {
...
+    // not sure if this is the correct transformation, based on the same funciton in VTK
+    mat4.copy(tmpMatrix, publicAPI.getViewMatrix());
+    mat4.invert(tmpMatrix, tmpMatrix);
+    mat4.scale(tmpMatrix, tmpMatrix, [
+      model.distance,
+      model.distance,
+      model.distance,
+    ]);
+    model.cameraLightTransform = mat4.create();
+    mat4.translate(model.cameraLightTransform, tmpMatrix, [0.0, 0.0, -1.0]);
```
martinken, line 368: Let's create this in the constructor and then we can identity() it here. That way we are not creating it on the heap every frame when the camera is moving.

Sources/Rendering/Core/Camera/index.js
```diff
@@ -348,7 +348,26 @@ function vtkCamera(publicAPI, model) {
   publicAPI.getFrustumPlanes = (aspect) => {
     // Return array of 24 params (4 params for each of 6 plane equations)
   };
-  publicAPI.getCameraLightTransformMatrix = () => {};
+
+  publicAPI.getCameraLightTransformMatrix = () => {
+    const matrix = mat4.create();
+    mat4.copy(matrix, model.cameraLightTransform);
+    return matrix;
+  };
```
martinken, line 356: I'm not seeing anything call this yet, so maybe we can change the signature. I'm trying to avoid heap allocations. So we could have it take matrix as an argument and then it just does the copy.

martinken:
Looks great. I didn't look carefully at the lighting code itself but the old to new images look the same.

A few of my comments have to do with trying to avoid heap allocs in the render loop. (and getTransformedPosition etc do this as well) This is because in the past with lots of actors that was the biggest performance issue. Creating objects/arrays causes the GC to run more often, with a large GC state, which slows down the rendering a lot. To fix it we removed a bunch of object/array/closure creations in the rendering loop. So places like getTransformedPosition which is called every frame for every light would be a problem in the polydata mapper as sometimes people create lots of polydata mappers. But for the volume mapper it is probably OK as people rarely create more than one volume mapper.  We probably should remove the new from those methods as eventually the polydata mapper may cal them as well and then it will be an issue. Just wanted to give some background to the comments.

floryst:
> Thanks! I agree with shader code composition part is difficult to read. I thought about relocating all related code into another index.js file and write up documentations after all photorealistic rendering parts are pushed. Putting everything in shader template might make the file very long with lots of #if#elseif#endif so I'm split between the two. @martinken @floryst thoughts?

The if/elseif/endif blocks are fine in my opinion, and if the shader template gets too unwieldy, we can break it down as needed.

Sources/Rendering/Core/Camera/index.js
JiayiXuu: I forgot to modify updateLightsGeometryToFollowCamera in Core/Renderer, which calls this function. Just added it to commit.

JiayiXuu:
> 

Sounds good! I'll make the change now (and in future commits).

martinken:
The shader code is a bit historical. In VTK C++ we built the shader almost entirely from C++ code. Some folks complained about this as it is hard to read etc. So for vtk.js we tried build very little in JS and rely on ifdefs as needed. Now that we have tried both approaches they each have their advantages and downfalls IMO.  But that is why the vtk.js volume mapper uses so many ifdefs and so little JS based building.

PR #2422, opened 2022-05-18
docs(breakingchanges): add breaking changes documentation

finetjul:
between 23 and 24

### Context
Not all breaking changes between 23 and 24 were documented

### Results
Give hints to switch to v24 

### Changes

### PR and Code Checklist

### Testing
- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -1,6 +1,6 @@
 ## From 23.x to 24
 
-All old-style widgets except OrientationMarkerWidget and PiecewiseGaussianWidget have been removed.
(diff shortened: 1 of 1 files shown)


floryst (approved):
LGTM.

I wonder if we should also publish a changelog generated from semantic-release, and if it is possible to auto-generate breaking changes entries. That way we get a base to start with, and can update those files with more detail as needed.

finetjul:
> I wonder if we should also publish a changelog generated from semantic-release, and if it is possible to auto-generate breaking changes entries. That way we get a base to start with, and can update those files with more detail as needed.

good idea

PR #2424, opened 2022-05-18

### LINKED ISSUE #2423: Old jsdoctypeparser version (via dox) can't handle typescript function types
### High-level description

ScalarBarActor.d.ts has a type that causes jsdoctypeparser v1.2.0 to fail:
```
          throw new Lexer.SyntaxError(e.message, typeExp, e.offset);
          ^
[Error [TypeLexerSyntaxError]: Expected "!", "#", ")", ".", ".<", "<", "=", "?", "[]", "|", "~", [ \t\r\n ] or [a-zA-Z0-9_$] but ":" found.:
(helper: any) => void
      ^]
```

### Steps to reproduce
- Run `npm run doc:generate-api`

### Detailed behavior

`dox` depends on `jsdoctypeparser@^1.2.0`, which fails at parsing `(helper: any) => void` in the param type of a jsdoc string [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ScalarBarActor/index.d.ts#L197).

FYI @daker there is a PR to dox to upgrade jsdoctypeparser in dox (https://github.com/tj/dox/pull/196).

### Expected behavior

Generating the API does not error out.

### Environment
- **vtk.js**: 24.12.1
- **OS**: N/A
- **Browser**: N/A

daker:
fix #2423

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/Coordinate/Constants.d.ts
+++ b/Sources/Rendering/Core/Coordinate/Constants.d.ts
@@ -1,0 +1,14 @@
+export declare enum Coordinate {
+  DISPLAY = 0,
+  NORMALIZED_DISPLAY = 1,
+  VIEWPORT = 2,
+  NORMALIZED_VIEWPORT = 3,
+  PROJECTION = 4,
+  VIEW = 5,
+  WORLD = 6,
+}
+
+declare const _default: {
+	Coordinate: typeof Coordinate;
+};
+export default _default;
--- a/Sources/Rendering/Core/ImageMapper/Constants.d.ts
+++ b/Sources/Rendering/Core/ImageMapper/Constants.d.ts
@@ -1,0 +1,14 @@
+export declare enum SlicingMode {
+	NONE = -1,
+	I = 0,
+	J = 1,
+	K = 2,
+	X = 3,
+	Y = 4,
+	Z = 5,
+}
+
+declare const _default: {
+	SlicingMode: typeof SlicingMode;
+};
+export default _default;
--- a/Sources/Rendering/Core/ImageProperty/Constants.d.ts
+++ b/Sources/Rendering/Core/ImageProperty/Constants.d.ts
@@ -1,0 +1,9 @@
+export declare enum InterpolationType {
+	NEAREST = 0,
+	LINEAR = 1,
+}
+
+declare const _default: {
+	InterpolationType: typeof InterpolationType;
+};
+export default _default;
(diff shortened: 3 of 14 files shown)


daker:
@Julesdevops or @floryst if you want to review it

Sources/Rendering/Core/Coordinate/Constants.d.ts
```diff
@@ -0,0 +1,14 @@
+export declare enum Coordinate {
```
Julesdevops, line 1: I would name it `CoordinateSystems` for consistency.

Sources/Rendering/Core/ImageMapper/Constants.d.ts
```diff
@@ -0,0 +1,14 @@
+export declare enum SlicingMode {
```
Julesdevops, line 1: `SlicingModes` if we want to be consistent.

Sources/Rendering/Core/ImageProperty/Constants.d.ts
```diff
@@ -0,0 +1,9 @@
+export declare enum InterpolationType {
```
Julesdevops, line 1: Same as above

Julesdevops (commented):
Looks cool overall

Sources/Rendering/Core/ImageProperty/index.d.ts
```diff
@@ -87,19 +83,19 @@ export interface vtkImageProperty extends vtkObject {
...
 	 */
-	getRGBTransferFunction(idx?: number): any;
+	getRGBTransferFunction(idx?: number): vtkColorTransferFunction;
 
 	/**
 	 * Alias to get the piecewise function (backwards compatibility)
 	 * @param {Number} [idx]
 	 */
-	getScalarOpacity(idx: number): number;
+	getScalarOpacity(idx?: number): any;
```
floryst, line 92: This should return a `vtkPiecewiseFunction`.

Sources/Rendering/Core/ImageMapper/Constants.d.ts
daker: well it should SlicingModes but the Constants.js we have SlicingMode

Sources/Rendering/Core/Coordinate/Constants.d.ts
daker: That's what we have in Constants.js

Julesdevops: fair enough

floryst:
LGTM

PR #2426, opened 2022-05-19
fix(scalarbaractor): check for renderable mtime when deciding if we should update the actor

Julesdevops:
### Context
Say you have a `scalarBarActor` instance and you want to sometimes update its styling; doing something like that:
```js
const newOverlayColor = "green";
scalarBarActor.setAxisTextStyle({ fontColor: newOverlayColor });
scalarBarActor.setTickTextStyle({ fontColor: newOverlayColor });
renderWindow.render();
```
Currently this code does not work.

### Results
This PR makes it so any changes done using the `vtkScalarBarActor` public API triggers a redraw of the actor.

### Changes
Check for the renderable modified time.

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: v19.4.0
  - **OS**: Archlinux
  - **Browser**: Firefox 100.0 (64-bit)

### DIFF
--- a/Sources/Rendering/Core/ScalarBarActor/index.js
+++ b/Sources/Rendering/Core/ScalarBarActor/index.js
@@ -221,8 +221,11 @@ function vtkScalarBarActorHelper(publicAPI, model) {
     // did something significant change? If so rebuild a lot of things
     if (
       model.forceUpdate ||
-      Math.max(scalarsToColors.getMTime(), publicAPI.getMTime()) >
-        model.lastRebuildTime.getMTime()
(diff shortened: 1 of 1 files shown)


Julesdevops:
thanks @luciemac

Julesdevops:
@finetjul What's your opinion on this?

finetjul:
I do not know enough the scalar bar actor to validate the PR.
Maybe @martinken has an idea ?

martinken:
Should be fine. +1

PR #2430, opened 2022-05-23
refactor(pako): replace with fflate

### LINKED ISSUE #2429: pako is bundled twice
### High-level description
pako is bundled twice when @kitware/vtk.js is imported in a project, one is a top-level dependency (pako@2.0.3) and the other is a dependency of jszip (pako@1.0.11).

### Steps to reproduce

https://bundle.js.org/?q=@kitware/vtk.js,@kitware/vtk.js/Common/Core/Math,@kitware/vtk.js/Common/DataModel/Plane,@kitware/vtk.js/Common/Core/Math,@kitware/vtk.js/Common/DataModel/ImageData,@kitware/vtk.js/Common/DataModel/ITKHelper,@kitware/vtk.js/Common/Core/Math,@kitware/vtk.js/Common/Core/Math,@kitware/vtk.js/Interaction/Manipulators/MouseCameraTrackballRotateManipulator,@kitware/vtk.js/Interaction/Manipulators/MouseCameraTrackballZoomToMouseManipulator,@kitware/vtk.js/Interaction/Manipulators,@kitware/vtk.js/Rendering/Core/Property2D/Constants,@kitware/vtk.js/Rendering/Core/Coordinate/Constants&treeshake=[*],[{+angleBetweenVectors+}],[{+default+as+z+}],[{+multiply3x3_vect3+}],[{+default+as+x+}],[{+default+as+y+}],[{++signedAngleBetweenVectors,cross,min,norm,}],[{+add,multiplyScalar,normalize,subtract,clampValue,dot+}],[{+default+as+m+}],[{+default+as+mm+}],[{+default+as+Manipulators+}],[{++DisplayLocation+}],[{+Coordinate+}]&config={"analysis":true}

### Current behavior

CleanShot 2022-05-20 at 4 12 41@2x

### Environment
- **vtk.js**: 
- **OS**: 
- **Browser**:

floryst:
### Context

- Replaces pako and jszip with fflate. Closes #2429 
- jszip uses a different major version of pako than vtk.js. This is an attempt to try to use fflate instead. The alternative would be to downgrade the vtk.js pako version to v1.

### Results

- Smaller bundle size when using vtk.js components that use pako and jszip.

### Changes

- [x] Change pako usages to fflate
- [x] Change jszip usages to fflate

### PR and Code Checklist

### Testing

- [x] test modules that used to use pako
- [x] test modules that used to use jszip


floryst:
I've updated this with all the necessary changes. This technically has breaking change behavior, since some APIs no longer pass around a JSZip object. If that's not a big deal, I'm okay with merging without a major bump, otherwise this can wait until the next major release. @jourdain any thoughts?

jourdain:
I'm good without a major bump but maybe we should have it as a `feat`?

`feat(jszip): use fflate and remove thirdparty export`

PR #2432, opened 2022-05-23
feat(add density gradient with gradient opacity support)

JiayiXuu:
### Context
This pull request adds support for normal calculation based on density. Previously normal is based on scalar values which might not represent surface orientation correctly. As shown in the cube sphere example, opacity & scalar normal directions are flipped, resulting in reversed shading. When this happens, only the shadow direction is wrong but the volume is usually decently illuminated. Therefore opacity normal is turned off by default. On the other hand, for photorealistic rendering, correct normal direction is important for stochastic sampling of scattering direction (to capture global shadow). 

image

### Results
Following images uses one light source, either directional or positional, that shines from the front towards the cavity. The first two rows compare two methods when `twoSidedLighting = Off `. They demonstrates the effect of density gradient nicely, because this dataset happens to have inverted scalar value in the cavity region just like the sphere example. With scalar normal, it appears that the light is coming from the back. 

For last two rows, we have `twoSidedLighting = On`. There are slight differences in some regions but the overall shadow effects are similar. 

image

### Changes
- [ ] Documentation and TypeScript definitions were updated to match those changes (not yet)
- Added `computeNormalFromOpacity` option in `VolumeMapper`
- Added functions `computeNormalForDensity` and `computeDensityNormal` to calculation normal and density normal in `vtkVolumeFS.vtk` and support gradient opacity
- Adjusted call sequence in `vtkVolumeFS.vtk` to account for different conditions
- Corrected errors in specular calculation from last push

### PR and Code Checklist

### Testing

- [X] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Windows 10  
  - **Browser**: Chrome Version 101.0.4951.54 (Official Build) (64-bit)


JiayiXuu:
1 test failed after modifying specular formula, can we change the test accordingly? 

image

finetjul:
> 1 test failed after modifying specular formula, can we change the test accordingly?
> 
> image

I prefer the "expected" image to the "output" image. In the output image, I find it odd to have "light" pixels at the border of the head.

JiayiXuu:
> I prefer the "expected" image to the "output" image. In the output image, I find it odd to have "light" pixels at the border of the head.

Yeah because there's a specular component in the test. I think it's pretty odd for medical volume to be specular (shiny/reflective) in the first place. When rendering reflective surface the new formula performs better at the border (hence border is brighter). 

old:
specular_old

new:
sepcular_new

The new formula is used in VTK. The default specular value is 0.0 so user has to turn it on manually.

JiayiXuu:
> Something feels off on that test. Where is the light on this test?

It's a default light. Directional headlight with 1.0 intensity.

JiayiXuu:
> OK I think the correct way to handle two sided lighting with a pure Phong model is to just remove the second twoSided lighting test. The two sided part is already handled by flipping ndotL with this approach. Can you give that a shot and see if it looks better?
> 
> ```
> float ndotL,vdotR;
> vertLightDirection = lightDirectionVC[i];
> ndotL = dot(normal.xyz, vertLightDirection);
> if (ndotL < 0.0 && twoSidedLighting)
> {
>   ndotL = -ndotL;
> }
> if (ndotL > 0.0)
> {
>   diffuse += ndotL * lightColor[i];
>   //specular
>   vdotR = dot(-rayDirVC, normalize(2.0 * ndotL * -normal.xyz + vertLightDirection));
>   if (vdotR > 0.0)
>   {
>     specular += pow(vdotR, vSpecularPower) * lightColor[i];
>   }
> }
> ```

Ah I see. Now both border and center parts are darker.
image

martinken:
Yes well the specular power means different things in a blinn model versus phong (ala https://nicolbolas.github.io/oldtut/Illumination/Tut11%20BlinnPhong%20Model.html) so that needs to be adjusted in the test. Try lowering the specular power and see if you find a value that works.

JiayiXuu:
> Yes well the specular power means different things in a blinn model versus phong (ala https://nicolbolas.github.io/oldtut/Illumination/Tut11%20BlinnPhong%20Model.html) so that needs to be adjusted in the test. Try lowering the specular power and see if you find a value that works.

Okay thanks. The lowest delta I got is 0.17%. So either we tune up the threshold to pass the test or revert back to blinn-phong. I'm leaning towards blinn-phong now because it handles >90 degree better...

JiayiXuu:
> Yes well the specular power means different things in a blinn model versus phong (ala https://nicolbolas.github.io/oldtut/Illumination/Tut11%20BlinnPhong%20Model.html) so that needs to be adjusted in the test. Try lowering the specular power and see if you find a value that works.

Yep it's working now. I'm leaning towards reverting back to blinn-phong though since it handles >90 degrees better. Although this may not happen often, but just in case.

martinken:
I'm fine with either replacing the valid image now that we've fixed the edge issue or keeping the old Blinn model.  So +1 either way

martinken (approved):

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -506,17 +619,13 @@ void applyLighting(inout vec3 tColor, vec4 normal)
             ndotL = -ndotL;
           }
           if (ndotL > 0.0)
-          {
+          {
```
floryst, line 622: Here's some extraneous whitespace that should be removed.

floryst (approved):
LGTM overall

PR #2433, opened 2022-05-24
fix(cellpicker): skip cells that have the VTK_EMPTY_CELL type

Julesdevops:
### Context
Currently there is a `vtkCellTypes.deleteCell(cellId)` public method which set the type of the cell with the given `cellIId` to `VTK_EMPTY_CELL`.
I think the expected behavior from calling such a method is that the cell is simply ignored in most logic within VTK.js, picking included.

### Results
```
const someCellIdIWantToDelete = 5234;
polyData.buildCells();
const cellTypes = polyData.getCells();
cellTypes.deleteCell(someCellIdIWantToDelete)
```
Doing this kind of things and then trying to pick at the position of the cell will ignore the said cell.
### Changes
Add a condition to skip empty cells in vtkCellPicker

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: v19.4.0
  - **OS**: linux
  - **Browser**: firefox 100.0 (64-bit)

### DIFF
--- a/Sources/Rendering/Core/CellPicker/index.js
+++ b/Sources/Rendering/Core/CellPicker/index.js
@@ -285,14 +285,20 @@ function vtkCellPicker(publicAPI, model) {
 
       const numberOfCells = data.getNumberOfCells();
 
+      /* eslint-disable no-continue */
       for (let cellId = 0; cellId < numberOfCells; cellId++) {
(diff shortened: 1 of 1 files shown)


Julesdevops:
@finetjul

finetjul:
LGTM

PR #2444, opened 2022-06-01
fix(ImageMapper): getBoundsForSlice half voxel larger

### LINKED ISSUE #2439: ImageMapper.getBoundsForSlice() does not contain slice like before
### High-level description
The itk-vtk-viewer's slice outline representation is slightly smaller after upgrading from vtk.js v23.x to v24.x.  

### vtk.js 23.x
image

### vtk.js 24.14.2
Screenshot from 2022-05-31 11-10-26

Suspect it is related to commit dcb2573a274197683ce001f4feb70969f247acde

Should imageMapper.getBoundsForSlice use image.getSpatialExtent here?
https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ImageMapper/index.js#L286

Happy to make a reproduction if helpful.

PaulHax:
### Context

ImageMapper.getBoundsForSlice gives bounds that are a half voxel small as described in #2439 

### Results
ImageMapper.getBoundsForSlice tightly contains the slice.

### Changes
`ImageMapper.getBoundsForSlice` calls `getSpatialExtent` on data instead of `getExtent`

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 24.16.0
  - **OS**: Ubuntu 22.04
  - **Browser**: Chrome

closes #2439


jadh4v:
LGTM.
Thanks for catching this. Also if possible update the documentation of the function here:
https://github.com/PaulHax/vtk-js/blob/2048b0134c12f18c644feee8e12fddf155547713/Sources/Rendering/Core/ImageMapper/index.d.ts#L50

Update the description of `thickness` as _The slice thickness in index space (unit voxel spacing)_.

The thickness is applied in both directions which makes the bounding box twice as thick. I wonder if it should be applied as `+- 0.5 * thickness` inside the function. This could be a breaking change though for applications.

PaulHax:
> Update the description of thickness as The slice thickness in index space (unit voxel spacing).

Done.

> The thickness is applied in both directions which makes the bounding box twice as thick.

Hmm, yeah.  Change parameter naming?  `thickness` -> `halfThickness`.  Naw, better to throw in the next major version small BREAKING_CHANGES buffer?

jadh4v:
> > Update the description of thickness as The slice thickness in index space (unit voxel spacing).
> 
> Done.
> 
> > The thickness is applied in both directions which makes the bounding box twice as thick.
> 
> Hmm, yeah. Change parameter naming? `thickness` -> `halfThickness`. Naw, better to throw in the next major version small BREAKING_CHANGES buffer?

Yea, we can go with the change in parameter name for now, and log an issue for the breaking change for later.

PR #2454, opened 2022-06-03
perf(vtktexture): speedup array data type conversion

finetjul:
### Context
I identified some potential performance hit in vtkTexture.

### Results
Should be faster to copy arrays.

### Changes
Rely on built-in copy function instead of a manual for loop

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: <master
  - **OS**: Windows 10
  - **Browser**: Chrome

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -610,11 +610,11 @@ function vtkOpenGLTexture(publicAPI, model) {
     ) {
       for (let idx = 0; idx < data.length; idx++) {
         if (data[idx]) {
-          const newArray = new Float32Array(pixCount);
-          for (let i = 0; i < pixCount; i++) {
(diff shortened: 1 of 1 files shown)


martinken:
Fine by me +1 Out of curiosity did you measure a performance difference?

martinken (approved):

martinken:
I meant did you measure a performance difference in vtk.js. Aka was there some slow code and you determined the problem was this and made the change and saw it get faster.  It sounds like this was just opportunistic, which is fine, no issues with the change. 

The background is that I do not trust toy timing tests, especially ones that hit the heap. I find that something that hits the heap a lot with a one object heap is not a great test for how it will work with a huge heap of interconnected objects. As the GC sweeps are far more costly in the latter case than the toy test.

floryst:
LGTM

PR #2456, opened 2022-06-04
fix(Shader): fix condition statement for shadingOff

JiayiXuu:
### Context
Shader code does not compile when switching `ShadingOn` from `true` to `false`. This is a fix for MR [#2432](https://github.com/Kitware/vtk-js/pull/2432).

### Results
Test on all flags related to shading: shading; opacity gradient; density gradient; parallel projection; two sided lighting. No further bugs found.

### Changes
Added additional condition `#if defined vtkGradientOpacityOn` so normal is computed either when there's lighting or when gradient opacity is on.

### PR and Code Checklist

### Testing

- [X] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Windows 10


martinken (approved):
Thanks!

PR #2457, opened 2022-06-06
docs(RemoteView): improve sample viewStream creation

eino:
Update sample to pass viewStream on vtkRemoteView creation,
rather than connecting it with the static connectImageStream,
allowing better lifecycle management.
Also update a couple of typescript definitions.

### DIFF
--- a/Sources/Rendering/Misc/RemoteView/index.d.ts
+++ b/Sources/Rendering/Misc/RemoteView/index.d.ts
@@ -11,6 +11,7 @@ interface IRemoteViewInitialValues {
   rpcMouseEvent?: string;
   rpcGestureEvent?: any;
   rpcWheelEvent?: any;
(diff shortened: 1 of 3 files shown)


daker (approved):
The typescript defs looks good

Sources/Rendering/Misc/RemoteView/index.d.ts
```diff
@@ -11,6 +11,7 @@ interface IRemoteViewInitialValues {
   rpcMouseEvent?: string;
   rpcGestureEvent?: any;
   rpcWheelEvent?: any;
+  viewStream: vtkViewStream;
```
jourdain, line 14: This is technically optional.

eino: Indeed! Fixed it :+1:

PR #2459, opened 2022-06-06
[WIP] Refactoring vtk.js internals into Typescript

floryst:
### Context

There has been interest to use Typescript internally in vtk.js, rather than just specifying the external API. Because of the exploratory nature of this MR, it will remain in draft form for now and serve as a starting point for discussion. Once interested parties are satisfied with the approach, we can begin incrementally converting parts of the codebase into typescript as needed. 

An overview of what this MR contains (will be updated):
- Fully typed `macros.ts`
- Fully typed `OutlineFilter`
- Fully typed `ClassHierarchy`

Some notes that can serve as points of discussion:
- This will by no means supplant the work done to specify the external API. In fact, we will still need it, since the ts static analyzer cannot handle runtime-based object construction, and so specifying the external API is still required.
- The looseness of typing the internal model (and public API) as a Record is required to support index-based access to properties in addition to dot notation.
- Prettier currently does not run on `*.ts` files. If we start incrementally refactoring into using typescript fully, we will begin running prettier over typescript files. I will leave it open to discussion whether we should prettify `*.d.ts` files.

@jourdain @finetjul @daker @FezVrasta


Sources/Filters/General/OutlineFilter/OutlineFilter.ts
```diff
@@ -0,0 +1,104 @@
+import macro, {
+  AlgoModel,
+  IndexableAPI,
+  IndexableModel,
+} from 'vtk.js/Sources/macros';
+import vtkPolyData from 'vtk.js/Sources/Common/DataModel/PolyData';
+import { vtkObject, vtkOutputPort } from 'vtk.js/Sources/interfaces';
```
daker, line 7: does the IntelliSense works with the absolute path?

DavidBerger98:
Just adding this from our discussion with @finetjul and @daker in #2532:

In case we want to keep the tabs in the `*.d.ts` files (would probably minimize the diffs if we lint them the first time), I suggest the following for `.prettier.config.js`:

```js
module.exports = {
  printWidth: 80,
  singleQuote: true,
  trailingComma: 'es5',
  arrowParens: 'always',
  overrides: [
    {
      files: '*.d.ts',
      options: {
        useTabs: true,
        tabWidth: 4,
        singleQuote: false,
      },
    },
  ],
};
```

However, it probably doesn't make sense to have different linting rules between `.js` and `.ts` files.

Sources/macros.ts
```diff
@@ -0,0 +1,2007 @@
...
+export type IndexableAPI<T> = T & Record<string, Function>;
+export type IndexableModel<T> = T & Record<string, unknown>;
+
+export interface ObjModel extends Record<string, unknown> {
+  mtime: number;
+  classHierarchy: Array<string>;
+  deleted?: boolean;
+}
+
+interface ObjPublicAPI extends vtkObject, Record<string, Function> {}
```
thewtex, line 256: How about `class`'es? :1234: :pleading_face:

PR #2460, opened 2022-06-07
fix: prevent unsafe model.previousPosition property access

FezVrasta:
### Context

Race conditions can lead to the `onButtonDown` method to be called after `onPointerLockMove`, this means that `previousPosition` will be set to `undefined` while `onPointerLockMove` tries to access its `x` and `y` properties.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

### Funding

### DIFF
--- a/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
@@ -225,6 +225,10 @@ function vtkMouseRangeManipulator(publicAPI, model) {
     // get a `onMouseMove` call after the pointer has been unlocked.
     if (!interactor.isPointerLocked()) return;
 
+    // previousPosition could be undefined if for some reason the
+    // `startPointerLockEvent` method is called before the `onButtonDown` one.
+    if (model.previousPosition == null) return;
+
     model.previousPosition.x += event.movementX;
     model.previousPosition.y += event.movementY;
 


Sources/Interaction/Manipulators/MouseRangeManipulator/index.js
```diff
@@ -9,6 +9,10 @@ function vtkMouseRangeManipulator(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkMouseRangeManipulator');
 
+  // Set a default value so that race conditions will not lead to
+  // attempting to access properties of an undefined model property
+  model.previousPosition = { x: 0, y: 0 };
```
finetjul, line 14: shouldn't x and y be undefined (or null) ?  because it is not quite true that they are at `(0,0)` at default

FezVrasta: If I set them to `null` then I also need to add checks on `onPointerLockMove` to avoid it using the value to perform math operations on it.

Honestly this doesn't matter because the value will be replaced with the correct one as soon the `mouseDown` method is called anyways.

FezVrasta: If you prefer I can add an early return on `onPointerLockMove` right before we attempt to access x/y, this will prevent the `onMouseMove` event from firing then.

FezVrasta: I just went ahead and made the change as it looks cleaner

finetjul: I think that makes more sense to me. Thanks !

floryst (approved):

PR #2462, opened 2022-06-08
Fix preventDefault usage and pointer lock target

floryst:
### Context

- Pointer lock target needs to be the interactor container rather than the canvas, especially in the ManyRenderers situation.
- unconditional preventDefault and stopPropagation causes issues for users. In particular, preventDefault with the new pointer event usage results in compatibility events (i.e. mouse events) from ever triggering, which breaks downstream applications that depend on it.

### Results

- pointer lock should work for ManyRenderers situations
- preventDefaultOnPointer* API gives flexibility to developers

### Changes

- [ ] pointer lock uses interactor container
- [ ] add preventDefaultOnPointerDown and preventDefaultOnPointerUp setters and getters
  - [ ] Change the default behavior from "true" to "false" (technically breaking, but I'm considering this a minor breakage since it helps avoid unintended breakages)

### PR and Code Checklist

### Testing

- [ ] Tried several examples (widgets, ManyRenderers, Cone, PiecewiseGaussianWidget) and nothing seems broken


jourdain (approved):
LGTM

floryst:
Recent force-push fixes an issue where `isPointerLocked()` mistakenly returns true if (1) model.container is null because no container was bound, and (2) document.pointerLockElement is null from no pointer actually being locked.

PR #2465, opened 2022-06-09
fix(scaleinpixels): uniformize scaleInPixels between widgets

tomsuchel:
In many widgets, handles did not have scaleInPixels set to `true`, so the handle size was constant
regardless of the size of the scene. This sometimes caused the handles to disappear completly. 

This PR fixes this issue by standardizing the value of `scaleInPixels` between all widgets using handles,
and by using more reasonable handle sizes.

`scaleInPixels` is also added in `methodsToLink` so its value can be easily changed.

### Changes

### PR and Code Checklist

### Testing


tomsuchel:
@finetjul

Sources/Widgets/Widgets3D/AngleWidget/state.js
```diff
@@ -8,7 +8,7 @@ export default function generateState() {
       mixins: ['origin', 'color', 'scale1', 'visible', 'manipulator'],
       name: 'moveHandle',
       initialValues: {
-        scale1: 0.1,
+        scale1: 50,
```
finetjul, line 11: 50 pixels seems rather large, have you tried on non-4k screens ?

finetjul:
How about changing the default sphere representation scaleInPixels  to true ?

Sources/Widgets/Widgets3D/AngleWidget/state.js
finetjul: ok pour 30px then.

floryst:
LGTM! Having this as a default is a good idea.

Sources/Widgets/Widgets3D/ImageCroppingWidget/state.js
```diff
@@ -62,7 +62,7 @@ export default function build() {
             ],
             name,
             initialValues: {
-              scale1: 10,
+              scale1: 20,
```
finetjul, line 65: why not 30 ?

finetjul:
Can you please flag it as BREAKING_CHANGES.

tomsuchel:
> Can you please flag it as BREAKING_CHANGES.

Done!

PR #2466, opened 2022-06-09
fix(dragging): uniformize dragging behavior between widgets

tomsuchel:
Uniformize dragging behavior to make it similar between all widgets: 
- always add the protected variable `_isDragging` in the model (if the widget can be dragged)
- do not change cursor & do not request animation if `model.dragable` is false
- refactor `handleLeftButtonRelease()` method to always return a value (`macro.VOID` or `macro.EVENT_ABORT`), and to correctly call `invokeEndInteractionEvent()`
- fix some cursors inconsistencies 

Also fixes a small bug in Angle/Distance/PolyLineWidget where we were calling `setOrigin` on a
PolyLineRepresentation.

### Changes

### PR and Code Checklist

### Testing

### DIFF
--- a/Sources/Widgets/Widgets3D/AngleWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/AngleWidget/behavior.js
@@ -5,7 +5,7 @@ const MAX_POINTS = 3;
 
 export default function widgetBehavior(publicAPI, model) {
   model.classHierarchy.push('vtkAngleWidgetProp');
-  let isDragging = null;
+  model._isDragging = false;
 
   const picker = vtkPointPicker.newInstance();
   picker.setPickFromList(1);
@@ -60,8 +60,8 @@ export default function widgetBehavior(publicAPI, model) {
       newHandle.setColor(moveHandle.getColor());
       newHandle.setScale1(moveHandle.getScale1());
       newHandle.setManipulator(manipulator);
-    } else {
-      isDragging = true;
+    } else if (model.dragable) {
+      model._isDragging = true;
       model._apiSpecificRenderWindow.setCursor('grabbing');
       model._interactor.requestAnimation(publicAPI);
     }
@@ -92,7 +92,9 @@ export default function widgetBehavior(publicAPI, model) {
 
       if (
         worldCoords.length &&
-        (model.activeState === model.widgetState.getMoveHandle() || isDragging)
+        (model.activeState === model.widgetState.getMoveHandle() ||
+          model._isDragging) &&
+        model.activeState.setOrigin // e.g. the line is pickable but not draggable
       ) {
         model.activeState.setOrigin(worldCoords);
         publicAPI.invokeInteractionEvent();
@@ -110,11 +112,26 @@ export default function widgetBehavior(publicAPI, model) {
   // --------------------------------------------------------------------------
 
   publicAPI.handleLeftButtonRelease = () => {
-    if (isDragging && model.pickable) {
+    if (
+      !model.activeState ||
+      !model.activeState.getActive() ||
+      !model.pickable
+    ) {
+      return macro.VOID;
+    }
+    if (
+      model.hasFocus &&
+      model.widgetState.getHandleList().length === MAX_POINTS
+    ) {
+      publicAPI.loseFocus();
+      return macro.VOID;
+    }
+
+    if (model._isDragging) {
       model._apiSpecificRenderWindow.setCursor('pointer');
       model.widgetState.deactivate();
       model._interactor.cancelAnimation(publicAPI);
-      publicAPI.invokeEndInteractionEvent();
+      model._isDragging = false;
     } else if (model.activeState !== model.widgetState.getMoveHandle()) {
       model.widgetState.deactivate();
     }
@@ -123,17 +140,12 @@ export default function widgetBehavior(publicAPI, model) {
       (model.hasFocus && !model.activeState) ||
       (model.activeState && !model.activeState.getActive())
     ) {
-      publicAPI.invokeEndInteractionEvent();
       model._widgetManager.enablePicking();
       model._interactor.render();
     }
 
-    // Don't make any more points
-    if (model.widgetState.getHandleList().length === MAX_POINTS) {
-      publicAPI.loseFocus();
-    }
-
-    isDragging = false;
+    publicAPI.invokeEndInteractionEvent();
+    return macro.EVENT_ABORT;
   };
 
   // --------------------------------------------------------------------------
--- a/Sources/Widgets/Widgets3D/DistanceWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/DistanceWidget/behavior.js
@@ -4,7 +4,7 @@ const MAX_POINTS = 2;
 
 export default function widgetBehavior(publicAPI, model) {
   model.classHierarchy.push('vtkDistanceWidgetProp');
-  let isDragging = null;
+  model._isDragging = false;
 
   // --------------------------------------------------------------------------
   // Display 2D
@@ -53,8 +53,8 @@ export default function widgetBehavior(publicAPI, model) {
       newHandle.setColor(moveHandle.getColor());
       newHandle.setScale1(moveHandle.getScale1());
       newHandle.setManipulator(manipulator);
-    } else {
-      isDragging = true;
+    } else if (model.dragable) {
+      model._isDragging = true;
       model._apiSpecificRenderWindow.setCursor('grabbing');
       model._interactor.requestAnimation(publicAPI);
     }
@@ -68,13 +68,6 @@ export default function widgetBehavior(publicAPI, model) {
   // --------------------------------------------------------------------------
 
   publicAPI.handleMouseMove = (callData) => {
-    if (
-      model.hasFocus &&
-      model.widgetState.getHandleList().length === MAX_POINTS
-    ) {
-      publicAPI.loseFocus();
-      return macro.VOID;
-    }
     const manipulator =
       model.activeState?.getManipulator?.() ?? model.manipulator;
     if (
@@ -92,7 +85,9 @@ export default function widgetBehavior(publicAPI, model) {
 
       if (
         worldCoords.length &&
-        (model.activeState === model.widgetState.getMoveHandle() || isDragging)
+        (model.activeState === model.widgetState.getMoveHandle() ||
+          model._isDragging) &&
+        model.activeState.setOrigin // e.g. the line is pickable but not draggable
       ) {
         model.activeState.setOrigin(worldCoords);
         publicAPI.invokeInteractionEvent();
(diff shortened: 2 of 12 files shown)


Sources/Widgets/Widgets3D/AngleWidget/behavior.js
```diff
@@ -92,7 +92,9 @@ export default function widgetBehavior(publicAPI, model) {
 
       if (
         worldCoords.length &&
-        (model.activeState === model.widgetState.getMoveHandle() || isDragging)
+        (model.activeState === model.widgetState.getMoveHandle() ||
+          isDragging) &&
+        model.activeState.setOrigin
```
finetjul, line 97: You might want to explain when setOrigin does not exist, e.g. `model.activeState.setOrigin // e.g. the line is pickable/activable but not draggable`

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
```diff
@@ -60,7 +60,7 @@ export default function widgetBehavior(publicAPI, model) {
       newHandle.setColor(moveHandle.getColor());
       newHandle.setScale1(moveHandle.getScale1());
       newHandle.setManipulator(manipulator);
-    } else {
+    } else if (model.dragable) {
```
finetjul, line 63: if not dragable, you still call `StartInteractionEvent` but you won't call `EndInteractionEvent`

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -354,6 +355,7 @@ export default function widgetBehavior(publicAPI, model) {
       model.activeState.activate();
       model._interactor.requestAnimation(publicAPI);
       publicAPI.invokeStartInteractionEvent();
+      window.moveHandle = model.widgetState.getMoveHandle();
```
finetjul, line 358: to not commit

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
tomsuchel: I agree we need to fix that, but this suggest we should change `handleLeftButtonRelease` method.
I found weird we always return `macro.VOID` or `macro.EVENT_ABORT` in `handleLeftButtonPress` and `handleMouseMove` but not in this one.

I suggest to standardize the `handleLeftButtonRelease` method using the template below (to adapt depending of the widgets)
```js
publicAPI.handleLeftButtonRelease = () => {
  // Discard useless events
  if (
    !model.activeState ||
    !model.activeState.getActive() ||
    !model.pickable
  ) {
    return macro.VOID;
  }
  
  if (model.pickable && model.dragable && isDragging) {
    // Finish drag
    model._apiSpecificRenderWindow.setCursor('pointer');
    model.widgetState.deactivate();
    model._interactor.cancelAnimation(publicAPI);
    isDragging = false;
  } else if (model.activeState !== model.widgetState.getMoveHandle()) {
    model.widgetState.deactivate();
  }
  
  if (
    (model.activeState && !model.activeState.getActive())
  ) {
    model._widgetManager.enablePicking();
    model._interactor.render();
  }
  
  publicAPI.invokeEndInteractionEvent();
  return macro.EVENT_ABORT;
}
```
WDYT?

finetjul: I wouldn't even trigger Start/End"Interaction"event if the line is not draggable.

tomsuchel: But if the widget is pickable but not dragable? We may want to trigger a start/end event so that the user can know when the widget is "clicked/selected"?

finetjul: fair enough, that's good with me then.

Sources/Widgets/Widgets3D/ImplicitPlaneWidget/index.js
```diff
@@ -43,18 +43,23 @@ function widgetBehavior(publicAPI, model) {
...
+      model._interactor.requestAnimation(publicAPI);
+    }
+
+    publicAPI.invokeStartInteractionEvent();
     return macro.EVENT_ABORT;
   };
 
   publicAPI.handleMouseMove = (callData) => {
-    if (isDragging && model.pickable) {
+    if (isDragging && model.pickable && model.dragable) {
```
finetjul, line 62: isn't `if (isDragging)` sufficient ?

Sources/Widgets/Widgets3D/LabelWidget/behavior.js
```diff
@@ -2,8 +2,7 @@ import macro from 'vtk.js/Sources/macros';
 
 export default function widgetBehavior(publicAPI, model) {
   model.classHierarchy.push('vtkLabelWidgetProp');
-
-  model.isDragging = null;
+  let isDragging = null;
```
finetjul, line 5: are you sure there is no derived widget that could need to access `isDragging` ?

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -334,12 +335,12 @@ export default function widgetBehavior(publicAPI, model) {
     }
 
     if (
-      model.isDragging === false &&
+      isDragging === false &&
```
finetjul, line 338: do you know why an exact test on false when `isDragging` is initialized to `null`. This may be something to provide comment for

Sources/Widgets/Widgets3D/SplineWidget/behavior.js
```diff
@@ -215,21 +216,21 @@ export default function widgetBehavior(publicAPI, model) {
...
-      model.isDragging = true;
+      model.freeHand = publicAPI.getAllowFreehand() && !isDragging;
+    } else if (model.dragable) {
+      isDragging = true;
       model._apiSpecificRenderWindow.setCursor('grabbing');
       model._interactor.requestAnimation(publicAPI);
-      publicAPI.invokeStartInteractionEvent();
     }
 
+    publicAPI.invokeStartInteractionEvent();
```
finetjul, line 233: be careful here, you may call invokeStartInteractionEvent twice

Sources/Widgets/Widgets3D/LabelWidget/behavior.js
tomsuchel: Indeed, derived widgets may need to access `isDragging` while overriding some methods. 
I have moved it to the model as a protected attribute (`model._isDragging`).

Sources/Widgets/Widgets3D/LineWidget/behavior.js
tomsuchel: The line in the condition (`publicAPI.rotateHandlesToFaceCamera();`) should be called each time the camera moves.
Even if this currently works (because we set `model.isDragging` to false after placing the first handle), I think it would be safer to use `!model.isDragging` instead of an exact test.

Sources/Widgets/Widgets3D/ImplicitPlaneWidget/index.js
tomsuchel: Yes indeed, I've removed `model.pickable && model.dragable` to make code cleaner

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
```diff
@@ -18,14 +18,15 @@ const EPSILON = 1e-6;
...
+  model._isDragging = null;
 
   model.keysDown = {};
 
   const superClass = { ...publicAPI };
 
   // --------------------------------------------------------------------------
   // Display 2D
-  // --------------------------------------------------------------------------
+  // -----------------------------------  ---------------------------------------
```
floryst, line 29: Found a randomly placed space that can be deleted.

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
```diff
@@ -5,7 +5,7 @@ const MAX_POINTS = 3;
 
 export default function widgetBehavior(publicAPI, model) {
   model.classHierarchy.push('vtkAngleWidgetProp');
-  let isDragging = null;
+  model._isDragging = null;
```
floryst, line 8: Instead of typing this as `null`, I would suggest typing it as `boolean` and having a default assignment of `false`. The same goes with all other instances of `_isDragging` instantiation,

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
```diff
@@ -75,6 +75,13 @@ export default function widgetBehavior(publicAPI, model) {
   // --------------------------------------------------------------------------
 
   publicAPI.handleMouseMove = (callData) => {
+    if (
+      model.hasFocus &&
+      model.widgetState.getHandleList().length === MAX_POINTS
+    ) {
+      publicAPI.loseFocus();
```
finetjul, line 82: I find it odd to loose focus when moving the mouse.
Can you explain why it makes sense to loose focus at that time ?

tomsuchel: I admit it was mainly to look like the DistanceWidget, which worked like that for 3 years.
But thinking about it, I would find it better to place this code at the beginning of the `handleLeftButtonRelease` method (so we place point + invoke startInteractionEvent during press, and lose focus + invoke endInteractionEvent during release).  WDYT?

finetjul: ok to have it in handleLeftButtonRelease. 

I do not have an opinion wether it is at the beginning or at the end.

Sources/Widgets/Widgets3D/DistanceWidget/behavior.js
```diff
@@ -108,11 +103,26 @@ export default function widgetBehavior(publicAPI, model) {
...
+      !model.pickable
+    ) {
+      return macro.VOID;
+    }
+    if (
+      model.hasFocus &&
+      model.widgetState.getHandleList().length === MAX_POINTS
+    ) {
+      publicAPI.loseFocus();
+      return macro.VOID;
```
finetjul, line 118: do not return yet

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -251,7 +252,7 @@ export default function widgetBehavior(publicAPI, model) {
   // --------------------------------------------------------------------------
 
   publicAPI.handleMouseMove = (callData) => {
-    if (model.hasFocus && publicAPI.isPlaced() && !model.isDragging) {
+    if (model.hasFocus && publicAPI.isPlaced() && !model._isDragging) {
       publicAPI.loseFocus();
       return macro.VOID;
     }
```
finetjul, line 258: to move in release event ?

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
```diff
@@ -110,11 +112,26 @@ export default function widgetBehavior(publicAPI, model) {
...
+      !model.pickable
+    ) {
+      return macro.VOID;
+    }
+    if (
+      model.hasFocus &&
+      model.widgetState.getHandleList().length === MAX_POINTS
+    ) {
+      publicAPI.loseFocus();
+      return macro.VOID;
```
finetjul, line 127: do not return yet

PR #2463, opened 2022-06-09
feat(volume shadow): add volume shadow support for all lighting conditions

JiayiXuu:
### Context
Pull request for adding volume shadow support as part of realistic rendering module, also referred to as secondary shadow ray/secondary scatter. This pull request aligns with recent changes in VTK, but adds optimizations and two extra parameters to further  speed up rendering. Four parameters are included: 
 - `volumetricScatteringBlending`: 
     * range [0.0,1.0]
     * 0.0: surface shadow
     * 1.0: volume/global shadow
 - `globalIlluminationReach`: 
     * range [0.0,1.0]
     * 0.0: maximum volume shadow ray length is equal to 1 unit of sampling distance
     * 1.0: maximum volume shadow ray length is equal to the largest possible value of sqrt(3), i.e. no restriction on shadow ray
 - `volumeShadowSamplingDistFactor`: 
     * range [1.0, Infinity]
     * modifies sampling distance of shadow ray (`SamplingDistanceISVS`) based on sampling distance of primary ray casting (`SamplingDistanceIS`), primarily to increase speed
     * 1.0: `SamplingDistanceISVS` = 1.0 * `SamplingDistanceIS`
     * n: `SamplingDistanceISVS` = n * `SamplingDistanceIS`
 - `anisotropy`: 
     * range (-1.0,1.0)
     * -0.99: light scatters backward
     * 0.99: light scatters forward
     * 0.0: light scatter is isotropic (uniform in all direction)

### Results
**0: Overview**

overview_go_compressed_cropped
_Overview for volume with gradient opacity enabled - this gif is compressed with 2X speed and frame drop_

overview_tf_cropped
_Overview for volume without gradient opacity - this gif has frame drop_

**1: volumetricScatteringBlending**

blending

Surface shadow seems to dominate the rendering result when it is on. So `volumetricScatteringBlending= 1.0` is drastically different from `volumetricScatteringBlending= 0.99`, but `volumetricScatteringBlending = 0.99` is pretty much similar to `volumetricScatteringBlending= 0.0`. I used a biased function so volume shadow is more prominent, though it didn't have a big effect. 

image

**2: volumeShadowSamplingDistFactor**
Shadow ray sampling distance is either equal to or larger than primary ray sampling distance.

samplingDist

[...]

PR #2470, opened 2022-06-12
docs(volume shadow): add example light and shadow ray example

JiayiXuu:
### Context
This PR adds an example for the recent volume shadow PR #2463. User can test the parameters with headsq.vti data. Simple instructions are embedded in index.md.
### Results

image

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Windows 10


martinken:
The example seems fine +1, but the behavior seems off. Some things I noted

- when I reset the camera the volume does not go to the middle of the view, it is like there is an offset somewhere in the matrices

- as you said, the surface to volume slider seems to behave really non-linearly which makes it twitchy to use

- some of the new methods are missing form the index.d.ts, is there a .md file or something explaining what they do? 

- the anisotropy seems really odd (or maybe I'm not getting it) It seems to mainly make the image lighter or darker to the point of black and full white except right near 0.0 it does something different. It seems like an odd adjustment so my understanding or something must be off.

Note there is a typo in master "publicAPI.setVlumeShadowSamplingDistFactor = (vsdf) => {" and it has a generated set method too.

martinken:
Ahh the off center volume is due to the light sphere.  You can call setUseBounds(false) on it so that it doesn't mess up the centering of the volume if you want to.

martinken (approved):

JiayiXuu:
> * as you said, the surface to volume slider seems to behave really non-linearly which makes it twitchy to use
   There's another stochastic model that uses `random()` to pick between surface/volume rendering each sample point, which makes blending a little bit smoother.
   But that reduces rendering speed and introduce more artifacts (associated with `random()` generator). I'll make another branch in the future with the stochastic method to compare.

> * the anisotropy seems really odd (or maybe I'm not getting it) It seems to mainly make the image lighter or darker to the point of black and full white except right near 0.0 it does something different. It seems like an odd adjustment so my understanding or something must be off.
   It might be due to thresholding, when `anisotropy` < 0.01, it becomes 0.0. Also the slider step is 0.1 so the transition looks odd.

PR #2475, opened 2022-06-14
fix(linewidget): moveHandle not visible and last point of polyline is ignored

tomsuchel:
Method `setMoveHandleVisibility` was called when the moveHandle origin was still `null`, causing it
to be filtered out (see ArrowHandleRepresentation:getRepresentationStates) and not rendered. 
Fix removing method `setMoveHandleVisibility` and using `updateHandleVisibility` to update moveHandle visibility.

---

When rendering a PolyLineRepresentation, the last point was ignored if not visible. On a LineWidget,
this was causing the line to disappear if the second point was hidden. 
Fix this by setting the origin of the moveHandle is set to null once the widget is placed.

### PR and Code Checklist

### DIFF
--- a/Sources/Widgets/Widgets3D/LineWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/LineWidget/behavior.js
@@ -278,6 +278,7 @@ export default function widgetBehavior(publicAPI, model) {
       ) {
         if (model.activeState.setOrigin) {
           model.activeState.setOrigin(worldCoords);
+          publicAPI.setMoveHandleVisibility(!publicAPI.isPlaced());
         } else {
           // Dragging line
           publicAPI
--- a/Sources/Widgets/Widgets3D/LineWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/LineWidget/behavior.js
@@ -26,6 +26,23 @@ export default function widgetBehavior(publicAPI, model) {
   publicAPI.getHandle = (handleIndex) =>
     model.widgetState[handleGetters[handleIndex]]();
 
+  /**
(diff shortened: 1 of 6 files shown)


tomsuchel:
@finetjul can you review? thanks

tomsuchel:
@finetjul can you review? thanks

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -281,6 +281,7 @@ export default function widgetBehavior(publicAPI, model) {
       ) {
         if (model.activeState.setOrigin) {
           model.activeState.setOrigin(worldCoords);
+          publicAPI.setMoveHandleVisibility(!publicAPI.isPlaced());
```
finetjul, line 284: it does not make sense to call setMoveHandleVisibility if the activeState is not a moveHandle.

I suggest instead to do: `updateHandleVisibility(getHandleIndex(model.activeState));`

(I'm not sure of the "render()" call in `updateHandleVisiblity()` though...

tomsuchel: I'm not sure to understand...

`updateHandleVisibility` does not update the `visible` mixin, so we would still have to use 
```js
moveHandle.setVisible(!publicAPI.isPlaced())
``` 
for this to work.

Also, the `getHandleIndex` function does not exists and I'm not sure how to compute the index in a smart way (but I can still do it with a `switch`)

Wouldn't it be easier to add a condition like this?
```js
if (model.activeState === model.widgetState.getMoveHandle()) {
    publicAPI.setMoveHandleVisibility(!publicAPI.isPlaced());
}
```

PR #2476, opened 2022-06-14
feat(linewidget): add reset function to LineWidget

tomsuchel:
The LineWidget was missing a `publicAPI.reset()` function, that can be usefull to reset the widget
without deleting it.


floryst:
This looks like there will be a slight dependence on #2466, since I think you'd want to also perform a `model._isDragging = false` on the off-chance that someone requests a widget reset while they are dragging a handle. (UX-wise, not sure how this would happen.)

floryst:
This looks like there will be a slight dependence on #2466, since I think you'd want to also perform a `model._isDragging = false` on the off-chance that someone requests a widget reset while they are dragging a handle. (UX-wise, not sure how this would happen.)

Sources/Widgets/Widgets3D/LineWidget/behavior.js
```diff
@@ -372,4 +372,17 @@ export default function widgetBehavior(publicAPI, model) {
...
+    model.widgetState.getMoveHandle().deactivate();
+
+    model.widgetState.getHandle1().setOrigin(null);
+    model.widgetState.getHandle2().setOrigin(null);
+    model.widgetState.getMoveHandle().setOrigin(null);
+    model.widgetState.getText().setOrigin(null);
+    model.widgetState.getText().setText('');
+
+    model.activeState = null;
+  };
```
finetjul, line 387: what about a more generic way by calling `generateState()` again ?

floryst: I like this idea from a high-level point of view. Problem is, I don't think we can just do `model.widgetState = generateState()`, since users may have kept a reference to the widget state externally (e.g. `model.widgetState.onModified(...)`.

finetjul: I guess we could add a `deepCopy()` convenient function in `Widgets/Core/WidgetState/index.js`.

floryst: I think we should work on that idea outside of this PR. `deepCopy()` does sound like a good idea, but I wonder if its potential complexity outweighs manually performing a reset.

finetjul: Good point

PR #2477, opened 2022-06-14
perf: cache publicAPI.getHardwareMaximumLineWidth result

FezVrasta:
### Context
Calling the webgl.getParameter is relatively slow, and if called many times can result in a bottleneck.

### Results
From my understanding the result of this call will not vary during a single session so it's safe to cache it.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/OpenGL/RenderWindow/index.js
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.js
@@ -732,9 +732,18 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
     });
   };
 
+  let hardwareMaximumLineWidth;
   publicAPI.getHardwareMaximumLineWidth = () => {
+    // We cache the result of this function because `getParameter` is slow
+    if (hardwareMaximumLineWidth != null) {
+      return hardwareMaximumLineWidth;
+    }
+
     const gl = publicAPI.get3DContext();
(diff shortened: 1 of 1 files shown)


Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -733,8 +733,16 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
   publicAPI.getHardwareMaximumLineWidth = () => {
+    // We cache the result of this function because `getParameter` is slow
+    if (model.hardwareMaximumLineWidth != null) {
+      return model.hardwareMaximumLineWidth;
+    }
+
     const gl = publicAPI.get3DContext();
     const lineWidthRange = gl.getParameter(gl.ALIASED_LINE_WIDTH_RANGE);
+
+    model.hardwareMaximumLineWidth = lineWidthRange[1];
```
finetjul, line 744: does not seem to make sense to "cache" it into model. Can you please use simple closure to cache it ?

FezVrasta: done!

finetjul:
LGTM

PR #2480, opened 2022-06-15
fix: Address incorrect opacity values when displaying RGB images in ImageMapper

swederik:
### Context
For large 2D image (i.e. bigger than the max 3D texture size) support in Cornerstone, we started using the ImageSlice actor and ImageMapper in our stack viewports (https://github.com/cornerstonejs/cornerstone3D-beta/pull/127). Almost everything is working fine, but we noticed that our RGB images were rendering at the wrong opacity. It looks like three component images are being sampled and the value of the alpha channel (scaled by cscale and cshift, based on the color transfer function) are producing the wrong opacity values in the final rendering.

Here's an example of what we are seeing (note the transparent image on the bottom compared to the same image produced with the volumeMapper above):
Screenshot 2022-06-15 at 10 04 03

### Changes

No docs changes necessary.

This tiny PR switches from `tColor.a` to `opacity` in images with three dependent components (i.e. RGB). The point here is that it looks like tColor is being sampled from the texture, which in our case is three components, but it's producing a vec4. The alpha channel value is then being mutated by cshift and cscale. As far as I understand it, the fourth component should just be fixed based on the opacity at the actor level.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: latest master (v24.18.5) 
  - **OS**: Mac OSX 12.13.1 
  - **Browser**: Chrome


martinken:
Don't we need to make different 3 and 4 component paths? Or 3 and default?  With a true RGBA image we do want the A to come from the image right?

swederik:
Yep that makes sense, I'll do that then. Thanks!

martinken (approved):

PR #2481, opened 2022-06-15
feat(Rendering): add coordinate systems for props

martinken:
This MR adds the noption of coordinate systems to vtkProp.
It currently adds WORLD as the default and DISPLAY as a
second option. DISPLAY coordinates are in pixels with 0,0 being
the lower left corner of the viewport and Z ranging from -1
near to 1 far. These coordinate systems are intended to be
implemented in the GPU.

This MR adds Actor2D and PolyDataMapper2D to WebGPU taking
advantage of the new coordinate systems.

This MR updates the ScalarBarActor to produce data in
DISPLAY coordinates. The old code used world coordinates
which meant that every time the camera changed the scalar
bar actor had to recompute it's data which is very expensive
resulting in hundreds of VBO's being created every second.
By using DISPLAY coordinate the data does not change with
camera changes, only with viewport changes or changes to
the source data/scalar bar actor itself which is far less
common.

With this MR the scalar bar actor functions on both WebGPU and
OpenGL and the Actor2D example works on both WebGPU and OpenGL.

### DIFF
--- a/Sources/Rendering/Core/Prop/index.d.ts
+++ b/Sources/Rendering/Core/Prop/index.d.ts
@@ -20,8 +20,8 @@ export interface IPropInitialValues {
 export interface vtkProp extends vtkObject {
 
     /**
-     * 
-     * @param estimatedRenderTime 
+     *
+     * @param estimatedRenderTime
      */
     addEstimatedRenderTime(estimatedRenderTime: number): void;
 
@@ -83,32 +83,32 @@ export interface vtkProp extends vtkObject {
     getNestedPickable(): boolean;
 
     /**
-     * Return the mtime of anything that would cause the rendered image to appear differently. 
-     * Usually this involves checking the mtime of the prop plus anything else it depends on such as properties, 
+     * Return the mtime of anything that would cause the rendered image to appear differently.
+     * Usually this involves checking the mtime of the prop plus anything else it depends on such as properties,
      * textures etc.
      */
     getRedrawMTime(): number
 
     /**
-     * 
+     *
      */
     getRendertimemultiplier(): number;
 
     /**
      * The value is returned in seconds. For simple geometry the accuracy may not be great
-     * due to buffering. For ray casting, which is already multi-resolution, 
-     * the current resolution of the image is factored into the time. We need the viewport 
+     * due to buffering. For ray casting, which is already multi-resolution,
+     * the current resolution of the image is factored into the time. We need the viewport
      * for viewing parameters that affect timing. The no-arguments version simply returns the value of the variable with no estimation.
      */
     getEstimatedRenderTime(): number;
 
     /**
-     * 
+     *
      */
     getAllocatedRenderTime(): number;
 
     /**
-     * 
+     *
      */
     getNestedProps(): any;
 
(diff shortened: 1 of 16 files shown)


Sources/Rendering/Core/Prop/index.d.ts
```diff
@@ -230,11 +230,25 @@ export interface vtkProp extends vtkObject {
 
     /**
      * This is used for culling and is a number between 0 and 1. It is used to create the allocated render time value.
-     * @param {Number} renderTimeMultiplier 
+     * @param {Number} renderTimeMultiplier
      */
     setRenderTimeMultiplier(renderTimeMultiplier: number): boolean;
 
     /**
+     * Indicate that this prop's data should be in world coordinates.
```
finetjul, line 238: what about the "userMatrix" ? in what coordinate system is it ?

martinken: who knows, only the user. But IIRC the user matrix just gets multiplied into everything else (or it replaces everything else, I forget which) But when it gets to the mapper the results of that will be interpreted by the mapper as WORLD or DISPLAY.

finetjul: Gotcha, so it might be good to precise in Prop3D/index.d.ts that `origin`, `position`, `orientation` and `userMatrix` are expressed in this coordinate system (world or display).

Sources/Rendering/Core/Prop/index.d.ts
```diff
@@ -37,7 +38,12 @@ export interface vtkProp extends vtkObject {
      */
     getActors2D(): vtkActor2D[];
 
-    /**
+	/**
+	 * Get the coordinate system this prop is defined in.
+	 */
+     getCoordinateSystem(): CoordinateSystem;
```
finetjul, line 44: bad indent

Sources/Rendering/Core/Prop/index.d.ts
```diff
@@ -169,24 +175,52 @@ export interface vtkProp extends vtkObject {
...
 
+	/**
+     * Set the coordinate system that this prop's data should be in.
+     * Once the prop has applied any modifiers such as position, orientation
+     * userMatrix the resulting values will be treated as in the specified
+     * coordinate system.
+     * Not all mappers support all coordinate systems.
+	 * @param {CoordinateSystem} coordinateSystem
+	 */
+     setCoordinateSystem(coordinateSystem: CoordinateSystem): void;
```
finetjul, line 191: bad indent

Sources/Rendering/Core/Prop/index.d.ts
```diff
@@ -230,11 +264,11 @@ export interface vtkProp extends vtkObject {
 
     /**
      * This is used for culling and is a number between 0 and 1. It is used to create the allocated render time value.
-     * @param {Number} renderTimeMultiplier 
+     * @param {Number} renderTimeMultiplier
      */
     setRenderTimeMultiplier(renderTimeMultiplier: number): boolean;
 
-    /**
+     /**
```
finetjul, line 271: bad indent

Sources/Rendering/WebGPU/PolyDataMapper2D/index.js
```diff
@@ -0,0 +1,107 @@
...
+  };
+}
+
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {
+  primitives: null,
+};
```
finetjul, line 79:
```
function defaultValues(initialValues) {
  return {primitives: [], ...initialValues};
}
...
```

PR #2483, opened 2022-06-16
fix(ResliceCursorWidget): Use correct moving factor

laurennlam:
### Changes
When computing the new center of the reslice cursor widget during mouse wheel, we previously applied twice the sign of the dirproj. Now it's applied only once.

@finetjul


finetjul:
LGTM

PR #2484, opened 2022-06-16
perf: drawArrays getParameter cache

FezVrasta:
### Context
Same as my other PR but for the drawArrays function.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

### Funding

### DIFF
--- a/Sources/Rendering/OpenGL/Actor/index.js
+++ b/Sources/Rendering/OpenGL/Actor/index.js
@@ -132,7 +132,7 @@ function vtkOpenGLActor(publicAPI, model) {
 
   publicAPI.opaquePass = (prepass, renderPass) => {
     if (prepass) {
-      model.context.depthMask(true);
+      model._openGLRenderWindow.setDepthWriteMask(true);
       publicAPI.activateTextures();
     } else if (model.activeTextures) {
       for (let index = 0; index < model.activeTextures.length; index++) {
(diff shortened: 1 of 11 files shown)


finetjul:
This does not seem cachable, I feel that DEPTH_WRITEMASK can have a varying value depending on the calling context.

@martinken

martinken:
Would require that all calls to get it go through that method for it to be safe in vtk.js  Otherwise some other class in vtk.js could change the state and the helper would not know it I think.

floryst:
@martinken @finetjul I've updated this PR to properly cache all calls to `gl.depthMask()`.

Aside: I personally think using a Proxy over the WebGLContext object would be better for implementing the cache. Since I think we still technically have IE11 support (at least until we drop webgl1), this is implemented as an added API to vtkOpenGLRenderWindow.

finetjul:
Please consider our discussions on #2146 and more specifically [that](https://github.com/Kitware/vtk-js/pull/2146#issuecomment-956466172) comment.

It seems that Option 1 does not work (due to performance issues), so we have to do Option 2, i.e. proxy WebGLContext by caching all calls to GL.

Who volunteers ? @FezVrasta ?

floryst:
I think we can reasonably drop IE11 at this point, so let's go with the Proxy approach.

floryst:
Closing this in favor of #2507, since I can't push to Federico's branch for some reason. (It worked before...)

PR #2490, opened 2022-06-24
feat(multiple scatter): support experimental multiple scatter option

JiayiXuu:
### Context
This PR adds multiple scatter option as an alternative to simple ray casting. If this option is turned on, gradient-based and volumetric shading are turned off by default due to computational limit. Multiple scatter accentuate the translucency of the volume and captures a little bit of shadow. Three parameters are added: 
- `multipleScatter`: toggles scatter on and off
- 'multipleScatterSamplingDistFactor': increase or decrease sample distance for higher order scatter ray
- 'sphericalSampleNumber': number of sampled in-scatters on uniform sphere

Parameters that are not exposed: 
- Amount of scatter sampled based on light direction versus phase function. Hard coded value is a half-half split. Gradient direction is not used sample scattered light to avoid gradient computation.
- Number of samples for evaluating transmittance at each sample location. Hard coded value is 50 samples to determine shade.

### Results
Because scattering contribution is added on top of regular rendering it increases the overall brightness level. The scattering effect is not obvious for solid, opaque volume. 

image

### Changes

- `vtkVolumeFS.glsl`: added path integration algorithm for multiple scatter
- `opengl/VolumeMapper`: added parameters and uniforms
- `core/VolumMapper`: expose get/set of new parameters

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Windows 10

### DIFF
--- a/Sources/Rendering/Core/VolumeMapper/index.js
+++ b/Sources/Rendering/Core/VolumeMapper/index.js
@@ -116,6 +116,10 @@ const DEFAULT_VALUES = {
   globalIlluminationReach: 0.0,
   volumeShadowSamplingDistFactor: 5.0,
   anisotropy: 0.0,
+  // multiple scatter parameters
+  sphericalSampleNumber: 5.0,
+  multipleScatterSamplingDistFactor: 1.0,
+  multipleScatter: false,
 };
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


finetjul:
It would be great to add a live example as well as a unit test to make sure it does not "break" with new commits.

JiayiXuu:
> It would be great to add a live example as well as a unit test to make sure it does not "break" with new commits.

Yep, I'll push a new example/test when this feature is better curated.

PR #2493, opened 2022-06-28
fix(Shader): fix condition statement for gradient-based shadow

JiayiXuu:
### Context
Shader doesn't compile when `VolumetricScatteringBlending` is set to 1.0 due to an error in conditional statement. 

### Results
Fixed issue

### Changes
Changed #ifdef statement for gradient based lighting function declaration in vtkVolumeFS.glsl

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Windows 10


JiayiXuu:
@floryst could you review the fix for this minor bug? Thank you!

floryst:
LGTM. Does `defined()` not work in an if statement?

JiayiXuu:
> LGTM. Does `defined()` not work in an if statement?

Thanks! The syntax is correct, it's just `#ifdef SurfaceShadowOn` need to enclose the entire `#if vtkLightComplexity < 3` statement, a slight logic difference.

floryst:
Ah, I see now. There's an else statement hidden in the collapsed code block. Thanks for the clarification!

PR #2498, opened 2022-06-29
fix(AbstractWidgetFactory): avoid race condition

floryst:
### Context

The setTimeout can cause a race condition, where a widget can be
instantiated and subsequently deleted, causing an error.

This changes the way widgets are to be written. Rather than assigning
behavior and widgetState members directly, we pass them into the
vtkAbstractWidgetFactory constructor.

### Results

- Removes a race condition in vtkAbstractWidgetFactory

### Changes

### PR and Code Checklist

### Testing
- [ ] Tested existing widgets to ensure functionality is not broken

### DIFF
--- a/Sources/Widgets/Widgets3D/AngleWidget/index.js
+++ b/Sources/Widgets/Widgets3D/AngleWidget/index.js
@@ -29,8 +29,6 @@ function vtkAngleWidget(publicAPI, model) {
     'defaultScale',
     'scaleInPixels',
   ];
-  model.behavior = widgetBehavior;
-  model.widgetState = stateGenerator();
 
   publicAPI.getRepresentationsForViewType = (viewType) => {
     switch (viewType) {
@@ -102,14 +100,17 @@ function vtkAngleWidget(publicAPI, model) {
 
 // ----------------------------------------------------------------------------
 
-const DEFAULT_VALUES = {
+const defaultValues = (initialValues) => ({
   // manipulator: null,
-};
+  behavior: widgetBehavior,
+  widgetState: stateGenerator(),
+  ...initialValues,
+});
 
 // ----------------------------------------------------------------------------
 
 export function extend(publicAPI, model, initialValues = {}) {
-  Object.assign(model, DEFAULT_VALUES, initialValues);
+  Object.assign(model, defaultValues(initialValues));
 
   vtkAbstractWidgetFactory.extend(publicAPI, model, initialValues);
   macro.setGet(publicAPI, model, ['manipulator']);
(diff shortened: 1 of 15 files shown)


floryst:
FYI @finetjul I was hoping to put this in the v25 major release, but forgot about it until now. It might be "close enough" that we can release this as a patch on v25, since this technically represents a breaking change in how widgets are instantiated.

Sources/Widgets/Widgets3D/AngleWidget/index.js
```diff
@@ -111,7 +109,11 @@ const DEFAULT_VALUES = {
 export function extend(publicAPI, model, initialValues = {}) {
   Object.assign(model, DEFAULT_VALUES, initialValues);
 
-  vtkAbstractWidgetFactory.extend(publicAPI, model, initialValues);
+  vtkAbstractWidgetFactory.extend(publicAPI, model, {
+    ...initialValues,
+    behavior: widgetBehavior,
```
finetjul, line 114: how about letting initialValues potentially overwrite behavior and widgetState ?
Move behavior and widgetSate in a `defaultValues()` function ?

finetjul:
LGTM

@tomsuchel WDYT ?

tomsuchel:
> LGTM
> 
> @tomsuchel WDYT ?

LGTM

floryst:
Update: users can now override the behavior and state objects in the constructor. Instead of passing them directly into the AbstractWidgetFactory, I am assigning it to the model directly, since the AbstractWidgetFactory will pick them up anyways. This model-setting approach is not as explicit as passing the objects to the Factory constructor, but I think it's okay.

Sources/Widgets/Widgets3D/ImageCroppingWidget/index.js
```diff
@@ -168,16 +166,19 @@ function vtkImageCroppingWidget(publicAPI, model) {
...
 
-const DEFAULT_VALUES = {
-  // cornerManipulator: null,
-  // edgeManipulator: null,
-  // faceManipulator: null
-};
+const defaultValues = (initialValues) => ({
+  cornerManipulator: null,
+  edgeManipulator: null,
+  faceManipulator: null,
```
finetjul, line 172: when they are set to null, we typically comment then with @jourdain

finetjul (approved):
LGTM

Sources/Widgets/Widgets3D/ImageCroppingWidget/index.js
floryst: That's fine, so long as we don't strictly check these values for null.

finetjul:
LGTM

PR #2502, opened 2022-07-05
fix(widgets): remove nullish coalescing operator

tomsuchel:
Nullish coalescing operator (`??`) are a recent feature and are sometimes not recognized by old babel
versions. As a precaution, this commit removes these operators as they are not necessary and can be
easily replaced by logical OR (`||`).

### PR and Code Checklist

### DIFF
--- a/Sources/Widgets/Widgets3D/AngleWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/AngleWidget/behavior.js
@@ -42,7 +42,7 @@ export default function widgetBehavior(publicAPI, model) {
     picker.initializePickList();
     picker.setPickList(publicAPI.getNestedProps());
     const manipulator =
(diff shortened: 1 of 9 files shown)


floryst:
Is this for folks who are using `vtk.js`, and not `@kitware/vtk.js`? Also, it's my understanding that optional chaining was introduced roughly around the time that nullish coalescing was introduced, so unless we remove optional chaining as well, I would propose that we require using the es2020 preset (or an updated preset-env) from babel. (The docs need updating for this.)

tomsuchel:
After some further research, I discovered that the problem I was facing was exactly this one: https://github.com/PaulLeCam/react-leaflet/issues/883

Indeed, this happen only with the UMD build (`vtk.js`) when using webpack <= 4. 

This PR can eventually be merged to prevent other users from encountering this bug, but we will probably use this operator one day, so I think we can close it.

Issue #2499, opened 2022-06-29
Potential typo in Utilities/ParaView/export-scene-macro.py


dasmy:
I assume, this file is deprecated. However, while trying to learn from the code, I came across the line https://github.com/Kitware/vtk-js/blob/6ae17f80c8496191d5439eaf746eac772a40450b/Utilities/ParaView/export-scene-macro.py#L244 .

Shouldn't it be `cd.GetNumberOfArrays()` instead of `pd.GetNumberOfArrays()`?

patrickoleary:
Shouldn't it be cd.GetNumberOfArrays() instead of pd.GetNumberOfArrays()?

dasmy:
> Shouldn't it be cd.GetNumberOfArrays() instead of pd.GetNumberOfArrays()?
> […](#)

Opps, fixed my typo. Thanks :-)

finetjul:
Hi dasmy, I think you are right. can you please make PR ?

dasmy:
Hi @finetjul,
As I needed some more fixes / additions to the macro (see https://discourse.paraview.org/t/trouble-exporting-scene-to-vtkjs/9829/10 for the full story), I did several additional modifications, see https://github.com/dasmy/export-scene-macro .

Does it make sense to discuss them and create a composite PR or should I for now only address the change mentioned above?

DavidBerger98:
Hi @dasmy,

Thank you for opening this issue and suggested changes.

You can create one PR with your changes from the discussion and the fixed typo. Linking the discourse thread and this issue in the PR should be enough.

I am pinging @jourdain in case he wants to add anything to the discussion.

Best,

David

jourdain:
I'm good pushing that macro forward. In general. It is a "deprecated path" now that we have a more mainstream C++ one. But it is definitely an easier path for the community to contribute and extend it to better fit their needs.

PR #2503, opened 2022-07-05
chore(export_scene_macro): fixes and improvements

dasmy:
### Context
Although ParaView supports exporting a scene to `*.vtkjs`, the `export_scene_macro` is still useful for rapid customization, prototyping, and educational purposes.

One of the examples is https://discourse.paraview.org/t/trouble-exporting-scene-to-vtkjs/9829/10, where the native export was incomplete, but fixing it in C++ was infeasible for me. This is why I looked at the macro, update
[SceneExplorer_node-18.x.html.zip](https://github.com/Kitware/vtk-js/files/9050061/SceneExplorer_node-18.x.html.zip)
d it to support Python 3, fixed some issues and added the features that were missing for my use case.
This PR also fixes https://github.com/Kitware/vtk-js/issues/2499 which I found while looking at the code.

### Results
* The macro runs again (didn't any more, tested in ParaView 5.10.1 on macOS)
* The macro does not only export arrays relevant for coloring but all existing arrays (and in contrast to the ParaView native export also `vtkStringArray`).

### Changes
The following changes have been made (see also git history):
* Fix errors (Python 3 compatibility, name of imported modules)
* Do not ignore vtkMultiBlockDataset when exporting
* Export all numeric arrays instead only the one needed for coloring the scene items
* Do not skip vtkStringArray when exporting to *.vtkjs
* Fixed export when names of scene items contain spaces or other special characters
* Support for coloring of scene items by non-numeric arrays (tested with string arrays)
* Python-style formatting

### Testing
- Run the macro from ParaView and inspect the produced `*.vtkjs` to see that all array data is now available.
- The exported scene can also be analyzed using a modified SceneExplorer, see attachment (from https://github.com/dasmy/SceneExplorer), where labels based on numeric and string-based arrays can be added to point data visualizations.
image

- Tested environment:
  - **vtk.js**: not relevant
  - **OS**: macOS
  - **Browser**: not relevant
  - **Paraview**: 5.10.1

### DIFF
--- a/Utilities/ParaView/export-scene-macro.py
+++ b/Utilities/ParaView/export-scene-macro.py
@@ -450,7 +450,7 @@ for rIdx in range(renderers.GetNumberOfItems()):
         dataArray = None
 
         if dsAttrs:
-            dataArray = dsAttrs.GetArray(colorArrayName)
+            dataArray = dsAttrs.GetAbstractArray(colorArrayName)
 
         if dataArray:
           # component = -1 => let specific instance get scalar from vector before mapping
--- a/Utilities/ParaView/export-scene-macro.py
+++ b/Utilities/ParaView/export-scene-macro.py
@@ -1,36 +1,41 @@
+import zipfile
+from urllib.parse import quote
+import hashlib
+import shutil
+import gzip
+import json
+import errno
+import time
+import os
+import sys
 
-### ----------------------------------------------------------------------- ###
-###                       Configure output location                         ###
-###
-### A top-level export directory will be created (if necessary) and used to
-### store all exported scenes.  Use the EXPORT_DIRECTORY pattern below to
-### customize where this directory should be.  Automatic replacement will
-### be done on the following variables if they appear in the pattern:
-###
-###   ${USER_HOME} : Will be replaced by the current user's home directory
-### ----------------------------------------------------------------------- ###
+from paraview.vtk import *
+from paraview import simple
+
+
+# ### ----------------------------------------------------------------------- ###
+# ###                       Configure output location                         ###
+# ###
+# A top-level export directory will be created (if necessary) and used to
+# store all exported scenes.  Use the EXPORT_DIRECTORY pattern below to
+# customize where this directory should be.  Automatic replacement will
+# be done on the following variables if they appear in the pattern:
+#
+# ${USER_HOME} : Will be replaced by the current user's home directory
+# ### ----------------------------------------------------------------------- ###
 
 EXPORT_DIRECTORY = '${USER_HOME}/vtkJsExport'
 FILENAME_EXTENSION = '.vtkjs'
 
-### ----------------------------------------------------------------------- ###
-###                   Convenience methods and definitions                   ###
-### ----------------------------------------------------------------------- ###
+# ### ----------------------------------------------------------------------- ###
+# ###                   Convenience methods and definitions                   ###
+# ### ----------------------------------------------------------------------- ###
 
-import sys, os, re, time, errno, json, math, gzip, shutil, argparse, hashlib
-
-from urllib.parse import quote
-import zipfile
-
-from paraview import simple
-from paraview.vtk import *
 
 try:
   from vtk.vtkFiltersGeometry import vtkCompositeDataGeometryFilter
-  from vtk.vtkCommonCore import vtkUnsignedCharArray
 except:
   from vtkFiltersGeometry import vtkCompositeDataGeometryFilter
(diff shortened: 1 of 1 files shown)


dasmy:
Btw.: I would love to see some proper output on where the exported file has been saved, but actually all `print()` statements are ignored if the Python console is closed. Is there any way of showing a message box (or even better: opening a file selection dialog so that the user can select where to store the output)?

finetjul:
Can you please revert indent spacing to 2 spaces instead of 4 ?

Utilities/ParaView/export-scene-macro.py
```diff
@@ -430,12 +440,13 @@ def mkdir_p(path):
...
       continue
+
     if hasattr(renProp, 'GetMapper'):
       mapper = renProp.GetMapper()
-      dataObject = mapper.GetInputDataObject(0, 0);
+      dataObject = mapper.GetInputDataObject(0, 0)
       dataset = None
 
-      if dataObject.IsA('vtkCompositeDataSet'):
+      if dataObject.IsA('vtkCompositeDataSet') or dataObject.IsA('vtkMultiBlockDataSet'):
```
jourdain, line 449: `vtkMultiBlockDataSet` is actually a `vtkCompositeDataSet` so the previous condition should work just fine.

jourdain:
the code looks good. Ideally we may want to re-work the commit message to follow semantic-release/commitizen while not triggering any new release as this code is only available here as reference.

jourdain:
```
0s
Run npx commitlint --from=[1](https://github.com/Kitware/vtk-js/runs/7212189718?check_suite_focus=true#step:6:1)[4](https://github.com/Kitware/vtk-js/runs/7212189718?check_suite_focus=true#step:6:5)aff403bdeba9edd4bc5b4e4ef911abf3227f9e
⧗   input: autopep8 formatting:
autopep8 --in-place --indent-size=2 --ignore=E121 export-scene-macro.py
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
⧗   input: also supoprt coloring by a non-numeric array
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
⧗   input: fix export of items with spaces (and other special chars) in their name
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
⧗   input: support exporting string arrays
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
⧗   input: export all standard arrays instead of only the color
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
⧗   input: fix errors, also export vtkMultiBlockDataSet
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

dasmy:
> Also, I suggest replacing the mutable default values with `None`, i.e. replace all the `root={}` with `root=None` in the function definitions and add the lines:
> 
> ```python
> if root is None:
>   root = {}
> ```
> 
> (As explanation: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

Definitely agree. Fixed it. Originally I did not want to touch to many things but as the change is a bit larger now anyway, this additional change does not make much of a difference :-D

Actually, the interface design of these routines seems bad anyway. From a first look, I would assume it might be better to drop the argument, always create a local root inside the routines and return it. Then, a caller could integrate the subtree it just received into its own local root. But didn't look very much in detail and would rather not rewrite the whole thing even more...

dasmy:
> Thanks for the commit rewrites but `feat()` and `fix()` will trigger new releases. To prevent any release, you have to stick with `style()` and `chore()`. Sorry for not mentioning before.

Ooops - didn't know that (and didn't read it anywhere). - Changed the messages again.

jourdain (approved):
LGTM

PR #2505, opened 2022-07-07
fix(OpenGL/Shader): membership with `in` keyword

floryst:
### Context
`Object.keys(obj).indexOf(name)` is more succinctly written as `name in obj`. Additionally, `name in obj` could be much faster than indexOf, which runs in O(n).

### Results
Mostly some maintenance, and a slight perf boost. 

### Changes
- [ ] Update `indexOf` usage on object keys to use `in`

### PR and Code Checklist

### Testing
No tests were modified.


finetjul:
LGTM

Issue #2438, opened 2022-05-26
Enabling picking on a vtkWidgetManager drops the frame count to 5 on simple mouse moves


miguelsmatos:
Hello, and thank you for working on the awesome project that vtk.js is!

I'm currently trying to setup a Vue app using vtk.js but am facing some performance issues even with a simple cone geometry. I wonder if anyone has 

### High-level description
Enabling picking for a `vtkWidgetManager` drops the frame count to ~5 when just moving the mouse (no click/drag).

### Steps to reproduce
Using vtk.js as an ES6 dependency of a vue project:
* Start with the vtk.js vue template
* Create a `vtkAnnotatedCubeActor` actor instance
* Create a `vtkInteractiveOrientationWidget` and place it on the cube
* Create a `vtkWidgetManager` 
* Enable picking
* Move the mouse around the vtkContainer and observe the fame count drop from 60 to ~5.
* When click and dragging, notice a delay of almost a second between the click and the movement (of the cone in the attached example).

Unfortunately I was not able to reproduce this with a CodePen, but created a simple repo at https://github.com/miguelsmatos/vite-vtk-js, with a single component https://github.com/miguelsmatos/vite-vtk-js/blob/main/src/components/HelloWorld.vue 

### Detailed behavior
* When picking is enabled, there seems to be a somewhat expensive operation that gets triggered with every MouseMove event. 
* There are a few warnings on the console. For Chrome:
```
Performance warning: clear() called with no buffers in bitmask
```
* For Firefox
```
vtkWidgetManager updateSelection() took 198ms on MouseMove
WebGL warning: clear: Calling gl.clear(0) has no effect
```
* This also affects the interaction with the part. When clicked and dragged, there is a significant delay until the part starts moving.

### Expected behavior

Some frame count drop, but not as pronounced.

### Environment
- **vtk.js**:  "@kitware/vtk.js": "^24.14.0"
- **OS**: Windows10 64bit
- **Browser**: Chrome 102.0.5005.61 and Firefox 101.0b9 (64-bit)

floryst:
This is related to #2435. I suspect that when two widget managers are present on different renderers in the same render window, they don't play nicely. FYI @martinken

Issue #2435, opened 2022-05-25
Tab main thread freezes when using multiple renderers with widget manager


Julesdevops:
### High-level description
This is about a tab crash when using WidgetManager with multiple renderers. (one instance of widgetmanager per renderer).
It was working fine before this commit: https://github.com/kitware/vtk-js/commit/5e32d4a0738dde13ef1a69a5893e9cac47f6a749
I'm not familiar with the widget infrastructure so maybe I'm doing something stupid in which case please accept my apologies.

### Steps to reproduce
A repo to reproduce the problem: https://github.com/Julesdevops/custom-vtkjs-app-template/tree/sphere-widget
Basically using a sphere widget with an InteractionOrientationWidget alongside with an OrientationMarkerWidget (legacy). I'm basically just mixing two examples:
- https://kitware.github.io/vtk-js/examples/InteractiveOrientationWidget.html
- https://kitware.github.io/vtk-js/examples/SphereWidget.html

### Detailed behavior
The tab just freezes/crash, UI is blocked forever.
I've tried investigating but this is just too complex for me. `handleEvent` seems to be re triggering the event over and over again: https://github.com/Kitware/vtk-js/blob/77ee125ae14a4fdfbc0e4f04727aec2ec4253fbf/Sources/Widgets/Core/WidgetManager/index.js#L317

### Expected behavior
We should be able to observe the behavior that existed before 5e32d4a0738dde13ef1a69a5893e9cac47f6a749 which is: no crash, working good.

### Environment
- **vtk.js**: v24.14.0
- **OS**: archlinux
- **Browser**: Chrome Version 101.0.4951.64 (Official Build) (64-bit) and Firefox 100.0.2 (64-bit)

floryst:
This is an issue with how we handle touch with widgets and how the WidgetManager has to interrupt the event chain in order to issue a picking operation. I have a better approach sketched out, but it ideally means that the event `callData` should be annotated with the pointing device, so we can handle touch in a special case and leave the mouse case as-is.

PR #2506, opened 2022-07-07
fix(WidgetManager): handle touch support

floryst:
### Context
- Touch widget support is broken. Related: https://discourse.vtk.org/t/reslicecursorwidget-does-not-operate-the-cross-selector-normally-on-the-mobile-touch-screen/8743/8
- Multiple widget managers cause freezes. Related: #2435, #2438

FYI @finetjul. I'm still investigating, but here is something that might work.

### Results
- touch should work on widgets
- multiple widget managers should not freeze

### Changes

### PR and Code Checklist

### Testing

- [x] Tested widgets:
  - polyline widget with an orientation marker widget

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -253,22 +253,34 @@ function vtkWidgetManager(publicAPI, model) {
   // API public
   // --------------------------------------------------------------------------
 
-  async function updateSelection({ position }) {
-    model._selectionInProgress = true;
+  async function updateSelection(callData, fromTouchEvent, callID) {
+    const { position } = callData;
     const { requestCount, selectedState, representation, widget } =
       await publicAPI.getSelectedDataForXY(position.x, position.y);
-    model._selectionInProgress = false;
 
-    if (requestCount) {
-      // Call activate only once
+    if (requestCount || callID !== model._currentUpdateSelectionCallID) {
+      // requestCount > 0: Call activate only once
+      // callID check: drop old calls
       return;
+    }
+
+    function activateHandle(w) {
+      if (fromTouchEvent) {
+        // release any previous left button interaction
+        model._interactor.invokeLeftButtonRelease(callData);
+      }
+      w.activateHandle({ selectedState, representation });
+      if (fromTouchEvent) {
+        // re-trigger the left button press to pick the now-active widget
+        model._interactor.invokeLeftButtonPress(callData);
+      }
     }
 
     // Default cursor behavior
     model._apiSpecificRenderWindow.setCursor(widget ? 'pointer' : 'default');
 
     if (model.widgetInFocus === widget && widget.hasFocus()) {
-      widget.activateHandle({ selectedState, representation });
+      activateHandle(widget);
       // Ken FIXME
       model._interactor.render();
       model._interactor.render();
@@ -276,7 +288,7 @@ function vtkWidgetManager(publicAPI, model) {
       for (let i = 0; i < model.widgets.length; i++) {
         const w = model.widgets[i];
         if (w === widget && w.getNestedPickable()) {
-          w.activateHandle({ selectedState, representation });
+          activateHandle(w);
           model.activeWidget = w;
         } else {
           w.deactivateAllHandles();
@@ -288,38 +300,12 @@ function vtkWidgetManager(publicAPI, model) {
     }
   }
 
-  const handleEvent = (eventName) => {
-    let guard = false;
-    return (callData) => {
-      if (
-        guard ||
-        model.isAnimating ||
-        !model.pickingEnabled ||
-        model._selectionInProgress
-      ) {
-        return macro.VOID;
-      }
-
-      const updatePromise = updateSelection(callData);
-      macro.measurePromiseExecution(updatePromise, (elapsed) => {
-        // 100ms is deemed fast enough. Anything higher can degrade usability.
-        if (elapsed > 100) {
-          macro.vtkWarningMacro(
-            `vtkWidgetManager updateSelection() took ${elapsed}ms on ${eventName}`
-          );
-        }
-      });
-
-      updatePromise.then(() => {
-        if (model._interactor) {
-          // re-trigger the event, ignoring our own handler
-          guard = true;
-          model._interactor[`invoke${eventName}`](callData);
-          guard = false;
-        }
-      });
-      return macro.EVENT_ABORT;
-    };
+  const handleEvent = async (callData, fromTouchEvent = false) => {
+    if (!model.isAnimating && model.pickingEnabled) {
+      const callID = Symbol('UpdateSelection');
+      model._currentUpdateSelectionCallID = callID;
+      await updateSelection(callData, fromTouchEvent, callID);
+    }
   };
 
   function updateWidgetForRender(w) {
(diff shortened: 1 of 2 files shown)


finetjul:
LGTM

floryst:
There is still a performance issue when multiple widget managers exist in a render window.

floryst:
Some of the performance issues stem from #2484. I have a fix for that, which should benefit this PR.

floryst:
I think this is good to go. I've tested this on a Chromium-based browser on a smartphone using a polyline wiget example that has been modified to include the orientation marker widget.

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -253,30 +253,40 @@ function vtkWidgetManager(publicAPI, model) {
...
 
     if (requestCount) {
       // Call activate only once
       return;
     }
 
+    function activateHandle(w) {
+      if (fromTouchEvent) {
+        // release any previous left button interaction
+        model._interactor.invokeLeftButtonRelease(callData);
```
finetjul, line 269: shouldn't it be conditional ? 
It seems dangerous to invoke a release event if no press event was previously called.
Some interactor observer may only have a callback to onLeftButonRelease.
Maybe that can be handled by the RWI by overloading the `invokeXXX` methods

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -288,38 +298,18 @@ function vtkWidgetManager(publicAPI, model) {
...
-          macro.vtkWarningMacro(
-            `vtkWidgetManager updateSelection() took ${elapsed}ms on ${eventName}`
-          );
-        }
-      });
+  const handleEvent = async (callData, fromTouchEvent = false) => {
+    if (
+      model.isAnimating ||
+      !model.pickingEnabled ||
+      model._selectionInProgress
```
finetjul, line 305: I'm not sure events should be discarded when processing one event. I guess it's important to handle the last event (debounce ? throttle?). 
Alternatively, we could discard only the events of the same type as the one being `in progress.

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -288,38 +298,18 @@ function vtkWidgetManager(publicAPI, model) {
...
-      const updatePromise = updateSelection(callData);
-      macro.measurePromiseExecution(updatePromise, (elapsed) => {
-        // 100ms is deemed fast enough. Anything higher can degrade usability.
-        if (elapsed > 100) {
-          macro.vtkWarningMacro(
-            `vtkWidgetManager updateSelection() took ${elapsed}ms on ${eventName}`
-          );
-        }
-      });
+  const handleEvent = async (callData, fromTouchEvent = false) => {
```
finetjul, line 301: this function should always return something

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -414,13 +404,18 @@ function vtkWidgetManager(publicAPI, model) {
...
-        // stay after widgets, but before default of 0
-        WIDGET_PRIORITY / 2
-      )
+      model._interactor.onLeftButtonPress((eventData) => {
+        const { deviceType } = eventData;
+        const touchEvent = deviceType === 'touch' || deviceType === 'pen';
+        // only try selection if the left button press is from touch.
+        if (touchEvent) {
+          handleEvent(eventData, touchEvent);
+        }
```
finetjul, line 417: this function should return something

Sources/Widgets/Core/WidgetManager/index.js
finetjul, line 301: shoudln't you also support `async` functions in the callback manager ?

Sources/Widgets/Core/WidgetManager/index.js
floryst: If `fromTouchEvent` is true, then there should have been a LeftButtonPress triggered just before. The way this works is that a touch event always triggers LeftButtonPress. Since the widget manager needs to do an (async) pick+selection, it can't abort the event chain, so the event may get picked up by an interactor style. Once the picking is finished and a widget is deemed to be selected, then we synthetically invoke LeftButtonRelease to release the focus on the interactor style so that we can grab focus on the now-selected widget.

Callback ordering for OnLeftButtonPress: `widgets -> widgetManager -> interactorStyle (if available)`.

This does have an issue where if picking takes too slow, then users may see out-of-order behavior. I'm open to suggestions on this.

Sources/Widgets/Core/WidgetManager/index.js
floryst: This is something we'll need to discuss further. Supporting serialization of async event listeners will change the behavior of `invokeXXX` calls (e.g. `obj.modified()`). We either need to ensure that usages of invoke/modified do not assume that all callbacks have run to completion (since async tasks could finish later), or annotate invoke/modified to be async as well.

Sources/Widgets/Core/WidgetManager/index.js
floryst: Good point. I toyed with the idea of handling the last event, discarding the results of previous events, but that didn't make it in this PR. I will re-visit that idea.

Sources/Widgets/Core/WidgetManager/index.js
floryst: handleEvent doesn't need to return anything, it has nothing to return. The event handling logic treats "undefined" returns the same as "macro.VOID".

finetjul: Even if undefined (nothing returned) is the same as "macro.VOID", I find it a good practice to have event callbacks return something. This gives some hint to the developer reading /modifying/copying your code.

I've seen many cases where nothing was returned but an ABORT should have been returned. And when I asked the developer, they typically said "I didn't know that callback could return something".

Sources/Widgets/Core/WidgetManager/index.js
finetjul: fair enough, the `if (fromTouchEvent)` block can behave as the `condition` I was looking for.

Sources/Widgets/Core/WidgetManager/index.js
floryst: Ah, well in that case I'm good with returning macro.VOID explicitly for clarity purposes.

floryst:
I've updated the PR with the following changes:
- event handlers now return macro.VOID. `handleEvent` doesn't since it's async, so I opted to wrap the call to handleEvent in a separate function.
- `model._currentUpdateSelectionCallID` is used to only allow the most recent updateSelection call to be processed.

Sources/Widgets/Core/WidgetManager/index.js
floryst: I'm pretty sure that if our event handling system supported serialization of async event handlers, it would eliminate the semi-hacky OnLeftButtonRelease call.

Maybe we can consider an `await invokeAndWait()` concept for instances where you really do want to wait for all callbacks to have been processed before continuing.

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -253,30 +253,43 @@ function vtkWidgetManager(publicAPI, model) {
   // API public
   // --------------------------------------------------------------------------
 
-  async function updateSelection({ position }) {
-    model._selectionInProgress = true;
+  model._updateSelectionCallID = null;
```
finetjul, line 256: does not seem the right place to initialize model variable

Sources/Widgets/Core/WidgetManager/index.js
floryst: This will have to be something we explore further as necessary. I won't cover it in the scope of this PR for now.

Sources/Widgets/Core/WidgetManager/index.js
floryst: I moved this into a comment in DEFAULT_VALUES.

finetjul: sounds good, but this variable is never read. are you sure there is no typo ?

floryst: Ah, that is a typo. I've updated it with the right name.

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -253,30 +253,41 @@ function vtkWidgetManager(publicAPI, model) {
...
+      // callID check: drop old calls
       return;
     }
 
+    function activateHandle(w) {
+      if (fromTouchEvent) {
+        // release any previous left button interaction
+        model._interactor.invokeLeftButtonRelease(callData);
+        w.activateHandle({ selectedState, representation });
+        model._interactor.invokeLeftButtonPress(callData);
```
finetjul, line 272: I typically prefer code factorization to avoid diverging in the future:
```
if (fromTouchEvent) {
  model._interactor.invokeLeftButtonRelease(callData);
}
w.activateHandle({ selectedState, representation });
if (fromTouchEvent) {
        model._interactor.invokeLeftButtonPress(callData);
}
```

floryst: Done!

finetjul:
LGTM

PR #2507, opened 2022-07-08
perf(OpenGLRenderWindow): caching proxy gl context

floryst:
### Context

This is the continuation of #2484, since I could not push to @FezVrasta's branch.

We use ES6 proxies to cache calls to getParameter and depthMask.

### Results

- improved performance on repeated calls to getParameter, which go through the cache rather than the webgl context.

### Changes

- [x] Implemented a Proxy around the webgl context.

### PR and Code Checklist

### Testing

### DIFF
--- a/Sources/Rendering/OpenGL/RenderWindow/index.js
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.js
@@ -8,6 +8,7 @@ import vtkOpenGLTextureUnitManager from 'vtk.js/Sources/Rendering/OpenGL/Texture
 import vtkOpenGLViewNodeFactory from 'vtk.js/Sources/Rendering/OpenGL/ViewNodeFactory';
 import vtkRenderPass from 'vtk.js/Sources/Rendering/SceneGraph/RenderPass';
 import vtkRenderWindowViewNode from 'vtk.js/Sources/Rendering/SceneGraph/RenderWindowViewNode';
+import { createContextProxyHandler } from 'vtk.js/Sources/Rendering/OpenGL/RenderWindow/ContextProxy';
 
 const { vtkDebugMacro, vtkErrorMacro } = macro;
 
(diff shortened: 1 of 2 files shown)


Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -90,6 +90,59 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
+    return {
+      get(gl, prop, receiver) {
+        let ret = Reflect.get(gl, prop, receiver);
+        if (ret instanceof Function) {
+          // prevents Illegal Invocation errors
+          ret = ret.bind(gl);
+        }
+
+        if (prop in propHandlers) {
+          return propHandlers[prop](gl, prop, receiver, ret);
```
finetjul, line 136: Performance being important here, instead of checking if `prop` belongs to `propHandlers` and then get the prop handler, how about:
```
const propHandler = propHandles[prop];
if (propHandler) {
  return propHandler(...)
}
```

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -90,6 +90,59 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkOpenGLRenderWindow');
 
+  function createContextHandler() {
```
finetjul, line 93: To move in a separate file ? (in the same folder though)

finetjul:
I have not doubt that it works as intended, but it is not easy to read (a lot of JS trickery).

As it is meant to be extended with critical gl calls, can you add some documentation on how and where to cache new gl calls ?

floryst:
I've added some explanation as comments. Most of the trickery comes from Proxy itself, but for the purposes of our usage, we need only concern ourselves with the get trap handler.

PR #2509, opened 2022-07-11
WebGPU Physically Based Rendering

andrewcombs:
### Context
Physically based rendering allows for better and more accurate visualization of certain types of data, in addition to being used to simply make certain visualizations more appealing. 

### Results
The PBR model gives much more customizable and accurate results than the previous lighting model, in addition to including support for lights. Below are comparisons between the previous and new lighting models in the same scenes. 
<table align="center">
  <tr>
    <td align="center">
      <h3>Old</h3>
      
      
      
    </td>
    <td align="center">
      <h3>New</h3>
      
      
      
    </td>
  </tr>
</table>

### Changes
- Actors using PBR will now be lit using a metallic roughness model
- Lights in the scene are now responsible for all aspects of the lighting. Specular color is no longer supported, as the color of reflections now comes from lights. 
- For simple lighting using directional lights, direction can be set instead of focal point using setDirectionVector.
- A cone falloff angle has been added to allow for smoother spotlights.
- Actor properties have been updated to include new values and textures, those being diffuse, metallic, roughness, normal, ambient occlusion, and emission.
- Two examples have been added, being PBR and SphereArray. The PBR example shows a fully textured and lit model, and the SphereArray example shows a grid of spheres with a range of roughness and metallic values.

### PR and Code Checklist

### Testing
The new included tests utilize almost every change made. If you would like to test, run both examples and ensure there are no errors and that nothing looks out of place.
- [x] Tested environment:
  - **vtk.js**: 25.1.3
  - **OS**: Windows 10
  - **Browser**: Chrome Canary 105.0.5173.0


Sources/Rendering/Core/Light/index.js
```diff
@@ -40,15 +40,20 @@ function vtkLight(publicAPI, model) {
...
+      vec3.sub(model.direction, model.focalPoint, model.position);
       vtkMath.normalize(model.direction);
       model.directionMTime = model.mtime;
     }
     return model.direction;
   };
 
+  // Sets the direction from a vec3 instead of a focal point
+  publicAPI.setDirectionVector = (directionVector) => {
+    const newFocalPoint = new Float64Array(3);
```
martinken, line 52: Should be just setDirection

Sources/Rendering/Core/Property/index.js
```diff
@@ -62,6 +62,20 @@ function vtkProperty(publicAPI, model) {
...
+      model.roughness !== roughness ||
+      model.specularPower !== specularPower
+    ) {
+      model.specularPower = specularPower; // Specular power still needs to be set as long as webgl is using it (otherwise testShaderReplacementsClear fails)
+      model.roughness = roughness;
+      publicAPI.modified();
+    }
+  };
+
+  // publicAPI.getSpecularPower = () => 1 / model.roughness;
```
martinken, line 77: Old code yank

Sources/Rendering/WebGPU/Renderer/index.js
```diff
@@ -35,6 +37,16 @@ fn main(
 }
 `;
 
+function getLightTypeIndex(light) {
```
martinken, line 40: Maybe a quick comment here explaining the 3 types and what it is used for

PR #2514, opened 2022-07-14
fix(WebGPU): fix two issues in WebGPU

martinken:
Fix an issue where the renderer UBO was being shared between
pipelines causing some multiviewport renderings to use an
incorrect UBO

Fix an issue in the ImageMapper where the world coordinates
and texture coordinates were not consistent with the new
definition of pixel ceneterd images.

Fix an issue where the range of a color transfer function
was being ignored in one place while used in building the table.

### DIFF
--- a/Sources/Rendering/WebGPU/ImageMapper/index.js
+++ b/Sources/Rendering/WebGPU/ImageMapper/index.js
@@ -158,6 +158,10 @@ function vtkWebGPUImageMapper(publicAPI, model) {
       // tmpMat4 is now SC -> Index, save this as we need it later
       mat4.invert(tmp3Mat4, tmpMat4);
 
+      // need translation and scale
+      mat4.fromTranslation(tmp2Mat4, [0.5, 0.5, 0.5]);
+      mat4.multiply(tmpMat4, tmp2Mat4, tmpMat4);
+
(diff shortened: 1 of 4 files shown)


Sources/Rendering/WebGPU/ImageMapper/index.js
```diff
@@ -236,7 +240,7 @@ function vtkWebGPUImageMapper(publicAPI, model) {
 
         const target = iComps ? i : 0;
         const cfun = actor.getProperty().getRGBTransferFunction(target);
-        if (cfun && actor.getProperty().getUseLookupTableScalarRange()) {
+        if (cfun) {
```
martinken, line 243: @sankhesh  Here is where I deleted the test. I think the solution is that this code here and the getTable() call far later down have to use the same range. Right now the getTable call uses the table range while in the old code this part did not. Same idea on the WebGL side.

martinken: The advantage of using the tfun range is that it tends to not change much. To use the window level to create the texture means rebuilding the texture every time the window or level changes. I'm not sure the details of the change you made so I'll leave it up to you in terms of how you want to change it.

Issue #2508, opened 2022-07-11
Cell Picking no longer working on vtk.js 25.X


WesleyTheGeolien:
### High-level description
After updating from vtk.js@24.3.1 (I think later versions of 24.X still works) to vtk.js@25.1.3 Cell picking for quads are no longer working.

This looks to be due to #2394 and in particular due to `createCellMap` in `Sources/Rendering/Core/CellPicker/index.js`. The new implementation create a cell mapping which only specifies Lines, PolyLines or Triangles, however in `Sources/Common/DataModels/Constants` there are a number of other possible cell types

### Steps to reproduce
Create a Poly Data object composed of vtkQuads.

### Expected behavior
Expected to be able to pick as before

### Environment
- **vtk.js**: 25.1.3
- **OS**: MacOS 12.4
- **Browser**: Firefox developper 103.0b2 (64-bit)

finetjul:
You can triangulate your mesh before doing picking.

But the best would be to port [vtkQuad](https://gitlab.kitware.com/vtk/vtk/-/blob/master/Common/DataModel/vtkQuad.h) into VTK.js. You can see how it was done for vtkTriangle.

Fancy contributing vtkQuad ?

finetjul:
Closed by #2515

PR #2515, opened 2022-07-14
Implement vtkQuad

WesleyTheGeolien:
### Context
This MR aims to implement the vtkQuad class as suggested by @finetjul in #2508 

### Changes
Created vtkQuad class. Type script updated but unsure on how to change documentation?

### PR and Code Checklist

### Testing
Added unit tests but two are failing for the pcoords, I have copied the behaviour in the cxx library but the tests are from the vtkTriangle vtk.js class so they may be incorrect, I am not familiar enough with how these are calculated as to whether they are correct. Maybe @finetjul you could take a look?
- [x] Tested environment:
  - **vtk.js**: latest
  - **OS**: Mac OS 12.4
  - **Browser**:  Chrome (through karma)

### DIFF
--- a/Sources/Common/DataModel/Quad/index.d.ts
+++ b/Sources/Common/DataModel/Quad/index.d.ts
@@ -1,0 +1,91 @@
+import { Vector3, Vector4 } from '../../../types';
+import vtkCell, { ICellInitialValues } from '../Cell';
+import { IIntersectWithLine } from '../Triangle';
+
+export interface IQuadInitialValues extends ICellInitialValues {}
+
+export interface vtkQuad extends vtkCell {
+  getCellType(): number;
+  /**
+   * Get the topological dimensional of the cell (0, 1, 2 or 3).
+   */
+  getCellDimension(): number;
+  getNumberOfEdges(): number;
+  getNumberOfFaces(): number;
+
+  /**
+   * Compute the intersection point of the intersection between quad and
+   * line defined by p1 and p2. tol Tolerance use for the position evaluation
+   * x is the point which intersect triangle (computed in function) pcoords
+   * parametric coordinates (computed in function) A javascript object is
+   * returned :
+   *
+   * ```js
+   * {
+   *    evaluation: define if the triangle has been intersected or not
+   *    subId: always set to 0
+   *    t: parametric coordinate along the line.
+   *    betweenPoints: Define if the intersection is between input points
+   * }
+   * ```
+   *
+   * @param {Vector3} p1 The first point coordinate.
+   * @param {Vector3} p2 The second point coordinate.
+   * @param {Number} tol The tolerance to use.
+   * @param {Vector3} x The point which intersect triangle.
+   * @param {Vector3} pcoords The parametric coordinates.
+   */
+  intersectWithLine(
+    p1: Vector3,
+    p2: Vector3,
+    tol: number,
+    x: Vector3,
+    pcoords: Vector3
+  ): IIntersectWithLine;
+
+  /**
+   * Determine global coordinate (x]) from subId and parametric coordinates.
+   * @param {Vector3} pcoords The parametric coordinates.
+   * @param {Vector3} x The x point coordinate.
+   * @param {Number[]} weights The number of weights.
+   */
+  evaluateLocation(pcoords: Vector3, x: Vector3, weights: number[]): void;
+
+  /*
+   *  Compute iso-parametric interpolation functions
+   * @param {Vector3} pcoords The parametric coordinates.
+   * @param {Vector4} sf out weights.
+   */
+  interpolationFunctions(pcoords: Vector3, sf: Vector4);
+}
+
+/**
+ * Method used to decorate a given object (publicAPI+model) with vtkQuad characteristics.
+ *
+ * @param publicAPI object on which methods will be bounds (public)
+ * @param model object on which data structure will be bounds (protected)
+ * @param {IQuadInitialValues} [initialValues] (default: {})
+ */
+export function extend(
+  publicAPI: object,
+  model: object,
+  initialValues?: IQuadInitialValues
+): void;
+
+/**
+ * Method used to create a new instance of vtkQuad.
+ * @param {IQuadInitialValues} [initialValues] for pre-setting some of its content
+ */
+export function newInstance(initialValues?: IQuadInitialValues): vtkQuad;
+
+/**
+ * vtkQuad is a cell which represents a quadrilateral. It may contain static
+ * methods to make some computations directly link to quads.
+ *
+ * @see vtkCell
+ */
+export declare const vtkQuad: {
+  newInstance: typeof newInstance;
+  extend: typeof extend;
+};
+export default vtkQuad;
--- a/Sources/Common/DataModel/Quad/index.js
+++ b/Sources/Common/DataModel/Quad/index.js
@@ -1,0 +1,260 @@
+import macro from 'vtk.js/Sources/macros';
+import vtkCell from 'vtk.js/Sources/Common/DataModel/Cell';
+import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
+import { CellType } from 'vtk.js/Sources/Common/DataModel/CellTypes/Constants';
+import vtkTriangle from 'vtk.js/Sources/Common/DataModel/Triangle';
+import vtkPoints from 'vtk.js/Sources/Common/Core/Points';
+
+function intersectionStruct() {
+  return {
+    intersected: false,
+    subId: -1,
+    x: [0.0, 0.0, 0.0],
+    pCoords: [0.0, 0.0, 0.0],
+    t: -1,
+  };
+}
+
+function vtkQuad(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkQuad');
+
+  publicAPI.getCellDimension = () => 2;
+  publicAPI.getCellType = () => CellType.VTK_QUAD;
+  publicAPI.getNumberOfEdges = () => 4;
+  publicAPI.getNumberOfFaces = () => 0;
+
+  publicAPI.intersectWithLine = (p1, p2, tol, x, pcoords) => {
+    let outObj = {
+      subId: 0,
+      t: Number.MAX_VALUE,
+      intersect: 0,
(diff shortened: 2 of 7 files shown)


Sources/Common/DataModel/Quad/index.d.ts
```diff
@@ -0,0 +1,121 @@
+import { Vector3 } from '../../../types';
+import vtkCell, { ICellInitialValues } from '../Cell';
+import { IIntersectWithLine } from '../Triangle';
+
+export interface ITriangleInitialValues extends ICellInitialValues {}
```
daker, line 5: IQuadInitialValues instead of ITriangleInitialValues

Sources/Common/DataModel/Quad/index.js
```diff
@@ -0,0 +1,252 @@
+import macro from 'vtk.js/Sources/macros';
+import vtkCell from 'vtk.js/Sources/Common/DataModel/Cell';
+import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
+import { CellType } from 'vtk.js/Sources/Common/DataModel/CellTypes/Constants';
+import vtkTriangle from '../Triangle';
+import vtkPoints from '../../Core/Points';
```
daker, line 6: shoud use the vtk.js/Sources/.. import

Sources/Common/DataModel/Quad/index.d.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export interface vtkQuad extends vtkCell {
+  GetCellType(): number;
```
daker, line 8: getCellType instead of GetCellType

daker (changes requested):

Sources/Common/DataModel/Quad/index.js
```diff
@@ -0,0 +1,252 @@
...
+export const STATIC = {};
+
+function vtkQuad(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkQuad');
+
+  publicAPI.getCellDimension = () => 2;
+  publicAPI.getCellType = () => CellType.VTK_QUAD;
+  publicAPI.GetNumberOfEdges = () => 4;
+  publicAPI.GetNumberOfFaces = () => 0;
```
daker, line 21: GetNumberOfEdges => getNumberOfEdges
GetNumberOfFaces => getNumberOfFaces

daker (changes requested):

Sources/Common/DataModel/Quad/index.js
```diff
@@ -0,0 +1,265 @@
...
+import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
+import { CellType } from 'vtk.js/Sources/Common/DataModel/CellTypes/Constants';
+import vtkTriangle from 'vtk.js/Sources/Common/DataModel/Triangle';
+import vtkPoints from 'vtk.js/Sources/Common/Core/Points';
+
+// ----------------------------------------------------------------------------
+// Static API
+// ----------------------------------------------------------------------------
+
+export const STATIC = {};
```
floryst, line 12: If there is no exportable static API, then you can leave this out. Were you planning on adding some static methods?

WesleyTheGeolien:
No I was leaving it more in there as a stub in case there ever was any static methods to export. 

Ca easily remove though.

floryst:
Let's remove the STATIC export then. No need to have this explicit placeholder if we don't have an intent on populating it any time soon.

WesleyTheGeolien:
Could someone please take a look at the `pCoords` in the tests that are failing? I am unsure what the "correct" values of these should be.

Sources/Common/DataModel/Quad/index.js
```diff
@@ -0,0 +1,259 @@
...
+
+    // Calculate the weights
+    publicAPI.interpolationFunctions(pcoords, weights);
+
+    x[0] = 0.0;
+    x[1] = 0.0;
+    x[2] = 0.0;
+
+    for (let i = 0; i < 4; i++) {
+      pt = model.points.getPoint(i);
```
finetjul, line 229: you might want to use instead the following notation:
```
const point = [];
for (...) {
  model.points.getPoint(i, point);
  ...
}
```

finetjul:
> Could someone please take a look at the `pCoords` in the tests that are failing? I am unsure what the "correct" values of these should be.

What line is `pCoords` failing ? what value do you have when it fails ?

WesleyTheGeolien:
Hey guys, wondering if someone could give me a hand with the pCoords so this can land?

Sources/Common/DataModel/Quad/index.js
```diff
@@ -0,0 +1,259 @@
...
+      subId: 0,
+      t: Number.MAX_VALUE,
+      intersect: 0,
+      betweenPoints: false,
+    };
+    let diagonalCase;
+    const d1 = vtkMath.distance2BetweenPoints(
+      model.points.getPoint(0),
+      model.points.getPoint(2)
+    );
```
finetjul, line 38: warning, getPoint() will use the same array for both calls.
you'd better do:
```
const d1 = vtkMath.distance2BetweenPoints(
      model.points.getPoint(0, []),
      model.points.getPoint(2, [])
    );
```
Alternatively, because you will reuse those points below:
```
cont point0 = model.points.getPoint(0, []);
cont point2 = model.points.getPoint(2, []);
const d1 = vtkMath.distance2BetweenPoints(
      point0,
      point2
    );
...
```

Sources/Common/DataModel/Quad/index.js
```diff
@@ -0,0 +1,259 @@
...
+
+    let secondIntersect;
+    const secondIntersectTmpObj = intersectionStruct();
+
+    let useFirstIntersection;
+    let useSecondIntersection;
+
+    switch (diagonalCase) {
+      case 0:
+        model.triangle.getPoints().setPoint(0, ...model.points.getPoint(0));
```
finetjul, line 91: store points before:
```
let points = null;
if (!model.triangle) {
  model.triangle = vtkTriangle.newInstance();
  points = vtkPoints.newInstance();
  ...
} else {
  points = model.triangle.getPoints();
}
```
points.setPoint(...)

DavidBerger98:
> Thanks @finetjul, I've taken into account your remarks. The `pCoords` test however are still off

On line 49 `[1,1,0]` is the correct value. So that should be the expected one.

The assertion on line 69 is not needed, since there was a noIntersection and thus `pcoords` was never recomputed. You can thus also remove the assertion on line 68.

Sources/Common/DataModel/Quad/test/testQuad.js
```diff
@@ -0,0 +1,71 @@
...
+  t.deepEqual(pcoords, [0, 0, 0]);
+
+  // Intersect on v3
+  p1 = [1, 1, 1];
+  p2 = [1, 1, 0];
+  intersection = quad.intersectWithLine(p1, p2, tol, x, pcoords);
+  t.equal(intersection.intersect, yesIntersection);
+  t.equal(intersection.t, 1);
+  t.deepEqual(x, p2);
+  t.deepEqual(pcoords, [0, 1, 0]);
```
DavidBerger98, line 49: Please change to `t.deepEqual(pcoords, [1, 1, 0]);`

Sources/Common/DataModel/Quad/test/testQuad.js
```diff
@@ -0,0 +1,71 @@
...
+  intersection = quad.intersectWithLine(p1, p2, tol, x, pcoords);
+  t.equal(intersection.intersect, noIntersection);
+
+  // Line coplanar to quad
+  p1 = [0.5, -1, 0];
+  p2 = [0.5, 1, 0];
+  intersection = quad.intersectWithLine(p1, p2, tol, x, pcoords);
+  t.equal(intersection.intersect, noIntersection);
+  t.equal(intersection.t, Number.MAX_VALUE);
+  t.deepEqual(pcoords, [0, 0, 0]);
```
DavidBerger98, line 69: Please remove these lines

Sources/Common/DataModel/Quad/test/testQuad.js
```diff
@@ -0,0 +1,71 @@
...
+test('Test vtkQuad instance', (t) => {
+  t.ok(vtkQuad, 'Make sure the class definition exists');
+  const instance = vtkQuad.newInstance();
+  t.ok(instance);
+  t.end();
+});
+
+test('Test vtkQuad intersectWithLine flat', (t) => {
+  const yesIntersection = 1;
+  const noIntersection = 0;
```
DavidBerger98, line 14: Please add a `Constants.js` like the one at `'vtk.js/Sources/Common/DataModel/Line/Constants';`. Name the enum `QuadWithLineIntersectionState` and use `QuadWithLineIntersectionState.NO_INTERSECTION` and `QuadWithLineIntersectionState.YES_INTERSECTION` instead of these constants.

WesleyTheGeolien:
Thanks @DavidBerger98  have taken into account your changes and the tests are now passing.

I believe this should be ready to merge now?

finetjul:
I believe you can now add support for vtkQuad cell picking. See [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/CellPicker/index.js#L19)

WesleyTheGeolien:
> I believe you can now add support for vtkQuad cell picking. See [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/CellPicker/index.js#L19)
> 
> If you can squash all the commits into a unique commit, it would be great.

I believe I have already implemented this and added the tests: https://github.com/Kitware/vtk-js/pull/2515/files#diff-dcd608a8f684922521c178b9a2965eccf3cadbedbfc27d680d70fe788e11b912

Just need ti squash all commits I guess

finetjul:
> I believe I have already implemented this and added the tests

My bad, for some reasons I forgot about it. You are correct, it's all fine !

> Just need to squash all commits I guess do I do this manually or do you guys do this on merge

We can do it automatically like you suggest but we prefer to do a merge commit. It's also better to squash manually because you have more controls on the commit title and body message.

WesleyTheGeolien:
@finetjul Right that should have squashed all commits.

finetjul:
Thanks @WesleyTheGeolien

PR #2516, opened 2022-07-14
fix(renderwindowinteractor): resolve WebXR AR touch event error on Android

tbirdso:
Short-term fix for issue where touching the screen on an Android device
during an AR session would crash the scene as a result of bad touch
event handling and XR Gamepad API polling. New behavior is no crash.
Future commit should add ability to parse data from event for interactivity.

### Context

Supports recent WebXR developments on mobile, including feedback from [WebXR examples](https://kitware.github.io/vtk-js/docs/develop_webxr.html).

### Results

AR scene no longer crashes on touch on Android device in local testing.

### Changes

- Fixes issue with bad access on touch-end event
- Fixes issue with unexpected handedness parameter for WebXR Gamepad data

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  `master`
  - **OS**:  Windows 10, Android 11
  - **Browser**:  Chrome 103.0.5060.114

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -552,6 +552,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
           model.lastGamepadValues[gp.index] = {
             left: { buttons: {} },
             right: { buttons: {} },
+            none: { buttons: {} },
           };
         }
         for (let b = 0; b < gp.buttons.length; ++b) {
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -762,7 +763,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
       if (pointers.length === 0) {
         const callData = {
           ...getModifierKeysFor(EMPTY_MOUSE_EVENT),
-          position: pointers[0].position,
+          position: undefined,
```
floryst, line 766: This should be `getScreenEventPositionFor(event)`. Thanks for catching this issue!

PR #2517, opened 2022-07-15
chore(WebGPU): improve testing coverage

martinken:
Add some missing tests to WebGPU

Add a baseline for a test that should pass

Move point picking into a different test that is
excluded from WebGPU for now as we do not have it implemented.


martinken:
@andrewcombs FYI

PR #2519, opened 2022-07-15
docs(webxrvolume): reslice default head volume to face the camera

tbirdso:
Adds manual reslice step to WebXRVolume example so that default head MR volume faces towards the
camera at startup. Note that the head is still rotated in other examples where it is easier to
interactively rotate the volume.

### Context

vtk.js WebXR interactivity support is still in its infancy and does not yet include rotating objects while in VR. As a result it is useful for the documented example volume to already have a "comfortable" orientation for demos.

### Results

Old behavior

before

New behavior

after

### Changes

Added reslice step in WebXRVolume example

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  `master`
  - **OS**:  Windows 10
  - **Browser**:  Chrome 103.0.5060.114

### Notes

The orientation issue seems to originate with the data and not the VR camera, other examples such as [https://kitware.github.io/vtk-js/examples/VolumeMapperLightAndShadow.html](https://kitware.github.io/vtk-js/examples/VolumeMapperLightAndShadow.html) also start off with the volume at a less optimal viewing orientation. Reslicing seemed like better practice than directly editing medical image metadata.

### DIFF
--- a/Examples/Volume/WebXRVolume/index.js
+++ b/Examples/Volume/WebXRVolume/index.js
@@ -1,3 +1,4 @@
+import { mat4 } from 'gl-matrix';
 import 'vtk.js/Sources/favicon';
 
 // Load the rendering pieces we want to use (for both WebGL and WebGPU)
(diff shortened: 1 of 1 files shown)


JiayiXuu:
> @JiayiXuu This change could also be applied to the in-dev cinematic tests to help make XR viewing more comfortable, though you may need to update the reslice axes for different datasets.

Sounds good thanks!

JiayiXuu (approved):

JiayiXuu:
@tbirdso The vr camera api is not exposed yet, right?

PR #2522, opened 2022-07-18
fix(volume-shader): disable cinematic rendering if 2+ components

### LINKED ISSUE #2520: TestVolumeTypes Example shader compile error with 2+ components
### High-level description
TestVolumeTypes example has shader compile error with 2+ components.

### Steps to reproduce
Go here:
https://kitware.github.io/vtk-js/examples/TestVolumeTypes.html

Increase components to 2.

### Detailed behavior
The error:
```
': ERROR: 0:1084: 'normalLight' : undeclared identifier
ERROR: 0:1084: 'applyLightingDirectional' : no matching overloaded function found
ERROR: 0:1084: '=' : dimension mismatch
ERROR: 0:1084: '=' : cannot convert from 'const mediump float' to 'highp 3-component vector of float'
```

Does not happen if Lighting unchecked.

### Environment
- **OS**: Ubuntu 22.04
- **Browser**: Chrome 103

PaulHax:
### Context
fix #2520 

### Results
No shader compile errors.

### Changes
Fallback to classic lighting when 2+ components

### PR and Code Checklist

### Testing

`npm run example TestVolumeTypes` and add components.
- [ ] Tested environment:
  - **vtk.js**: master
  - **OS**: Ubuntu 22.04
  - **Browser**: Chrome 103

### DIFF
--- a/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
+++ b/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
@@ -914,6 +914,7 @@ vec4 getColorForValue(vec4 tValue, vec3 posIS, vec3 tstep)
       mat4 normalMat = computeMat4Normal(posIS, tValue, tstep);
       #if !defined(vtkComponent0Proportional)
         vec4 normal0 = normalMat[0];
+        vec4 normalLight = computeNormal(posIS, tValue.a, tstep); // for case: vtkLightComplexity > 0 && defined(vtkIndependentComponentsOn) && (vtkNumComponents > 1) && !defined(vtkComponent0Proportional)
       #endif
       #if !defined(vtkComponent1Proportional)
         vec4 normal1 = normalMat[1];
--- a/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
+++ b/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
@@ -914,7 +914,6 @@ vec4 getColorForValue(vec4 tValue, vec3 posIS, vec3 tstep)
       mat4 normalMat = computeMat4Normal(posIS, tValue, tstep);
       #if !defined(vtkComponent0Proportional)
         vec4 normal0 = normalMat[0];
-        vec4 normalLight = computeNormal(posIS, tValue.a, tstep); // for case: vtkLightComplexity > 0 && defined(vtkIndependentComponentsOn) && (vtkNumComponents > 1) && !defined(vtkComponent0Proportional)
       #endif
       #if !defined(vtkComponent1Proportional)
         vec4 normal1 = normalMat[1];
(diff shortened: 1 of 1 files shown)


JiayiXuu:
@PaulHax 
You can swap #1059-1078 with:
```
  #if vtkLightComplexity > 0
    #if !defined(vtkComponent0Proportional) 
      #if vtkNumComponents == 1
    
        #ifdef SurfaceShadowOn
            #if vtkLightComplexity < 3
                vec3 tColorS = applyLightingDirectional(tColor.rgb, normalLight);
            #else
                vec3 tColorS = applyLightingPositional(tColor.rgb, normalLight, IStoVC(posIS));
            #endif
        #endif

        #ifdef VolumeShadowOn
          vec3 tColorVS = applyShadowRay(tColor.rgb, posIS, rayDirVC);
          #ifdef SurfaceShadowOn
            float vol_coef = volumetricScatteringBlending * (1.0 - tColor.a / 2.0) * (1.0 - atan(normalLight.w) * INV4PI);
            tColor.rgb = (1.0-vol_coef) * tColorS + vol_coef * tColorVS;
          #else
            tColor.rgb = tColorVS;
          #endif
        #else
            tColor.rgb = tColorS;
        #endif
        
      #else
        applyLighting(tColor.rgb, normal0);
      #endif
    #endif
```
This snippet makes it so cinematic rendering only supports 1 component. If numComponent >= 2 we fall back to old lighting models. 

With your fix, cinematic rendering will be applied to the first component. Both work but just to be consistent (all or none), I think it's the best if we disable cinematic for all components. Thoughts?

PaulHax:
No strong opinion.  Do users explicitly enable cinematic rendering?  Would mixing cinematic rendering on first component and "classic" lighting of other components look too strange?  One use case: a "label map" component to modulate the cinematic component.

JiayiXuu:
> No strong opinion. Do users explicitly enable cinematic rendering? Would mixing cinematic rendering on first component and "classic" lighting of other components look too strange? One use case: a "label map" component to modulate the cinematic component.

Cinematic rendering (cvr) is off by default. I'm not sure if it'll look strange because there are all sorts of things we can mix and it depends on the data. I haven't tested anything myself. 

I can see how cvr can be useful for the first component, say if we want a very detailed bone rendering with not-very-important tissue around it. But since my time is limited, I'm leaning towards leaving cvr+multivolume as another topic to explore with whoever that picks it up.

PaulHax:
Copied in JiayiXuu's changes to fallback to classic lighting when more than 1 component.

thewtex:
@JiayiXuu @PaulHax ,

Addressing an important use case, can we support at 1 component intensity, 1 component label image (the label image does not have lighting applied)?

Multi-intensity-component lighting would excellent, but we could start with one.

CC @aylward @bnmajor

JiayiXuu:
> Addressing an important use case, can we support at least 1 component intensity, 1 component label image (the label image does not have lighting applied)?

So having cinematic rendering for 1st component and not the rest, or just lighting? @thewtex

thewtex:
> cinematic rendering for 1st component and not the rest, or just lighting?

@JiayiXuu what do you mean?

JiayiXuu:
> > cinematic rendering for 1st component and not the rest, or just lighting?
> 
> @JiayiXuu what do you mean?

Could you elaborate on "1 component intensity"? Direct volume rendering with gradient shadow is applied to the component which has `SetShade = true`. We can enable cinematic rendering for the first component - is that the feature you are requesting?

thewtex:
> Could you elaborate on "1 component intensity"? Direct volume rendering with gradient shadow is applied to the component which has SetShade = true. We can enable cinematic rendering for the first component - is that the feature you are requesting?

Yes :-)

However, [according to the documentation](https://kitware.github.io/vtk-js/api/Rendering_Core_VolumeProperty.html),

> Shading can be independently turned on/off per component.


By "label image component", I mean the [VolumeProperty OpacityMode is set to PROPORTIONAL](https://kitware.github.io/vtk-js/api/Rendering_Core_VolumeProperty.html#Introduction) for that component, in which case we would not perform shading, per:

> - in some cases shading does not apply (for example, in a maximum intensity projection) and therefore shading will not be performed even if this flag is on

JiayiXuu:
@PaulHax Looks good to go!

thewtex:
Per discussion with @JiayiXuu , we are merging now to fix volume rendering when cinematic rendering is not used. @JiayiXuu will create an issue to discuss how to implement cinematic volume rendering for multiple components in the future.

PR #2523, opened 2022-07-19
docs(Stick/Sphere): make stick and sphere examples surface

martinken:
The old code had them as wireframe which isn't what
we want.


jourdain:
Since it is just examples and does not affect the release, we should just do `docs(Stick/SphereMapper): Fix example to use surface`

PR #2525, opened 2022-07-19
Local ambient occlusion

JiayiXuu:
### Context
Add local ambient occlusion feature as part of the cinematic rendering module. Volumetric shading often return sharp shadows that completely occlude parts of the volume. LAO renders soft shadow and provides an alternative to work around the issue.

### Results
Images taken with different kernel size and kernel radius, only ambient lighting is on:

Screenshot 2022-07-19 152114

Two other interactive samples are available:
[Ultrasound data](https://kitwaremedical.github.io/vtk.js-render-tests/LAODemoFetus.html)
[Head CT](https://kitwaremedical.github.io/vtk.js-render-tests/LAODemoHead.html)

Random directions are sampled by combining `Math.random()` function and shader's jitter texture. Final light translucency is thresholded to be greater than 0.3 to reduce variance. To speed up rendering, full brightness is returned at sample positions with `gradient magnitude = 0` or `opacity < 0.05`, hence fog-like tissue can be seen in above image. Fully brightness is also returned at clipping plane (not shown in image). 

Speckle is still visible at some parameter settings, which can be improved by increasing the diffuse factor.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Windows 10

### DIFF
--- a/Sources/Rendering/Core/VolumeMapper/index.d.ts
+++ b/Sources/Rendering/Core/VolumeMapper/index.d.ts
@@ -95,6 +95,24 @@ export interface vtkVolumeMapper extends vtkAbstractMapper {
 	getAnisotropy(): number;
 
 	/**
+	 * Get local ambient occlusion flag
+	 * @default false
+	 */
+	getLocalAmbientOcclusion(): boolean;
+
+	/**
+	 * Get kernel size for local ambient occlusion
+	 * @default 15
+	 */
+	getKernelSize(): number;
+
+	/**
+	 * Get kernel radius for local ambient occlusion
+	 * @default 7
+	 */
+	getKernelRadius(): number;
+
+	/**
 	 * 
 	 * @param x 
 	 * @param y 
@@ -186,7 +204,28 @@ export interface vtkVolumeMapper extends vtkAbstractMapper {
 	 * Value of -1.0 means light scatters backward, value of 1.0 means light scatters forward.
 	 * @param anisotropy
 	 */
-	setAnisotropy(anisotropy: number): number;
+	setAnisotropy(anisotropy: number): void;
+
+	/**
+	 * Set whether to turn on local ambient occlusion (LAO). LAO is only effective if shading is on and volumeScatterBlendCoef is set to 0.
+	 * LAO effect is added to ambient lighting, so the ambient component of the actor needs to be great than 0.
+	 * @param localAmbientOcclusion
+	 */
+	setLocalAmbientOcclusion(localAmbientOcclusion: boolean): void;
+
+	/**
+	 * Set kernel size for local ambient occlusion. It specifies the number of rays that are randomly sampled in the hemisphere.
+	 * Value is clipped between 1 and 32.
+	 * @param kernelSize
+	 */
+	setKernelSize(kernelSize: number): void;
+
+	/**
+	 * Set kernel radius for local ambient occlusion. It specifies the number of samples that are considered on each random ray.
+	 * Value must be greater than or equal to 1.
+	 * @param kernelRadius
+	 */
+	setKernelRadius(kernelRadius: number): void;
 
 	/**
 	 * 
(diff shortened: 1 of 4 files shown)


Sources/Rendering/Core/VolumeMapper/index.d.ts
```diff
@@ -94,6 +94,24 @@ export interface vtkVolumeMapper extends vtkAbstractMapper {
...
+	 * @default 15
+	 */
+	getKernelSize(): number;
+
+	/**
+	 * Get kernel radius for local ambient occlusion
+	 * @default 7
+	 */
+	getKernelRadius(): number;
+
```
floryst, line 114: How about `getLAOKernelSize` and `getLAOKernelRadius` for clarity? (or `getLaoKernelSize` if you want the model property name to be `laoKernelSize`).

Sources/Rendering/Core/VolumeMapper/index.js
```diff
@@ -93,6 +93,16 @@ function vtkVolumeMapper(publicAPI, model) {
     model.anisotropy = vtkMath.clampValue(at, -0.99, 0.99);
     publicAPI.modified();
   };
+
+  publicAPI.setKernelSize = (ks) => {
+    model.kernelSize = vtkMath.floor(vtkMath.clampValue(ks, 1, 32));
```
floryst, line 98: If you have a `const superClass = { ...publicAPI }` at the beginning of the vtkVolumeMapper func block, you can instead do the following:

```
publicAPI.setKernelSize = (ks) => 
  superClass.setKernelSize(vtkMath.floor(vtkMath.clampValue(ks, 1, 32)));`
```

This will use the setGet auto-generated method, which calls modified() for you and returns a boolean as to whether the value changed.

Sources/Rendering/Core/VolumeMapper/index.d.ts
JiayiXuu: Sounds good. Just changed the names.

Sources/Rendering/Core/VolumeMapper/index.d.ts
```diff
@@ -94,6 +94,24 @@ export interface vtkVolumeMapper extends vtkAbstractMapper {
...
+	 * Get local ambient occlusion flag
+	 * @default false
+	 */
+	getLocalAmbientOcclusion(): boolean;
+
+	/**
+	 * Get kernel size for local ambient occlusion
+	 * @default 15
+	 */
+	getLaoKernelSize(): number;
```
martinken, line 107: Super minor but generally we capitalize abbreviations so it would be LAO

martinken (approved):
LGTM other than Lao to LAO

Sources/Rendering/Core/VolumeMapper/index.js
martinken: Funny how people like different things. I almost added a comment saying to not use superClass like this because what you are doing does not invoke the superclass, but they I saw your comment suggesting to do it this way :-) All good with me, just funny.

floryst: Yeah, technically there isn't anything involving the superclass, but I never did think of a better name. I just preferred making a copy of the publicAPI up front when augmenting the macro-added functions, and "superclass" was the closest I could get.

Sources/Rendering/Core/VolumeMapper/index.js
```diff
@@ -74,24 +76,29 @@ function vtkVolumeMapper(publicAPI, model) {
...
-    model.globalIlluminationReach = vtkMath.clampValue(gl, 0.0, 1.0);
-    publicAPI.modified();
+    superClass.setGlobalIlluminationReach(vtkMath.clampValue(gl, 0.0, 1.0));
   };
 
   publicAPI.setVolumetricScatteringBlending = (vsb) => {
-    model.volumetricScatteringBlending = vtkMath.clampValue(vsb, 0.0, 1.0);
-
-    publicAPI.modified();
+    superClass.setVolumetricScatteringBlending(
```
floryst, line 83: Best to return the return value of the inner `set` calls, since setters return a boolean whether the value changed or not.

bruyeret:
Hello @JiayiXuu,
I am working on the multi volume mapper and this requires a big reformat of the volume mapper shader.
I need to understand this part of LAO:
> Random directions are sampled by combining Math.random() function and shader's jitter texture.
Why do you combine these two functions instead of only using the shader's jitter texture (the custom `random` function)?

PR #2526, opened 2022-07-19
fix(InteractorStyleTrackballCamera): null guard

floryst:
### Context
- Starting a dolly before mouse move or mouse press and then moving the mouse results in usage of `model.previousPosition` when it's not yet set.

### Results

- No more console errors for this corner case.

### Changes

- [x] Guard against `model.previousPosition` being null/undefined.

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


finetjul:
LGTM

PR #2529, opened 2022-07-20
feat(WebGPU): Added environment mapping

andrewcombs:
Added environment mapping to WebGPU. The background is controlled by the renderer's BackgroundTexture, and the strength of it's diffuse and specular contributions can be set with setBackgroundTextureDiffuseStrength and setBackgroundTextureSpecularStrength respectively (both default to 0.5). In order for surface roughness to work properly, mipmaps should be specified for the BackgroundTexture by setting its mipLevel. Diffuse lighting is specified by the last mipmap, so generating until a 1x1 texture is not recommended.

### Context
Although physically based rendering allows for more realistic and interesting rendering of data, it is incomplete without environment mapping. Extremely shiny, metallic, or reflective materials are often flat or misrepresented due to being heavily weighted with reflections, of which there were none. Currently, fully metallic portion of a model appears completely black except for the small reflection(s) on it's surface, as seen below.

<div align="center">
issue
</div>

This makes setting up a scene to accurately illuminate certain objects very tedious.

### Results
To fix these problems, the renderer's backgroundTexture has been used to provide both specular and diffuse illumination to objects within a scene. Below is the same scene as seen above in context, this time with environment mapping.

<div align="center">
result
</div>

By using an equirectangular image as lighting, the object appears both more interesting and more realistic. Reflective and metallic surfaces are now actually reflective, and the image feels in general more grounded.

### Changes
- When present, environment maps (set by renderer's backgroundTexture) will contribute light to all actors of that renderer which are using PBR lighting.
- vtkTexture now includes a mipLevel property, which, when set, will result in the texture having that many levels of mipmaps generated. **It is strongly recommended that texture dimensions are powers of two, as having anything else may cause the mipmap generation to fail.**
- Renderer now includes backgroundTextureDiffuseStrength and backgroundTextureSpecularStrength which set their respective values for the background texture.
- Renderer includes a useTextureAsViewBackground option which shows the background texture behind the scene
- vtkWebGPUSampler now includes support for setting a texture's address modes for U, V, and W properties, along with mipmap filter mode.
- the PBR example has been updated to include an environment map with sliders to control the diffuse and specular intensity, as well as one for field of view.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 25.2.3
  - **OS**: Windows 10
  - **Browser**: Chrome Canary 105.0.5189.0


package.json
```diff
@@ -110,7 +110,7 @@
     "rollup-plugin-polyfill-node": "0.9.0",
     "rollup-plugin-postcss": "4.0.2",
     "rollup-plugin-string": "3.0.0",
-    "rollup-plugin-svgo": "1.1.0",
+    "rollup-plugin-svgo": "^2.0.0",
```
floryst, line 113: Why were these packages updated/version strings changed? If these are not related to this PR, we should move them to a separate PR.

Sources/Rendering/Core/Texture/index.js
```diff
@@ -132,15 +267,21 @@ export function extend(publicAPI, model, initialValues = {}) {
...
 
 // ----------------------------------------------------------------------------
 
 export const newInstance = macro.newInstance(extend, 'vtkTexture');
+export const STATIC = { generateMipmaps };
 
 // ----------------------------------------------------------------------------
 
-export default { newInstance, extend };
+export default { newInstance, extend, ...STATIC };
```
floryst, line 287: This addition should also be reflected in `Texture/index.d.ts`.

Sources/Rendering/WebGPU/CellArrayMapper/index.js
```diff
@@ -699,7 +703,27 @@ function vtkWebGPUCellArrayMapper(publicAPI, model) {
         '  var PBR: vec3<f32> = mapperUBO.DiffuseIntensity*kD*diffuse + kS*specular;',
         '  PBR += emission;',
         '  computedColor = vec4<f32>(PBR, mapperUBO.Opacity);',
+        '  //VTK::Light::Impl', // TODO: Find a better way of doing this, because it's ugly
```
floryst, line 706: You can consider conditionally including (or appending) the below shader code here, if you don't want to use this temporary substitution marker.

Examples/Rendering/PBR/index.js
```diff
@@ -98,12 +116,18 @@ reader.setUrl(`${__BASE_PATH__}/data/pbr/helmet.obj`).then(() => {
   const metallicTex = createTexture(`${__BASE_PATH__}/data/pbr/metallic.jpg`);
   const emissionTex = createTexture(`${__BASE_PATH__}/data/pbr/emission.jpg`);
   const normalTex = createTexture(`${__BASE_PATH__}/data/pbr/normal.jpg`);
-
+  const environmentTex = createTextureWithMipmap(
+    `${__BASE_PATH__}/data/pbr/kiara_dawn_4k.png`,
```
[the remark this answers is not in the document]
andrewcombs: Because of the fact that this is being used as the background for the scene (and normally only about 1/10th of it is being shown due to field of view), it's very important that it stays as being high quality. Would it be better to store the image somewhere else and then download it when the example starts?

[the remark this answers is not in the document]
martinken: Another option would be to try keeping the resolution but increasing the compression? That might work fine.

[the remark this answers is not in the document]
floryst: In that case, I would add it to https://github.com/Kitware/vtk-js-datasets and have the example download the image from there.

Examples/Rendering/PBR/controller.html
```diff
@@ -35,3 +35,27 @@ <h3>PBR Properties</h3>
     </td>
   </tr>
 </table>
+<table>
+  <h3>Environment Properties</h3>
+  <tr>
+    <td>
+      <label>FOV: </label>
+      <input class='angle' type='range' min='30' max='120' value='30' step='1' oninput="this.nextElementSibling.value = this.value"/>
```
martinken, line 43: I would default this to 75 to get a better looking background by default

Sources/Rendering/Core/Renderer/index.d.ts
```diff
@@ -35,6 +35,9 @@ export interface IRendererInitialValues extends IViewportInitialValues {
 	occlusionRatio?: number;
 	maximumNumberOfPeels?: number;
 	texturedBackground?: boolean;
+	backgroundTextureDiffuseStrength?: number;
+	backgroundTextureSpecularStrength?: number;
+	useTextureAsViewBackground?: boolean;
```
martinken, line 40: Thinking about this some more what do you think of
- environmentTexture
- environmentTextureDiffuseStrength
- environmentTextureSpecularStrength
- useEnvironmentTextureAsBackground

On the C++ side we support using a texture map as a static background image and I'm thinking this may help clarify the difference.

Sources/Rendering/Core/Texture/index.d.ts
```diff
@@ -5,6 +5,7 @@ export interface ITextureInitialValues {
 	interpolate?: boolean;
 	edgeClamp?: boolean;
 	imageLoaded?: boolean;
+	mipLevel?: number;
```
martinken, line 8: Missing setter getter in the typescript file I think

Sources/Rendering/Core/Texture/index.js
```diff
@@ -100,20 +100,155 @@ function vtkTexture(publicAPI, model) {
     const dimensionality = (width > 1) + (height > 1) + (depth > 1);
     return dimensionality;
   };
+
+  publicAPI.getImageData = () => {
```
martinken, line 104: Probably need to call this getJSImageData as people who use VTK will assume it is returning vtkImageData

Sources/Rendering/Core/Texture/index.js
```diff
@@ -100,20 +100,155 @@ function vtkTexture(publicAPI, model) {
...
+    }
+
+    return null;
+  };
 }
 
+// Use nativeArray instead of self
+const generateMipmaps = (nativeArray, width, height, level) => {
+  // TODO: FIX UNEVEN TEXTURE MIP GENERATION:
+  // When textures don't have standard ratios, higher mip levels
```
martinken, line 145: Not an expert. I believe the standard approach is to filter first, then sample. (averaging the four pixels is a center point linear sampling) As written the level 0 texture is not being filtered prior to sampling and the other resolutions are being stored in their post filtered form. Might be worth allocating memory and filtering first.

Sources/Rendering/WebGPU/CellArrayMapper/index.js
```diff
@@ -616,11 +617,14 @@ function vtkWebGPUCellArrayMapper(publicAPI, model) {
 
   // we only apply lighting when there is a "var normal" declaration in the
   // fragment shader code. That is the lighting trigger.
+  // TODO: Fix frame tanking when zooming in close to models
```
martinken, line 620: Is this still a TODO? I am getting solid 60fps on my laptop regardless of zoom but I have a fairly good GPU

Sources/Rendering/WebGPU/Renderer/index.js
```diff
@@ -301,18 +356,63 @@ function vtkWebGPURenderer(publicAPI, model) {
...
+          addressModeV: 'clamp-to-edge',
+          addressModeW: 'repeat',
+          minFilter: interpolate,
+          magFilter: interpolate,
+          mipmapFilter: 'linear',
+        });
+      }
+      model.clearFSQ.setPipelineHash('clearfsqwithtexture');
+    }
+    const keyMats = model.webgpuCamera.getKeyMatrices(publicAPI);
```
martinken, line 405: I think you need an else clause here (or similar) in case they use textured background and then turn it off

martinken (changes requested):
Should remove the package* files from this MR as well since they are not changing

Sources/Rendering/WebGPU/CellArrayMapper/index.js
andrewcombs: Running it now on my main computer (which actually has a graphics unit) it runs flawlessly, so it was probably the laptop

Sources/Rendering/WebGPU/Renderer/index.js
andrewcombs: Fixed. New PBR example includes a checkbox to enable and disable background at runtime

Sources/Rendering/Core/Texture/index.js
andrewcombs: The way I have it set up right now, the level zero texture is the initial unchanged input data which should be unfiltered. Any new maps that are generated go through the linear down sample and then blur, and I haven't seen much difference by switching it the other way around

Sources/Rendering/Core/Renderer/index.d.ts
andrewcombs: that would make more sense to me, but I didn't change it because the renderer already had a backgroundTexture property and I didn't want to change the names of any values that people may already be using. If renaming it is okay then I'm all for it

Examples/Rendering/PBR/controller.html
andrewcombs: I'm having a strange issue where if I change the field of view before the initial rendering, zooming out immediately causes the entire screen to turn one color. I'm not sure why it's happening or how to fix it

Sources/Rendering/Core/Texture/index.js
andrewcombs: JsImageData is already a property in texture, is there something else you'd prefer it to be called?

Sources/Rendering/Core/Renderer/index.d.ts
martinken: Yeh, let''s use EnvironmentTexture that is what it is called in C++, backgroundTexture is a static 2D wallpaper type texture .

Sources/Rendering/Core/Texture/index.js
martinken: getInputAsJSImageData

martinken: though it doesn't seem to handle the vtkImageData case, is that intended?

Sources/Rendering/Core/Texture/index.js
martinken: you make a copy of the level 0, filter that and then sample it to get level 1. But this isn't my area of expertise. Do you have a reference that says to sample first then smooth? From wikipedia "A lowpass pyramid is made by smoothing the image with an appropriate smoothing filter and then subsampling the smoothed image, usually by a factor of 2 along each coordinate direction." But if you have a reference to sampling first then smoothing (or if someone who's background is image processing knows better speak up).

Sources/Rendering/Core/Texture/index.js
andrewcombs: I implemented it mostly so I could easily get raw image data from canvas and html images while testing mipmap generation.  It's not being used right now but I thought it would be a nice convenience feature for the two types of images I intended it for since they're pretty annoying to get data from

Examples/Rendering/PBR/index.js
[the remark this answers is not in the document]
martinken: An 80 quality jpg is 1.1 MB and looks really good FWIW

Sources/Rendering/Core/Texture/index.js
martinken: And honestly I might even just skip the filtering if you don;t see much of a difference, especially if it takes a while

andrewcombs: I couldn't find anything really strong supporting one or the other after looking through a host of different resources. The method of mipmapping the environment map is an approximation to begin with, but a good approximation that saves time over actually sampling the map with progressively more scattered samples. I did some tests on sampling then blurring vs blurring then sampling, and the main difference is that the texture gets blurred faster when it's sampled first vs blurred first. The other difference is that it over doubles the time required to compute the mipmaps, since the base texture alone has more data than all the other mipmaps combined. Additionally, less blur may result in needing to use a kernel that's larger than 3x3 which would slow it down further. I'm fine with either, but from a speed and ease perspective sampling-then-blurring makes more sense.

Examples/Rendering/PBR/index.js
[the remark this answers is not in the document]
andrewcombs: Ive updated it to be a pre-tonemapped compressed jpeg, now with a size of 1.6MB

Sources/Rendering/WebGPU/Renderer/index.js
```diff
@@ -301,18 +356,77 @@ function vtkWebGPURenderer(publicAPI, model) {
...
+      ubo.addEntry('BackgroundColor', 'vec4<f32>');
+      model.clearFSQ.setUBO(ubo);
+    }
+
+    const keyMats = model.webgpuCamera.getKeyMatrices(publicAPI);
     const background = model.renderable.getBackgroundByReference();
+
     model.clearFSQ.getUBO().setArray('BackgroundColor', background);
+    const fsqMatrix = [];
+    const transposedNormals = [];
```
martinken, line 424: Let's allocate these two as statics so that we are not allocating and destroying them each frame. Ala 

const _fsqClearMat4 = new Float64Array(16);

Examples/Rendering/PBR/controller.html
martinken: Weird.  I change the html to value=75 and it just seems to work, but maybe a graphics driver thing?

martinken:
Something odd with the occlusion edges, like the lighting equation is blowing up or something. See the bright pixels at the edges of the tubes

image

Sources/Rendering/WebGPU/Renderer/index.js
andrewcombs: Should this be put somewhere that's only run once or is it fine to just directly swap [] with new Float64Array(16)? I'm not super familiar with javascript allocation

martinken: Look at WebGPU\ImageMapper for an example

andrewcombs:
> Something odd with the occlusion edges, like the lighting equation is blowing up or something. See the bright pixels at the edges of the tubes
That's a result of the poor normals. They work well when head on, but around edges they get progressively more skewed and incorrect. In this case they're stretching out what should be only a few pixels to wrap around the tube

martinken:
> 

That color (full white and overblown orange) should not be coming from that tube (high roughness) for any normal vector that I can think of. It really feels like an overflow, bad divide, missing min/max or similar.

andrewcombs:
The full blown color turned out to be a clamping which was just a little too high. The issue is that pitch black zones are now there in the overblown areas place. I've narrowed it down to the fact that the dot product between the normal and view vector is wrong, and I have no clue why. The first image is normal dot view in vtkjs and the second is the same thing (both vectors are in view space) in blender. For some reason, it is significantly darker in vtkjs and I cant figure out why. I've tried color space corrections, clamping, normalizing, etc. and nothing fixes it.

ndotv-webgpu
ndotv-blender

martinken:
The blender image looks tone mapped or something. The darks are lighter and the lights are darker. Just less contract in it.

So I think you are right that the added max commented below
```
fn smithSurfaceRoughness(N: vec3<f32>, V: vec3<f32>, L: vec3<f32>, k: f32) -> f32 {
  // var ggx1: f32 = max(0.01, schlickGGX(N, V, k)); // Prevents void zones at the cost of some accuracy
  var ggx1: f32 = schlickGGX(N, V, k);
  var ggx2: f32 = schlickGGX(N, L, k);
  return ggx1*ggx2;
}
```
is the issue. When I remove that the sparkles go away and I get why. When the V*N is zero or less then we are setting the numerator to 0.01 while limiting the denominator to 0.0000001 which causes the resulting hot spots.

andrewcombs:
The max was the problem with the sparkles, but there is still the issue that the dot product between them is not working correctly. If a fragment is visible and not back facing, which the ones visible in the example are, there should be no possible way that the dot product between the view and normal is zero. However, using an if statement, it is very clear that the dark values are zero. This means that it can't be tone mapping (which I tried anyway, just to make sure). I would send screenshots, but chrome canary has decided to remove unsafe webgpu so I can't do anything for now.

martinken:
Normals do not have to match the geometry. And often don't.  This model uses point normals which are interpolated across the surface primitives. And there are cases in this model where the normal is quite a bit off from the primitive's actual flat normal. Think of a sphere modeled using a cube with shared point normals sticking out from the 8 corners. With normal interpolation the shading will look sort of sphere-like. But you can easily get views where a face is visible, but some of the interpolated normals end up pointing away from the view vector. In real life that cannot happen, but when we use a cube to low res simulate a sphere it happens a lot. As a test if you take the normals out of the OBJ file you will see those odd conditions pretty much disappear. The rendering becomes flat shaded with cell normals and the impossible case goes away.

andrewcombs:
Ok, I'm glad it is not an underlying issue with how the rendering system is set up. Do you have any suggestions on how to deal with those areas? If not, is there anything else I should change before I push?

martinken:
If just removing the max from smithSurfaceRoughness works then I'd just do that. If not then I think we decide how we want to handle those odd cases where the normal is pointing away from view. I would either
1) add V into the normal until VdotN is some small value 0.001 say and then use that normal for all lighting eqns
2) limit VdotN early on to be some non zero minimum and use that everywhere without modifying normal

martinken:
Also note that here 
```
          '  normal = mix(input.normalVC, normal, mapperUBO.NormalStrength);',
```
that mix is mixing the non-front-face correct normal with a front-face-corrected value

andrewcombs:
I've fixed the front facing normals, but am still working on the normals issue. The most promising way I've found so far is with a smooth clamp from [-1, 1] to [0, 1], which looks like this:

$$ NdotV = (\frac{(N\cdot V) + 1}{2})^{2.6} $$

The 2.6 can be swapped with 2.45 which is better optimized for values 0.35 and above or 2.8 which is better optimized for values below 0.25, however 2.6 is overall the closest.

martinken (approved):
LGTM +1

PR #2544, opened 2022-08-04
feat(Core): Various additions to core functionalities

DavidBerger98:
CellArray
- Implemented insertNextCell
- Implemented initialize
- getNumberOfCells and getCellSizes now use publicAPI.getData() instead of model.values

DataArray
- Implemented getValue and setValue
- Implemented insertTuple
- Implemented insertNextTuple
- Implemented deepCopy
- Implemented interpolateTuple
- getRange now uses publicAPI.getData() instead of model.values
- getNumberOfValues and getNumberOfTuples now use model.size instead of model.values.length
- model.size defaults to model.values.length if not specified during initialization

Points
- Implemented insertNextPoint
- Added tests

FieldData
- Implemented interpolateData
- Changes to passData and a bugfix
- addArray now checks if an array with the given name already exists and replaces it in that case.

DataSetAttributes
- Implemented getAttributes
- Implemented copyScalarsOn
- Implemented copyVectorsOn
- Implemented copyNormalsOn
- Implemented copyTCoordsOn
- Implemented copyTensorsOn
- Implemented copyGlobalIdsOn
- Implemented copyPedigreeIdsOn

Co-authored-by: Julien Finet <julien.finet@kitware.com>


DavidBerger98:
@finetjul

Sources/Common/Core/DataArray/index.d.ts
```diff
@@ -199,7 +282,7 @@ export function getMaxNorm(dataArray: vtkDataArray): number
  *
  * @param publicAPI object on which methods will be bounds (public)
  * @param model object on which data structure will be bounds (protected)
- * @param {object} [initialValues] (default: {})
+ * @param {object} [initialValues] (default: {}) Must pass a number > 0 for `size` except if `empty: true` is also passed or a non-empty typed array for `values`. 
  */
 export function extend(publicAPI: object, model: object, initialValues?: object): void;
```
daker, line 287: initialValues?: IDataArrayInitialValues

Sources/Common/DataModel/DataSetAttributes/index.d.ts
```diff
@@ -112,6 +112,14 @@ export interface vtkDataSetAttributes extends vtkFieldData {
 	 */
 	getActiveAttribute(attType: string): any;
 
+	/**
+	 * Get a list of attribute names that the given array 
+	 * is for this vtkDataSetAttributes.
+	 * @param {vtkDataArray} arr
+	 * @returns {String[]}
+	 */
+	getAttributes(arr: vtkDataArray): string[]
```
daker, line 121: ; at the end

daker (changes requested):

finetjul:
> @finetjul I have some additional suggestions:
> 
> - change `vtkDataArray.getTuple` to use the switch-statement like in `interpolateTuple`:
> 
> ```js
> switch (numberOfComponents) {
>   case 4:
>     tupleToFill[3] = model.values[offset + 3];
>   // eslint-disable-next-line no-fallthrough
>   case 3:
>     tupleToFill[2] = model.values[offset + 2];
>   // eslint-disable-next-line no-fallthrough
>   case 2:
>     tupleToFill[1] = model.values[offset + 1];
>   // eslint-disable-next-line no-fallthrough
>   case 1:
>     tupleToFill[0] = model.values[offset];
>     break;
>   default:
>     for (let i = 0; i < numberOfComponents; i++) {
>       tupleToFill[i] = model.values[offset + i];
>     }
> }
> ```

Sounds good to me.
> 
> - `vtkPoints.setPoint` does not work as "advertised" in `index.d.ts`. In fact the second `t.deepEqual` in the test I added in this commit fails. Either we remove the second annotation of `setPoint` in `index.ts`:
> 
> ```js
> /**
>  * Insert point into object.
>  * @param {Number} idx The index of point.
>  * @param {Number[]} coord The point's coordinates.
>  */
> setPoint(idx: number, coord: number[]): void;
> ```
> 
>   and set `publicAPI.setPoint = (idx,x,y,z) => publicAPI.setTuple(idx,[x,y,z])`, since it pretty much re-implements `vtkDataArray.setTuple` at this state.
> Or we set `publicAPI.setPoint = publicAPI.setTuple` and remove the first annotation?
> 

I say we expose both signatures in setPoint.

> - set `recompute = true` as a default value for `vtkCellArray.getNumberOfCells` and `vtkCellArray.getCellSizes`.
> It is easily forgotten to do `cellArray.getNumberOfCells(true)` so sometimes we can get the wrong value when expecting another one. By default it should return the correct value. Setting `cellArray.getNumberOfCells(false)` is an optimization for advanced users.

No, `recompute=true` as default would be too expensive. There are only 3 case where the number of cells can change (at construction, when `setData` is called and when `insertNextCell`). I think all cases are covered and I don't see when getNumberofCells can be wrong.

DavidBerger98:
> I say we expose both signatures in setPoint.

How would you go about doing this? I suspect checking the type of `x` being either `Array` or `Number` would be too slow.

PR #2532, opened 2022-07-22
Add vtkClipClosedSurface

DavidBerger98:
Implement vtkClipClosedSurface with vtkContourTriangulator, vtkIncrementalOctreePointLocator and vtkIncrementalOctreeNode.

- [x] vtkClipClosedSurface
  - [x] Implementation
  - [x] Documentation
  - [x] Tests
  - [x] Example
- [x] vtkContourTriangulator
  - [x] Implementation
  - [x] Documentation
  - [x] Tests
- [x] vtkLocator
  - [x] Implementation
  - [x] Documentation
  - [x] Tests
- [x] vtkAbstractPointLocator
  - [x] Implementation
  - [x] Documentation
  - [x] Tests
- [x] vtkIncrementalOctreePointLocator
  - [x] Implementation
  - [x] Documentation
  - [x] Tests
- [x] vtkIncrementalOctreeNode
  - [x] Implementation
  - [x] Documentation
  - [x] Tests

PR #2547, opened 2022-08-10
Centroid computation

joannacirillo:
### Context

Translation of C++ functions of vtkPolygon.
Note: this PR is to be merged after PR #2532 as it uses functions implemented in the latter.

### Results

Implementation of
[x] computeArea
[x] computeCentroid
[x] isConvex
[x] interpolateFunctions
[x] intersectWithLine
[x] intersectConvex2DCell

Each of them are also exported as static functions.

### Changes

The `computeNormal` function has been replaced by the static function `getNormal` exported in the publicAPI as `computeNormal`.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  commit 82d04acdd95c7c3d218a67c1c920cb6cd7307b55
  - **OS**: Windows 10
  - **Browser**: Firefox 103.0.1


joannacirillo:
@finetjul The `pointInPolygon` function implemented in #2532 accepts vtkPoints along with an id list as parameter, I was wondering whether it should also be the case for the newly implemented static functions.

Sources/Common/DataModel/Polygon/index.js
```diff
@@ -174,45 +193,629 @@ export function getBounds(poly, points, bounds) {
...
- * but in a more generic way. Maybe we can keep the public
- * static method somehow and have the private method use it instead.
- *
- * @param {Array<Number>|TypedArray<Number>} poly
+ * @param {Vector3[]} poly
  * @param {vtkPoints} points
  * @param {Vector3} normal
  * @returns {Number}
  */
 export function getNormal(poly, points, normal) {
```
finetjul, line 203: let's keep the API simple, only support `vtkPoints` instead of a points/poly pair.

So you need to modify setPoints to optionally take in a list of pointIds: `setPoints(points : [[],[],[]]|[], pointIds = null|[])`

Sources/Common/DataModel/Polygon/Constants.js
```diff
@@ -6,6 +6,13 @@ export const PolygonWithPointIntersectionState = {
...
-  INTERSECTION: 2,
-  ON_LINE: 3,
 };
+export const VTK_DBL_EPSILON = 2.2204460492503131e-16;
+
+export const PolygonWithPolygonIntersectionState = {
+  NO_INTERSECTION: 0,
+  POINT_INTERSECTION: 1,
+  LINE_INTERSECTION: 2,
+};
```
finetjul, line 16: why can't you have failure in this case ?
what about the case where the whole polygon is within another polygon

Sources/Common/DataModel/Polygon/Constants.js
```diff
@@ -6,6 +6,13 @@ export const PolygonWithPointIntersectionState = {
   FAILURE: -1,
   OUTSIDE: 0,
   INSIDE: 1,
-  INTERSECTION: 2,
-  ON_LINE: 3,
```
finetjul, line 10: can't a point be both OUTSIDE AND INSIDE, hence ON_LINE ?

joannacirillo: In the C++ implementation, this case returns inside the polygon. It could indeed differentiate that case however.

Sources/Common/DataModel/Polygon/index.js
```diff
@@ -174,45 +193,629 @@ export function getBounds(poly, points, bounds) {
...
-  return vtkMath.normalize(normal);
+  const norm2 = vtkMath.dot(nn, nn);
+  if (norm2 > 0) {
+    const norm = Math.sqrt(norm2);
+    normal[0] = nn[0] / norm;
+    normal[1] = nn[1] / norm;
+    normal[2] = nn[2] / norm;
+  }
+
+  return norm2;
```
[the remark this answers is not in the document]
joannacirillo: Here the norm is already computed anyway, so `norm` can be returned. I think checking if `norm2` is positive does not make sense since this could not happen with real vectors and vectors having values such as `NaN` or `undefined` or `null` would anyway throw while computing the dot product.

Sources/Common/DataModel/Polygon/Constants.js
joannacirillo: One polygon coincident with the other is now implemented.

joannacirillo: The failure case is when the normal is the zero vector, hence all points are on one line. The computation done would still work on that case since it loops over all edges of one polygon, see if it intersects with the other polygon and does the same on the second polygon if needed.
I would say whether we consider a failure possible depends on whether we want to consider all points being on one line acceptable.
For `pointInPolygon`, the normal being zero vector causes a failure since one cannot computed the winding number if all points are on one line.

Sources/Common/DataModel/Polygon/index.js
```diff
@@ -334,122 +1127,162 @@ function vtkPolygon(publicAPI, model) {
...
 
 // ----------------------------------------------------------------------------
 // Object factory
 // ----------------------------------------------------------------------------
 
 const DEFAULT_VALUES = {
-  firstPoint: null,
-  pointCount: 0,
-  tris: [],
+  points: [],
```
finetjul, line 1269: model.points is actually defined in vtkCell as vtkPoints. I doubt it should be [] here

Sources/Common/DataModel/Polygon/index.js
```diff
@@ -334,122 +1127,162 @@ function vtkPolygon(publicAPI, model) {
...
-      };
-      currentPoint = currentPoint.next;
-    }
-
-    model.firstPoint.previous = currentPoint;
-    currentPoint.next = model.firstPoint;
+  publicAPI.setPoints = (points, pointsIds) => {
+    let pointsData = points;
+    if (Array.isArray(pointsData[0])) {
+      pointsData = points.flat(2);
```
finetjul, line 1214: shouldn't it be `points.flat()` instead?

Sources/Common/DataModel/Polygon/index.js
```diff
@@ -334,122 +1127,162 @@ function vtkPolygon(publicAPI, model) {
...
+    }
+    model.points = vtkPoints.newInstance({ values: pointsData });
+    if (!pointsIds) {
+      model.pointsIds = new Array(pointsData.length / 3);
+      for (let i = 0; i < points.length; i++) model.pointsIds[i] = i;
+    } else {
+      model.pointsIds = pointsIds;
+    }
+    publicAPI.computeNormal();
```
finetjul, line 1223: should probably call `modified()`

Sources/Common/DataModel/Polygon/index.js
```diff
@@ -334,122 +1127,162 @@ function vtkPolygon(publicAPI, model) {
...
+          );
         }
       }
     }
 
-    return model.pointCount <= 2;
+    return idsToHandle.length <= 2 ? triangles : null;
   }
 
+  publicAPI.computeNormal = () => getNormal(model.points, model.normal);
```
finetjul, line 1199: here you ignore `model.pointIds`, are you sure it is intended ?

PR #2556, opened 2022-08-24
fix(dataarray): getTuple

joannacirillo:
### Context

On current master:
Assuming `da` is a vtkDataArray
```
const p1 = da.getTuple(0);
const p2 = da.getTuple(10);

const p3 = [];
const p4 = [];
da.getTuple(0, p3);
da.getTuple(10, p4);

console.log(p1 === p2) // prints true
console.log(p3 === p4) // prints false
```
The function currently allow to not give a tuple to fill or not and uses a tuple holder if no tuple is given. The tuple holder is however common to the hole class, which causes calling of `getTuple` twice with no tuple on the same instance of the dataArray not working as intended

### Results

Fix the `getTuple` function. 

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: v25.7.2
  - **OS**: Windows 10
  - **Browser**: Chromium 100.0.4852.0

### DIFF
--- a/Sources/Common/Core/DataArray/index.js
+++ b/Sources/Common/Core/DataArray/index.js
@@ -4,7 +4,6 @@ import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
 
 const { vtkErrorMacro } = macro;
 const { DefaultDataType } = Constants;
-const TUPLE_HOLDER = [];
 
 // ----------------------------------------------------------------------------
 // Global methods
@@ -103,8 +102,10 @@ function getDataType(typedArray) {
 function getMaxNorm(normArray) {
   const numComps = normArray.getNumberOfComponents();
   let maxNorm = 0.0;
+  const tuple = new Array(numComps);
   for (let i = 0; i < normArray.getNumberOfTuples(); ++i) {
-    const norm = vtkMath.norm(normArray.getTuple(i), numComps);
+    normArray.getTuple(i, tuple);
+    const norm = vtkMath.norm(tuple, numComps);
     if (norm > maxNorm) {
       maxNorm = norm;
     }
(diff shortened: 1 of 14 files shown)


Sources/Common/Core/DataArray/index.js
```diff
@@ -299,8 +298,10 @@ function vtkDataArray(publicAPI, model) {
     return publicAPI.insertTuples(idx, tuples);
   };
 
-  publicAPI.getTuple = (idx, tupleToFill = TUPLE_HOLDER) => {
+  publicAPI.getTuple = (idx, tupleToFill) => {
```
floryst, line 301: How about replacing this with `tupleToFill = []`? It will create a new array instance every invocation, unlike the previous implementation.

floryst:
Thanks for catching this bug!

Sources/Common/Core/DataArray/test/testDataArray.js
```diff
@@ -29,6 +30,30 @@ test('Test vtkDataArray getRange function with single-channel data.', (t) => {
...
+    values: newArray,
+  });
+
+  t.ok(vtkMath.areEquals(da.getTuple(0), [0, 1, 2]), 'get first tuple');
+  t.ok(vtkMath.areEquals(da.getTuple(10), [30, 31, 32]), 'get 10th tuple');
+  const tupleToFill1 = [];
+  const tupleToFill2 = [];
+  da.getTuple(0, tupleToFill1);
+  da.getTuple(10, tupleToFill2);
+  t.ok(!(tupleToFill1 === tupleToFill2), 'getTuple twice');
```
finetjul, line 53: Shouldn't you also verify (which is the reason of this bug fix):
```
t.ok(da.getTuple(10) !== da.getTuple(0)), 'getTuple twice');
```

finetjul:
Adding @jourdain the original author of the `TUPLE_HOLDER`

Sources/Common/Core/DataArray/index.js
jourdain: which could kill the perf when done in a loop.

finetjul: This is indeed a "perf" vs "hard-to-debug" dilemna. Here are some real-life code examples that would be buggish:
```
foo(da.getTuple(0), da.getTuple(1), da.getTuple(2));
vtkMath.areEquals(da1.getTuple(0), da2.getTuple(0));
```

To me, if performance is an issue, `tupleToFill` MUST be provided by caller (to be documented in index.d.ts).

floryst: We could add docs strongly recommending `tupleToFill` be passed in by the user for perf reasons, but leave the signature as `tupleToFill = []` to have a sane default. Additionally, we should rewrite all internal invocations of `getTuple()` to pass in an array.

jourdain: Let's go with no bug... And perf can be fixed if arg is passed... I guess it is better that way

PR #2560, opened 2022-08-25
fix(tubeFilter): update npts and change to global func

### LINKED ISSUE #2559: TubeFilter over allocating output points
### High-level description
It appears that in certain cases (haven't pin pointed the exact cases) that the vtkTubeFilter overallocates output points array. This leaves trailing 0's at the end of the array, consequently causing problems on the `getBounds()` call

### Steps to reproduce
As stated before I am not entirely sure on what inputs  / where the difference comes between the two calculations (I havent looked in too much detail either though ... ) but I can provide points and lines that cause this issue if needed.

### Detailed behavior
There appears to be a difference between the calculation of newNumPts: line: https://github.com/Kitware/vtk-js/blob/24de4f815199a6ae1cab05a06b5e1593877d771a/Sources/Filters/General/TubeFilter/index.js#L634 and the filling of the array in generatePoints. I propose to change the calculations in the forLoop to simply setting the offset and incrementing it (the same that is done in generatePoints (I'd be happy to put together the PR for this)

### Expected behavior
Not to have trailing 0s at the end of the Array that havent been allocated

### Environment
- "@kitware/vtk.js": "^25.7.2",
- **OS**: MacOSX 12.5
- **Browser**: Firefox dev 104

WesleyTheGeolien:
### Context
Fixes #2559 

### Changes
Updated when npts was allocated and changed to use computeOffset which does the same calcualtion but if ever needs changing at least it is only in one place

### PR and Code Checklist

### Testing

Edit and uncomment the section below if relevant

### DIFF
--- a/Sources/Filters/General/TubeFilter/test/testTubeFilter.js
+++ b/Sources/Filters/General/TubeFilter/test/testTubeFilter.js
@@ -181,3 +181,59 @@ test.onlyIfWebGL('Test vtkTubeFilter rendering', (t) => {
   });
   renderWindow.render();
 });
+
+test('Test vtkTubeFilter numberOfPoints', (t) => {
+  const numberOfSides = 3;
+  const numberOfLines = 2;
+  const sidesShareVertices = 1;
+
+  const polyData = vtkPolyData.newInstance();
+
+  const points = new Uint32Array([
+    0, 0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4, 5, 5, 5,
+  ]);
+  const lines = new Uint32Array([2, 0, 1, 4, 2, 3, 4, 5]);
+
+  polyData.getPoints().setData(points);
+  polyData.getLines().setData(lines);
+
+  const tubeFilter = vtkTubeFilter.newInstance();
+  tubeFilter.setCapping(false);
+  tubeFilter.setNumberOfSides(numberOfSides);
+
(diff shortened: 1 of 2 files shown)


WesleyTheGeolien:
MAybe should add a test to catch this?

floryst:
> MAybe should add a test to catch this?

Yes, if you can add a test, it would be greatly appreciated.

Sources/Filters/General/TubeFilter/test/testTubeFilter.js
```diff
@@ -181,3 +181,59 @@ test.onlyIfWebGL('Test vtkTubeFilter rendering', (t) => {
...
+  // Each segment should have numberOfPoints * sidesShareVertices * numberOfSides
+  // sidesShareVertices = 1
+  // numberOfSides = 3
+  // 2 * 1 * 3 + 4 * 1 * 3
+  t.ok(
+    tubeOutput.getPoints().getNumberOfPoints() ===
+      2 * sidesShareVertices * numberOfSides +
+        4 * sidesShareVertices * numberOfSides,
+    'Make sure the output number of points is correct.'
+  );
```
WesleyTheGeolien, line 217: @floryst This test should be sufficient, in the previous implementation the `npts` was incremented after calculating the number of points in the line. This would lead to the size of the first line (2) being used twice.

floryst:
LGTM. If you can rebase onto master, we will be good to merge.

Issue #2567, opened 2022-09-06
vtkDataSetAttributes not exporting all methods


WesleyTheGeolien:
### High-level description
Functions from FieldData are not appearing on the vtkDataSetAttributes

### Steps to reproduce
- Use intellisense or run `npx tsc -b`
- create some polyData
- `polyData.getPointData().getArrays()` 
```bash
error TS2339: Property 'getArrays' does not exist on type 'vtkDataSetAttributes'.
```

### Detailed behavior
Looking through my local node modules `node_modules/@kitware/vtk.js/Common/DataModel/DataSetAttributes.d.ts` line 2 `import { IFieldDataInitialValues, vtkFieldData } from '././FieldData';` however FieldData should be at `./DataSetAttributes/FieldData` so I think there is a problem in the handling during the build phase? Looking at: https://github.com/Kitware/vtk-js/blob/master/Sources/Common/DataModel/DataSetAttributes/index.d.ts it appears to be correct.

### Expected behavior
No typescript error

### Environment
- **vtk.js**: @kitware/vtk.js 24.3.1
- **OS**: Mac OSX 12.5

WesleyTheGeolien:
Also on a similar note the typescript for certain functions seems vague: `getArrays(): any`

daker:
Fixed in v25.8.1

@jourdain

PR #2568, opened 2022-09-07
chore(vtkFieldData): update types

WesleyTheGeolien:
### Context
updated types for vtkFielddata see #2567 

### Results
- typescript for field data is more expressive

### Changes
 updated `vtkFieldData.arrays` to be of type `vtkDataArray` instead of `any` I believe this is the only type the arrays should be

### PR and Code Checklist

### Testing
- [ ] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Mac OSX 12.5

### DIFF
--- a/Sources/Common/DataModel/DataSetAttributes/FieldData.d.ts
+++ b/Sources/Common/DataModel/DataSetAttributes/FieldData.d.ts
@@ -1,12 +1,12 @@
 import { vtkObject } from "../../../interfaces" ;
 import { Nullable } from "../../../types";
-import vtkDataArray from "../../Core/DataArray";
+import type vtkDataArray from "../../Core/DataArray";
 
 /**
  * 
  */
 export interface IFieldDataInitialValues {
-	arrays?: Array<any>;
+	arrays?: vtkDataArray[];
 	copyFieldFlags?: Array<any>;
 	doCopyAllOn?: boolean;
 	doCopyAllOff?: boolean;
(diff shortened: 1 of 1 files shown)


Sources/Common/DataModel/DataSetAttributes/FieldData.d.ts
```diff
@@ -6,7 +6,7 @@ import vtkDataArray from "../../Core/DataArray";
  * 
  */
 export interface IFieldDataInitialValues {
-	arrays?: Array<any>;
+	arrays?: Array<vtkDataArray> | vtkDataArray[];
```
daker, line 9: `vtkDataArray[]` is the short form of `Array<vtkDataArray>` , so you can just use `vtkDataArray[]`

Sources/Common/DataModel/DataSetAttributes/FieldData.d.ts
```diff
@@ -75,13 +75,13 @@ export interface vtkFieldData extends vtkObject {
 	/**
 	 * Get all arrays.
 	 */
-	getArrays(): any;
+	getArrays(): Array<vtkDataArray> | vtkDataArray[];
```
daker, line 78: same here

daker (changes requested):

daker:
@WesleyTheGeolien also the commit message needs have a scope of docs(xxx) to prevent generating a new release

WesleyTheGeolien:
> @WesleyTheGeolien also the commit message needs have a scope of docs(xxx) to prevent generating a new release

sure @daker was not sure which category it all fell under! I guess the next release shouldn't be too far away and this will get into that anyway?

PR #2573, opened 2022-09-10
fix(wslinkclient): changed type for protocols to access protocol as…

### LINKED ISSUE #2569: Type error for getRemote function on vtkWSLClient
## Context

I am trying to connect to a paraview server using `RemoteView` (here is the [RemoteView example](https://kitware.github.io/vtk-js/examples/RemoteView.html)). I have added a few protocols (ProxyManager, ViewPort, etc.) and I am trying to access these once it gets connected to the paraview server.

```js
  const client = vtkWSLinkClient.newInstance({
    protocols: {
      proxyManager: ProxyManager,
      fileManager: FileListing,
      viewManager: ViewPort,
    },
  });
```

```js
  client
    .connect(serverConfig)
    .then((validClient) => {
      console.log('connected image stream');
      connectImageStream(validClient.getConnection().getSession());
      const s = validClient.getConnection().getSession();
      const pm = validClient.getRemote().proxyManager;
      const vm = validClient.getRemote().viewManager;
```

## Description

Now I am trying to add types in this file. Declaring types for the protocols would lead me to this error

image

Could we change the type for `getRemote` to fix this ?

https://github.com/Kitware/vtk-js/blob/0b1de784be7cad010a1ab17f840d78839a07f628/Sources/IO/Core/WSLinkClient/index.d.ts#L133

```ts
  getRemote: () => Record<string, any>
```

Nikhil216:
… property on remote

fix #2569

### DIFF
--- a/Sources/IO/Core/WSLinkClient/index.d.ts
+++ b/Sources/IO/Core/WSLinkClient/index.d.ts
@@ -76,12 +76,12 @@ export interface vtkWSLinkClient extends vtkObject {
    * Assign protocols to the client. Those will only be used at connect time and therefore needs to be set before being connected otherwise `registerProtocol` should be used instead.
    * @returns {Boolean} true if the set method modified the object
    */
-  setProtocols(protocols: object): boolean;
+  setProtocols(protocols: Record<string, any>): boolean;
 
   /**
    * Get protocols that were either provided in `newInstance` or via its set
    */
-  getProtocols(): object;
+  getProtocols(): Record<string, any>;
 
   /**
    * Update the list of methods that should be ignore from the busy state monitoring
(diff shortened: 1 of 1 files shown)


jourdain:
Since wslink releases are not that common, you may want to trigger one so you can get your definition. Just replace `docs(...` by `fix(...`. 

Otherwise everything looks good. Thanks for contributing!

jourdain:
You need to amend your commit and force push. If I merge your change right now, nothing will happen as semantic-release will only see a `docs(....`

Nikhil216:
Thanks for all the help @jourdain. I guess this should do the job.

jourdain:
Thanks! I'll merge it once the CI comes back.

PR #2577, opened 2022-09-20
perf: substitute

FezVrasta:
### Context

this new implementation is 40-60% faster (depending if the `all` flag is set to true or not)

https://jsbench.me/kil8a6dx31/1

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

### Funding

### DIFF
--- a/Sources/Rendering/OpenGL/ShaderProgram/index.js
+++ b/Sources/Rendering/OpenGL/ShaderProgram/index.js
@@ -6,19 +6,23 @@ const { vtkErrorMacro } = macro;
 // perform in place string substitutions, indicate if a substitution was done
 // this is useful for building up shader strings which typically involve
 // lots of string substitutions. Return true if a substitution was done.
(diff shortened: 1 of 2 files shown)


finetjul:
Can you please avoid nested ternary operators and maybe create a test to make sure there is no regresion.

FezVrasta:
I added some tests but I'm not sure how to run them, `npm run test` doesn't seem to catch them

floryst:
> I added some tests but I'm not sure how to run them, `npm run test` doesn't seem to catch them

The tests should be picked up by karma via `npm run test`. I think you need to call `t.end()` after all of your assertions, or use `t.plan(numOfAssertions)`.

finetjul:
LGTM

FezVrasta:
Everything should be okay now

PR #2580, opened 2022-09-24
Remove redundant `delete` operator

volrath:
### Context

I tried to compile the code base with the Google Closure Compiler, which recognizes the affected line as an error:

```
Failed to inspect file
  [...]/node_modules/@kitware/vtk.js/Rendering/OpenGL/RenderWindow.js

it was required from
  [...]/node_modules/@cornerstonejs/core/dist/umd/index.js

Errors encountered while trying to parse file
  [...]/node_modules/@kitware/vtk.js/Rendering/OpenGL/RenderWindow.js
  {:line 521, :column 0, :message "Invalid delete operand. Only properties can be deleted."}
```

Out of curiosity, I tried compiling the code with webpack and, unsurprisingly, it didn't raise any errors. I guess webpack doesn't complain here because `delete` of anything other than an object property is just ignored.

In any case, I noticed that `model._textureResourceIds` is a `Map` object, so calling it's `delete` method is enough. I suppose the `delete` operand is just a leftover from a time where `model._textureResourceIds` was an object.

### Results

No more errors when compiling with GCP.

### Changes

- [x] ~Documentation and TypeScript definitions were updated to match those changes~ N/A

### PR and Code Checklist

### Testing

- [x] ~This change adds or fixes unit tests~ 
- [x] ~Tested environment:~

Change should be simple enough as to not require testing, but let me know if you disagree.

### DIFF
--- a/Sources/Rendering/OpenGL/RenderWindow/index.js
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.js
@@ -513,7 +513,7 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
     const result = model._textureResourceIds.get(texture);
     if (result !== undefined) {
       publicAPI.getTextureUnitManager().free(result);
-      delete model._textureResourceIds.delete(texture);
+      model._textureResourceIds.delete(texture);
     }
   };
 


finetjul:
LGTM

volrath:
I think my commit message was missing a colon. I'm not familiar with semantic release commit messages and/or commitizen. Also for some reason installing this program is failing on my system, so hopefully I followed the format correctly this time.

PR #2581, opened 2022-09-25
Add object set

### LINKED ISSUE #2575: Setting a model variable as object always trigger modified()
### High-level description
If `macro.set(publicAPI, model, ['foo']);` and `foo` is an `Object` (e.g. `{a: 1, b: 2}`, then calling `setFoo({a: 1, b:2})` will always trigger `modified()`.

### Steps to reproduce
```
extend(...){
  macro.set(publicAPI, model, ['foo']);
}
...
myobj.setFoo({a: 1, b: 2});
myobj.getMTime(); // =>1234
myobj.setFoo({a: 1, b: 2});
myobj.getMTime(); // =>1235
```

See [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Core/WidgetManager/index.js#L239) for a performance hit that modifies a WidgetRepresentation each time the mouse is moved which will always `requestData()` on the widget representation

### Detailed behavior
This is because `macro.setter()` does a shallow equal comparison `if (model[field] !== value) {` instead of a deep equal.

### Expected behavior
I can see 2 options:
 - Add deep equal check if `value` is an object in `macro.setter()`
 - Create `macro.setObject(publicAPI, model, ['foo']);` convenient setter similar to `macro.setArray()`

### Environment
- **vtk.js**:  master

finetjul:
### Context
fix #2575 

### Results
Improve rendering performances when having widgets in the renderer.

### Changes
Create a new 'object' type in `macro.setGet()` or `macro.set()`

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 10
  - **Browser**: Chrome

### DIFF
--- a/Sources/Rendering/Core/Prop3D/index.js
+++ b/Sources/Rendering/Core/Prop3D/index.js
@@ -99,8 +99,12 @@ function vtkProp3D(publicAPI, model) {
   };
 
   publicAPI.setUserMatrix = (matrix) => {
+    if (vtkMath.areMatricesEqual(model.userMatrix, matrix)) {
+      return false;
+    }
     mat4.copy(model.userMatrix, matrix);
     publicAPI.modified();
+    return true;
   };
 
   publicAPI.getMatrix = () => {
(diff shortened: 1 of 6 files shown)


floryst (approved):
Overall, LGTM. My only unresolved discussion point has to do with the shallow vs deep copy approach, but I think shallow copy is okay for now. @jourdain, what do you think about the API?

PR #2593, opened 2022-10-11
Fix hardware selector on edges

bruyeret:
### Context

Bug in the hardware selector, when in edge + surface mode and in FIELD_ASSOCIATION_CELLS mode.
The hardware selector doesn't find the cell ID when picking exactly on an edge.

### Results

The cell ID is now right when picking exactly on an edge.

### Changes

- Fixed and improved the hardware selector example.
- Fixed a call to `creatVBO` in the PolyDataMapper.

### Testing

No test added or changed.

### DIFF
--- a/Sources/Rendering/Core/HardwareSelector/example/index.js
+++ b/Sources/Rendering/Core/HardwareSelector/example/index.js
@@ -5,6 +5,7 @@ import 'vtk.js/Sources/favicon';
 
 // Load the rendering pieces we want to use (for both WebGL and WebGPU)
 import 'vtk.js/Sources/Rendering/Profiles/Geometry';
+import 'vtk.js/Sources/Rendering/OpenGL/Glyph3DMapper';
 
 import { throttle } from 'vtk.js/Sources/macros';
 import vtkActor from 'vtk.js/Sources/Rendering/Core/Actor';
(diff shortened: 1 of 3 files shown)


finetjul (approved):

sankhesh (approved):
LGTM +1 

Might as well remove the TODO comment in https://github.com/Kitware/vtk-js/blob/7f8b1c24f18b16a5a820312d47d3c701b54227ce/Sources/Rendering/OpenGL/PolyDataMapper/index.js#L1965

finetjul:
> Might as well remove the TODO comment in
> 
> https://github.com/Kitware/vtk-js/blob/7f8b1c24f18b16a5a820312d47d3c701b54227ce/Sources/Rendering/OpenGL/PolyDataMapper/index.js#L1965

Good point, tracked [here](https://github.com/Kitware/vtk-js/issues/2595).

floryst:
Two HardwareSelector tests are failing in `vtkHardwareSelector TestHardwareSelectorPoints`. If the promise execution is the same as the order in the file, it's these two assertions that fail:
- https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/HardwareSelector/test/testHardwareSelectorPoints.js#L98
- https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/HardwareSelector/test/testHardwareSelectorPoints.js#L115

bruyeret:
I checked on VTK C++ to see if selecting edges when displaying surface with edges does select the cell.
And indeed it does, so I fixed the tests and added 2 tests for polylines.

floryst:
LGTM

Issue #1754, opened 2021-02-25
SVGLandmarkRepresentation does not respect sized viewports


llangholz89:
I'm not sure if this is already being covered in the other work going on around widgets, but I've noticed that the SVGLandmarkRepresentation (and likely all other SVG widgets) is not positioned properly when working with renderers that are not full-sized (i.e. `renderer.setViewport([0.1, 0.2, 0.9, 08.])`)

To reproduce, simply add a line in the `LineWidget` example to size the viewport:
```
...
const fullScreenRenderer = vtkFullScreenRenderWindow.newInstance({
  background: [0, 0, 0],
});
const renderer = fullScreenRenderer.getRenderer();
renderer.setViewport([0.1, 0.2, 0.9, 0.8]); // <-- set a specific size on the viewport
...
```

I looked into it only a little bit, but got stuck on how to properly calculate the appropriate "y" coordinate for the SVG, and started wondering if more info was needed from the PixelSpaceCallbackMapper's callback. But I'm not super familiar with that mechanism so I figured I'd post this here before going down that rabbit hole...

![ezgif com-gif-maker](https://user-images.githubusercontent.com/16705837/109104607-224b2000-76e1-11eb-8c82-81c29e2bfe10.gif)

jourdain:
I think @finetjul fix that problem but I might be wrong.

finetjul:
Nope, this has not been fixed. This is unrelated to the recent fixes regarding HighDPI. Should be related to  "world to display coordinate" conversion.

Issue #1750, opened 2021-02-24
Import svg measurement widgets from paraview-glance


agirault:
- https://github.com/Kitware/paraview-glance/blob/master/src/vtk/SVGLineRepresentation/index.js
- https://github.com/Kitware/paraview-glance/tree/master/src/vtk/SVGCircleHandleRepresentation

PR #2596, opened 2022-10-11
Remove SVG representations

### LINKED ISSUE #2206: How should I monitor the events of the SVG widgets in vtk.js, such as mouse clicks, movements, etc.

floryst:
### Context
Completely removes the SVG representations, since we no longer have a need to provide an implementation of this ourselves.

### Results

SVG representations and support are completely removed, reducing maintenance on code that is better left to a third party SVG manipulation library.

Closes #2206, #1963, #1754, and #1750.

### Changes
- [x] SVGRepresentations are removed
- [ ] Existing widgets using the SVG representations are updated to use a custom SVG helper

### PR and Code Checklist

### Testing
- [x] Widgets are tested manually to ensure the SVG behavior in the examples are correct


floryst:
FYI @finetjul I haven't figured out how best to map SVG interactions to WidgetManager state interactions. In particular, the LabelWidget used the SVGRepresentation hover state. One way to solve this is to consider how we can let the WidgetManager support external state "hover"/selection.

floryst:
I found a solution/workaround to mapping events on SVG elements to the corresponding state. The relevant code for that resides in the `vtkLabelWidget` commit.

Sources/Rendering/Misc/RemoteView/example/index.js
```diff
@@ -46,9 +46,7 @@ const config = {
 clientToConnect
   .connect(config)
   .then((validClient) => {
-    const viewStream = this.clientToConnect
-      .getImageStream()
-      .createViewStream('-1');
+    const viewStream = clientToConnect.getImageStream().createViewStream('-1');
```
jourdain, line 49: was that not already fixed?

jourdain (approved):

Sources/Rendering/Misc/RemoteView/example/index.js
floryst: This branch is merging into beta. I will rebase my changes onto beta so master commits don't get mixed in. Thanks for catching that!

Examples/Widgets/Utilities/SVGHelpers.js
```diff
@@ -0,0 +1,193 @@
+import {
```
finetjul, line 1: does it mean that if it is in the Examples folder it won't be exported in the ESM/UMD  module? 
If so, we might consider to keep it for a while to give time for VTK.js developers to adapt to the change.

finetjul (approved):
LGTM

Examples/Widgets/Utilities/SVGHelpers.js
jourdain: I don't think we should have it within vtk.js but we should definitely document/explain what is required if someone needs to reproduce that example. And mention, that is just an example on how to possibly do it but other framework like vue/react could be use to streamline such usage as well and that is beyond the scope of vtk.js to provide a svg infrastructure.

finetjul: I agree, documentation would be useful, the update example is not easy to read.

It would be worth mentioning the other (duplicate? ) SVG approach with `setDisplayCallback` (e.g. in `ImageCroppingWidget` example).

floryst: I've updated the documentation in `concepts_widgets.md` and the comment block for `bindSVGRepresentation` to help guide folks who happen to use the SVGHelpers file.

finetjul (approved):

PR #2598, opened 2022-10-12
perf: avoid GC calls

FezVrasta:
### Context

Small perf improvement that avoids a quite frequent GC event from happening.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Common/Core/ClassHierarchy/index.js
+++ b/Sources/Common/Core/ClassHierarchy/index.js
@@ -1,7 +1,14 @@
+/* eslint-disable prefer-rest-params */
 export default class ClassHierarchy extends Array {
-  push(...args) {
-    // no perf issue since args.length should be small
-    const newArgs = args.filter((arg) => !this.includes(arg));
+  push() {
+    // using a for loop because a filter would trigger the garbage collector
+    const newArgs = [];
+    for (let i = 0; i < arguments.length; i++) {
+      if (!this.includes(arguments[i])) {
+        newArgs.push(arguments[i]);
+      }
+    }
+
     return super.push(...newArgs);
   }
 }


floryst:
An `npm run reformat` is needed here. Also, is this due to a hot loop that creates/destroys a bunch of vtk.js objects? How visible/measurable is this issue+improvement under your tests?

FezVrasta:
I see this called a couple times per render during scroll, it can take 8ms for each of these calls when the GC kicks in (most of the times).

floryst:
Great, thanks for the info. What's the timing with this patch?

LGTM once we get some numbers of before/after!

FezVrasta:
Before:

CleanShot 2022-10-12 at 2 38 12@2x

After:
CleanShot 2022-10-12 at 4 25 42@2x

Sources/Common/Core/ClassHierarchy/index.js
```diff
@@ -1,7 +1,14 @@
...
 export default class ClassHierarchy extends Array {
-  push(...args) {
-    // no perf issue since args.length should be small
-    const newArgs = args.filter((arg) => !this.includes(arg));
+  push() {
+    // using a for loop because a filter would trigger the garbage collector
+    const newArgs = [];
+    for (let i = 0; i < arguments.length; i++) {
+      if (!this.includes(arguments[i])) {
+        newArgs.push(arguments[i]);
```
finetjul, line 8: why don't you call `super.push(arguments[i])` instead of creating a temp `newArgs`?

FezVrasta: I didn't measure any performance issue with that part of code. I'm also not sure if multiple push will result in better performance.

finetjul: `.push()` is almost always called with 1 argument each time.

floryst: Initially I was okay with this since `.push()` does have a return value (which is the new length of the array), but now that @finetjul brings this up it would be better to actually call `super.push(arguments[i])` as mentioned and avoid the extraneous newArgs allocation, ending with a `return this.length`.

finetjul: and also it is more correct: it will remove duplicates if they exist in `arguments`

PR #2602, opened 2022-10-14
Fix widgets with scaleInPixels by adding RenderWindowViewNode.getComputedDevicePixelRatio

PaulHax:
### Context
Widget handles with `scaleInPixels` appear smaller when window.devicePixelRatio is more than 1.  Expected them to be the same CSS pixel size.

Smaller when screen "scale" setting is 200%:
image

### Results
100% scale
image

200% scale
image

### Changes

Adds vtkRenderWindowViewNode.getComputedDevicePixelRatio using DOM container CSS pixel size and user set rendering pixel size.

### PR and Code Checklist

### Testing

`npm run example -- ImageCroppingWidget` and change display settings or browser zoom.

- [x] Tested environment:
  - **vtk.js**: 25.9.1
  - **OS**: Linux
  - **Browser**: Chrome 105

### DIFF
--- a/Sources/Rendering/OpenGL/RenderWindow/index.d.ts
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.d.ts
@@ -327,7 +327,18 @@ export interface vtkOpenGLRenderWindow extends vtkOpenGLRenderWindowBase {
 	setViewStream(stream: vtkViewStream): boolean;
 
 	/**
-	 *
+	 * Sets the pixel width and height of the rendered image.  
+	 * 
+	 * WebGL and WebGPU render windows apply these values to 
+	 * the width and height attribute of the canvas element.
+	 * 
+	 * To match the device resolution in browser environments, 
+	 * multiply the container size by `window.devicePixelRatio`
+	 * `apiSpecificRenderWindow.setSize(Math.floor(containerWidth * devicePixelRatio), Math.floor(containerHeight * devicePixelRatio));
+	 * See the VTK.js FullscreenRenderWindow class for an example.
+	 * 
+	 * @see getComputedDevicePixelRatio()
+	 * 
 	 * @param {Vector2} size 
 	 */
 	setSize(size: Vector2): void;
@@ -363,14 +374,15 @@ export interface vtkOpenGLRenderWindow extends vtkOpenGLRenderWindowBase {
 	getVrResolution(): Vector2;
 
 	/**
-	 * Use to scale the size of a rendered canvas pixel to a browser CSS pixel.  
-	 * `const cssPixelWidth = renderedPixelWidth / apiRenderWindow.getComputedDevicePixelRatio()`
-	 * Rather than using window.devicePixelRatio directly, the device pixel ratio is inferred
-	 * from the container CSS pixel size and rendered image pixel size. The user can reduce the rendered size.
-	 * `apiSpecificRenderWindow.setSize(Math.floor(container.width * window.devicePixelRatio / 2), Math.floor(container.height * window.devicePixelRatio / 2))`
+	 * Scales the size of a browser CSS pixel to a rendered canvas pixel.  
+	 * `const renderedPixelWidth = cssPixelWidth * apiRenderWindow.getComputedDevicePixelRatio()`
+	 * Use to scale rendered objects to a consistent perceived size or DOM pixel position.
 	 * 
+	 * Rather than using window.devicePixelRatio directly, the device pixel ratio is inferred
+	 * from the container CSS pixel size and rendered image pixel size. The user directly sets the rendered pixel size.
+	 * 
+	 * @see setSize()
 	 * @see getContainerSize()
-	 * @see getSize()
 	 */
 	getComputedDevicePixelRatio(): number;
 }
--- a/Sources/Rendering/OpenGL/RenderWindow/index.d.ts
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.d.ts
@@ -361,6 +361,18 @@ export interface vtkOpenGLRenderWindow extends vtkOpenGLRenderWindowBase {
 	 *
 	 */
 	getVrResolution(): Vector2;
+
+	/**
+	 * Use to scale the size of a rendered canvas pixel to a browser CSS pixel.  
+	 * `const cssPixelWidth = renderedPixelWidth / apiRenderWindow.getComputedDevicePixelRatio()`
+	 * Rather than using window.devicePixelRatio directly, the device pixel ratio is inferred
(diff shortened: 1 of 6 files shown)


finetjul:
Last time we talked about it, we decided that it (devicePixelRatio) should be handled at renderwindow level.

@jourdain or @floryst can comment on that.

finetjul:
[Here](https://github.com/Kitware/vtk-js/pull/2396) is a bit of history context:
 - this is the same PR that got abandoned for some unclear reason. According to [this comment](https://github.com/Kitware/vtk-js/pull/2396#issuecomment-1117340554), it should make a test fail. I can't remember why.
 - [Sebastien's comment](https://github.com/Kitware/vtk-js/pull/2396#discussion_r864907809) also mention graphics should also be scaled.

PaulHax:
Just wanted to sneak another `window` reference in 😄

But seems to be a well trodden path.  Pushed a sketch adding `WebGPU/OpenGLRenderWindow.getDevicePixelRatio()`  and modifying `WidgetManager.updateDisplayScaleParams`.  Perhaps we rehash the devicePixelRatio discussion this Thursday?

PaulHax:
Looks like three.js `WebGLRenderer` has `setPixelRatio` and takes CSS units in `setSize`

https://github.com/mrdoob/three.js/blob/034343d97134c55559aa4edbcce3d405eae308da/examples/webgl_postprocessing_sao.html#L58-L59

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -242,6 +242,8 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
     return model.size;
   };
 
+  publicAPI.getDevicePixelRatio = () => window.devicePixelRatio;
```
finetjul, line 245: how about having it in `RenderWindowViewNode` instead ?
It would be a simply `model` variable with `get/set` accessors but initialized to `window.devicePixelRatio || 1` in `RenderWindowViewNode.extend()`

PaulHax: One wrinkle, window.devicePixelRatio can change after initializing.

finetjul: Interesting... 
that means we should probably also observe the change and propagate it by invoking a new type of event ?

floryst: https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio#monitoring_screen_resolution_or_zoom_level_changes

finetjul (commented):
I like your approach.

If implemented that way, it means that `vtkFullScreenRenderWindow`, `vtkGenericRenderWindow`, `ViewProxy`, `SynchronizableRenderWindow` should now observe when `devicePixelRatio` changes to "resize()" the container size (e.g. when `CTRL` + `+`/`-`/`0` or when switching monitor), right ?

Can't we make that `observation` automatic ? e.g. `vtkRenderWindow.setFixedPixelRatio(boolean)`., if `false` then it observes and updates automatically `size` ?

Sources/Rendering/OpenGL/RenderWindow/index.d.ts
```diff
@@ -361,6 +361,11 @@ export interface vtkOpenGLRenderWindow extends vtkOpenGLRenderWindowBase {
 	 *
 	 */
 	getVrResolution(): Vector2;
+
+	/**
+	 * Uses container CSS pixel size and rendered image pixel size
+	 */
```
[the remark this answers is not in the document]
PaulHax: Updated to 
```
/**
	 * Use to scale the size of a rendered canvas pixel to a browser CSS pixel.  
	 * `const cssPixelWidth = renderedPixelWidth / apiRenderWindow.getComputedDevicePixelRatio()`
	 * Rather than using window.devicePixelRatio directly, the device pixel ratio is inferred
	 * from the  container CSS pixel size and rendered image pixel size. The user can reduce the rendered size.
	 * `apiSpecificRenderWindow.setSize(Math.floor(container.width * window.devicePixelRatio / 2), Math.floor(container.height * window.devicePixelRatio / 2))`
	 * 
	 * @see getContainerSize()
	 * @see getSize()
	 */
	getComputedDevicePixelRatio(): number;
```

finetjul (approved):

PaulHax:
> If implemented that way, it means that `vtkFullScreenRenderWindow`, `vtkGenericRenderWindow`, `ViewProxy`, `SynchronizableRenderWindow` should now observe when `devicePixelRatio` changes to "resize()" the container size (e.g. when `CTRL` + `+`/`-`/`0` or when switching monitor), right ?

They probably should.  However, I have not found a case yet where their resize callbacks have not been triggered.  Have not tested dragging window between monitors with different pixel density case.

> Can't we make that `observation` automatic ? e.g. `vtkRenderWindow.setFixedPixelRatio(boolean)`., if `false` then it observes and updates automatically `size` ?

RenderWindow should then take on the "resize" responsibility in the `setFixedPixelRatio(true)` case as well.  Or there will be overlap with `vtkFullScreenRenderWindow` resize callbacks.

finetjul:
> However, I have not found a case yet where their resize callbacks have not been triggered

Fair enough, it should be good enough then.

floryst:
The commit should be a `feat` since there is an API addition. I would also recommend adding a new commit that replaces usages of `window.devicePixelRatio` with your new API.

PaulHax:
> The commit should be a `feat` since there is an API addition. I would also recommend adding a new commit that replaces usages of `window.devicePixelRatio` with your new API.

👍️  

`window.devicePixelRatio` is still in SVGLandmarkRepresentation but that's going away I hear.

floryst:
There are still usages of `devicePixelRatio` that can be updated, such as `[Generic/FullScreen]RenderWindow`, the `SynchronizableRenderWindow` example, and `ViewProxy`. And yes, the SVG stuff doesn't have to be updated, since that will be deleted.

PaulHax:
> There are still usages of `devicePixelRatio` that can be updated, such as `[Generic/FullScreen]RenderWindow`, the `SynchronizableRenderWindow` example, and `ViewProxy`. And yes, the SVG stuff doesn't have to be updated, since that will be deleted.

The higher level render windows (FullScreen, etc) call `apiSpecificRenderWindow.setSize`.  So this is the appropriate place for `window.devicePixelRatio` if I understand the current approach (internalDPR = renderedSize / containerSize) Perhaps changes to the docs to make this clear?

PaulHax:
New doc copy pushed.

PR #2603, opened 2022-10-19
feat: added nonlinear param to buildFunctionFromTable

Ouwen:
### Context

In order to support sigmoid LUT based on dicom standard, non-linear LUT's need to be created on the fly any time a user changes the WW/WL. The current buildFunctionFromTable only support linearly spaced x input which is inefficient sampling for sigmoid functions. This change simply allows users to build functions from the long format as used in RGBLong.

### Results

This change simply allows users to build functions from the long format as used in RGBLong.

### Changes

### PR and Code Checklist

### Testing

Because this is a surgical change, no new tests have been added.

### DIFF
--- a/Sources/Rendering/Core/ColorTransferFunction/index.d.ts
+++ b/Sources/Rendering/Core/ColorTransferFunction/index.d.ts
@@ -1,3 +1,4 @@
+import vtkDataArray from '../../../Common/Core/DataArray';
 import { vtkObject } from '../../../interfaces';
 import { ColorSpace, Scale } from "./Constants";
 
@@ -200,6 +201,25 @@ export interface vtkColorTransferFunction extends vtkObject {
     size: number,
     withAlpha: boolean
   ): Float32Array;
+
+  /**
+   * Construct a color transfer function from a vtkDataArray.
+   * The order of values depends on the number of components
+   * of the array.
+   * 3 -> RGB
+   * 4 -> XRGB
+   * 5 -> RGBMS
+   * 6 -> XRGBMS
+   * 
+   * X represents the input value to a function
+   * RGB represents the red, green, and blue value output
+   * M represents the midpoint
+   * S represents sharpness
+   * @param {vtkDataArray} array
+   */
+   buildFunctionFromArray(
+    array: vtkDataArray,
+  ): void;
 
   /**
    * Construct a color transfer function from a table.
--- a/Sources/Rendering/Core/ColorTransferFunction/index.js
+++ b/Sources/Rendering/Core/ColorTransferFunction/index.js
@@ -871,6 +871,63 @@ function vtkColorTransferFunction(publicAPI, model) {
     return model.table;
   };
 
+  publicAPI.buildFunctionFromArray = (array) => {
+    publicAPI.removeAllPoints();
+    const numComponents = array.getNumberOfComponents();
+    for (let i = 0; i < array.getNumberOfTuples(); i++) {
+      switch (numComponents) {
+        case 3: {
+          model.nodes.push({
+            x: i,
+            r: array.getComponent(i, 0),
+            g: array.getComponent(i, 1),
+            b: array.getComponent(i, 2),
+            midpoint: 0.5,
+            sharpness: 0.0,
+          });
+          break;
+        }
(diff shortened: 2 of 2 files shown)


Ouwen:
@floryst hope you are doing well, this is my first time contributing to vtkjs so please let me know if there are any additional modifications I need to make!

Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -871,6 +871,21 @@ function vtkColorTransferFunction(publicAPI, model) {
...
+        r: table[i * 6 + 1],
+        g: table[i * 6 + 2],
+        b: table[i * 6 + 3],
+        midpoint: table[i * 6 + 4],
+        sharpness: table[i * 6 + 5],
+      };
+      model.nodes.push(node);
+    }
+    publicAPI.sortAndUpdateRange();
+  };
```
finetjul, line 888: If `table` was a `vtkDataArray`, we could get the number of components per tuple:
- 3 ->RGB
- 4 ->XRGB
- 5 -> RGBMS
- 6 -> XRGBMS

You would then have something very generic.
(here is how you could convert `table` into a `vtkDataArray`:
```
ctf.buildFunctionFromArray(vtkDataArray.newInstance({values: table, numberOfComponents: 6})
```

Sources/Rendering/Core/ColorTransferFunction/index.d.ts
```diff
@@ -257,6 +267,11 @@ export interface vtkColorTransferFunction extends vtkObject {
    */
   setMappingRange(min: number, max: number): void;
 
+  /**
+   * Get the range of mapped scalars.
+   */
+  getRange(): number[];
```
finetjul, line 273: should be defined in superclass vtkScalarsToColors

Sources/Rendering/Core/ColorTransferFunction/index.js
Ouwen: makes sense, I will make that change

Sources/Rendering/Core/ColorTransferFunction/index.d.ts
finetjul: It should be fixed in vtkScalarsToColors, the `min` and `max` parameters are ignored by the function. 

I think they should be removed from `ScalarsToColors/index.d.ts` but also from `ScalarsToColors/index.js` (in a different commit, same PR)

Ouwen:
@finetjul ready for your review

Sources/Rendering/Core/ColorTransferFunction/index.d.ts
```diff
@@ -201,6 +202,14 @@ export interface vtkColorTransferFunction extends vtkObject {
     withAlpha: boolean
   ): Float32Array;
 
+  /**
+   * Construct a color transfer function from a vtkDataArray.
+   * @param {vtkDataArray} array
+   */
```
finetjul, line 208: Would be nice to document how tuple size is used by the function

finetjul (approved):
Looking good. I have just a minor final comment regarding documentation.

Sources/Rendering/Core/ColorTransferFunction/index.d.ts
Ouwen: The documentation has been moved here

Ouwen: good catch!

finetjul:
Looking good.
Final comment: can you please squash your commits in a unique commit, e.g. `"feat(colortransferfunction): add buildFunctionFromArray()"`

Ouwen:
Squashed the commits and pushed

floryst (approved):
LGTM. Thanks, @Ouwen!

finetjul:
LGTM

floryst:
@Ouwen can you rebase onto master? I'm given no options to update your branch prior to merging, so having you do that would allow us to merge.

Ouwen:
Sure I’ll rebase to master and force push

Ouwen:
@floryst branch is now rebased with master

PR #2605, opened 2022-10-21
fix(Texture): update internalFormat when component count changes

PaulHax:
### Context

When changing the number of components in the TestVolumeTypes example the ImageMapper would not update and there would be a WebGL warning.  Bug introduced in #1400 to allow user to set `internalFormat` manually. (Why would one want to?)

### Results

ImageMapper updates when component count changes.  (Here is a hack for getting 4 components to work for ImageMapper: https://github.com/Kitware/itk-vtk-viewer/blob/7c88e1e055dc7c793561c4cdb106f9815c05a799/src/Rendering/VTKJS/vtk/OpenGLImageMapperFractional.js#L426-L446 )

### Changes

Adds `_forceInternalFormat` to track if `internalFormat` should be updated with each call to `getInternalFormat`.  Calling `publicAPI.setInternalFormat` sets `_forceInternalFormat = true`

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 25.9.2
  - **OS**: Linux
  - **Browser**: Chrome 105


Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -350,7 +350,7 @@ function vtkOpenGLTexture(publicAPI, model) {
 
   //----------------------------------------------------------------------------
   publicAPI.getInternalFormat = (vtktype, numComps) => {
-    if (!model.internalFormat) {
+    if (!model._forceInternalFormat) {
```
floryst, line 353: I'm not sure if I fully follow the purpose of this. The only difference I see is that if `publicAPI.setInternalFormat(0)` is called, then we still set `_forceInternalFormat` to be true. The changes don't seem to affect much other than that.

PaulHax: Without this change, `internalFormat` does not change between `create3DFilterableFromRaw` calls.  So if we change the component count of the input data, the `internalFormat` is not updated and there is an WebGL error.

Your right, hacky.  Maybe better to refactor this "getter" that sets and change lots of thing...

floryst: Ah, I now understand what you are trying to do. @finetjul do you know the purpose of adding [this conditional in getInternalFormat](https://github.com/Kitware/vtk-js/pull/1400/files#diff-18c5d83bebc14325d84e520000b95e06dadffa088fe27eef138af64744171b89R316-R321)? I think that will help determine what functionality we should expect from calling `getInternalFormat`. Paul's proposed changes effectively revert that conditional, except when the user explicitly sets the internal format (which afaik nobody ever does).

PaulHax: Without that conditional, earlier calls to publicAPI.setInternalFormat would get overwritten by all the `create*` functions' calls of `getInternalFormat`

floryst:
This seems fine to me. @jourdain?

jourdain:
All good for me

PR #2613, opened 2022-10-31
fix(ViewNode): delete scene graph nodes

floryst:
### Context
Scene graph node destructors are not invoked when pruned. As a result, certain references may be retained (e.g. in [vtkOpenGLVolumeMapper](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/VolumeMapper/index.js#L1115)), resulting in a memory leak.

### Results
- Scene graph nodes have their `.delete()` method called appropriately.

### Changes

- [x] vtkViewNode overrides `delete()` to call delete on children

### PR and Code Checklist

### Testing
- No tests are changed. Test results are checked by CI.

### DIFF
--- a/Sources/Rendering/SceneGraph/ViewNode/index.js
+++ b/Sources/Rendering/SceneGraph/ViewNode/index.js
@@ -162,6 +162,7 @@ function vtkViewNode(publicAPI, model) {
           deleted = [];
         }
         deleted.push(child);
+        child.delete();
       } else {
         child.setVisited(false);
       }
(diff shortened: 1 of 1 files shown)


jourdain (approved):
LGTM

Sources/Rendering/SceneGraph/ViewNode/index.js
```diff
@@ -184,6 +185,12 @@ function vtkViewNode(publicAPI, model) {
     }
     return ret;
   };
+
+  const parentDelete = publicAPI.delete;
+  publicAPI.delete = () => {
+    model.children.forEach((child) => child.delete());
```
finetjul, line 191: for performance (and GC) reasons, I would suggest to use a for loop instead.

floryst: Switched to using a for loop. I figured delete wouldn't be called frequently enough for it to matter, but might as well not have to find out and fix it later.

finetjul (approved):

PR #2623, opened 2022-11-08
perf(WidgetManager): reduce capture/render calls

floryst:
### Context
The WidgetManager triggers unnecessary renders which can be expensive while performing the picking operation.

### Results
- The WidgetManager triggers a render only when a widget selection changes.

### Changes
- [x] WidgetManager behavior changes to reduce renders.

### PR and Code Checklist

### Testing

- [x] Test widgets to verify nothing changes

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -299,9 +299,10 @@ function vtkWidgetManager(publicAPI, model) {
           w.deactivateAllHandles();
         }
       }
-      if (wantRender) {
-        model._interactor.render();
-      }
+    }
+
+    if (wantRender) {
+      model._interactor.render();
     }
   }
 
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -283,24 +283,27 @@ function vtkWidgetManager(publicAPI, model) {
     // Default cursor behavior
     model._apiSpecificRenderWindow.setCursor(widget ? 'pointer' : 'default');
 
+    let wantRender = false;
     if (model.widgetInFocus === widget && widget.hasFocus()) {
       activateHandle(widget);
-      // Ken FIXME
-      model._interactor.render();
-      model._interactor.render();
+      wantRender = true;
     } else {
       for (let i = 0; i < model.widgets.length; i++) {
         const w = model.widgets[i];
         if (w === widget && w.getNestedPickable()) {
           activateHandle(w);
           model.activeWidget = w;
+          wantRender = true;
         } else {
+          wantRender ||= !!w.get().activeState;
           w.deactivateAllHandles();
         }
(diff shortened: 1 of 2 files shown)


Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -283,24 +283,25 @@ function vtkWidgetManager(publicAPI, model) {
...
+      wantRender = true;
     } else {
       for (let i = 0; i < model.widgets.length; i++) {
         const w = model.widgets[i];
         if (w === widget && w.getNestedPickable()) {
           activateHandle(w);
           model.activeWidget = w;
+          wantRender = true;
         } else {
+          wantRender ||= !!w.get().activeState;
```
floryst, line 298: If the widget had an active state prior to deactivating, we request a render. `vtkAbstractWidget` doesn't have a `getActiveState()` API; is this something worth adding?

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -283,24 +283,25 @@ function vtkWidgetManager(publicAPI, model) {
     // Default cursor behavior
     model._apiSpecificRenderWindow.setCursor(widget ? 'pointer' : 'default');
 
+    let wantRender = false;
     if (model.widgetInFocus === widget && widget.hasFocus()) {
       activateHandle(widget);
-      // Ken FIXME
-      model._interactor.render();
-      model._interactor.render();
+      wantRender = true;
```
finetjul, line 289: it is useless to be set to true, `wantRender` is not used afterward

Sources/Widgets/Core/WidgetManager/index.js
jourdain: probably making that call in get() is not great...

jourdain (approved):
Not tested but seems reasonable

Sources/Widgets/Core/WidgetManager/index.js
finetjul: I think we should avoid using `get()`

Sources/Widgets/Core/WidgetManager/index.js
floryst: Whoops, I didn't pay attention to my indents. The `if (wantRender)` should be outside the if/else.

floryst: Fixed.

Sources/Widgets/Core/WidgetManager/index.js
floryst: Agreed. I think exposing `activeState` is preferred. It does mean that if by some chance someone decides to activate a widget substate manually (i.e. not through widget.activateHandle), then it won't be rendered.

floryst:
I've added `getActiveState`. Please take a look again.

finetjul (approved):
LGTM

PR #2624, opened 2022-11-08
Update example to use `@kitware/vtk.js`

davidshq:
The main documentation (e.g. https://kitware.github.io/vtk-js/docs/vtk_vanilla.html, https://kitware.github.io/vtk-js/docs/intro_vtk_as_es6_dependency.html) indicates we should use `@kitware/vtk.js` instead of `vtk.js`, this updates the code to use the recommended package.

Note: I think the code previously would only work if one placed the source for vtk.js within the application folder, e.g., something like: `/my-app/vtk.js/Sources/...` and not in any manner using ESM.

After having isuses with `@kitware/vtk.js` and this code I attempted installing `vtk.js` instead, but this code still produced errors, so I went back to `@kitware/vtk.js` and made these changes.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Examples/Rendering/ManyRenderers/index.js
+++ b/Examples/Rendering/ManyRenderers/index.js
@@ -1,19 +1,19 @@
-import 'vtk.js/Sources/favicon';
+import '@kitware/vtk.js/favicon';
 
 // Load the rendering pieces we want to use (for both WebGL and WebGPU)
-import 'vtk.js/Sources/Rendering/Profiles/Geometry';
-import 'vtk.js/Sources/Rendering/Misc/RenderingAPIs';
+import '@kitware/vtk.js/Rendering/Profiles/Geometry';
+import '@kitware/vtk.js/Rendering/Misc/RenderingAPIs';
 
-import vtkActor from 'vtk.js/Sources/Rendering/Core/Actor';
-import vtkConeSource from 'vtk.js/Sources/Filters/Sources/ConeSource';
-import vtkSphereSource from 'vtk.js/Sources/Filters/Sources/SphereSource';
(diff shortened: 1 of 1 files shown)


daker:
@davidshq Thanks for the PR! i think changing the imports only will break the build of this example because using `@kitware/vtk.js` requires the use of rollup while the `npm run example` command [requires webpack](https://github.com/Kitware/vtk-js/blob/master/Utilities/ExampleRunner/example-runner-cli.js#L96).
To switch to the kitware package we need to switch the examples to rollup.

davidshq:
@daker hmmm. I'm actually using webpack, not rollup:

https://github.com/davidshq-learning/vtk/tree/main/js/many-renderers

I didn't know about the example runner before your mention of it, I'd just been following the directions found at these locations:
- https://kitware.github.io/vtk-js/docs/intro_vtk_as_es6_dependency.html
- https://kitware.github.io/vtk-js/docs/vtk_vanilla.html
- https://kitware.github.io/vtk-js/examples/ManyRenderers.html

daker:
@floryst what do you think?

floryst:
I think it's a good idea to change all examples to use `@kitware/vtk.js` imports, as that's the recommended package. `vtk.js` is still there in the off chance you need it (e.g. for the UMD build). FYI @jourdain if you have any input on this.

I'm pretty sure the only change we will need to support this is to have the example runner alias `@kitware/vtk.js` to `Sources/...`.

jourdain:
I like your suggestion @floryst

daker:
@davidshq i have proposed a new PR #2631

floryst:
Closing in favor of #2631.

PR #2629, opened 2022-11-10
perf(OpenGLTexture): conditional getExtension call

floryst:
### Context
`gl.getExtension()` calls generally are very fast, but we can skip these calls entirely in webgl2 codepaths.

### Results
Slight perf improvement.

### Changes
- [x] OpenGLTexture now conditionally calls getExtension() in webgl2 code pathways. While the half_float extension result can technically be cached, I'm not going to introduce that here, since it's only relevant in webgl1 pathways. #2212 is related here.

### PR and Code Checklist

### Testing
CI will run tests.

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -644,10 +644,14 @@ function vtkOpenGLTexture(publicAPI, model) {
 
     // if the opengl data type is half float
     // then the data array must be u16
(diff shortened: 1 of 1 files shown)


finetjul:
@bruyeret

finetjul (approved):
LGTM

FezVrasta:
Can confirm performance is greatly improved when a Float32 input is provided.

Issue #2628, opened 2022-11-10
TestVolumeTypes does not render 4-component image texture


floryst:
### High-level description
The [TestVolumeTypes](https://kitware.github.io/vtk-js/examples/TestVolumeTypes.html) example does not render a 4-component image texture.

@thewtex 

### Steps to reproduce
1. Go to https://kitware.github.io/vtk-js/examples/TestVolumeTypes.html
2. Increase components to 4
3. Notice that the image disappears. The volume renders correctly.

### Detailed behavior
Switching to 4 components results in the following invalid operation warning:

```
GL_INVALID_OPERATION: Texture format does not support mipmap generation.
```

The call that I think causes the issue is when texImage2D is invoked with the following parameters (`internalFormat, format, type`), and then `generateMipmap(target)` is called.
```
'RGBA16F' 'RGBA' 'FLOAT'
```

The texImage2D docs indicate that the above combination is filterable, so I'm not entirely sure what else is wrong.

### Expected behavior

The 4-component image texture should render. Rendering with interpolation set to NEAREST works.

### Environment
- **vtk.js**: v25.13.0
- **OS**: macOS M1
- **Browser**: Chrome 107

PR #2635, opened 2022-11-17
fix(ImageMapper): Only generate mipmap for RGBA uint8 texture

thewtex:
Addresses #2628

To generate a mipmap, the texture type must be color-renderable and texture-filterable according to:

  https://stackoverflow.com/questions/44754479/issue-with-rgba32f-texture-format-and-mipmapping-using-opengl-es-3-0
  https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glGenerateMipmap.xhtml
  https://registry.khronos.org/OpenGL/specs/es/3.0/es_spec_3.0.pdf

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


sankhesh:
What about other types that are color-renderable and texture-filterable? like R8 and RGB8?

floryst:
@thewtex bump on sankhesh's comment.

thewtex:
We could also generate mipmaps for `numComp` 1, 3.

floryst (approved):
LGTM. I'm not sure if there is a case where GC on the Set allocation will impact performance, so I'll leave that as-is for now.

PR #2648, opened 2022-11-30
Avoid side effects of getNeedToRebuildShaders

bruyeret:
In PolyDataMapper and ImageMapper, `lastRenderPassShaderReplacement` was sometimes not updated.

It should have been updated in `getNeedToRebuildShaders`, but when this function was not called, the update didn't occur.

Fixed this behaviour by moving the update of `lastRenderPassShaderReplacement` in the render function.


floryst:
`updateShaders` calls `getNeedToRebuildShaders` before `buildShaders`. Under what conditions are you seeing `lastRenderPassShaderReplacement` not being updated properly?

bruyeret:
When someone creates a class extending PolyDataMapper and overrides `getNeedToRebuildShaders`, he may not call the original `getNeedToRebuildShaders` of PolyDataMapper.

PR #2652, opened 2022-12-05
Add null to findPokedRenderer return type

bruyeret:
Edit findPokedRenderer to handle a case when `rc` is an empt list (used to return undefined)
Edit typescript header as findPokedRenderer sometimes return null
Edit a call to this function in OrientationMarkerWidget accordingly, which sometimes caused errors

### PR and Code Checklist
- [x] Run `npm run test`

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
@@ -1,4 +1,5 @@
 import { vtkObject, vtkSubscription } from "../../../interfaces";
+import { Nullable } from "../../../types";
 import vtkRenderer from "../Renderer";
 import { Axis, Device, Input } from "./Constants";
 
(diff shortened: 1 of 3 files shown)


Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
```diff
@@ -1152,7 +1152,7 @@ export interface vtkRenderWindowInteractor extends vtkObject {
 	 * @param {Number} x 
 	 * @param {Number} y 
 	 */
-	findPokedRenderer(x: number, y: number): vtkRenderer;
+	findPokedRenderer(x: number, y: number): vtkRenderer | null;
```
daker, line 1155: import Nullable then use it `Nullable<vtkRenderer>`

floryst:
LGTM

PR #2655, opened 2022-12-05
feat(SphereMapper): add scaleFactor to scale radius of sphere

### LINKED ISSUE #2036: SetScaleFactor in SphereMapper
### Need
I'm Currently using `vtkSphereMapper` by setting a point scalar array (`setScaleArray()`). 

However, once that's set, one can no longer interact with how big are or small is the scale therefore, there is no way to change the equivalent of `setScaleFactor()` function available in `vtkGlyph3DMapper`.

It would be nice to have the possibility to easily set a scaleFactor.

### Feature
```
const actor = vtkActor.newInstance()
const mapper = vtkSphereMapper.newInstance()
.
.
.
mapper.setScaleArray('scale')
// Can be eventually available in the vtkSphereMapper class
mapper.setScaleFactor(30)
```

jspanchu:
### Context
- fix #2036 

### Changes
- Added `scaleFactor` property in `SphereMapper`

### PR and Code Checklist

### Testing
Run the `SphereMapper` example, try to modify `ScaleFactor` and observe the sphere radii change.
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: ubuntu
  - **Browser**: Chrome 107.0.5304.121

### DIFF
--- a/Sources/Rendering/Core/SphereMapper/index.d.ts
+++ b/Sources/Rendering/Core/SphereMapper/index.d.ts
@@ -18,6 +18,11 @@ export interface vtkSphereMapper extends vtkMapper {
 
 	/**
 	 * 
+	 */
+	getScaleFactor(): number;
+
+	/**
+	 * 
 	 * @param {Number} radius 
 	 */
 	setRadius(radius: number): boolean;
@@ -27,6 +32,13 @@ export interface vtkSphereMapper extends vtkMapper {
 	 * @param scaleArray 
 	 */
 	setScaleArray(scaleArray: any): boolean;
+
+/**
+	 * Factor multiplied with scale array elements. Radius is used when no scale array is given.
+	 * @param scaleFactor number to multiply with when a scale array is provided. 1 by default.
+	 * @see getScaleFactor(), setScaleArray(), setRadius()
+	 */ 
+	setScaleFactor(scaleFactor: number): boolean;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


Sources/Rendering/Core/SphereMapper/index.d.ts
```diff
@@ -27,6 +27,12 @@ export interface vtkSphereMapper extends vtkMapper {
 	 * @param scaleArray 
 	 */
 	setScaleArray(scaleArray: any): boolean;
+
+	/**
+	 *
```
finetjul, line 32: Add more doc such as:
```
/**
 * Factor multiplied with scale array elements. Radius is used when no scale array is given.
 * @param scaleFactor number to multiply with when a scale array is provided. 1 by default.
 * @see getScaleFactor(), setScaleArray(), setRadius()
 */ 
```

Sources/Rendering/Core/SphereMapper/index.d.ts
```diff
@@ -27,6 +27,12 @@ export interface vtkSphereMapper extends vtkMapper {
...
+	 * @param scaleFactor
+	 */
+	setScaleFactor(scaleFactor: any): boolean;
```
finetjul, line 35: not `any` but `number`

finetjul, line 35: it is also missing `getScaleFactor()`,

finetjul (changes requested):

finetjul (approved):

PR #2662, opened 2023-01-03
docs(HardwareSelector): add a new example actor

bruyeret:
Add an actor to the HardwaerSelector example.
It is a polydata containing a vert, a line and a poly.
This highlights a bug with actors containing multiple primitive types.

### Context

There is a bug in the HardwareSelector which makes it pick the wrong cell when multiple cell primitives (vert, line, poly) are used.

### Results

This pull request only provides an example highlighting the bug.

### Changes
Add an actor to the HardwareSelector example.

### PR and Code Checklist


finetjul:
LGTM

Shouldn't you create an issue to track the bug hence reference the issue in this commit ?

bruyeret:
Thanks for the review, good idea, here is the issue for this bug: #2664

PR #2666, opened 2023-01-04
feat(ImageArrayMapper): New mapper for array of mixed images

jadh4v:
### Context

Add a new class `vtkImageArrayMapper` that extends `vtkAbstractImageMapper`. This class will render a collection of images that may have variable properties such as pixel-type, number of channels, and width/height. This feature is required, since the current `vtkImageMapper` assumes such properties to be consistent across slices. Therefore, the current image mapper (`vtkImageMapper`) only supports a specific kind of image series' that have consistent properties and can be defined on a regular grid and a single pixel array (all pixels of single type). Hence, some radiological diagnostic workflows that require acquiring images along different planes as well as mixing of grayscale and RGB images in a single series is not supported in the current mapper.

### Results

ImageArrayMapper-Scrolling2DMixedImages

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: master 
  - **OS**:  Windows 11 
  - **Browser**: Chrome 109.0.5414.120 (Official Build) (64-bit)

### DIFF
--- a/Sources/Common/DataModel/Collection/index.d.ts
+++ b/Sources/Common/DataModel/Collection/index.d.ts
@@ -1,5 +1,5 @@
-import vtkDataSetAttributes from '../DataSetAttributes';
-import { vtkObject } from "../../../interfaces" ;
+import { vtkObject } from "../../../interfaces";
+import { Nullable } from "../../../types";
 
 /**
  *
@@ -62,7 +62,7 @@ export interface vtkCollection extends vtkObject {
 	/**
 	 * get the current item and provided index, returns null if index is out of bounds
 	 */
-	getItem(idx: number): vtkObject | null;
+	getItem(idx: number): Nullable<vtkObject>;
 
 	/**
 	 * Execute a passed function on every item in the collection
--- a/Sources/Common/DataModel/Collection/index.d.ts
+++ b/Sources/Common/DataModel/Collection/index.d.ts
@@ -68,7 +68,7 @@ export interface vtkCollection extends vtkObject {
 	 * Execute a passed function on every item in the collection
 	 * @param callbackfn callback function to be executed on every item in the collection
 	 */
-	forEach(callbackfn: (value: T, index: number, array: readonly T[]) => void): void;
+	forEach<T>(callbackfn: (value: T, index: number, array: readonly T[]) => void): void;
 
 	/**
 	 * Execute a passed function on every item in the collection, in order to calculate an aggregated return value.
@@ -81,7 +81,7 @@ export interface vtkCollection extends vtkObject {
 	 * Similar to forEach, but returns an array of resulting values.
 	 * @param callbackfn callback function to execute on each item in the collection, that returns a value.
 	 */
-	map(callbackfn: (value: T, index: number, array: readonly T[]) => void): void;
+	map<T>(callbackfn: (value: T, index: number, array: readonly T[]) => void): void;
 
 	/**
 	 * Check each element for modified time and update the collection's
--- a/Sources/Common/DataModel/Collection/index.d.ts
+++ b/Sources/Common/DataModel/Collection/index.d.ts
@@ -1,0 +1,118 @@
+import vtkDataSetAttributes from '../DataSetAttributes';
+import { vtkObject } from "../../../interfaces" ;
+
+/**
+ *
+ */
+export interface ICollectionInitialValues {
+	collection?: vtkObject[];
+}
+
+
+export interface vtkCollection extends vtkObject {
+
+	/**
+	 * Add item (vtkObject) to the collection
+	 * @param {vtkObject} item item to be added to the collection
+	 */
+  	addItem(item: vtkObject): void;
+
+	/**
+	 * Add item (vtkObject) to the collection _at_ the given index.
+	 * This differs from VTK-C++ where insertItem inserts at position
+	 * after the provided index value.
+	 * @param {number} idx index where the new item should be inserted.
+	 * @param {vtkObject} item item to be inserted
+	 */
+	insertItem(idx: number, item: vtkObject): void;
+
+	/**
+	 * Replace an existing item (vtkObject) with a new one
+	 * @param {number} idx index of item to be replaced
+	 * @param {vtkObject} item
+	 */
+	replaceItem(idx: number, item: vtkObject): void;
+
+	/**
+	 * Remove an existing item from the collection
+	 * @param {number|vtkObject} inValue index or reference of an item to be removed
+	 */
+	removeItem(inValue: number|vtkObject): void;
+
+	/**
+	 * Remove all items from the collection
+	 */
+	removeAllItems(): void;
+
+	/**
+	 * Check if a provided item is already present in the collection
+	 */
+	isItemPresent(item: vtkObject): boolean;
+
+	/**
+	 * Get the total number of items in the collection
+	 */
+	getNumberOfItems(): number;
+
+	/**
+	 * Check if the collection is empty
+	 */
+	empty(): boolean;
+
+	/**
+	 * get the current item and provided index, returns null if index is out of bounds
+	 */
+	getItem(idx: number): vtkObject | null;
+
+	/**
+	 * Execute a passed function on every item in the collection
+	 * @param callbackfn callback function to be executed on every item in the collection
+	 */
+	forEach(callbackfn: (value: T, index: number, array: readonly T[]) => void): void;
+
+	/**
+	 * Execute a passed function on every item in the collection, in order to calculate an aggregated return value.
+	 * @param callbackfn callback function to be used for aggregating a single value from each item in the collection
+	 * @param initialValue starting value to calculate aggregation
+	 */
+	reduce<T>(callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: readonly T[]) => T, initialValue: T): T;
+
+	/**
+	 * Similar to forEach, but returns an array of resulting values.
+	 * @param callbackfn callback function to execute on each item in the collection, that returns a value.
+	 */
+	map(callbackfn: (value: T, index: number, array: readonly T[]) => void): void;
+
+	/**
+	 * Check each element for modified time and update the collection's
+	 * MTime to the latest timestamp from individual items in the collection.
+	 */
+	updateMTimeWithElements(): void;
+}
+
+/**
+ * Method used to decorate a given object (publicAPI+model) with vtkCollection characteristics.
+ *
+ * @param publicAPI object on which methods will be bounds (public)
+ * @param model object on which data structure will be bounds (protected)
+ * @param {ICollectionInitialValues} [initialValues] (default: {})
+ */
+export function extend(publicAPI: object, model: object, initialValues? : ICollectionInitialValues): void;
+
+/**
+ * Method used to create a new instance of vtkCollection.
+ * @param {ICollectionInitialValues} [initialValues] for pre-setting some of its content
+ */
+export function newInstance(initialValues? : ICollectionInitialValues): vtkCollection;
+
+/**
+ * vtkCollection is a container of multiple vtkObject items.
+ * This can be useful for encapsulating multiple vtkObjects such as images
+ * into a single vtkObject (vtkCollection instance) to be passed as input
+ * to other filters and mappers as a single unit.
+ */
+export declare const vtkCollection: {
+	newInstance: typeof newInstance;
+	extend: typeof extend;
+};
+export default vtkCollection;
--- a/Sources/Common/DataModel/Collection/index.js
+++ b/Sources/Common/DataModel/Collection/index.js
@@ -1,0 +1,105 @@
+import macro from 'vtk.js/Sources/macros';
+
+const { vtkErrorMacro } = macro;
+
+// ----------------------------------------------------------------------------
+// vtkCollection methods
+// ----------------------------------------------------------------------------
+
+function vtkCollection(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkCollection');
+
+  publicAPI.addItem = (item) => {
+    model.collection.push(item);
+    publicAPI.modified();
+  };
+
+  publicAPI.insertItem = (idx, item) => {
+    if (idx < 0 || model.collection.length < idx) {
+      vtkErrorMacro('idx out of bounds for insertion.');
+    }
+    model.collection.splice(idx, 0, item);
+    publicAPI.modified();
+  };
+
+  publicAPI.replaceItem = (idx, item) => {
+    model.collection.splice(idx, 1, item);
+    publicAPI.modified();
+  };
+
+  publicAPI.removeItem = (inValue) => {
+    const idx =
+      typeof inValue === 'number' ? inValue : model.collection.indexOf(inValue);
+    if (idx >= 0 && idx < model.collection.length) {
+      model.collection.splice(idx, 1);
+      publicAPI.modified();
+    } else {
+      vtkErrorMacro('idx out of bounds for removeItem.');
+    }
+  };
+
+  publicAPI.removeAllItems = () => {
+    model.collection = [];
+    publicAPI.modified();
+  };
+
+  publicAPI.isItemPresent = (item) => model.collection.includes(item);
+
+  publicAPI.getNumberOfItems = () => model.collection.length;
+
+  publicAPI.empty = () => model.collection.length === 0;
+
+  publicAPI.getItem = (idx) => model.collection[idx];
+
+  publicAPI.forEach = (ftor) => {
+    model.collection.forEach(ftor);
+    // Call modified() for the collection, since ftor could have modified the elements.
+    publicAPI.updateMTimeWithElements();
+  };
+
+  publicAPI.reduce = (ftor, initialValue) =>
+    model.collection.reduce(ftor, initialValue);
+
+  publicAPI.map = (ftor) => model.collection.map(ftor);
+
+  publicAPI.updateMTimeWithElements = () => {
+    let maxMTimeOfItems = 0; // we expect time values to be positive numbers
+    for (let i = 0; i < model.collection.length; ++i) {
+      const elem = model.collection[i];
+      maxMTimeOfItems = Math.max(maxMTimeOfItems, elem.getMTime());
+    }
+
+    if (maxMTimeOfItems > publicAPI.getMTime()) {
+      publicAPI.modified();
+    }
+  };
+}
+
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {
+  collection: [],
+};
+
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  // Object methods
+  macro.obj(publicAPI, model);
+
+  // Object specific methods
+  vtkCollection(publicAPI, model);
+}
+
+// ----------------------------------------------------------------------------
+
+export const newInstance = macro.newInstance(extend, 'vtkCollection');
+
+// ----------------------------------------------------------------------------
+
+export default { newInstance, extend };
--- a/Sources/Rendering/Core/AbstractImageMapper/index.d.ts
+++ b/Sources/Rendering/Core/AbstractImageMapper/index.d.ts
@@ -1,0 +1,51 @@
+import vtkAbstractMapper, { IAbstractMapperInitialValues } from "../AbstractMapper";
+import vtkImageData from "../../../Common/DataModel/ImageData";
+
+
+export interface IAbstractImageMapperInitialValues extends IAbstractMapperInitialValues {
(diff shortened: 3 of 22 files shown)


jadh4v:
> 
@finetjul Hi Julien, the image array (or series of images) we expect can be large (say as small as 5 and as large as 40-50 images). So having that many setinputconnection ports felt sub-optimal to me.

finetjul:
If you look at how [addInputConnection()](https://github.com/Kitware/vtk-js/blob/master/Sources/macros.js#L688-L912) is implemented, you will see there is not much overhead. 
To me it does not seem worth it to not rely on the built-in mechanism.
WDYT @jourdain ?

jourdain:
I think so.

jadh4v:
I am still looking into the pipeline errors, and have to add unit tests. But I would like to start the review for the PR.
Also tagging: @floryst @sankhesh

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -786,7 +786,7 @@ function vtkOpenGLTexture(publicAPI, model) {
     flip = false
   ) => {
     // Now determine the texture parameters using the arguments.
-    publicAPI.getOpenGLDataType(dataType);
+    publicAPI.getOpenGLDataType(dataType, true);
```
jadh4v, line 789: If this function is called to construct a new texture, feel like the dataType check should always be forced.

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -64,10 +64,12 @@ function vtkOpenGLImageMapper(publicAPI, model) {
       model.openGLCamera = model._openGLRenderer.getViewNodeFor(
         ren.getActiveCamera()
       );
+      /*
       // is slice set by the camera
       if (model.renderable.getSliceAtFocalPoint()) {
         model.renderable.setSliceFromCamera(ren.getActiveCamera());
       }
+      */
```
jadh4v, line 72: I have removed/disabled this as it feels like an interaction setup, and shouldn't be part of a mapper. Ideally a backend-mapper shouldn't modify user/app facing properties of a core mapper like this. Such events can be easily setup using application side callbacks, etc.

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -64,10 +64,6 @@ function vtkOpenGLImageMapper(publicAPI, model) {
       model.openGLCamera = model._openGLRenderer.getViewNodeFor(
         ren.getActiveCamera()
       );
-      // is slice set by the camera
-      if (model.renderable.getSliceAtFocalPoint()) {
-        model.renderable.setSliceFromCamera(ren.getActiveCamera());
-      }
```
jadh4v, line 70: I have removed/disabled this as it feels like an interaction setup, and shouldn't be part of a mapper. Ideally a backend-mapper shouldn't modify user/app facing properties of a core mapper like this. Such events can be easily setup using application side callbacks, etc.

Sources/Rendering/Core/ImageArrayMapper/index.js
```diff
@@ -0,0 +1,171 @@
...
+      publicAPI.modified();
+    } else {
+      vtkErrorMacro(
+        `Slice number out of range. Acceptable range is: [0, ${totalSlices}]`
+      );
+    }
+  };
+
+  publicAPI.computeSlice = (imageNumber, subSlice) =>
+    model.sliceToSubSliceMap.findIndex((x) => x === { imageNumber, subSlice });
```
finetjul, line 130: shouldn't it be a deepEqual instead ?

Sources/Rendering/Core/ImageArrayMapper/index.d.ts
```diff
@@ -0,0 +1,223 @@
...
+export function extend(publicAPI: object, model: object, initialValues?: IImageArrayMapperInitialValues): void;
+
+/**
+ * Method use to create a new instance of vtkImageArrayMapper
+ * @param {IImageArrayMapperInitialValues} [initialValues] for pre-setting some of its content
+ */
+export function newInstance(initialValues?: IImageArrayMapperInitialValues): vtkImageArrayMapper;
+
+/**
+ * vtkImageArrayMapper provides display support for a collection single/multi-frame images.
```
finetjul, line 211: collection **of** single/multi-frame

Sources/Rendering/Core/AbstractImageMapper/index.d.ts
```diff
@@ -0,0 +1,58 @@
...
+	/**
+	 * @returns currently active image data object
+	 */
+	getCurrentImage(): vtkImageData | null;
+
+	/**
+	 * To be overridden by deriving types
+	 * @param slice return the image data object corresponding to a slice number.
+	 */
+	getImage(slice: number): vtkImageData | null;
```
finetjul, line 25: This function seems very specific to ImageArrayMapper. Do you really need it in such superclass ?

finetjul: This is suprising to me especially since ImageArrayMapper does not derive from ImageMapper.

Sources/Common/DataModel/Collection/index.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+  };
+
+  publicAPI.removeAllItems = () => {
+    model.collection = [];
+    publicAPI.modified();
+  };
+
+  publicAPI.isItemPresent = (item) =>
+    model.collection.findIndex((elem) => elem === item) < 0;
```
floryst, line 49: You can just do `model.collection.includes(item)` here.

Sources/Rendering/Core/ImageArrayMapper/index.js
```diff
@@ -0,0 +1,171 @@
...
+  //------------------
+  // Private functions
+
+  const _computeSliceToSubSliceMap = () => {
+    const inputCollection = publicAPI.getInputData();
+    if (!inputCollection || inputCollection.empty()) {
+      return;
+    }
+
+    const perImageMap = inputCollection.map((image, index) => {
```
floryst, line 30: You can also do `flatMap` here. Up to you if you think a `map` then a `flat` is more readable.

Sources/Rendering/Core/ImageArrayMapper/index.js
```diff
@@ -0,0 +1,171 @@
...
+  publicAPI.setInputData = (inputData) => {
+    _superSetInputData(inputData);
+    _computeSliceToSubSliceMap();
+  };
+
+  // Overrides vtkAbstractImageMapper.getImage()
+  publicAPI.getImage = (slice = publicAPI.getSlice() ?? 0) => {
+    const imageCollection = publicAPI.getInputData();
+    if (!imageCollection) {
+      vtkWarningMacro(`No input set.`);
```
floryst, line 57: Don't need backticks if you aren't using a template.

Sources/Rendering/Core/ImageArrayMapper/index.js
floryst: Or at least write out the comparison: `x.imageNumber === imageNumber && x.subSlice === subSlice`. As it's written, it will always eval to false.

Sources/Rendering/OpenGL/ImageMapper/index.js
floryst: Agreed. This functionality should be re-added somewhere else then, which means having to install a listener on the camera in the core mapper.

Sources/Common/DataModel/Collection/index.d.ts
```diff
@@ -0,0 +1,123 @@
+import vtkDataSetAttributes from '../DataSetAttributes';
+import { vtkObject } from "../../../interfaces" ;
+// import vtkDataSet from '../DataSet';
```
floryst, line 3: Unused comment?

Sources/Common/DataModel/Collection/index.d.ts
```diff
@@ -0,0 +1,123 @@
...
+	/**
+	 * get the current item and provided index, returns null if index is out of bounds
+	 */
+	getItem(idx: number): vtkObject | null;
+
+	/**
+	 * Execute a passed function on every item in the collection
+	 * @param callbackfn callback function to be executed on every item in the collection
+	 */
+	forEach(callbackfn: (value: T, index: number, array: readonly T[]) => void): void;
```
floryst: Need to specify the generic name: `forEach<T>(...`

floryst: Same with other instances below.

Sources/Rendering/Core/ImageArrayMapper/index.d.ts
```diff
@@ -0,0 +1,222 @@
...
+	 * @param imageNumber The image number is the index of the vtkImageData object in the input collection.
+	 * @param subSlice The subSlice number is the k-index of a slice within a vtkImageData object in the input collection.
+	 */
+	computeSlice(imageNumber: number, subSlice: number): number;
+
+	/**
+	 * Get the vtkImageData index corresponding to the provided global slice number.
+	 * @param slice global slice number
+	 */
+	getImageNumber(slice?: number): number;
```
floryst, line 79: Specify what `getImageNumber` does when not given a parameter.

floryst:
Looking awesome so far!

Sources/Rendering/Core/ImageArrayMapper/index.d.ts
```diff
@@ -0,0 +1,222 @@
...
+	 * Given a global slice number, identify the subSlice number (slice k-index within a vtkImageData).
+	 * @param slice global slice number
+	 */
+	getSubSlice(slice?: number): number;
+
+
+	/**
+	 *
+	 */
+	getResolveCoincidentTopology(): ICoincidentTopology
```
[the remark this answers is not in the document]
jadh4v: Yea, I think its a good idea. I have created a new issue from this comment.

Sources/Rendering/OpenGL/ImageMapper/index.js
jadh4v: Ideally, I was thinking the application should hook up the camera update events with the call to update the slice. Otherwise, the mapper is too tightly coupled with interaction code.

But if you want to retain the previous functionality (i.e. the `SliceAtFocalPoint` flag), we could do something like this to force the caller to provide the camera which will be tracked:
```js
  publicAPI.setSliceAtFocalPoint = (flagValue, camera) => {
    model.sliceAtFocalPoint = _superSetSliceAtFocalPoint(flagValue);
    if(flagValue)
        camera.onModified(() => setSliceFromCamera(camera));
  };
```

Sources/Rendering/Core/ImageArrayMapper/index.js
jadh4v: I will have to add `flatMap()` to `vtkCollection` first. I think the current form is fine if we want to keep utility functions on `vtkCollection` to a smaller set.

Sources/Rendering/OpenGL/ImageMapper/index.js
jadh4v: Ok I got an implementation working, I will push it to the branch so you can re-review it.

Sources/Common/DataModel/Collection/index.js
```diff
@@ -0,0 +1,104 @@
...
+  // Set our className
+  model.classHierarchy.push('vtkCollection');
+
+  publicAPI.addItem = (item) => {
+    model.collection.push(item);
+    publicAPI.modified();
+  };
+
+  publicAPI.insertItem = (idx, item) => {
+    model.collection.splice(idx, 0, item);
```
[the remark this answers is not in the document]
jadh4v: Good catch, I will fix that. Additionally, in VTK-C++, the `insertItem` function inserts the item _after_ the provided idx rather than _at_ idx. It therefore also requires that the list be non-empty. I think this behavior is not very intuitive. Do we still want to retain the same behavior in vtk-js?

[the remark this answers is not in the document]
floryst: Let's insert *at* the index, and document it as such.

Sources/Rendering/Core/ImageArrayMapper/index.js
```diff
@@ -0,0 +1,171 @@
...
+      model.slice = slice;
+      publicAPI.modified();
+    } else {
+      vtkErrorMacro(
+        `Slice number out of range. Acceptable range is: [0, ${totalSlices}]`
+      );
+    }
+  };
+
+  publicAPI.computeSlice = (imageNumber, subSlice) =>
```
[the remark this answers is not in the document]
jadh4v: Yes its basically the index of the image. I will change the name of all functions and variable names related to `imageNumber` to `imageIndex`.

Sources/Rendering/Core/AbstractImageMapper/index.d.ts
jadh4v: yes, the function is not needed in `AbstractImageMapper` or `ImageMapper`. I will remove it.

floryst:
Does this support incomplete/holey collections? Say that I have a stack of 20 DICOM slices pending download. The download order can be arbitrary, so when the `ith` slice downloads, I want to insert it in the collection at the `ith` spot. The goal is to be able to show images in the viewer as they are downloaded.

jadh4v:
> Does this support incomplete/holey collections? Say that I have a stack of 20 DICOM slices pending download, which corresponds to an empty collection with size 20. The download order can be arbitrary, so when the `ith` slice downloads, I want to insert it in the collection at the `ith` spot. The goal is to be able to show images in the viewer as they are downloaded.

Not in its current state. However, I can adapt it to handle _holes_. How should the missing images be represented? `null` or empty instance of `vtkImageData`?
Secondly, should the missing images be rendered as empty slices or only render available slices (so slice number jumps as you scroll). Or perhaps it could be a mode/flag to switch between `continuous slice numbering` vs `jumping to next available`.

floryst:
> Not in its current state. However, I can adapt it to handle holes. How should the missing images be represented? null or empty instance of vtkImageData?

After some more thought on this, maybe prepopulating with empty vtkImageData would work without any extra changes to your PR. It could be something like a 1x1x1 image representing a single slice. If you think that would work without extra changes, then I would settle for that approach. No need to make a collection with holes.

> Secondly, should the missing images be rendered as empty slices or only render available slices (so slice number jumps as you scroll). Or perhaps it could be a mode/flag to switch between continuous slice numbering vs jumping to next available.

If you know the total number of slices up front, continuous slice numbering makes sense. Jumping to the next available might make sense for unknown slice counts, though it wouldn't make much sense spatially for the user.

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -427,17 +425,34 @@ function vtkImageMapper(publicAPI, model) {
...
+      model.cameraSubscription = null;
+    }
+
+    // Setup new subscription to listen to camera modifications.
+    if (flagValue && camera) {
+      model.cameraSubscription = camera.onModified(() =>
+        publicAPI.setSliceFromCamera(camera)
+      );
+    }
+  };
```
jadh4v, line 444: @floryst Does this change look good? Basically I am directly hooking up the `setSliceFromCamera()` function with the `onModified` event of the Camera. This is to compensate for removing the call from the OpenGLImageMapper.

Sources/Common/DataModel/Collection/index.d.ts
```diff
@@ -0,0 +1,118 @@
...
+
+	/**
+	 * Check if the collection is empty
+	 */
+	empty(): boolean;
+
+	/**
+	 * get the current item and provided index, returns null if index is out of bounds
+	 */
+	getItem(idx: number): vtkObject | null;
```
daker, line 65: `Nullable<vtkObject>`
import Nullable from types file

Sources/Rendering/Core/AbstractImageMapper/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+	/**
+	 *
+	 */
+	getIsOpaque(): boolean;
+
+	/**
+  	 * Return currently active image for the mapper. Overridden by deriving classes.
+	 */
+	getCurrentImage(): vtkImageData | null;
```
daker, line 19: `Nullable<vtkObject>`

Sources/Rendering/Core/ImageArrayMapper/index.d.ts
```diff
@@ -0,0 +1,252 @@
...
+	 *  Get vtkImageData corresponding to the provided (global) slice number.
+	 * @param slice (global) slice number. If a slice number is not provided,
+	 * the function uses the current slice number (i.e. the output of getSlice()).
+	 */
+	getImage(slice?: number): vtkImageData;
+
+	/**
+  	 * Return currently active image. This depends on the currently active slice number.
+	 */
+	getCurrentImage(): vtkImageData | null;
```
daker, line 43: `Nullable<vtkObject>`

Sources/Rendering/Core/ImageMapper/index.d.ts
```diff
@@ -98,6 +99,12 @@ export interface vtkImageMapper extends vtkAbstractMapper {
 	 */
 	getSlice(): number;
 
+	/**
+  	 * Return currently active image. By default, there can only be one image
+	 * for this mapper, if an input is set.
+	 */
+	getCurrentImage(): vtkImageData | null;
```
daker, line 106: `Nullable<vtkObject>`

daker (changes requested):

Sources/Rendering/Core/ImageMapper/index.js
floryst: @jourdain @finetjul what do you think about pulling out the setSliceAtFocalPoint logic from the OpenGLImageMapper? I think we are going to keep the original functionality, but it does mean that the OpenGLImageMapper must handle special cases when dealing with either ImageMapper or ImageArrayMapper.

finetjul: Generally speaking, I do not like when a mapper modifies the camera. It is prone to hard-to-debug bugs.

floryst: In this case, the mapper is dependent on the camera for information. We have [the OpenGL mapper setting properties on the Core mapper](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/ImageMapper/index.js#L67-L70), and Raj was looking into doing this update via a camera onModified listener.

Sources/Rendering/Core/AbstractImageMapper/index.js
```diff
@@ -0,0 +1,41 @@
...
+  publicAPI.getCurrentImage = () => null;
+}
+
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {
+  slice: 0,
+  customDisplayExtent: [0, 0, 0, 0],
```
sankhesh, line 22: Shouldn't this be a 6-tuple?

Sources/Rendering/Core/AbstractImageMapper/index.js
```diff
@@ -0,0 +1,41 @@
...
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  // Build VTK API
+  vtkAbstractMapper.extend(publicAPI, model, initialValues);
+
+  macro.setGet(publicAPI, model, ['slice', 'useCustomExtents']);
+  macro.setGetArray(publicAPI, model, ['customDisplayExtent'], 4);
```
sankhesh, line 35: Array size - 6?

Sources/Rendering/OpenGL/Texture/index.js
sankhesh: It's not forced so callers can set the datatype depending on the use case.

sankhesh (approved):
LGTM with one query.

Sources/Rendering/OpenGL/Texture/index.js
jadh4v: The `create2DFromRaw` function takes a dataType as an argument but silently rejects it if `openGLDataType` is already set in the texture. I suppose, I can just override that behavior by re-setting the `openGLDataType = 0` before the call to `create2DFromRaw`.

jadh4v: Looks like I'm running into a new issue after a rebase with master:
```
[.WebGL-000055D400392300] GL_INVALID_OPERATION: Texture is immutable.
```
So I might have to delete/create the texture in the ImageMapper anyway.

Sources/Rendering/Core/ImageMapper/index.js
jadh4v: I guess, I will just revert to previous behavior. It's not directly related to this PR, so we can address this later.

jadh4v:
I believe this is GTG.

floryst (approved):
LGTM

PR #2671, opened 2023-01-13
Clean OpenGL RenderWindow eventListeners

### LINKED ISSUE #2670: OpenGL RenderWindow leaks memory though event listeners
### High-level description

OpenGL RenderWindow subscribes the canvas to two events but never unsubscribes.

### Steps to reproduce

Using Google Chrome
* Create and navigate to any view with a RenderView using openGL
* Take a memory snapshot
* Navigate to some other view where the RenderView is not visible and `renderWindow.delete()`s are called.
* Take another memory snapshot and notice that there's a leak.

### Detailed behavior

vtkJSRenderWindow_leaking
(There's a lot of leaks from application code in the screenshot but Objects content should have vtkJS objects.

### Expected behavior

No leaks

### Environment
- **vtk.js**: 25.8.5
- **OS**: 12.6.2
- **Browser**: Chrome 108.0.5359.124

Antti-Palola:
Remove `webglcontextlost` and `webglcontextrestored` eventListeners added for canvas when deleting the RenderWindow.

### Context

fix #2670

### Results

It is no longer leaking memory through these event listeners.

### Changes

No changes to API, clean up internally.

### PR and Code Checklist

### Testing

- [X] Tested environment:
 - **vtk.js**: 25.8.5
- **OS**: 12.6.2
- **Browser**: Chrome 108.0.5359.124

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -1101,6 +1101,9 @@ function vtkRenderWindowInteractor(publicAPI, model) {
         publicAPI.handleVisibilityChange
       );
     }
+    if (model.container) {
+      publicAPI.unbindEvents();
+    }
     superDelete();
   };
 
--- a/Sources/Rendering/OpenGL/RenderWindow/index.js
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.js
@@ -86,6 +86,10 @@ export function popMonitorGLContextCount(cb) {
   return GL_CONTEXT_LISTENERS.pop();
 }
 
+function _preventDefault(e) {
+  e.preventDefault();
+}
+
 // ----------------------------------------------------------------------------
 // vtkOpenGLRenderWindow methods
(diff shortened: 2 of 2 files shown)


Antti-Palola:
One thing to consider: The base RenderWindowInteractor also has `bindEvents` that registers multiple event listeners. One has to know and remember to call `unbindEvents` which for example in our project was left out.

Should the class clear the events itself on delete too? It's safe to call removeEventsListener even if user remembered to unbind explicitly as well.

floryst:
> Should the class clear the events itself on delete too? It's safe to call removeEventsListener even if user remembered to unbind explicitly as well.

I agree that the RenderWindowInteractor should unbind events (if a container exists) on delete, as that's a safer default.

These changes look good to me! If you want to, it would be great to add the RenderWindowInteractor fixes to this PR, and then rebase onto master. I can't seem to update and merge your branch.

Antti-Palola:
Updated with automatic delete and rebased.

floryst (approved):
LGTM!

PR #2675, opened 2023-01-18
fix(VolumeMapper): use the correct clipping range for key matrices for labelOutline

sedghi:
### Context
This pull request aims to fix an issue with the correct rendering of the [volumeOutline](https://kitware.github.io/vtk-js/examples/VolumeOutline.html) by ensuring the correct computation of the world to [projection matrix in the shader](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl#L512). The projection matrix calculation depends on an accurate clipping range within the camera. Previously, the `interactorMPRSlice` was [correctly setting the clipping range to be](https://github.com/Kitware/vtk-js/blob/master/Sources/Interaction/Style/InteractorStyleMPRSlice/index.js#L296) `(model.distance, model.distance + 0.1)`, but for use cases that do not use the interactorSlice, the projection matrix was being calculated incorrectly (with the wrong clipping range). This pull request addresses this issue by temporarily setting the clipping range for the wcpc calculation and then resetting it.

### Results
- Makes the use cases that renders volumeOutline without interactorMPRSlice work
 
### Changes
- Temporarily set the clipping range for wcpc computation

### PR and Code Checklist

### Testing

- [X] Tested environment:
  - **vtk.js**:  latest master -->
  - **OS**:  macOS Monterey 12.6
  - **Browser**: 109.0.5414.87 (Official Build) (x86_64)

### DIFF
--- a/Sources/Rendering/OpenGL/VolumeMapper/index.js
+++ b/Sources/Rendering/OpenGL/VolumeMapper/index.js
@@ -809,8 +809,28 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
 
       program.setUniformMatrix('vWCtoIDX', worldToIndex);
 
+      const camera = ren.getActiveCamera();
+      const [cRange0, cRange1] = camera.getClippingRange();
+      const distance = camera.getDistance();
+
+      // set the clipping range to be model.distance and model.distance + 0.1
+      // since we use the in the keyMats.wcpc (world to projection) matrix
+      // the projection matrix calculation relies on the clipping range to be
+      // set correctly. This is done inside the interactorStyleMPRSlice which
+      // limits use cases where the interactor style is not used.
+
+      camera.setClippingRange(distance, distance + 0.1);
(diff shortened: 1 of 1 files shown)


sedghi:
I'm not sure why the tests run for node 16 but not for node 18 ...

floryst:
> I'm not sure why the tests run for node 16 but not for node 18 ...

Flaky test runner on github actions. I still haven't tracked down the issue...

floryst:
I'm curious: is 0.1 an arbitrary "sufficiently correct" constant here? IIRC for the MPRSlice interactor style, that was an arbitrary slab thickness.

sedghi:
@floryst Yeah that is as you say. But since the correct projection matrix is needed to make the `labelmapOutline` work correctly it need to be set at the mapper level too, since we have the use case that we don't use the mpr interactor, but the camera is using the thin slab using clipping planes

floryst:
Seems fine for now. I think this is good to go.

PR #2678, opened 2023-01-19
perf(rendering): use texStorage for WebGL2 texture allocations

sankhesh:
Preferring texStorage2D/3D over texImage2D/3D inherently comes with a performance improvement because they ensure that the texture is complete at allocation.

### Context

Using texStorage ensures that textures are complete at allocation time

### PR and Code Checklist


jadh4v:
This appears to make the webGL textures _immutable_. Looks like it disallows any kind of reuse of textures (for e.g. in case of imageMapper slicing).
image

finetjul:
See #2706 for a related fix...

PR #2680, opened 2023-01-20
refactor(mousewheel): normalize mouse wheel event spin values

finetjul:
This attempts to fix the problem of not being able to scroll a volume by 1 slice for each mouse wheel.
Depending on the OS and the browser, doing a mouse wheel event typically procuces a spin value > 1 (always <1 for trackpad). This can be explained because mouse wheel events are designed to "scroll a page by multiple lines at once".
In VTK.js specific domain, having a scroll spin value of 1 makes sense by default, hence the normalization factor.

### Context
With most OS/browser configurations, we can't browse a volume slice by slice with the mouse wheel.

### Results
The first mouse wheel event now scrolls always by 1 (e.g. if spin was 3 it now worths 1), subsequent scrolls are a factor of the original scroll value (e.g. if spint was 6 it now worths 2 because original first spin was 3).
 
Behavior for trackpads is unchanged.

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 10
  - **Browser**: Chrome, Firefox

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -94,6 +94,9 @@ function vtkRenderWindowInteractor(publicAPI, model) {
   // map from pointerId to { pointerId: number, position: [x, y] }
   const pointerCache = new Map();
 
+  // Factor to apply on wheel spin.
+  let wheelCoefficient = 1;
+
   // Public API methods
 
   //----------------------------------------------------------------------
@@ -676,6 +679,25 @@ function vtkRenderWindowInteractor(publicAPI, model) {
       position: getScreenEventPositionFor(event),
       deviceType: getDeviceTypeFor(event),
     };
+
+    // Wheel events are thought to scroll pages (i.e. multiple lines at once).
+    // See normalizeWheel() documentation for more context.
+    // While trackpad wheel events are many small (<1) wheel spins,
+    // mouse wheel events have absolute spin values higher than 1.
+    // Here the first spin value is "recorded", and used to normalize
+    // all the following mouse wheel events.
+    if (model.wheelTimeoutID === 0) {
+      // 0.4 is roughly half-way between a large trackpad first event and small
+      // mouse wheel first event.
+      if (Math.abs(callData.spinY) > 0.4) {
+        // Event is coming from mouse wheel
+        wheelCoefficient = Math.abs(callData.spinY);
+      } else {
+        // Event is coming from trackpad
+        wheelCoefficient = 1;
+      }
+    }
+    callData.spinY /= wheelCoefficient;
 
     if (model.wheelTimeoutID === 0) {
       publicAPI.startMouseWheelEvent(callData);


floryst:
I will test this on Chrome + Firefox on Mac. I would suggest testing IE, but IIRC we technically don't support it anymore...

floryst (commented):
Chrome's spinY is now set to 1/-1! Nice!

Firefox on Mac treats my mouse as having spinY values of less than 1. For the mouse wheel, one wheel stop has a 0.6 delta to the spinY parameter.

floryst:
For one tick of my mouse wheel, `event.wheelDeltaY` (deprecated property) is ±240 on Chrome, ±48 on Firefox, ±12 on Safari. `event.deltaY` (standard property) is ±8.0004... on Chrome, ±16 on Firefox, ±4.0002... on Safari.

floryst:
Could look at [`jquery-mousewheel`](https://github.com/jquery/jquery-mousewheel/blob/main/jquery.mousewheel.js). The code is roughly the same structure-wise, with some extra normalizing steps.

finetjul:
Do you confirm that "first spin value" of your trackpad is always < 0.1 ?

Maybe I could do: 
```
if (abs(spinY) > 0.4)  { // 0.4 is roughly half-way between a large trackpad first event and small mouse wheel first event
  // it's coming from a mouse wheel, normalize
  wheelCoefficient = Math.abs(callData.spinY);
} else {
  // it's coming from a trackpad, no normalization
  wheelCoefficient = 1;
}
```

floryst:
Yep, my trackpad has a spin of 0.05. That threshold seems sufficient enough for now from the limited set of tested devices.

floryst:
@finetjul any update on this?

FezVrasta:
I'd love to see this merged 🙏

FezVrasta:
One of our devs reported a spin value of 0.34 for their mouse wheel, do you think the default value could be tweaked or made configurable?

finetjul:
I think we can lower the threshold to 0.3, no problem. 
Can you please create a MR and detail on which platform + device you have 0.34 ?

On Windows, is it sensitive to the following OS settings:
image

FezVrasta:
CleanShot 2023-03-08 at 6 26 35@2x

Is it this one on mac?

finetjul:
I think so...
It would be interesting to see what scroll range you can get.
and if your deveveloper has 0.34 by not having something by default in such settings.

WillCMcC:
Hey @finetjul! I am on mac using a logitech mx ergo. It reports a spinY of .3 on initial scroll event and 3.25 for all subsequent events. I have been unable to find a setting that can change this.

WillCMcC:
Trying out a logitech g305 and am getting consistent `1`

WillCMcC:
@finetjul https://github.com/Kitware/vtk-js/pull/2742 made the change here and tested with various periphs. Thanks for the guidance!

PR #2683, opened 2023-01-23
perf(rendering): only allocate texture buffers when data is available

sankhesh:
### Changes

### PR and Code Checklist


Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -815,24 +815,26 @@ function vtkOpenGLTexture(publicAPI, model) {
...
-        0,
-        0,
-        0,
-        model.width,
-        model.height,
-        model.format,
-        model.openGLDataType,
-        scaledData[0]
-      );
+      if (scaledData[0] != null) {
```
floryst, line 818: Does texImage2D not care if it's passed null/undefined?

floryst:
LGTM!

floryst:
Looks good! Is that the last change for this PR?

sankhesh:
Yes, this is good to go. Thanks for reviewing @floryst

PR #2685, opened 2023-01-25
docs(geometryviewer): add option to not map scalars to colors in Geom…

finetjul:
…etryViewer example

### Context
Expose the MapScalarsToColors option in the Geometry viewer.

### Results
It is then possible to use a scalar point data array that is RGB or RGBA.

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master -->
  - **OS**: Windows 11
  - **Browser**: Chrome


jourdain (approved):
LGTM

PR #2697, opened 2023-01-31
Fix HardwareSelector picking of actors with both polylines and polys

### LINKED ISSUE #2664: HardwareSelector picks the wrong cell
### High-level description

HardwareSelector picks the wrong cell when an actor contains different types of primitive.

### Steps to reproduce

Create an actor with at least two types of primitive, for example a line and a poly.
Try to pick the poly using the HW selector and FIELD_ASSOCIATION_CELLS.
It will give the line's cell ID instead of the poly's cell ID.

PR #2662 adds an example which reproduces the bug.

bruyeret:
Take cellOffset into account in `selectionMaps` when creating a VBO with `createVBO()`
Disable rendering of edges when rendering for picking
Added a test for this bug

Closes #2664

### DIFF
--- a/Sources/Rendering/OpenGL/PolyDataMapper/index.js
+++ b/Sources/Rendering/OpenGL/PolyDataMapper/index.js
@@ -1747,7 +1747,11 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
         model.drawingEdges =
           drawSurfaceWithEdges &&
           (i === primTypes.TrisEdges || i === primTypes.TriStripsEdges);
-        if (!model.drawingEdges || !model.renderDepth) {
+        // don't draw edges when rendering depth or rendering for picking
+        if (
+          !model.drawingEdges ||
+          !(model.renderDepth || model.lastSelectionState >= 0)
+        ) {
           model.lastBoundBO = model.primitives[i];
           model.primitiveIDOffset += model.primitives[i].drawArrays(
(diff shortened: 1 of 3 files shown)


Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -1747,7 +1747,11 @@ function vtkOpenGLPolyDataMapper(publicAPI, model) {
         model.drawingEdges =
           drawSurfaceWithEdges &&
           (i === primTypes.TrisEdges || i === primTypes.TriStripsEdges);
-        if (!model.drawingEdges || !model.renderDepth) {
+        // don't draw edges when rendering depth or rendering for picking
```
finetjul, line 1750: if a polydata has only polylines, shouldn't they be rendered for picking ?

bruyeret: Polylines don't share the same buffer as edges:
```
export const primTypes = {
  Start: 0,
  Points: 0,
  Lines: 1,
  Tris: 2,
  TriStrips: 3,
  TrisEdges: 4,
  TriStripsEdges: 5,
  End: 6,
};
```
Polylines are `Lines` and edges are `TrisEdges` or `TriStripsEdges`, hence the definition of `model.drawingEdges` (l. 1747-1749)

PR #2699, opened 2023-01-31
fix(radon): fix radon transform following half-float support

### LINKED ISSUE #2658: Radon Transformation Scaling Issue
### High-level description
The Radon Transform blend mode does not work if the texture data is Float32Array.

image
 

### Steps to reproduce
In the [`VolumeMapperBlendMode` example](https://github.com/Kitware/vtk-js/blob/master/Examples/Volume/VolumeMapperBlendModes/index.js#L45), comment out the line that forces the scalar texture be HalfFloat (setPreferSizeOverAccuracy makes the scalar texture half float)

### Detailed behavior
It seems like the code written for radon transform has some wrong assumption for the scaling. 

The [code in the shader](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl#L1468) for the Radon transform uses some scaling which would fail if the data type is Float32Array. 

I tried changing the 

`gl_FragData[0] = texture2D(ctexture, vec2(normalizedRayIntensity * cscale0 + cshift0, 0.5));` 

to 

`gl_FragData[0] = texture2D(ctexture, vec2(normalizedRayIntensity, 0.5));` 

and it works. So there is something going on with the color texture scale and offset.

### Expected behavior
The Radon transform should work regardless of data type

finetjul:
Regression from a6d9d32d. Closes #2658

fix #2658

### Context
Radon transform example was displaying a fully white volume.

### Results
With change the volume is now back to a grayscale volume.

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 11
  - **Browser**: Chrome

### DIFF
--- a/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
+++ b/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
@@ -1438,7 +1438,7 @@ void applyBlend(vec3 posIS, vec3 endIS, vec3 tdims)
     {
       tValue = getTextureValue(posIS);
       normalizedRayIntensity = normalizedRayIntensity - sampleDistance*texture2D(otexture, vec2(tValue.r * oscale0 + oshift0, 0.5)).r;
-      gl_FragData[0] = texture2D(ctexture, vec2(normalizedRayIntensity * cscale0 + cshift0, 0.5));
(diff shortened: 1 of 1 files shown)


floryst:
@finetjul is this good to go?

finetjul:
Yes, good to go !

floryst (approved):
LGTM!

PR #2709, opened 2023-02-07
Fix texture from image cropping when size is not a power of two

bruyeret:
When creating a texture from an image which dimensions are not powers of 2, vtk.js used to:
- create a canvas bigger than the image, which dimensions are powers of 2
- paint and stretch the image to fill the canvas
- use as texture a piece of the canvas which has the same size as the image

Fixed this behaviour by using the size of the canvas instead of the size of the image to create the texture.

To illustrate the bug, here is an example with a 1500x700 image applied to a plane
Before the fix, the image was stretched to 2048x1024 and then cropped to 1500x700

Before

image

After

image

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -1071,8 +1071,6 @@ function vtkOpenGLTexture(publicAPI, model) {
 
     model.target = model.context.TEXTURE_2D;
     model.components = 4;
-    model.width = image.width;
-    model.height = image.height;
     model.depth = 1;
     model.numberOfDimensions = 2;
     model._openGLRenderWindow.activateTexture(publicAPI);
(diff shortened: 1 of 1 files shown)


bruyeret:
It looks like this code with the powers of 2 is here to be able to generate mipmaps with `generateMipmap` or a `level` > 0 in `texImage2D`.
Should we stretch the texture only if using webgl 1, and don't do it for webgl 2? Or keep this behaviour for both ? @sankhesh

sankhesh (approved):
LGTM 👍 for now. We need to update this logic for webgl2 textures.

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1085,14 +1083,20 @@ function vtkOpenGLTexture(publicAPI, model) {
 
     // Scale up the texture to the next highest power of two dimensions (if needed) and flip y.
     const needNearestPowerOfTwo =
-      !vtkMath.isPowerOfTwo(image.width) || !vtkMath.isPowerOfTwo(image.height);
+      !model._openGLRenderWindow.getWebgl2() &&
```
sankhesh, line 1086: 👍

PR #2718, opened 2023-02-17
fix(rendering): Use texImage2D/3D on Mac WebGL2 for Norm16 textures

sankhesh:
### Context

On Apple platforms, WebKit doesn't doesn't seem to allocate textures of type short and unsigned short with `texStorage` and the `EXT_texture_norm16` extension. This change ensures that when this combination is used, texture allocation reverts to using `texImage2D/3D` calls with the norm16 extension.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: latest master
  - **OS**: macOS 
  - **Browser**: Chrome, Safari, Firefox

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -779,6 +779,34 @@ function vtkOpenGLTexture(publicAPI, model) {
   }
 
   //----------------------------------------------------------------------------
+  function useTexStorage(dataType) {
+    if (model._openGLRenderWindow) {
+      if (model.resizable) {
+        // Cannot use texStorage if the texture is supposed to be resizable.
+        return false;
+      }
+      if (model._openGLRenderWindow.getWebgl2()) {
+        const webGLInfo = model._openGLRenderWindow.getGLInformations();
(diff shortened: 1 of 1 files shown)


sankhesh:
@floryst Could you please try this branch on your mac?

floryst:
This works on my mac.

We will also need to test intel macs, since the issue does manifest there as well. I can run a test later today.

billhoffman:
Is there a way to have CI test this type of thing with web development?

sedghi:
This worked fine for me, can we merge it? 
image

sedghi:
@billhoffman Much interested in your question, I haven't personally found any services that offer different browsers on different **hardwares**. We used to use browserStack for having different browser on different OS, and https://percy.io/

floryst:
> Is there a way to have CI test this type of thing with web development?

It would be interesting to see if this would have been caught in a Github Actions macOS runner. Unfortunately, it is somewhat hardware specific.

> @floryst I have Intel mac and this worked fine for me, can we merge it?

Yep! Now that I have confirmation, I think we are good to go.

PR #2720, opened 2023-02-21
fix(picking): point picking Z-values with hardware selector

sankhesh:
When picking points, the hardware selector ensures that the depthbuffer is not written to in the selection render pass. This change removes the additional preserve depth buffer call that is unnecessary and was generating wrong z-buffer values between successive selections.

### Context

Fixes https://github.com/Kitware/react-vtk-js/issues/93

### PR and Code Checklist

### DIFF
--- a/Examples/Geometry/Picking/controller.html
+++ b/Examples/Geometry/Picking/controller.html
@@ -1,0 +1,19 @@
+<table>
+  <tr>
+    <td>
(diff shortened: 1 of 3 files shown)


finetjul:
@bruyeret

jourdain:
Tested with vue-vtk-js and it does indeed solve the picking issue...

PR #2723, opened 2023-02-22
Bypass mouse scroll debounce

ev-elucidbio:
…ce for mouse scroll events

Exposed a function to allow the consuming system to bypass the debounce on the scroll. The bypass needs to be called explicitly, allowing backwards compatability

### Context
We were encountering an issue with the scroll events being debounced. Our back end wanted every scroll from the remote view.

### Results

This change exposes a function that will allow the consumer to bypass the debounce. It must be called explicitly. This decision was made so that it would be impossible to accidentally turn off the debounce and potentially overload the server.
### Changes

### PR and Code Checklist

### Testing

I manually tested this change. Can be verified by calling the following function from the `RemoteView` 
```
setMouseScrollDebounceBypass(true);
```
and observing the traffic on the webserver for mouse wheel events.

- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Ubuntu 22.04.2 LTS
  - **Browser**: Chrome 110.0.5481.100, FireFox 110.0

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
@@ -167,6 +167,11 @@ export interface vtkRenderWindowInteractor extends vtkObject {
 	getPreventDefaultOnPointerUp(): boolean;
 
 	/**
+	 *  @default false
+	 */
+	getMouseScrollDebounceByPass(): boolean;
+
+	/**
 	 * 
 	 * @param {IRenderWindowInteractorEvent} callData 
 	 */
@@ -851,6 +856,13 @@ export interface vtkRenderWindowInteractor extends vtkObject {
 	 */
 	setPreventDefaultOnPointerUp(preventDefault: boolean): boolean;
 
+	/**
+	 * Allow system to bypass scrolling debounce. This function must be called to allow the debounce to be bypassed
+	 * @param mouseScrollDebounceByPass
+	 */
+
+	setMouseScrollDebounceByPass(mouseScrollDebounceByPass:boolean): void;
+
    /**
 	 *
 	 * @param recognizeGestures
@@ -1181,13 +1193,6 @@ export interface vtkRenderWindowInteractor extends vtkObject {
 	 * Stop animating if the renderWindowInteractor is deleted.
 	 */
 	delete(): void;
-
-	/**
-	 * Allow system to bypass scrolling debounce. This function must be called to allow the debounce to be bypassed
-	 * @param mouseScrollDebounceByPass
-	 */
-
-	setMouseScrollDebounceByPass(mouseScrollDebounceByPass:boolean): void;
 }
 
 
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
@@ -1193,6 +1193,12 @@ export interface vtkRenderWindowInteractor extends vtkObject {
 	 * Stop animating if the renderWindowInteractor is deleted.
 	 */
 	delete(): void;
+
+	/**
+	 * Allow system to bypass scrolling debounce. This function must be called to allow the debounce.
+	 */
+
+	setMouseScrollDebounceBypass(bypass:boolean): void;
 }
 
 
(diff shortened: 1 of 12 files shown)


jourdain:
Any reason for not using the macro for the set/get?

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -1108,6 +1114,10 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     superDelete();
   };
 
+  publicAPI.setMouseScrollDebounceBypass = (byPass) => {
+    model.bypassDebounce = byPass;
+  };
```
ev-elucidbio, line 1119: @jourdain - if I understand what you're asking for https://github.com/Kitware/vtk-js/pull/2723#issuecomment-1441073973 correctly, you're asking if I can remove this and do something like
```
vtkRenderWindowInteractor.extend('vtkRenderWindowInteractor', 'bypassDebounce', true);
```
from the `RemoteView` function?

jourdain: No sorry, I meant in `Sources/Rendering/Core/RenderWindowInteractor/index.js` why you are not using the (set) macro that will trigger a `modified()` when `setMouseScrollDebounceBypass` get called. (That means you do not need to write that function. 

Then for the other classes, can we just leave it as-is and rely on the fact that you get the interactor from getters and set it on the interactor only?

jourdain: Basically just doing the following change (+updating the TS).

__Sources/Rendering/Core/RenderWindowInteractor/index.js__

```javascript

const DEFAULT_VALUES = {
	// ...
}

	// ...
	macro.setget(publicAPI, model, [
            'mouseScrollDebounceByPass',
        ]);
	// ...
}

```

jourdain: And that part that I forgot

```js
    if (model. mouseScrollDebounceByPass) {
      publicAPI.extendAnimation(600);
      publicAPI.endMouseWheelEvent();
      model.wheelTimeoutID = 0;
    } else {
      // start a timer to keep us animating while we get wheel events
      model.wheelTimeoutID = setTimeout(() => {
        publicAPI.extendAnimation(600);
        publicAPI.endMouseWheelEvent();
        model.wheelTimeoutID = 0;
      }, 200);
    }
```

ev-elucidbio: will implement. Thank you!

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
```diff
@@ -1182,9 +1184,10 @@ export interface vtkRenderWindowInteractor extends vtkObject {
 
 	/**
 	 * Allow system to bypass scrolling debounce. This function must be called to allow the debounce to be bypassed
+	 * @param mouseScrollDebounceByPass
 	 */
 
-	setMouseScrollDebounceBypass(bypass:boolean): void;
+	setMouseScrollDebounceByPass(mouseScrollDebounceByPass:boolean): void;
```
jourdain, line 1190: Need the get as well

Sources/Rendering/Misc/RemoteView/index.d.ts
```diff
@@ -126,7 +126,7 @@ export interface vtkRemoteView extends vtkObject {
    * 
    *  @param bypass
    */
-  setMouseScrollDebounceBypass(bypass: boolean): void;
+  setMouseScrollDebounceByPass(bypass: boolean): void;
```
jourdain, line 129: Can we just skip that method and call `remoteView.getInteractor().setMouseScrollDebounceByPass(true)` instead?

Sources/Rendering/Misc/RemoteView/index.js
```diff
@@ -93,8 +93,8 @@ function vtkRemoteView(publicAPI, model) {
     publicAPI.modified();
   };
 
-  publicAPI.setMouseScrollDebounceBypass = (bypass) => {
-    model.interactor.setMouseScrollDebounceBypass(bypass);
+  publicAPI.setMouseScrollDebounceByPass = (bypass) => {
+    model.interactor.setMouseScrollDebounceByPass(bypass);
```
jourdain, line 97: Can we just skip that method and call `remoteView.getInteractor().setMouseScrollDebounceByPass(true)` instead?

jourdain:
Did you test it? Does that still work fine for you?

ev-elucidbio:
> Did you test it? Does that still work fine for you?

Yes

ev-elucidbio:
image

image

ev-elucidbio:
closing - opened another pr due to an issue with a commit message

PR #2729, opened 2023-02-27
Add sphere mapper picking support

### LINKED ISSUE #2659: HardwareSelector "Error linking shader fragment..."
### High-level description
I have a bunch of spheres produced by `vtkSphereMapper`. I'm trying to use `hardwareSelector` to pick single spheres. An error is returned in the console `Error linking shader FRAGMENT varying vertexIDVSOutput does not match any VERTEX varying`

### Steps to reproduce
See [this codesandbox](https://codesandbox.io/s/vtk-react-mst-f1xyml?file=/src/scene/SceneManager.tsx:2954-2969)

### Detailed behavior
image

### Environment
- **vtk.js**: 25.15.0
- **OS**: Mac Ventura 13.0
- **Browser**: Safari Version 16.1 (18614.2.9.1.12)

bruyeret:
Add picking support for the sphere mapper
Add picking to the `SphereMapper` example

Closes #2659

### DIFF
--- a/Sources/Rendering/Core/SphereMapper/example/index.js
+++ b/Sources/Rendering/Core/SphereMapper/example/index.js
@@ -11,7 +11,12 @@ import vtkPlaneSource from '@kitware/vtk.js/Filters/Sources/PlaneSource';
 import vtkSphereMapper from '@kitware/vtk.js/Rendering/Core/SphereMapper';
 
 import { AttributeTypes } from '@kitware/vtk.js/Common/DataModel/DataSetAttributes/Constants';
-import { FieldDataTypes } from '@kitware/vtk.js/Common/DataModel/DataSet/Constants';
+import {
+  FieldDataTypes,
+  FieldAssociations,
+} from '@kitware/vtk.js/Common/DataModel/DataSet/Constants';
+
+import { throttle } from '@kitware/vtk.js/macros';
 
 import controlPanel from './controlPanel.html';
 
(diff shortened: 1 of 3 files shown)


Sources/Rendering/Core/SphereMapper/example/index.js
```diff
@@ -87,11 +92,50 @@ renderer.addActor(actor);
...
+hwSelector.setFieldAssociation(FieldAssociations.FIELD_ASSOCIATION_CELLS);
+
+function eventToWindowXY(event) {
+  // We know we are full screen => window.innerXXX
+  // Otherwise we can use pixel device ratio or else...
+  const { clientX, clientY } = event;
+  const [width, height] = apiSpecificRenderWindow.getSize();
+  const x = Math.round((width * clientX) / window.innerWidth);
+  const y = Math.round(height * (1 - clientY / window.innerHeight)); // Need to flip Y
+  return [x, y];
```
finetjul, line 108: can't you use `displayToNormalizedDisplay()` or some other related function?

bruyeret: I've used the same function as in the HardwareSelector example
The y coordinates needs to be flipped, I don't know if there is a function which does this kind of things
Using `displayToNormalizedDisplay()` doesn't work as expected, the x coordinate isn't 1 when my mouse is at the right border of the page

PR #2730, opened 2023-02-27
Fix widgets

finetjul:
### Context
Improve the ResliceCursorWidget following the recent breaking change in v26

### Results
Allow fine control over visiblity and pickability of the ResliceCursorWidget handles.

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 11
  - **Browser**: Chrome latest


sedghi:
@finetjul Can we know what bug(s) does this PR fixes regarding the lookup table and polyData? thanks

finetjul:
Sure thing, this is this commit : https://github.com/Kitware/vtk-js/pull/2730/commits/87e90142b9e05c1ba8012bb2c47c9bad1d3c51ac

The problem was that if you had a polydata with a point data array that contains RGBA  values, the alpha values were not considered to decide whether the polydata is an opaque geometry or a translucent geometry. Hence the polydata was always rendered in the opaque render pass. This could be fixed by forcing `setForceTranslucent` on the actor, but the default behavior is now correct with this commit.

PR #2736, opened 2023-03-05
value can be optionally processed when call set-model-field

jinjilynn:
to be honest, i really need to pass a param through 'modified()', so this is an alternative

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/macros.js
+++ b/Sources/macros.js
@@ -512,7 +512,11 @@ const objectSetterMap = {
   object(publicAPI, model, field) {
     return (value) => {
       if (!DeepEqual(model[field.name], value)) {
-        model[field.name] = value;
+        var _value = value;
+        if(typeof field.processed === 'function'){
+            _value = field.processed(value);
+        }
+        model[field.name] = _value;
         publicAPI.modified();
         return true;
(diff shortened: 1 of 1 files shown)


finetjul:
Can you please add tests to `testMacro.js`.

Alternatively, we could maybe make it more generic (i.e. works for all setters) as such:
```
if (!DeepEqual(model[field.name], value)) {
  if(publicAPI.[`_${capitalize(field.name)}AboutToBeChanged`] != null) {
    publicAPI.[`_${capitalize(field.name)}AboutToBeChanged`](_value);
  }
  const oldValue = model[field.name];
  model[field.name] = _value;
  if(publicAPI.[`_${capitalize(field.name)}Changed`] != null) {
    publicAPI.[`_${capitalize(field.name)}Changed`](oldValue);
  }
  publicAPI.modified();
}
```
WDYT @floryst @jourdain ?

jourdain:
A test will definitely be great along with some explanation of the use case.

PR #2738, opened 2023-03-07
feat: extend objectSetterMap for macros 

jinjilynn:
add process check and run it if available

### DIFF
--- a/Sources/macros.js
+++ b/Sources/macros.js
@@ -513,8 +513,8 @@ const objectSetterMap = {
     return (value) => {
       if (!DeepEqual(model[field.name], value)) {
         let _value = value;
-        if (typeof field.process === 'function') {
-          _value = field.process(value);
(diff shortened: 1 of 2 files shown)


Sources/macros.js
```diff
@@ -512,7 +512,11 @@ const objectSetterMap = {
   object(publicAPI, model, field) {
     return (value) => {
       if (!DeepEqual(model[field.name], value)) {
-        model[field.name] = value;
+        let _value = value;
+        if (typeof field.process === 'function') {
```
finetjul, line 516: I wouldn't bother checking the type of `process`, simply check if it exists. It would be expected to be a function.
Maybe rename `process` with `aboutToBeChanged` or some more explicit naming as such.

jinjilynn: thank you for your advice

finetjul:
Hi @jinjilynn , could you please explain why you closed your PR ?

PR #2740, opened 2023-03-08
Fix LIC and translucency

bruyeret:
Order independant translucent pass uses draw buffer location 1.
Surface LIC was also using location 1 making it impossible to use in a scene with translucent objects.
Moved locations 1 and 2 to locations 2 and 3.

The order independant translucent pass used an 'opaquePass' to get the Z-buffer.
This pass doesn't work for objects rendered with LIC and is not optimized for Z buffer rendering (no need to compute colors, or render edges). As a result, translucent objects were renderd on top of LIC objects.
Replaced this 'opaquePass' with an 'opaqueZBufferPass'.


finetjul:
LGTM

finetjul (approved):

sankhesh:
LGTM :+1:

sankhesh (approved):

PR #2742, opened 2023-03-09
fix(renderwindowinteractor): adjust the threshold for spinY values when detecting trackpads

WillCMcC:
Discussion here: https://github.com/Kitware/vtk-js/pull/2680

### Context
vtkjs uses a spinY threshold of `> .4 ` to decide if a wheel event is sent from the trackpad. 
Instead we want it to use `>=.3` to detect some common logitech mice

### Results
Wheel scroll on logitech mice now allows slice by slice scrolling

### Changes
a change to a hardcoded value in the conditional logic of RenderWindowInteractor

### PR and Code Checklist

### Testing
Test various mice to make sure that scroll still feels responsive. I have tested a logitech mx ergo (the culprit), a logitech g305, a logitech g203, and an apple magic trackpad.

### Funding
This contribution is funded by [Sirona Medical](https://sironamedical.com/).

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -687,9 +687,10 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     // Here the first spin value is "recorded", and used to normalize
     // all the following mouse wheel events.
     if (model.wheelTimeoutID === 0) {
-      // 0.4 is roughly half-way between a large trackpad first event and small
-      // mouse wheel first event.
-      if (Math.abs(callData.spinY) > 0.4) {
+      // we attempt to distinguish between trackpads and mice
+      // .3 will be larger than the first trackpad event,
+      // but small enough to detect some common edge case mice
+      if (Math.abs(callData.spinY) >= 0.3) {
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -687,9 +687,9 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     // Here the first spin value is "recorded", and used to normalize
     // all the following mouse wheel events.
     if (model.wheelTimeoutID === 0) {
-      // 0.4 is roughly half-way between a large trackpad first event and small
+      // >=0.3 will permit us to detect mouse wheel events from certain mice with low spinY values
```
finetjul, line 690: this line does not make sense with next line.

It is better to use the passive form (will permit us to detect -> allows to detect).

I would keep the information that the threshold is meant to split between "mice" and "trackpads"

WillCMcC: Gotcha. I've update the comment 👍

finetjul (approved):

jadh4v:
I have cleaned the branch to do a rebase on the latest master. I will merge once the pipelines clear.

PR #2745, opened 2023-03-09
fix(MouseRangeManipulator): asymmetric scroll wheel behaviour

jadh4v:
Fix a bug where the `scrollListener` behaves in an asymmetric manner when comparing increasing value against decreasing value. This specifically occurs when using fractional `scale` values such as `0.5`. Bug is due to the use of `Math.round()` function and its application to the independently calculated `newValue`. Additionally, `Math.round()` also behaves asymmetrically when applied to the value of 0.5, as it always rounds towards `+infinity`.

The correct approach is to calculate the number of accumulated steps from the `newDelta` and round down to the closest lower integer using `Math.floor()`. This new approach also requires that the `listener.step` value be always positive. A required check has been added to the input value. If the caller needs to flip the scroll wheel direction, they should use negative `scale` values instead.

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  latest master
  - **OS**:  Windows 11 Pro, 22H2 22621.1265
  - **Browser**:  Firefox 110.0.1 (64-bit)


jadh4v:
cc: @WillCMcC

PR #2746, opened 2023-03-13
Rename opaqueZBufferPass to zBufferPass and create a real opaqueZBufferPass

bruyeret:
The old opaqueZBufferPass rendered the depth of translucent actors too
Renamed it zBufferPass and created a new opaqueZBufferPass which only renders the depth of opaque actors

opaqueZBufferPass is used by the OrderIndependantTranslucentPass to occlude fragments hidden by opaque actors
zBufferPass is used when picking or rendering depth for volume occlusion

### DIFF
--- a/Sources/Rendering/OpenGL/Actor/index.js
+++ b/Sources/Rendering/OpenGL/Actor/index.js
@@ -44,7 +44,8 @@ function vtkOpenGLActor(publicAPI, model) {
     }
   };
 
-  publicAPI.traverseOpaqueZBufferPass = (renderPass) => {
+  // render both opaque and translucent actors
+  publicAPI.traverseZBufferPass = (renderPass) => {
     if (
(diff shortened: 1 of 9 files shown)


Sources/Rendering/OpenGL/Actor/index.js
```diff
@@ -60,6 +61,11 @@ function vtkOpenGLActor(publicAPI, model) {
     publicAPI.apply(renderPass, false);
   };
 
+  // only render opaque actors
+  publicAPI.traverseOpaqueZBufferPass = (renderPass) => {
+    publicAPI.traverseOpaquePass(renderPass);
+  };
```
finetjul, line 67: remove `{` and `};` ?

sankhesh (approved):
LGTM :+1:

PR #2760, opened 2023-03-21
Fix the issues where Chrome times out when running tests

floryst:
### Context
Chrome was hanging due to `window.__karma__.complete` not being called.
This had two root causes:

1. Chrome's timer task queue is no longer deprioritized to the end of
DOM loading (due to deprecation of a field test)
2. BufferedObjectPipe doesn't drain the existing buffer when the reader
is set. For some reason, Chrome sets the reader on the buffered pipe
later than Firefox does.

### Results
- Chrome should now report completed statuses when running the browser tests.
- Instead of editing `karma.conf.js` to modify the test targets, you now must edit `Sources/Testing/index.js`.

### Changes

- [x] karma testing configuration updated

### PR and Code Checklist

### Testing
Passing CI likely means this worked.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -70,5 +70,9 @@ module.exports = {
     '**/example_/*.js',
     // ignore js files in utilities
     'Utilities/**/*.js',
+    // ignore configs
+    'karma.conf.js',
+    'webpack.*.js',
+    '.eslintrc.js',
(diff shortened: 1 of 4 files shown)


finetjul:
Did not try but I'm pretty glad you figured it out ! Well done !

PR #2764, opened 2023-03-22
fix(RemoteMouse): send position in wheel event

psavery:
This modifies the remote mouse events to use `createRemoteEvent()` similarly to the other mouse events. One thing this adds is the `x` and `y` positions of the mouse, which can be used to determine which renderer the mouse is hovering over.

This is needed as a first step for fixing pyvista/pyvista#4020

The next step is to fix the web protocol to use the mouse position to determine which renderer the mouse is over.

I have tested this locally and it appears to work properly.

### PR and Code Checklist


jourdain (approved):
LGTM

PR #2766, opened 2023-03-23
refactor(macros): add support for _onPropChanged callback

finetjul:
By adding a function _onPropChanged into the model, the function will be called each time prop is changed but before modified() is called.

### Context
Sometimes it is useful to call extra code whenever a prop changes.

### Results
Each time an XXX prop is modified via the `macro.set`, `macro.setGet` or `macro.setArray` and if `model` contains `_onXXXChanged`, then it will be called automatically after changing the value but before `modified` is called.

Closes #2765 

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 11
  - **Browser**: Chrome latest

### DIFF
--- a/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
+++ b/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
@@ -35,6 +35,8 @@ function vtkOrientationMarkerWidget(publicAPI, model) {
     }
   }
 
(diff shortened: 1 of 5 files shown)


jourdain (approved):
LGTM

I was just debating for the param order
(publicAPI, model, new, old) vs (publicAPI, model, old, new)

PR #2770, opened 2023-03-30
fix(lights): do not add duplicate lights

psavery:
In trame-vtk, we have a call to [add a light](https://github.com/Kitware/trame-vtk/blob/e2c09edfbfc28ce57959174d72e1eaa6527d4f86/trame_vtk/modules/vtk/serializers/render_windows.py#L48-L50).

Sometimes, this would be triggered multiple times, like if the user toggled
local view on/off.

It doesn't really make sense to add the same light twice (you can just modify
the intensity instead). So, similar to [the view prop check](https://github.com/Kitware/vtk-js/blob/24d26e040e3454f61b882ab7d95982d084f03a9d/Sources/Rendering/Core/Viewport/index.js#L19-L25),
add a check to prevent the same light from being added twice.

### PR and Code Checklist

### DIFF
--- a/Sources/Rendering/Core/Renderer/index.js
+++ b/Sources/Rendering/Core/Renderer/index.js
@@ -181,11 +181,10 @@ function vtkRenderer(publicAPI, model) {
     publicAPI.modified();
   };
 
-  publicAPI.hasLight = (light) =>
-    !!model.lights.filter((item) => item === light).length;
+  publicAPI.hasLight = (light) => model.lights.includes(light);
   publicAPI.addLight = (light) => {
     if (light && !publicAPI.hasLight(light)) {
-      model.lights = model.lights.concat(light);
+      model.lights.push(light);
       publicAPI.modified();
     }
   };
--- a/Sources/Rendering/Core/Renderer/index.js
+++ b/Sources/Rendering/Core/Renderer/index.js
@@ -181,9 +181,13 @@ function vtkRenderer(publicAPI, model) {
     publicAPI.modified();
   };
 
+  publicAPI.hasLight = (light) =>
+    !!model.lights.filter((item) => item === light).length;
   publicAPI.addLight = (light) => {
-    model.lights = [].concat(model.lights, light);
-    publicAPI.modified();
+    if (light && !publicAPI.hasLight(light)) {
+      model.lights = model.lights.concat(light);
+      publicAPI.modified();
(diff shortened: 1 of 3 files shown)


Sources/Rendering/Core/Renderer/index.js
```diff
@@ -181,9 +181,13 @@ function vtkRenderer(publicAPI, model) {
     publicAPI.modified();
   };
 
+  publicAPI.hasLight = (light) =>
+    !!model.lights.filter((item) => item === light).length;
```
floryst, line 185: This can be written as `model.lights.includes(light)`.

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -181,9 +181,13 @@ function vtkRenderer(publicAPI, model) {
...
   publicAPI.addLight = (light) => {
-    model.lights = [].concat(model.lights, light);
-    publicAPI.modified();
+    if (light && !publicAPI.hasLight(light)) {
+      model.lights = model.lights.concat(light);
```
floryst, line 188: `model.lights.push(light)` is more succinct.

Sources/Rendering/Core/Renderer/index.js
psavery: Cool. I was copying this from the [the view prop one](https://github.com/Kitware/vtk-js/blob/24d26e040e3454f61b882ab7d95982d084f03a9d/Sources/Rendering/Core/Viewport/index.js#L19-L25). Should I change that one too?

floryst: Yeah sure, that'll be good to simplify as well.

psavery: Done.

Sources/Rendering/Core/Renderer/index.js
psavery: Done.

psavery:
> I'm surprised this wasn't the case for addLight, so thanks!
> 
> Since you are adding a `vtkRenderer.hasLight()` method, it'll be great if you can add the method signature to the types [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/Renderer/index.d.ts#L66-L70).

Done!

jourdain (approved):

PR #2773, opened 2023-03-31
fix(SynchronizableRenderWindow): do not delete state properties

psavery:
Before, the state properties that were being converted to an array were being deleted. This meant that if the same state was loaded again (which can happen, for instance, if it gets mounted again by the visibility being toggled), the behavior in the second mount would be different.

In particular, we saw that a normals array was being removed on the second mount because it essentially thought that the normals array was no longer a part of the data (even though it was - it was just moved out of the properties and into the array).

To fix this, do not delete state properties, but instead, create a local copy that has the properties removed before proceeding (the properties still need to be removed for some further steps). This fixes the round-trip issue.

### PR and Code Checklist


jourdain (approved):
LGTM

PR #2774, opened 2023-04-02
feat(interactorstyleimage): allow camera dollying in Image3D mode

sankhesh:
This change lets the user dolly the camera in vtkInteractorStyleImage's Image3D mode with Ctrl+Shift or Alt+Shift keyboard modifiers.

### Changes

### PR and Code Checklist

### Testing

No new testing is needed for this change as it is related to user interaction.


Sources/Interaction/Style/InteractorStyleImage/index.js
```diff
@@ -52,7 +52,11 @@ function vtkInteractorStyleImage(publicAPI, model) {
       publicAPI.startWindowLevel();
     } else if (model.interactionMode === 'IMAGE3D' && callData.shiftKey) {
       // If shift is held down, do a rotation
```
floryst, line 54: This comment should be updated to reflect this change in controls, or just remove the comment entirely since the code is self-explanatory.

Do we have any other docs that reference this? If so, we would need to update them.

floryst (approved):
LGTM!

PR #2784, opened 2023-04-11
perf(Texture): cache scale/offset using dataArray

bruyeret:
computeScaleOffsets used to iterate over all pixels of the volume to find min and max
Using a vtkDataArray instead of raw data enables caching of min/max values per component
create3DFilterableFromRaw uses create3DFilterableFromDataArray and create a wrapper array around the raw data
create3DFilterableFromRaw is not used anymore (deprecate ?)

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -1,9 +1,7 @@
 import Constants from 'vtk.js/Sources/Rendering/OpenGL/Texture/Constants';
 import HalfFloat from 'vtk.js/Sources/Common/Core/HalfFloat';
 import * as macro from 'vtk.js/Sources/macros';
-import vtkDataArray, {
-  STATIC as dataArrayHelpers,
-} from 'vtk.js/Sources/Common/Core/DataArray';
+import vtkDataArray from 'vtk.js/Sources/Common/Core/DataArray';
 import * as vtkMath from 'vtk.js/Sources/Common/Core/Math';
 import vtkViewNode from 'vtk.js/Sources/Rendering/SceneGraph/ViewNode';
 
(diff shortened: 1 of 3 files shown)


Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1340,17 +1329,65 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+    // Original is read only, copy is read/write
+    // Use the copy as volumeInfo.scale and volumeInfo.offset
+    const scaleOffsetsCopy = {
+      offset: [],
+      scale: [],
+    };
+    for (let c = 0; c < numComps; ++c) {
+      scaleOffsetsCopy.offset[c] = scaleOffsets.offset[c];
+      scaleOffsetsCopy.scale[c] = scaleOffsets.scale[c];
+    }
```
finetjul, line 1389: use deepcopy ?

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1186,19 +1184,10 @@ function vtkOpenGLTexture(publicAPI, model) {
...
-
-    const offset = [];
-    const scale = [];
+  // Compute scale and offset per component from min and max per component
+  function computeScaleOffsets(min, max, numComps) {
+    const offset = new Array(numComps);
+    const scale = new Array(numComps);
     for (let c = 0; c < numComps; ++c) {
       if (min[c] === max[c]) {
         max[c] = min[c] + 1.0;
```
finetjul, line 1193: warning, this modifies input `max` . There's probably something safer to do...

Sources/Rendering/OpenGL/Texture/index.js
bruyeret: Thanks, I didn't know about `structuredClone`!

sedghi:
This is a great PR, thanks indeed.
I'm just curious about the potential memory leak that might happen because of the vtkDataArray creation. When is that getting cleaned up?

CC @bruyeret @finetjul

finetjul:
to me that vtkDataArray is being GCed after calling the function (because no one relies on it). To make sure you can check when "delete()" is being called.

PR #2786, opened 2023-04-12
fix(imageMapper): not exact half float should be under a flag

sedghi:
### Context

Half float in ImageMapper was always used even if there was potential inaccurate representation. Since half float uses 11 bits compared to short which has 15 bits, there is a potential that the values are above 2048 which makes the rendering inaccurate. See https://github.com/Kitware/vtk-js/pull/2046#discussion_r699478690 for more information

### Results and Changes

Add `preferSizeOverAccuracy` for imageMapper similar to Volume mapper PR for half float under a flag which was previously done in https://github.com/Kitware/vtk-js/pull/2046 

Although not sure where to put the new publicAPI docs, since there is no index.d.ts for the openGL texture 

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: Latest
  - **OS**: macOS
  - **Browser**: Chrome

### DIFF
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -1062,17 +1062,21 @@ function vtkOpenGLImageMapper(publicAPI, model) {
         vtkErrorMacro('Reformat slicing not yet supported.');
       }
 
-      const { offset, scale } = model.openGLTexture.computeScaleOffsets(
-        numComp,
-        scalars
-      );
+      // For now, fix the image mapper, but we should probably fix the
+      // vtkImageArrayMapper as well
+      if (!model.renderable.isA('vtkImageArrayMapper')) {
+        const { offset, scale } = model.openGLTexture.computeScaleOffsets(
+          numComp,
+          scalars
+        );
 
-      model.openGLTexture.setUseHalfFloat(
-        dataType,
-        offset,
-        scale,
-        model.renderable.getPreferSizeOverAccuracy()
-      );
+        model.openGLTexture.setUseHalfFloat(
+          dataType,
+          offset,
+          scale,
+          model.renderable.getPreferSizeOverAccuracy()
+        );
+      }
 
       model.openGLTexture.create2DFromRaw(
         dims[0],
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -1062,6 +1062,18 @@ function vtkOpenGLImageMapper(publicAPI, model) {
         vtkErrorMacro('Reformat slicing not yet supported.');
       }
 
+      const { offset, scale } = model.openGLTexture.computeScaleOffsets(
+        numComp,
+        scalars
+      );
+
+      model.openGLTexture.setUseHalfFloat(
+        dataType,
+        offset,
+        scale,
+        model.renderable.getPreferSizeOverAccuracy()
+      );
+
       model.openGLTexture.create2DFromRaw(
         dims[0],
         dims[1],
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -1186,7 +1186,7 @@ function vtkOpenGLTexture(publicAPI, model) {
     return true;
   };
 
-  function computeScaleOffsets(numComps, numPixelsIn, data) {
+  publicAPI.computeScaleOffsets = (numComps, data) => {
     // compute min and max values per component
     const min = [];
     const max = [];
@@ -1207,7 +1207,7 @@ function vtkOpenGLTexture(publicAPI, model) {
       scale[c] = max[c] - min[c];
     }
     return { scale, offset };
-  }
+  };
 
   // HalfFloat only represents numbers between [-2048, 2048] exactly accurate,
   // for numbers outside of this range there is a precision limitation
@@ -1224,7 +1224,12 @@ function vtkOpenGLTexture(publicAPI, model) {
     return true;
   }
 
-  function checkUseHalfFloat(dataType, offset, scale, preferSizeOverAccuracy) {
+  publicAPI.setUseHalfFloat = (
+    dataType,
+    offset,
+    scale,
+    preferSizeOverAccuracy
+  ) => {
     publicAPI.getOpenGLDataType(dataType);
 
     let useHalfFloat = false;
@@ -1236,18 +1241,15 @@ function vtkOpenGLTexture(publicAPI, model) {
         halfFloatExt && model.openGLDataType === halfFloatExt.HALF_FLOAT_OES;
     }
 
-    if (!useHalfFloat) {
-      return false;
+    if (
+      useHalfFloat &&
+      (hasExactHalfFloat(offset, scale) || preferSizeOverAccuracy)
+    ) {
+      model.useHalfFloat = true;
+    } else {
+      model.useHalfFloat = false;
     }
-
-    // Don't consider halfFloat and convert back to Float when the range of data does not generate an accurate halfFloat
-    // AND it is not preferable to have a smaller texture than an exact texture.
-    if (!hasExactHalfFloat(offset, scale) && !preferSizeOverAccuracy) {
-      return false;
-    }
-
-    return true;
-  }
+  };
 
   //----------------------------------------------------------------------------
   publicAPI.create3DFromRaw = (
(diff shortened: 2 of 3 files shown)


Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -1047,6 +1047,22 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
+      if (!model.renderable.isA('vtkImageArrayMapper')) {
+        const { offset, scale } = model.openGLTexture.computeScaleOffsets(
+          numComp,
+          scalars
+        );
+
+        model.openGLTexture.setUseHalfFloat(
+          dataType,
+          offset,
+          scale,
```
finetjul, line 1061: shouldn't you simply have to pass `model.renderable.getPreferSizeOverAccuracy()` to `create2DFromRaw` below ?

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1186,7 +1186,7 @@ function vtkOpenGLTexture(publicAPI, model) {
     return true;
   };
 
-  function computeScaleOffsets(numComps, numPixelsIn, data) {
+  publicAPI.computeScaleOffsets = (numComps, data) => {
```
finetjul, line 1189: meanwhile the signature of this function has changed. You might want to rebase

Sources/Rendering/OpenGL/ImageMapper/index.js
finetjul: Or maybe add a `createFilterable2DFromDataArray` or something like that...

sedghi:
@finetjul It is ready for your re-review. There is only one problem that I see that using the preferSizeOverAccuracy to true now has images upside down for some reason. I thought you might now what is going on

image

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1222,19 +1222,103 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+
+    if (!model.useHalfFloat) {
+      publicAPI.getOpenGLDataType(dataType, true);
+    }
+
+    return {
+      numComps,
+      dataType,
+      data,
+      scaleOffsets: structuredClone(scaleOffsets),
```
finetjul, line 1277: I do not think the `structuredClone` is called at the right time. It should be done AFTER setting scaleOffsets to the volumeInfo (in `create3DFilterableFromDataArray`)

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1222,19 +1222,103 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+      (hasExactHalfFloat(offset, scale) || preferSizeOverAccuracy)
+    ) {
+      model.useHalfFloat = true;
+    } else {
+      model.useHalfFloat = false;
     }
+  }
 
-    // Don't consider halfFloat and convert back to Float when the range of data does not generate an accurate halfFloat
-    // AND it is not preferable to have a smaller texture than an exact texture.
```
finetjul, line 1230: why getting rid of this comment ?

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1222,19 +1222,103 @@ function vtkOpenGLTexture(publicAPI, model) {
...
 
-    if (!useHalfFloat) {
-      return false;
+    if (
+      useHalfFloat &&
+      (hasExactHalfFloat(offset, scale) || preferSizeOverAccuracy)
+    ) {
+      model.useHalfFloat = true;
+    } else {
+      model.useHalfFloat = false;
```
finetjul, line 1231: model.useHalfFloat = useHalfFloat && ...

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1401,21 +1461,6 @@ function vtkOpenGLTexture(publicAPI, model) {
...
-    // https://bugs.chromium.org/p/chromium/issues/detail?id=1408247
-    model.useHalfFloat = checkUseHalfFloat(
-      dataType,
-      scaleOffsets.offset,
-      scaleOffsets.scale,
-      preferSizeOverAccuracy
-    );
-
-    // since our default is to use half float, in case that we can't use it
-    // we need to use another type
```
finetjul, line 1414: why getting rid of this comment ?

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1401,21 +1461,6 @@ function vtkOpenGLTexture(publicAPI, model) {
       depth,
     };
 
-    // preferSizeOverAccuracy will override norm16 due to bug with norm16 implementation
-    // https://bugs.chromium.org/p/chromium/issues/detail?id=1408247
```
finetjul, line 1405: why getting rid of this comment ?

Sources/Rendering/OpenGL/Texture/index.js
sedghi: Don't know why or how, will put them back

Sources/Rendering/OpenGL/Texture/index.js
sedghi: ready for your review again

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1630,8 +1672,8 @@ function vtkOpenGLTexture(publicAPI, model) {
                 newArray,
                 outIdx,
                 data[inOffset + xstride * tileX * numComps + nc],
-                scaleOffsetsCopy.offset[nc],
-                scaleOffsetsCopy.scale[nc]
+                scaleOffsets.offset[nc],
+                scaleOffsets.scale[nc]
```
finetjul, line 1676: why not `scaleOffsetsCopy` ?

sedghi: sorry missed that one

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1222,19 +1222,98 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+    const minArray = new Array(numComps);
+    const maxArray = new Array(numComps);
+    for (let c = 0; c < numComps; ++c) {
+      const [min, max] = dataArray.getRange(c);
+      minArray[c] = min;
+      maxArray[c] = max;
     }
 
-    return true;
+    // Create a copy of scale and offset to avoid aliasing issues
```
finetjul, line 1248: technically that comment should not be here but before `structuredClone`

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1222,19 +1222,98 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+    // Original is read only, copy is read/write
+    // Use the copy as volumeInfo.scale and volumeInfo.offset
+    const scaleOffsets = computeScaleOffsets(minArray, maxArray, numComps);
+
+    const offset = [];
+    const scale = [];
+    for (let c = 0; c < numComps; ++c) {
+      offset[c] = 0.0;
+      scale[c] = 1.0;
+    }
```
finetjul, line 1258: Optional: if offset/scale is not used by `create2DFilterableFromDataArray`, then it does not seem to be meaningful to create those offset/scale parameters in `processDataArray` but leave it in  `create3DFilterableFromDataArray`

Sources/Rendering/OpenGL/Texture/index.js
sedghi: thanks

Sources/Rendering/OpenGL/Texture/index.js
sedghi: it is not heavy computation, it is a initialization as I see, I can put it under a flag something like `initializeScaleOffset` ? But then we are computing the scaleOffset line above so I think that is confusing? But I can do that if you want

finetjul: I only meant to move away those lines from `processDataArray()` along with their returned values.
And simply have those variables being declared+initialized in `create3DFilterableFromDataArray()`.
If you don't see what I mean, it's ok to leave it as is. No worries.

sedghi: sure, makes sense

floryst:
I'll merge #2786 and #2787. They both look ready.

Issue #2761, opened 2023-03-22
Unnecessary re-computation of 2D texture in ImageMapper during window level


sedghi:
### High-level description
Seems like the `model.openGLTexture.create2DFromRaw` in the imageMapper is getting called on any window level manipulation. This is not necessary since the scalar data has not changed.

### Steps to reproduce

`npm run example -- ImageMapper`

Put some console logs inside the `create2DFromRaw` and do Window leveling and you see it is running over and over

### Detailed behavior
Seems like the `toString` function 

```js
  const toString = `${slice}A${image.getMTime()}A${imgScalars.getMTime()}B${publicAPI.getMTime()}C${model.renderable.getSlicingMode()}D${actor.getProperty().getMTime()}`;
```

As you see `actor.getProperty().getMTime()` (which I'm not sure why included and why window levelling change it either, interested to know the answer here) and that causes the check on `VBOBuildString` to fail and run the openGL textrue create2DFromRaw again. 

see video captured below

https://user-images.githubusercontent.com/7490180/226780887-e079d494-1c63-4b99-a5be-61ff5272245b.mp4

### Expected behavior
It should not reconfigure the openGLTexture since the scalarData has not changed. This does not happen in the VolumeMapper since there we are just checking the  

```
 toString = `${image.getMTime()}`;
 if (model.scalarTextureString !== toString) {
```

### Environment
- **vtk.js**: vtk latest 
- **OS**: MacOS 
- **Browser**: Chrome latest

PR #2787, opened 2023-04-13
perf(texture): should not rely on actor Mtime to re create texture

sedghi:
### Context

resolves https://github.com/Kitware/vtk-js/issues/2761

Basically the 2D texture was getting created over and over again since the check relies on the actor Mtime which should not. It caused performance problems for windowleveling on large images, reducing the fps even by 10 points since it was re uploading the scalar texture over and over again.

### Results

Remove the actor MTime check for the texture to be updated

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: Latest
  - **OS**: macOS
  - **Browser**: Chrome Latest


finetjul:
Looking at the history why this `actor.getProperty().getMTime()` was added in the first place, it seems to be useful if the "minificationFilter/magnificationFilter" properties have changed:

https://github.com/Kitware/vtk-js/commit/64f1cb501a2f0c66e1f0eb43ca43192373aa9e5e

@thewtex

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -920,9 +920,7 @@ function vtkOpenGLImageMapper(publicAPI, model) {
     }
 
     // rebuild the VBO if the data has changed
-    const toString = `${slice}A${image.getMTime()}A${imgScalars.getMTime()}B${publicAPI.getMTime()}C${model.renderable.getSlicingMode()}D${actor
-      .getProperty()
-      .getMTime()}`;
+    const toString = `${slice}A${image.getMTime()}A${imgScalars.getMTime()}B${publicAPI.getMTime()}C${model.renderable.getSlicingMode()}`;
```
sankhesh, line 923: This is testing if the property has been updated and for good measure. The texture filtering changes based on property interpolation. The property should not be changing each frame. If you see that the mapper is re-creating the texture each frame because of the property, check to see if something else in your application (or vtk-js) is changing the property.

sedghi: it is not my application, it can be replicated in the ImageMapper demo as I showed in the issue. basically window levelling should not re create scalar texture as it doesn't for VolumeMapper either. So maybe windowLevelling should not update the actor properties?

sedghi:
@finetjul Does it makes sense for minificationFilter/magnificationFilter to be on the actor? Shouldn't it be on the mapper?

Sources/Rendering/OpenGL/ImageMapper/index.js
sankhesh: Yes, or better yet - this `if` check should just test whether the property's interpolation has changed.

sedghi: @sankhesh How do you suggest I filter the changed properties? is there any way you recommend?

finetjul:
> @finetjul Does it makes sense for minificationFilter/magnificationFilter to be on the actor? Shouldn't it be on the mapper?

This is how it is done in VTK C++, so it does make sense. 
The idea is to have display properties to be stored in the actor/volume property. The mapper consumes them.

To fix the performance issue there should be more checks than the mtime.

sedghi:
ready for your re-review @finetjul @sankhesh

finetjul:
LGTM

sankhesh (approved):

PR #2789, opened 2023-04-18
feat(WidgetManager): add getActiveWidget()

floryst:
### Context
Adds `widgetManager.getActiveWidget()`, which allows users to query the currently active widget, if any. This is most useful if you want to perform operations that depend on whether a widget is currently selected.

### Results
`widgetManager.getActiveWidget()` returns the currently active widget, or null.

### Changes

### PR and Code Checklist

### Testing

No testing is needed. This further develops and exposes an already-used property.

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -153,9 +153,11 @@ function vtkWidgetManager(publicAPI, model) {
     // Default cursor behavior
     model._apiSpecificRenderWindow.setCursor(widget ? 'pointer' : 'default');
 
+    model.activeWidget = null;
     let wantRender = false;
     if (model.widgetInFocus === widget && widget.hasFocus()) {
       activateHandle(widget);
+      model.activeWidget = widget;
       wantRender = true;
     } else {
       for (let i = 0; i < model.widgets.length; i++) {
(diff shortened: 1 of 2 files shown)


Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -485,6 +487,7 @@ const DEFAULT_VALUES = {
   // _currentUpdateSelectionCallID: null,
   viewId: null,
   widgets: [],
+  activeWidget: null,
```
finetjul, line 490: we typically do not declare in DEFAULT_VALUES parameters that are null

finetjul (commented):
LGTM

Sources/Widgets/Core/WidgetManager/index.js
floryst: I've typed `getActiveWidget()` as `Nullable<vtkAbstractWidget>`, so I'd like to initialize `activeWidget` on construction to `null` rather than the implicit `undefined`. Would it be better to instead do `model.activeWidget = null` in the constructor function above? (Note that the current null assignment line is inside `updateSelection()`.)

finetjul: What you say makes sense to me. 

Shall we add `undefined` to the `Nullable` type. (probably not a good idea).

@jourdain a suggestion ?

floryst: I'd avoid adding undefined to `Nullable` b/c `null !== undefined`, even though `null == undefined`.

jourdain: I think that is fine especially with TS definition.

PR #2790, opened 2023-04-18
perf(OpenGL/Texture): reduce lookups in hot loop

floryst:
### Context

For large array inputs, there is a measurable difference in total execution time of the containing loop (observed initially on Chrome).

### Results

- Faster half float loop
### Changes

- [x] Updated OpenGL/Texture.js

### PR and Code Checklist

### Testing

Timing results on Chrome provided below. The relevant timing lines are described.
* "toHalf module local fn fewer lookups": this PR
* "toHalf module": original behavior
image

Timing on Firefox:

image


Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -656,8 +657,9 @@ function vtkOpenGLTexture(publicAPI, model) {
       for (let idx = 0; idx < data.length; idx++) {
         if (data[idx]) {
           const newArray = new Uint16Array(pixCount);
```
finetjul, line 659: What if you create a very large array outside the for loop and use subarray at each iteration, could you save even more time ?

PR #2794, opened 2023-04-18
fix(AbstractWidget): clear activeState

floryst:
### Context

activeState should not be set after calling deactivateAllHandles.

### Results

activeState is cleared after calling `deactivateAllHandles`.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Widgets/Core/AbstractWidget/index.js
+++ b/Sources/Widgets/Core/AbstractWidget/index.js
@@ -32,6 +32,7 @@ function vtkAbstractWidget(publicAPI, model) {
 
   publicAPI.deactivateAllHandles = () => {
     model.widgetState.deactivate();
+    model.activeState = null;
   };
 
   // --------------------------------------------------------------------------


jourdain (approved):
Don't recall the logic but LGTM

floryst:
The only weird case I can think of is what should happen when you do the following:

```
widgetManager.grabFocus(widget);
widget.deactivateAllHandles();
```

Though if you are doing manual activate/deactivate of handles, maybe you don't intend on using grabFocus/loseFocus.

Sources/Widgets/Core/AbstractWidget/index.js
```diff
@@ -32,6 +32,7 @@ function vtkAbstractWidget(publicAPI, model) {
 
   publicAPI.deactivateAllHandles = () => {
     model.widgetState.deactivate();
+    model.activeState = null;
```
finetjul, line 35: `deactivateAllHandles()` being a convenient function and not the only way to deactivate handles, I wonder if a more robust fix wouldn't be : 
 - for the widget to observe the activeState, and if it becomes inactive,then set activeState to null.
 - or when deactivating a substate in the widget state, it could check if it is the same as `widget.getActiveState()` of the associated widget and if so, set it to null.

finetjul: Can the widget in focus be different than the `activeState` ?
if not, then deactivating the widget in focus could also make it loose focus.

floryst: I certainly think so. An example would be a widget that processes mouse events while in focus but there is no handle state. A concrete example of this would be the PolyLineWidget, but there's no move handle.

The concern is mostly for existing widgets, e.g. a focused PolyLineWidget means you're drawing, but no active state means nothing happens. This is more of a "developer beware" note that I think would be good to document.

PR #2796, opened 2023-04-20
fix(scalarbaractor): support ScalarBarActor resize

finetjul:
The texture size may change over time. WebGL2 functions texSubImage2D can't therefore be used. The 'resizable' member variable should be considered. It was considered for von vtkOpenGLTexture but ignored on vtkTexture.

### Context
ScalarBarActor example was having resize issues when in full screen

### Results
No more resize issue with the ScalarBarActor.

### Changes
Do not use WebGL2 functions `texSubImage2D` when texture is resizable.

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 11
  - **Browser**: latest Chrome

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -781,7 +781,7 @@ function vtkOpenGLTexture(publicAPI, model) {
   //----------------------------------------------------------------------------
   function useTexStorage(dataType) {
     if (model._openGLRenderWindow) {
-      if (model.resizable) {
+      if (model.resizable || model.renderable?.getResizable()) {
         // Cannot use texStorage if the texture is supposed to be resizable.
         return false;
       }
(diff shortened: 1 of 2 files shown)


Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -969,7 +969,7 @@ function vtkOpenGLTexture(publicAPI, model) {
         if (j <= model.maxLevel) {
           tempData = invertedData[6 * j + i];
         }
-        if (model._openGLRenderWindow.getWebgl2() && !model.resizable) {
+        if (useTexStorage(dataType)) {
```
jourdain, line 972: I don't get that but glad you fixed it...

finetjul: that particular line is just factorization to be consistent with the rest of the file.

The other change is more important.

PR #2799, opened 2023-04-21
fix: directly assign the value from getDirection to mat3

jinjilynn:
According to the implementation of the multiply3x3_vect3 function, mat_3x3 should be a one-dimensional array

```javascript
export function multiply3x3_vect3(mat_3x3, in_3, out_3) {
  const x = mat_3x3[0] * in_3[0] + mat_3x3[1] * in_3[1] + mat_3x3[2] * in_3[2];
  const y = mat_3x3[3] * in_3[0] + mat_3x3[4] * in_3[1] + mat_3x3[5] * in_3[2];
  const z = mat_3x3[6] * in_3[0] + mat_3x3[7] * in_3[1] + mat_3x3[8] * in_3[2];

  out_3[0] = x;
  out_3[1] = y;
  out_3[2] = z;
}
```

i am not saying  it is not my image "direction", i mean multiplying an array by a number may be a NaN or 0

### DIFF
--- a/Sources/Rendering/Core/ImageMapper/index.js
+++ b/Sources/Rendering/Core/ImageMapper/index.js
@@ -120,12 +120,7 @@ function vtkImageMapper(publicAPI, model) {
 
   publicAPI.getSlicingModeNormal = () => {
     const out = [0, 0, 0];
-    const a = publicAPI.getCurrentImage().getDirection();
-    const mat3 = [
-      [a[0], a[1], a[2]],
-      [a[3], a[4], a[5]],
-      [a[6], a[7], a[8]],
-    ];
+    const mat3 = publicAPI.getCurrentImage().getDirection();
 
     switch (model.slicingMode) {
       case SlicingMode.X:


finetjul (approved):
LGTM

Sources/Rendering/Core/ImageMapper/index.js
```diff
@@ -120,12 +120,7 @@ function vtkImageMapper(publicAPI, model) {
 
   publicAPI.getSlicingModeNormal = () => {
     const out = [0, 0, 0];
-    const a = publicAPI.getCurrentImage().getDirection();
-    const mat3 = [
-      [a[0], a[1], a[2]],
-      [a[3], a[4], a[5]],
-      [a[6], a[7], a[8]],
-    ];
+    const mat3 = publicAPI.getCurrentImage().getDirection();
```
floryst, line 123: Something doesn't seem right to me. `getDirection()` returns a flattened matrix in column-major order. But in the code below, `vtkMath.multiply3x3_vect3(mat3, [1,0,0], out)` treats `mat3` as a row-major matrix. @finetjul @jourdain did we intend to change the ordering of vtkMath matrix operations to be column major, or did we keep it as row major?

finetjul: Good catch !

My current understanding on the row/column-major order in vtk-js is the following:
 - vtkMath functions should behave like VTK C++ (hence row-major). This is what's expected when porting C++ code to JS
 - Mapper code should compute/store matrices in column-major and use glmatrix for computations

To me it means that ImageData must store `direction` in row-major order.

floryst: Ok, so then we have a mismatch in terms of what parts of vtk.js should be row-major and column-major. [vtkImageData.computeTransforms](https://github.com/Kitware/vtk-js/blob/master/Sources/Common/DataModel/ImageData/index.js#L196) treats direction as column-major, as indexToWorld (and worldToIndex) is given directly to gl-matrix (which is column-major).

We discussed this a while back (https://github.com/Kitware/vtk-js/issues/612), but didn't get anywhere in terms of resolution. As it stands, vtk.js mixes matrix ordering without a clear documented boundary, and this is going to be more confusing as we go along. Looping in @sankhesh and @jadh4v for any opinions.

I would like for vtk.js to tend to one or the other, otherwise we'll just have a codebase littered with comments like "image direction is column major, so transpose to row major for vtkMath" (in the best case).

sankhesh: That's right. Let's stick to row-major order everywhere like vtk-cpp. The case that @floryst pointed out, is just a shortcut to avoid transposing the matrix before passing it as a uniform to the shader. Ideally, we should avoid having such shortcuts.

sankhesh: Mappers directly set uniforms, etc on the shaders so that can be left alone for now. Anyways, most of that should be private to the mappers.

floryst: @sankhesh do you think this is good to go? Or do we need to make some other changes? I think we might run into an issue if `getDirection()` is interpreted as row major here but column major elsewhere.

sankhesh (approved):
LGTM

Sources/Rendering/Core/ImageMapper/index.js
sankhesh: Yes, just approved it.

PR #2810, opened 2023-04-26
Add curved planar reformation mapper

bruyeret:
### Context

Add a vtkImageCPRMapper for curved planar reformation from a centerline and a volume
Adapt reslice cursor to work with vtkImageCPRMapper

### Results

image
Aorta viewed using the vtkImageCPRMapper example
There is a cross section view and a stretch view

### Changes

### PR and Code Checklist

### Todo

- [ ] Add tests for ImageCPRMapper
- [ ] Add typescript header for OpenGL implementation of ImageCPRMapper


Sources/Rendering/Core/AbstractMapper3D/index.js
```diff
@@ -1,38 +1,27 @@
...
-    let l = 0.0;
-    publicAPI.getBounds();
-    for (let i = 0; i < 3; i++) {
-      diff = model.bounds[2 * i + 1] - model.bounds[2 * i];
-      l += diff * diff;
-    }
-
-    return Math.sqrt(l);
+    const bounds = publicAPI.getBounds();
+    return vtkBoundingBox.getDiagonalLength(bounds) || 0;
```
finetjul, line 24: what's wrong with returning null ?

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -0,0 +1,378 @@
...
+	getWidth(): number;
+
+	/**
+	 * @see getWidth
+	 * @param width
+	 */
+	setWidth(width: number): boolean;
+
+	/**
+	 * Use @see getUseUniformOrientation to use the this uniform orientation instead of the orientation specified by the centerline
```
finetjul, line 40: Typo: "to use the this"

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -0,0 +1,378 @@
...
+	/**
+	 * @see getOrientationArrayName
+	 * @param arrayName
+	 */
+	setOrientationArrayName(arrayName: string | null): boolean;
+
+	/**
+	 * For each point on the oriented centerline, the tangent direction is the direction in which the mapper will sample
+	 * Let O (a mat3) be the orientation at a point on a centerline, and N (a vec3) the tangent direction
+	 * Then the mapper will sample along O * N
```
finetjul, line 97: Please document the default value

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -0,0 +1,378 @@
...
+
+	/**
+	 * @see getTangentDirection
+	 * @param tangent
+	 */
+	setTangentDirection(tangent: vec3): boolean;
+
+	/**
+	 * For each point on the oriented centerline, the bitangent direction forms with the normal and the tangent direction a new basis
+	 * It is used by the vtkCPRManipulator which can be used with this mapper
```
finetjul, line 109: this comment sentence is not useful. to be removed

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -0,0 +1,378 @@
...
+	 * Let O (a mat3) be the orientation at a point on a centerline, and N (a vec3) the tangent direction
+	 * Then the mapper will sample along O * N
+	 */
+	getTangentDirection(): vec3;
+
+	/**
+	 * @see getTangentDirection
+	 * @param tangent
+	 */
+	setTangentDirection(tangent: vec3): boolean;
```
finetjul, line 105: should this be renamed `setIJKTangentDirection` ?

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -0,0 +1,378 @@
...
+	 */
+	preRenderCheck(): boolean;
+
+	/**
+	 * Set the polydata used as a centerline
+	 * You can also use `publicAPI.setInputData(centerlineData, 1);`
+	 * Use all the segments of all the polylines (the centerline can be in multiple pieces)
+	 * The polydata can contain a PointData DataArray to specify the direction in which the mapper should sample for each point of the centerline ( @see getDirectionArrayName @see getDirectionArrayOffset )
+	 * If no such point data is specified, a uniform direction can be used instead ( @see getUniformDirection @see getUseUniformOrientation )
+	 * The points of the centerline are in model coordinates of the volume used as input ( @see setImageDataData ) and not index coordinates (see MCTC matrix of the OpenGL ImageCPRMapper)
```
finetjul, line 198: typo: setImageDataData

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -0,0 +1,378 @@
...
+	setCenterlineConnection(centerlineConnection: vtkOutputPort): void;
+
+	/**
+	 * Set the volume which should be sampled by the mapper
+	 * You can also use `publicAPI.setInputData(imageData, 0);`
+	 * The model coordinates of this imageData are used by this mapper when specifying points, vectors or width (see MCTC matrix of the OpenGL ImageCPRMapper)
+	 * You can use `imageData.getWorldToIndex();` or `imageData.getIndexToWorld();` to go from this model coordinates to index coordinates or the other way around
+	 * @param imageData
+	 */
+	setImageDataData(imageData: vtkImageData): void;
```
finetjul, line 219: setImageData

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -0,0 +1,378 @@
...
+
+	/**
+	 * Set the connection for the volume
+	 * You can also use `publicAPI.setInputConnection(imageData, 0);`
+	 * @see setImageDataData
+	 * @param imageData
+	 */
+	setImageDataConnection(imageData: vtkOutputPort): void;
```
finetjul, line 227: setImageConnection

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -0,0 +1,378 @@
...
+
+	/**
+	 *
+	 */
+	setResolveCoincidentTopologyToOff(): boolean;
+
+	/**
+	 *
+	 */
+	setResolveCoincidentTopologyToPolygonOffset(): boolean;
```
finetjul, line 327: aren't those derived from superclass ?

Sources/Rendering/Core/ImageCPRMapper/index.js
```diff
@@ -0,0 +1,350 @@
...
+      // Get dependencies of centerline
+      const orientationDataArray = publicAPI.getOrientationDataArray();
+      const linesDataArray = inputPolydata.getLines();
+      const pointsDataArray = inputPolydata.getPoints();
+
+      if (!model.useUniformOrientation && !orientationDataArray) {
+        vtkErrorMacro(
+          'Failed to create oriented centerline from polydata: no orientation'
+        );
+        model._orientedCenterline = vtkPolyLine.newInstance();
```
finetjul, line 59: It seems better to create a polyline only once in constructor, and reset it here.

Sources/Rendering/Core/ImageCPRMapper/index.js
```diff
@@ -0,0 +1,350 @@
...
+        centerlineTime > linesDataArray.getMTime() &&
+        centerlineTime > pointsDataArray.getMTime() &&
+        (model.useUniformOrientation ||
+          centerlineTime > orientationDataArray.getMTime())
+      ) {
+        return model._orientedCenterline;
+      }
+
+      // Otherwise, recompute
+      model.orientedCenterlineTime.modified();
```
finetjul, line 76: call modified at the end, not before computing.

Sources/Rendering/Core/ImageCPRMapper/index.js
```diff
@@ -0,0 +1,350 @@
...
+      }
+      const nPoints = linesData[0];
+      if (nPoints <= 1) {
+        // Empty centerline
+        model._orientedCenterline = vtkPolyLine.newInstance();
+        return model._orientedCenterline;
+      }
+      // Copy the points to a new data array
+      const pointIndices = linesData.subarray(1, 1 + nPoints);
+      const centerlinePoints = new Float32Array(3 * nPoints);
```
finetjul, line 93: you should create it with the same type of pointsDataArray

Sources/Rendering/Core/ImageCPRMapper/index.js
```diff
@@ -0,0 +1,350 @@
...
+      for (let i = 0; i < nPoints; ++i) {
+        const pointIdx = pointIndices[i];
+        orientations[i] = getOrientation(pointIdx);
+      }
+
+      // Done recomputing
+      model._orientedCenterline = vtkPolyLine.newInstance({
+        points,
+        orientations,
+      });
```
finetjul, line 159: a `vtkPolyLine` has a constructor (actually  it is `vtkCell`) that takes a `points` and an optional `pointIdsList`. You did not have to compute points yourself.

Sources/Rendering/Core/ImageCPRMapper/index.js
```diff
@@ -0,0 +1,350 @@
...
+
+    // No polydata: return previous centerline
+    return model._orientedCenterline;
+  };
+
+  publicAPI.setOrientedCenterline = (centerline) => {
+    model._orientedCenterline = centerline;
+    // To trigger rebuild of distances
+    if (centerline) {
+      centerline.modified();
```
finetjul, line 171: does not seem right. 
If the centerline has not change, then it should not rebuild distances.

Sources/Rendering/Core/ImageCPRMapper/index.js
```diff
@@ -0,0 +1,350 @@
...
+    }
+    publicAPI.modified();
```
finetjul, line 173: should return true if centerline is different from existing centerline, false otherwise

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -66,13 +66,76 @@ function vtkPolyLine(publicAPI, model) {
...
+  publicAPI.getTotalDistanceArray = () => {
+    if (model.distancesTime.getMTime() < model.points.getMTime()) {
+      const numPoints = publicAPI.getNumberOfPoints();
+      model.distances = new Array(numPoints);
+      if (numPoints > 0) {
+        const previousPoint = new Array(3);
+        const currentPoint = new Array(3);
+        let totalDistance = 0;
+        model.points.getPoint(0, previousPoint);
+        for (let i = 0; i < numPoints; ++i) {
```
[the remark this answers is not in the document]
bruyeret: The start at 0 is intended
From the doc for this function:
```
  /**
   * Returns an array containing for each pointIdx between 0 (included) and
   * numberOfPoints (excluded) the distance from the first point of the
   * polyline to the point at pointIdx
   * In particular if tda = publicAPI.getTotalDistanceArray(), then tda[0] = 0
   * and tda[tda.length - 1] is the total length of the polyline
   */
  getTotalDistanceArray(): number[]
```

Sources/Common/Transform/Transform/index.js
```diff
@@ -22,9 +44,66 @@ function vtkTransform(publicAPI, model) {
     return out;
   };
 
+  publicAPI.transformPoints = (points, out) => {
+    const arrayIn = getSlices(points, 3);
+    const arrayOut = getSlices(out, 3);
```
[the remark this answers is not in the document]
bruyeret: `vtkMatrixBuilder.apply` updates the array in place
If I want to use it in transform points, I will have to copy the `points` values into `out`
I fixed the problem by using a similar algorithm as in `vtkMatrixBuilder.apply`

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
bruyeret: Why `IJK`?

Sources/Common/DataModel/PolyLine/index.js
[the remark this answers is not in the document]
finetjul: I only meant that:
```
model.distances[0] = 0;
for (let i = 1; i < numPoints; ++i) {
   ...
```

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
finetjul: IJK is the local coordinate system of the volume (i.e. before applying volume `Direction` or actor `UserMatrix` )

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
bruyeret: No, they are added for each mapper

Sources/Rendering/Core/ImageCPRMapper/index.js
bruyeret: Should I replace what I did with a call to `vtkPolyline.initialize`?
I will still need to compute orientations
I also have a question:
`vtkCell` has an attribute `pointsIds` but what is the point of this attribute as the points are reordered using this list of ids in `initialize`? It looks like `vtkCell.deepCopy` won't work as expected if `pointIds` is not null

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
bruyeret: Then it is not the tangent direction in the IJK coordinate system
The tangent direction is in the oriented centerline coordinate system
To get the tangent direction in world coodinates:
WorldTangent = UserMatrix * CenterlineMatrix * TangentDirection
For example, if you want the tangent direction to follow the y axis of the centerline, set tangent direction to [0, 1, 0]

Sources/Rendering/Core/ImageCPRMapper/index.js
finetjul: > Should I replace what I did with a call to vtkPolyline.initialize?
Yes

> vtkCell has an attribute pointsIds but what is the point of this attribute as the points are reordered using this list of ids in initialize?

I guess it is to know the original point Ids.
This is the behavior in VTK C++

> It looks like vtkCell.deepCopy won't work as expected if pointIds is not null

`vtkCell.deepCopy` is bugged, indeed.

Sources/Widgets/Manipulators/CPRManipulator/index.js
```diff
@@ -0,0 +1,126 @@
...
+      model.mapper.getCenterlinePositionAndOrientation(clampedDistance);
+
+    let worldAxes;
+    if (orientation) {
+      const modelDirections = mat3.fromQuat([], orientation);
+      const baseDirections = model.mapper.getDirectionMatrix();
+      worldAxes = mat3.mul([], modelDirections, baseDirections);
+    }
+
+    model.lastDistance = clampedDistance;
```
finetjul, line 77: do not serialize lastDistance -> _lastDistance

Sources/Widgets/Manipulators/CPRManipulator/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+export function extend(publicAPI, model, initialValues = {}) {
+  vtkAbstractManipulator.extend(publicAPI, model, defaultValues(initialValues));
+
+  macro.setGet(publicAPI, model, [
+    'angle',
+    'distance',
+    'distanceStep',
+    'lastDistance',
+    'stretchActor',
```
finetjul, line 113: from the actor you can get the mapper. so no need to expose a setter to the mapper.

Sources/Widgets/Manipulators/CPRManipulator/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+    let worldAxes;
+    if (orientation) {
+      const modelDirections = mat3.fromQuat([], orientation);
+      const baseDirections = model.mapper.getDirectionMatrix();
+      worldAxes = mat3.mul([], modelDirections, baseDirections);
+    }
+
+    model.lastDistance = clampedDistance;
+    return { worldPosition: position, worldAxes };
```
finetjul, line 78: rename `worldAxes` into `worldDirection` to be consistent with the other namings.

Sources/Widgets/Widgets3D/ResliceCursorWidget/cprBehavior.js
```diff
@@ -0,0 +1,116 @@
...
+      model.widgetState,
+      model._factory.getScaleInPixels(),
+      model._factory.getRotationHandlePosition()
+    );
+  };
+
+  publicAPI.translateCenterOnPlaneDirection = (nbSteps) => {
+    const stretchState = model.widgetState.getStatesWithLabel(
+      `lineIn${model.stretchPlaneName}`
+    )?.[0];
```
finetjul, line 109: I guess, scrolling should be performed on the manipulator of the widget and not on the manipulator of a "line" handle.
It should then be handled in the superclass behavior.

Sources/Widgets/Widgets3D/ResliceCursorWidget/cprBehavior.js
```diff
@@ -0,0 +1,116 @@
...
+  model.crossPlaneViewType = planeNameToViewType[model.crossPlaneName];
+  model.previousData = null;
+
+  publicAPI.getActiveInteraction = () => {
+    if (
+      model.widgetState
+        .getStatesWithLabel(`lineIn${model.stretchPlaneName}`)
+        .includes(model.activeState)
+    ) {
+      return InteractionMethodsName.TranslateCenterAndUpdatePlanes;
```
finetjul, line 28: I though you could return `TranslateLine` and make RCW/behavior.js support a manipulator that returns a `worldDirection`. 
Wouldn't it work ?

Sources/Widgets/Manipulators/TrackballManipulator/index.js
```diff
@@ -55,29 +55,18 @@ function vtkTrackballManipulator(publicAPI, model) {
...
-    );
-    prevX = callData.position.x;
-    prevY = callData.position.y;
-    return newDirection;
-  };
-
-  publicAPI.reset = (callData) => {
-    prevX = callData.position.x;
-    prevY = callData.position.y;
-  };
```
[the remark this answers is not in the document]
bruyeret: You are right, this is some code left from older ideas

Sources/Rendering/Core/ImageCPRMapper/index.js
```diff
@@ -0,0 +1,343 @@
...
+      for (let i = 0; i < nPoints; ++i) {
+        const pointIdx = pointIndices[i];
+        orientations[i] = getOrientation(pointIdx);
+      }
+
+      // Done recomputing
+      publicAPI.resetOrientedCenterline();
+      model._orientedCenterline.initialize(pointsDataArray, pointIndices);
+      model._orientedCenterline.setOrientations(orientations);
+      return model._orientedCenterline;
```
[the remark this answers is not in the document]
bruyeret: I chose to let this return as IMO it doesn't degrade performance and improve code clarity

Sources/Widgets/Manipulators/CPRManipulator/index.js
bruyeret: `lastDistance` can be set or get
The manipulator has a "state" to know where it currently is
It should then be serializable

Sources/Widgets/Manipulators/CPRManipulator/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+function defaultValues(initialValues) {
+  return {
+    angle: 0,
+    distance: 0,
+    distanceStep: 1,
```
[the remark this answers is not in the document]
bruyeret: I can't use max norm as the direction of projection depends on the point on the centerline
This would make scrolling speed dependent of point on the centerline

Sources/Widgets/Widgets3D/ResliceCursorWidget/cprBehavior.js
bruyeret: I need to handle the scroll in the cprManipulator because it holds the `currentDistance`

Sources/Widgets/Widgets3D/ResliceCursorWidget/cprBehavior.js
bruyeret: I could eventually use InteractionMethodsName.TranslateCenter but it actually doing:
`newCenter = currentCenter + currentWorldCoords - previousWorldCoords`
Which is not suitable for the CPR

Sources/Widgets/Manipulators/CPRManipulator/index.js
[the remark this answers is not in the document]
finetjul: I got it wrong, I meant `distanceStep= min(spacingInX, spacingInY, spacingInZ)`.
Maybe that could be a good default value or the value to consider if `distanceStep` is 0, null or undefined.

Sources/Widgets/Widgets3D/ResliceCursorWidget/cprBehavior.js
finetjul: The manipulator of the widget can be a CPRManipulator. (and the manipulator of the line handle is also a CPRManipulator... and maybe they are the same instances...)

Sources/Widgets/Widgets3D/ResliceCursorWidget/cprBehavior.js
finetjul: Can't `newCenter = currentCenter + currentWorldCoords - previousWorldCoords` be moved in the plane manipulator ?

Sources/Widgets/Manipulators/CPRManipulator/index.js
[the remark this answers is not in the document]
bruyeret: Good idea!
This is actually what I use in the example: `cprManipulator.setDistanceStep(Math.min(...imageSpacing));`

Sources/Widgets/Widgets3D/ResliceCursorWidget/cprBehavior.js
bruyeret: This would mean that the PlaneManipulator doesn't return the clicked position anymore, which is a breaking change and many widgets rely on this manipulator
Furthermore, a manipulator shouldn't be stateful (the cprManipulator being an exception now)

finetjul: Ok to leave it as is for now. It can be simplified in a different PR.

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -66,13 +66,77 @@ function vtkPolyLine(publicAPI, model) {
 
     return line.evaluateLocation(pcoords, x, weights);
   };
+
+  publicAPI.evaluateOrientation = (subId, pcoords, q, weights) => {
+    if (model.orientations) {
+      line.setOrientations([
+        model.orientations[subId],
+        model.orientations[subId + 1],
```
finetjul, line 74: what if subId is the last from the polyline ?

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -66,13 +66,77 @@ function vtkPolyLine(publicAPI, model) {
...
+
+  publicAPI.getDistancesToFirstPoint = () => {
+    if (model.distancesTime.getMTime() < model.points.getMTime()) {
+      const numPoints = publicAPI.getNumberOfPoints();
+      model.distances = new Array(numPoints);
+      if (numPoints > 0) {
+        model.distances[0] = 0;
+        const previousPoint = new Array(3);
+        const currentPoint = new Array(3);
+        let totalDistance = 0;
```
finetjul, line 91:
```
let totalDistance = 0;
model.distances[0] = totalDistance;
```

Sources/Common/DataModel/PolyLine/index.js
```diff
@@ -66,13 +66,77 @@ function vtkPolyLine(publicAPI, model) {
...
+      line.setOrientations(null);
+    }
+
+    return line.evaluateOrientation(pcoords, q, weights);
+  };
+
+  publicAPI.getDistancesToFirstPoint = () => {
+    if (model.distancesTime.getMTime() < model.points.getMTime()) {
+      const numPoints = publicAPI.getNumberOfPoints();
+      model.distances = new Array(numPoints);
```
finetjul, line 86: Reuse the array:
```
if (!model.distances) {
 model.distances = new Array(numPoints);
}
else {
  model.distances.length = numPoints;
}
```

Sources/Rendering/Core/AbstractMapper3D/index.js
```diff
@@ -1,38 +1,27 @@
...
   publicAPI.getCenter = () => {
-    publicAPI.getBounds();
-    for (let i = 0; i < 3; i++) {
-      model.center[i] = (model.bounds[2 * i + 1] + model.bounds[2 * i]) / 2.0;
-    }
+    const bounds = publicAPI.getBounds();
+    model.center = vtkBoundingBox.isValid(bounds)
+      ? vtkBoundingBox.getCenter(bounds)
+      : null;
     return model.center.slice();
```
finetjul, line 19: `model.center?.slice();`

Sources/Rendering/Core/AbstractMapper3D/index.js
```diff
@@ -41,10 +30,10 @@ function vtkAbstractMapper3D(publicAPI, model) {
 // ----------------------------------------------------------------------------
 
 const DEFAULT_VALUES = {
-  bounds: [1, -1, 1, -1, 1, -1],
+  bounds: [...vtkBoundingBox.INIT_BOUNDS],
   center: [0, 0, 0],
 
-  viewSpecificProperties: null,
+  viewSpecificProperties: {},
```
finetjul, line 36: this is not equivalent to initializing it in `extend()`.

Here ALL vtkAbstractMapper instanes will share the same `viewSpecificProperties`.

You can fix it by using the `defaultValues()` function notation.

Sources/Rendering/Core/ImageCPRMapper/example/controller.html
```diff
@@ -0,0 +1,7 @@
+<table>
+  <tr>
+    <td>
+      <select id='centerline' style="width: 100%"></select>
```
finetjul, line 4: you could have added the `angle` input instead of an empty `select`

Sources/Rendering/Core/ImageCPRMapper/example/index.js
```diff
@@ -0,0 +1,335 @@
...
+    0,
+    -worldBitangent[0],
+    -worldBitangent[1],
+    -worldBitangent[2],
+    0,
+    worldActorTranslation[0],
+    worldActorTranslation[1],
+    worldActorTranslation[2],
+    1
+  );
```
finetjul, line 163: you can also use spread operators to simplify notation...

Sources/Widgets/Manipulators/CPRManipulator/index.js
```diff
@@ -0,0 +1,140 @@
...
+// ----------------------------------------------------------------------------
+
+function vtkCPRManipulator(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkCPRManipulator');
+
+  publicAPI.handleEvent = (callData, glRenderWindow) => {
+    const mapper = model.cprActor?.getMapper();
+    if (!mapper) {
+      return { worldCoords: [0, 0, 0] };
```
finetjul, line 32: don't return a valid coord, return null instead.

Sources/Widgets/Manipulators/CPRManipulator/index.js
```diff
@@ -0,0 +1,140 @@
...
+    const height = mapper.getHeight();
+    const distance = height - modelPlanePicking[1];
+
+    return publicAPI.distanceEvent(distance);
+  };
+
+  publicAPI.distanceEvent = (distance) => {
+    const mapper = model.cprActor?.getMapper();
+    if (!mapper) {
+      return { worldCoords: [0, 0, 0] };
```
finetjul, line 68: return a null worldCoords

Sources/Widgets/Manipulators/CPRManipulator/index.js
```diff
@@ -0,0 +1,140 @@
...
+  };
+  // Find default distanceStep from image spacing
+  // This only works if the mapper in the actor already has an ImageData
+  if (!initialValues.distanceStep) {
+    const imageSpacing = initialValues.cprActor
+      ?.getMapper()
+      ?.getInputData(0)
+      ?.getSpacing?.();
+    if (imageSpacing) {
+      values.distanceStep = Math.min(...imageSpacing);
```
finetjul, line 113: this should not be evaluated at initialization time, but each time distanceStep is required. Because a `setDistanceStep()` could set `model.distanceStep` to 0 or null at anytime.

Sources/Common/DataModel/PolyLine/index.js
bruyeret: `subId` can't be the last point of the polyline

`evaluateOrientation` interpolates the orientation between two points
The subId has to be in `[0, nPoints - 1[` (in `[0, nSegments[`)
This is part of the contract of the function, I can make it clear in the ts header if you want

Also `findPointIdAtDistanceFromFirstPoint` will never return `nPoints - 1` because it is meant to be used with this function
This is why it used to be called `findSubId`

finetjul: fair enough. leave it as is.

Sources/Rendering/Core/ImageCPRMapper/example/controller.html
bruyeret: The select is filled with the dirrent centerlines in the JS code
It was not trivial but I added an `angle` input

finetjul (approved):
LGTM.

You can remove the Draft status.

Sources/Common/DataModel/Line/index.d.ts
```diff
@@ -29,6 +32,18 @@ export interface vtkLine extends vtkCell {
 	 */
 	getCellDimension(): number;
 
+	/**
+	 * Get the list of orientations (a list of quat) for each point of the line.
+	 * Can be null if the line is not oriented
+	 */
+	getOrientations(): quat[] | null;
```
floryst, line 39: Instances of `quat[] | null` ( or any `| null`) should be switched to `Nullable<quat[]>` (or `Nullable<Array<quat>>`) for consistency.

Sources/Widgets/Manipulators/LineManipulator/index.js
```diff
@@ -41,15 +41,16 @@ function vtkLineManipulator(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkLineManipulator');
 
-  publicAPI.handleEvent = (callData, glRenderWindow) =>
-    projectDisplayToLine(
+  publicAPI.handleEvent = (callData, glRenderWindow) => ({
+    worldCoords: projectDisplayToLine(
```
floryst, line 45: This is an API-breaking change, and so the commit must indicate as such with "BREAKING CHANGE:"  in the commit body. Additionally, this must be documented in the BREAKING_CHANGES.md doc.

PR #2816, opened 2023-05-05
fix(webgpu): Fix support of WebGPU render backend

finetjul:
### Context
Running the LineWidget example with WebGPU backend generated an error because WidgetManager requires onWindowResizeEvent public function.
Moreover in many places `getOpenGLRenderWindow()` was still called (even in a WebGPU context)

### Results
Expose onWindowResizeEvent and trigger it when the size of the window is modified.
Rename `getOpenGLRenderWindow()` into `getApiSpecificRenderWindow()`

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 11
  - **Browser**: Chrome Canary

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -1,3 +1,6 @@
+## From 28.x to 29
+
+- **getOpenGLRenderWindow**: `getOpenGLRenderWindow` has been renamed to `getApiSpecificRenderWindow` in `vtkFullScreenRenderWindow`, `vtkGenericRenderWindow` and `vtkViewProxy` to support WebGL and WebGPU backend.
 ## From 27.x to 28
 
 - **vtkManipulator.handleEvent**: Change all `handleEvent` signatures of manipulators. Used to be `handleEvent(callData, glRenderWindow): vec3`, it is now `handleEvent(callData, glRenderWindow): { worldCoords: Nullable<vec3>, worldDirection?: mat3 }`.
--- a/Sources/Rendering/OpenGL/RenderWindow/index.js
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.js
@@ -1308,15 +1308,6 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
     model.activeFramebuffer = newActiveFramebuffer;
   };
 
-  const superSetSize = publicAPI.setSize;
-  publicAPI.setSize = (width, height) => {
-    const modified = superSetSize(width, height);
-    if (modified) {
-      publicAPI.invokeWindowResizeEvent({ width, height });
-    }
-    return modified;
-  };
-
   publicAPI.getGraphicsResourceForObject = (vtkObj) => {
     if (!vtkObj) {
       return null;
@@ -1369,6 +1360,12 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
       glRen?.releaseGraphicsResources();
     });
   };
+
+  model._onSizeChanged = (_publicAPI, _model, newValue) =>
+    publicAPI.invokeWindowResizeEvent({
+      width: newValue[0],
+      height: newValue[1],
+    });
 }
 
 // ----------------------------------------------------------------------------
--- a/Sources/Rendering/OpenGL/RenderWindow/index.js
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.js
@@ -1308,6 +1308,15 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
     model.activeFramebuffer = newActiveFramebuffer;
   };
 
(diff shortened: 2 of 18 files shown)


finetjul:
@floryst there is still an error when mousing over the render window. There are many calls to `getSelectedDataForXY`, eventually `model._capturedBuffers` is null and `model._capturedBuffers.generateSelection()` fails.

You must have had that error that would explain why you added the check `callID !== model._currentUpdateSelectionCallID`.
Any chance you know what is missing here ?

floryst:
[`captureBuffers`](https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Core/WidgetManager/index.js#L203) is the issue here. The first call to captureBuffers sets `_captureInProgress` to true. Subsequent calls (e.g. mouse movement) will see that `_captureInProgress == true`, so they return early. Importantly, they return _when there are no captured buffers_.

A quick fix would be the following implementation. I don't know if there are any unintended effects, however.

```diff
@@ -202,22 +202,23 @@ function vtkWidgetManager(publicAPI, model) {
 
   async function captureBuffers(x1, y1, x2, y2) {
     if (model._captureInProgress) {
+      await model._captureInProgress;
       return;
     }
-    model._captureInProgress = true;
     renderPickingBuffer();
 
     model._capturedBuffers = null;
-    model._capturedBuffers = await model._selector.getSourceDataAsync(
+    model._captureInProgress = model._selector.getSourceDataAsync(
       model._renderer,
       x1,
       y1,
       x2,
       y2
     );
+    model._capturedBuffers = await model._captureInProgress;
     model.previousSelectedData = null;
+    model._captureInProgress = null;
     renderFrontBuffer();
-    model._captureInProgress = false;
   }
```

I did try it out with WebGPU + Chrome Canary, and I don't see anymore getSelection errors. However, the LineWidget cannot be interacted with once placed, and the spheres are not correctly colored. I haven't investigated that bit yet.

finetjul:
Thanks for your tweak, I added it to the PR. 

In WebGPU, there are many rendering warnings which might be related:
image

In WebGL, it is working the same way as the [existing example](https://kitware.github.io/vtk-js/examples/LineWidget.html).

floryst:
LGTM after the RenderWindow conflict is resolved. Seems like there's more work to do with the WebGPU implementation.

BREAKING_CHANGES.md
```diff
@@ -1,3 +1,6 @@
+## From 28.x to 29
+
+- **getOpenGLRenderWindow**: `getOpenGLRenderWindow` has been renamed to `getApiSpecificRenderWindow` in `vtkFullScreenRenderWindow`, `vtkGenericRenderWindow` and `vtkViewProxy` to support WebGL and WebGPU backend.
```
floryst, line 3: Will you mark the commit as a breaking change?

Also, we have some other breaking changes that should go into v29. I've created a v29 milestone to track those, so that we can either merge them all at once or update the beta branch.

finetjul: Oops, I forgot to flag it as a breaking change, I'll do that.

I've no rush for that commit to be merged, so it can be merged whenever

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -1278,14 +1278,11 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
-    if (modified) {
-      publicAPI.invokeWindowResizeEvent({ width, height });
-    }
-    return modified;
-  };
+  model._onSizeChanged = (_publicAPI, _model, newValue) =>
+    publicAPI.invokeWindowResizeEvent({
+      width: newValue[0],
+      height: newValue[1],
+    });
```
floryst, line 1285: @finetjul I was going to fix the conflicts and realized that `model._onSizeChanged` is not wired up anywhere. Is there a pathway for invoking the window resize event now that `publicAPI.setSize()` no longer directly invokes the event?

floryst: I'm partially reverting this such that setSize directly invokes the window resize event in both the OpenGL and WebGPU RenderWindows. The changes are all internal, so if you want to continue with the idea you had for _onSizeChanged, you can do that without a breaking change.

finetjul: `_onSizeChanged` was being called since this following commit 
https://github.com/Kitware/vtk-js/commit/63787d550a76fc9f27a6be074a76d08fcefad639 

TBH, I do not remember anything from this current PR though...

floryst: Ah, that's good to know. I forgot about that change. We should document it somewhere.

PR #2839, opened 2023-05-26
Reslice mapper pd

sankhesh:
### Context

The vtkImageResliceMapper allows reslicing the volume with custom polydata. However, when switching between a vtkPlane and vtkPolyData as the slicing function, the mapper was failing to update graphics buffers. 

Also fixed a regression in the vtkImplicitPlaneWidget due to recent changes in handleEvents API of widgets.
### Results

Screenshot from 2023-05-26 12-47-05

Screenshot from 2023-05-26 12-47-21

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/ImageResliceMapper/example/index.js
+++ b/Sources/Rendering/Core/ImageResliceMapper/example/index.js
@@ -4,6 +4,7 @@ import '@kitware/vtk.js/favicon';
 import '@kitware/vtk.js/Rendering/Profiles/All';
 
 import vtkColorTransferFunction from '@kitware/vtk.js/Rendering/Core/ColorTransferFunction';
+import vtkCylinderSource from '@kitware/vtk.js/Filters/Sources/CylinderSource';
 import vtkFullScreenRenderWindow from '@kitware/vtk.js/Rendering/Misc/FullScreenRenderWindow';
 import vtkImageProperty from '@kitware/vtk.js/Rendering/Core/ImageProperty';
 import vtkImageResliceMapper from '@kitware/vtk.js/Rendering/Core/ImageResliceMapper';
@@ -41,6 +42,14 @@ const slicePlane = vtkPlane.newInstance();
 slicePlane.setNormal(0, 1, 0);
 mapper.setSlicePlane(slicePlane);
 mapper.setSlabType(SlabTypes.MAX);
+
+const slicePolyDataSource = vtkCylinderSource.newInstance({
+  height: 100,
+  radius: 100,
+  resolution: 20,
+  capping: 1,
+  center: [100, 100, 100],
+});
 
 const actor = vtkImageSlice.newInstance();
 actor.setMapper(mapper);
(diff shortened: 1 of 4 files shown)


Sources/Rendering/Core/ImageResliceMapper/example/index.js
```diff
@@ -98,6 +107,11 @@ function setImage(im) {
     0.5 * (bds[2] + bds[3]),
     0.5 * (bds[5] + bds[4])
   );
+  slicePolyDataSource.setCenter(
+    0.5 * (bds[0] + bds[1]),
+    0.5 * (bds[2] + bds[3]),
+    0.5 * (bds[5] + bds[4])
```
finetjul, line 113: `0.5 * (bds[4] + bds[5])` or `vtkBoundingBox.getCenter(bds)` or even better `im.getCenter()`

Sources/Rendering/Core/ImageResliceMapper/example/index.js
```diff
@@ -108,6 +122,12 @@ function setImage(im) {
   planeState.onModified(() => {
     slicePlane.setOrigin(planeState.getOrigin());
     slicePlane.setNormal(planeState.getNormal());
+    slicePolyDataSource.setCenter(planeState.getOrigin());
+    slicePolyDataSource.setDirection(planeState.getNormal());
+    if (mapper.getSlicePolyData()) {
+      slicePolyDataSource.update();
+      mapper.setSlicePolyData(slicePolyDataSource.getOutputData());
```
finetjul, line 129: shouldn't this work with connections: `setInputConnection(1, slicePolyDataSource.getOutputPort())`

finetjul (approved):

finetjul (approved):

Sources/Rendering/Core/ImageResliceMapper/example/index.js
```diff
@@ -189,6 +209,19 @@ for (let idx = 0; idx < slabTypes.length; ++idx) {
     renderWindow.render();
   };
 }
+const sliceFunction = document.querySelectorAll('.spd');
+for (let idx = 0; idx < sliceFunction.length; ++idx) {
+  const st = sliceFunction[idx];
+  st.onchange = (e) => {
+    if (e.target.value === 'plane') {
+      mapper.setSlicePolyData(null);
```
jadh4v, line 217: Looks like the mapper switches back to Plane slicing if the polydata is set to `null`. This kinda feels too subtle. Class design wise, would it be better to have an explicit flag or integer to set the modes to implicit function vs polydata slicing? or handle the type internally and only have a single setter like `setSlicingGeometry`?

jadh4v (approved):
One comment, but otherwise LGTM

Sources/Rendering/Core/ImageResliceMapper/example/index.js
sankhesh: This PR doesn't change that behavior and that's documented here: https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ImageResliceMapper/index.d.ts#L206-L214

When designing the mapper, I considered adding a flag to switch between the two but figured preferring the polydata input should be fine for most cases.

Sources/Rendering/Core/ImageResliceMapper/example/index.js
sankhesh: It doesn't work as is but yeah, that could prove useful.

PR #2842, opened 2023-06-06
fix(ScalarsToColors): direct scalars checked against 1 instead of 255

TimDeBauwReLu:
### Context
With the upgrade to VTK 27.6.0, I encountered the problem that some polydata meshes where opaque, while other were not. I found that this only occured on the polydata meshes with DIRECT_SCALARS as color mode. 

### Results
Polydata are always shown with its correct opacity

### Changes
For me, changing `range[0] === 255` to `range[0] === 1.0` solved the issue


finetjul:
Did you try to create an `Uint8array` and call `setTuple(idx, [255,0,0,255])`?

PR #2848, opened 2023-06-15
Add CPR mapper support for stretched and projected CPR

bruyeret:
Add support for these mode of CPR by using:
- `mapper.setCenterPoint(...)`
- `mapper.setUseUniformOrientation(...)`
- `mapper.getOrientedCenterline().setDistanceFunction(...)`

Add an example of stretched mode and a spine centerline.

image


Sources/Common/DataModel/PolyLine/index.d.ts
```diff
@@ -67,6 +67,29 @@ export interface vtkPolyLine extends vtkCell {
...
+  getOrientations(): Nullable<quat[]>;
+
+  /**
+   * @see getOrientations
+   * @param orientations
+   */
+  setOrientations(orientations: Nullable<quat[]>): boolean;
+
+  /**
+   * The function used in getDistancesToFirstPoint and in findPointIdAtDistanceFromFirstPoint
```
finetjul, line 84: give information about the default

Sources/Common/DataModel/PolyLine/index.d.ts
```diff
@@ -67,6 +67,29 @@ export interface vtkPolyLine extends vtkCell {
    * @param distance The distance from the first point of the polyline
    */
   findPointIdAtDistanceFromFirstPoint(distance: number): number;
+
+  /**
+   * An array of quaternions used to orient the polyline at each of its point
+   * The length of the array has to be the same size as the number of points
```
finetjul, line 73: what's the default ?

Sources/Rendering/Core/ImageCPRMapper/example/index.js
```diff
@@ -350,6 +354,54 @@ angleEl.addEventListener('input', () =>
...
+  mapper.setCenterPoint(null);
+  mapper.setUseUniformOrientation(false);
+  mapper.getOrientedCenterline().setDistanceFunction(vec3.dist);
+  updateDistanceAndDirection();
+}
+
+function useStretchedMode() {
+  mapper.setCenterPoint(centerline.getPoints().getPoint(0));
+  mapper.setUseUniformOrientation(true);
+  mapper.getOrientedCenterline().setDistanceFunction((a, b) => {
```
finetjul, line 367: should probably be set automatically by the mapper in setUseUniformOrientation( and not manually by the application

bruyeret: No, the distance function depends on which result is expected: a stretched or a projected CPR won't have the same distance function but will both use a uniform orientation
Also, a straightened CPR could use a uniform orientation

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -62,6 +62,21 @@ export interface vtkImageCPRMapper extends vtkAbstractMapper3D {
...
 	setUseUniformOrientation(useUniformOrientation: boolean): boolean;
 
+	/**
+	 * A point used to offset each line of pixel in the rendering
+	 * The line of pixel is offseted such as the center of the line is as close as possible to the center point
+	 * This can be used in combination with @see getUseUniformOrientation and a custom distance function for @see getOrientedCenterline to visualize a CPR in projected mode or stretched mode
+	 * Defaults to null.
+	 * @returns the center point
+	 */
+	getCenterPoint(): Nullable<vec3>;
```
daker, line 72: `Nullable<Vector3>` ?

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -62,6 +62,21 @@ export interface vtkImageCPRMapper extends vtkAbstractMapper3D {
...
+	
+	/**
+	 * @see getCenterPoint
+	 * @param point
+	 */
+	setCenterPoint(point: Nullable<vec3>): boolean;
```
daker, line 78: `Nullable<Vector3>` ?

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -191,6 +206,17 @@ export interface vtkImageCPRMapper extends vtkAbstractMapper3D {
...
+	/**
+	 * Configure the mapper and the centerline to be in straightened CPR mode
+	 */
+	useStraightenedMode(): void;
+
+	/**
+	 * Configure the mapper and the centerline to be in strectched CPR mode
+	 * @param centerPoint The center point, optional, default to the first point of the centerline or [0, 0, 0]
+	 */
+	useStretchedMode(centerPoint?: vec3): void;
```
daker, line 218: `Nullable<Vector3>` ?

daker (changes requested):

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
bruyeret: The difference between the two is:
`type Vector3 = [number, number, number];`
and
`type vec3 = [number, number, number] | Float32Array;`

The code I made accepts both Float32Array and Array and `vec3` are already used in a lot of typescript headers
Why do you think I should use `Vector3` instead of `vec3` ?

bruyeret: Is keeping `Nullable<vec3>` fine @daker?

daker: Yes, sorry i didn't see your reply

daker (approved):
TS defs LGTM

bruyeret:
Thanks @daker!
Is it good for you too @finetjul?

PR #2856, opened 2023-06-22
Lookuptable fixes

sankhesh:
### Before

download (1)

### After

download

### Changes

### PR and Code Checklist


jourdain:
LGTM

Sources/Common/Core/LookupTable/index.js
```diff
@@ -272,6 +272,11 @@ function vtkLookupTable(publicAPI, model) {
     // Add "special" colors (NaN, below range, above range) to table here.
     const { numberOfColors } = model;
 
+    // Resize the array to hold three additional colors
+    model.table.push(0, 0, 0, 0);
+    model.table.push(0, 0, 0, 0);
+    model.table.push(0, 0, 0, 0);
+
```
jourdain, line 279: Might not be needed since doing that `array[array.length] = 5` is valid and will technically extend the size of the array. ;-)

jourdain (approved):

Sources/Common/Core/LookupTable/index.js
sankhesh: Ah ok. I'll remove it then.

jourdain:
LUT test is failing on firefox. You may need to add an extra baseline.

download

sankhesh:
@jourdain Thanks for pointing it out. It is the text labels in the baseline that are getting messed up. I'll just remove it.

PR #2869, opened 2023-07-06
Remove call to modified() in setSelectionWebGLIdsToVTKIds()

bruyeret:
When creating VBOs, the call to modified() induced by the setter caused a performance hit
The mtime of the mapper doesn't need to be modified when calling this setter
`selectionWebGLIdsToVTKIds` is only used when processing a selection made with the hardware selector and the mtime of the mapper doesn't change this behavior

### DIFF
--- a/Sources/Rendering/Core/Mapper/index.js
+++ b/Sources/Rendering/Core/Mapper/index.js
@@ -53,6 +53,13 @@ function vtkMapper(publicAPI, model) {
     // make sure we do NOT call modified()
   };
 
+  publicAPI.setSelectionWebGLIdsToVTKIds = (selectionWebGLIdsToVTKIds) => {
+    model.selectionWebGLIdsToVTKIds = selectionWebGLIdsToVTKIds;
+    // make sure we do NOT call modified()
+    // this attribute is only used when processing a selection made with the hardware selector
+    // the mtime of the mapper doesn't need to be changed
+  };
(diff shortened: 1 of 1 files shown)


finetjul (approved):
LGTM

floryst:
I would provide more comment context for why we don't need to update the mtime.

bruyeret:
I updated the comment of the PR to add more context @floryst

sankhesh (approved):
LGTM

floryst:
LGTM!

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -53,6 +53,11 @@ function vtkMapper(publicAPI, model) {
     // make sure we do NOT call modified()
   };
 
+  publicAPI.setSelectionWebGLIdsToVTKIds = (selectionWebGLIdsToVTKIds) => {
+    model.selectionWebGLIdsToVTKIds = selectionWebGLIdsToVTKIds;
+    // make sure we do NOT call modified()
```
floryst, line 58: Oh wait, I meant that you should add more context here, since someone reading the code might not know the reason for not calling modified().

bruyeret: Oh sorry for misunderstanding your review!
I added the comment to the code

PR #2872, opened 2023-07-13
fix(DataArray): scalar data implies index 0

floryst:
### Context
Calling `getRange()` and then `getRange(0)` on scalar data computes the range twice, when it should be computed once.

### Results

- Avoids recomputing the range of scalar data

### Changes

- [x] DataArray.getRange now correctly caches the range

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


jourdain (approved):

finetjul (approved):
LGTM

PR #2873, opened 2023-07-13
perf(ImageMapper): do not use subarray

floryst:
### Context

TypedArray.subarray is slower than just looping through values and copying them.

### Results

- ImageMapper.buildBufferObjects is faster for large arrays.

### Changes

- [x] ImageMapper updated

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


finetjul (approved):
LGTM

Issue #2537, opened 2022-07-27
OpenGL RenderWindow in sandboxed iframe console error


PaulHax:
### High-level description
When using OpenGL RenderWindow in a iframe (without xr-spatial-tracking feature policy?) get this error on first render:
```
Uncaught (in promise) DOMException: Failed to execute 'isSessionSupported' on 'XRSystem': Access to the feature "xr" is disallowed by permissions policy.
    at vtkOpenGLRenderWindow (webpack-internal:///./node_modules/vtk.js/Sources/Rendering/OpenGL/RenderWindow/index.js:221:75)
```

### Steps to reproduce
Use vtk.js in "sandboxed" `iframe`
```
<iframe
  id="inlineFrameExample"
  title="Inline Frame Example"
  width="300"
  height="200"
  src="./viewer.html"
  sandbox="allow-scripts"
>
</iframe>
```
https://codesandbox.io/s/volume-rendering-bounds-forked-846gd7?file=/index.html:119-304

### Expected behavior
No XR related errors unless I opt into trying to activate XR.

### Environment
- **vtk.js**: 25.4
- **OS**: Linux
- **Browser**: Chrome 103

### Reported here:
https://github.com/Kitware/itk-vtk-viewer/issues/534

### Approach?
Looks like `xr.isSessionSupported` returns a Promise, so maybe something more like this:
```
if (navigator.xr) {
    navigator.xr
      .isSessionSupported('immersive-vr')
      .then(isSupported => {
        isImmersiveVrSupported = isSupported
      })
      .catch(error => {
        if (error.name !== 'SecurityError') {
          throw error
        }
      })
  }
```
Or remove this feature and lift this to user responsibility.  I'm not seeing in repo use of `HaveVRDisplay` event.

YaoDongcai:
In fact, you can handle it according to the following:

Object.defineProperty(navigator, "xr", {
  get: function () {
    return undefined;
  },
});

PR #2879, opened 2023-07-27
fix(vtkOpenGLRenderWindow): Remove unused `HaveVRDisplay` event and outdated TypeScript definitions

tbirdso:
Removes `vtkOpenGLRenderWindow.HaveVRDisplay` event. This event is unused internally and was previously left over from WebVR to WebXR integration to avoid API breakage.

Under previous behavior the `navigator.xr` check was called eagerly at construction, which could result in an unhandled exception. This change removes that eager check so that `navigator.xr` is referenced only if an XR-related method such as `startXR` is called.

Applications should query `navigator.xr` directly to determine whether a given session type may be supported.

Resolves https://github.com/Kitware/vtk-js/issues/2537

Also removes outdated WebVR-related TypeScript definitions from `vtkOpenGLRenderWindow`. Affected method implementations were previously removed in transition from WebVR to WebXR.

### Context

`navigator.xr` is only available where WebXR is supported, and sometimes appears available but is disabled for security. Queries should be limited to instances where XR is explicitly being used, or at least exceptions should be handled otherwise.

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  `master`
  - **OS**:  Windows 10
  - **Browser**: Chrome 115.0.0.0

### DIFF
--- a/Sources/Rendering/OpenGL/RenderWindow/index.d.ts
+++ b/Sources/Rendering/OpenGL/RenderWindow/index.d.ts
@@ -378,24 +378,6 @@ export interface vtkOpenGLRenderWindow extends vtkOpenGLRenderWindowBase {
 	getSize(): Vector2;
 
 	/**
-	 *
-	 * @param {Vector2} size
-	 */
-	setVrResolution(size: Vector2): void;
-
-	/**
-	 *
-	 * @param {Number} x
-	 * @param {Number} y
-	 */
-	setVrResolution(x: number, y: number): void;
-
-	/**
-	 *
-	 */
-	getVrResolution(): Vector2;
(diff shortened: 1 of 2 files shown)


tbirdso:
@PaulHax @jourdain Please test and verify whether these changes resolve your reported IFrame issues.

floryst:
LGTM code-wise.

PaulHax (approved):
Tested, fix works for the sandbox in the issue!

tbirdso:
I see a thumbs up from @jourdain so I believe this closes #2537. Please re-open that issue if problems persist.

PR #2884, opened 2023-08-01
fix(ColorTransferFunction): misuse of return value

### LINKED ISSUE #2881: [Bug] Error in ColorTransferFunction when indexed Lookup Mode is used
### Bug description

In getColor() function   there is a call to getNodeValue function where returned value is passed by function argument (array). 
https://github.com/Kitware/vtk-js/blob/188c7fb379a87b97f39262659d4a66aedc2e88d3/Sources/Rendering/Core/ColorTransferFunction/index.js#L395-L398

But getNodeValue is setting 6-element array instead of object with r,g,b properties as above.
https://github.com/Kitware/vtk-js/blob/188c7fb379a87b97f39262659d4a66aedc2e88d3/Sources/Rendering/Core/ColorTransferFunction/index.js#L962-L976

So the code in getColor should be:
```
 rgb[0] = nodeVal[1];
 rgb[1] = nodeVal[2];
 rgb[2] = nodeVal[3];
```

### Steps to reproduce

Any use of `ColorTransferFunction` with `setIndexedLookup(true)`

### Detailed Behavior

_No response_

### Expected Behavior

The getColor function should return color indiced by annotation, instead undefined values are returned.

### Environment

- vtk.js version: 17.17.x
- Browsers: any
- OS: any

floryst:
### Context

Fixes #2881 
- getNanColor returns an array and accepts no parameters.
- getNodeValue sets nodeVal to an array, not object.

### Results

- Correct usage of getNanColor and getNodeValue

### Changes

- [x] Updates to vtkColorTransferFunction

### PR and Code Checklist

### Testing

### DIFF
--- a/Sources/Rendering/Core/ColorTransferFunction/index.js
+++ b/Sources/Rendering/Core/ColorTransferFunction/index.js
@@ -389,13 +389,17 @@ function vtkColorTransferFunction(publicAPI, model) {
       // todo
       const idx = publicAPI.getAnnotatedValueIndexInternal(x);
       if (idx < 0 || numNodes === 0) {
-        publicAPI.getNanColor(rgb);
+        const nanColor = publicAPI.getNanColorByReference();
+        rgb[0] = nanColor[0];
+        rgb[1] = nanColor[1];
+        rgb[2] = nanColor[2];
       } else {
         const nodeVal = [];
         publicAPI.getNodeValue(idx % numNodes, nodeVal);
-        rgb[0] = nodeVal.r;
-        rgb[1] = nodeVal.g;
-        rgb[2] = nodeVal.b;
+        // nodeVal[0] is the x value. nodeVal[1...3] is rgb.
+        rgb[0] = nodeVal[1];
+        rgb[1] = nodeVal[2];
+        rgb[2] = nodeVal[3];
       }
       return;
     }
@@ -1030,7 +1034,10 @@ function vtkColorTransferFunction(publicAPI, model) {
       rgba[3] = 1.0; // NodeColor is RGB-only.
       return;
     }
-    publicAPI.getNanColor(rgba);
+    const nanColor = publicAPI.getNanColorByReference();
+    rgba[0] = nanColor[0];
+    rgba[1] = nanColor[1];
+    rgba[2] = nanColor[2];
     rgba[3] = 1.0; // NanColor is RGB-only.
   };
 


Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -1030,7 +1033,10 @@ function vtkColorTransferFunction(publicAPI, model) {
       rgba[3] = 1.0; // NodeColor is RGB-only.
       return;
     }
-    publicAPI.getNanColor(rgba);
+    const nanColor = publicAPI.getNanColor();
```
jourdain, line 1036: does that call do a new allocation? Maybe use getNanColorAsReference() if available instead?

Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -389,13 +389,16 @@ function vtkColorTransferFunction(publicAPI, model) {
...
+        rgb[0] = nanColor[0];
+        rgb[1] = nanColor[1];
+        rgb[2] = nanColor[2];
       } else {
         const nodeVal = [];
         publicAPI.getNodeValue(idx % numNodes, nodeVal);
-        rgb[0] = nodeVal.r;
-        rgb[1] = nodeVal.g;
-        rgb[2] = nodeVal.b;
+        rgb[0] = nodeVal[1];
```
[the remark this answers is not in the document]
floryst: Yep. I can add a comment indicating such.

Sources/Rendering/Core/ColorTransferFunction/index.js
floryst: Ah good point. Yes, I think there's a copy here. I forgot about the AsReference methods.

jourdain (approved):

Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -389,13 +389,17 @@ function vtkColorTransferFunction(publicAPI, model) {
       // todo
       const idx = publicAPI.getAnnotatedValueIndexInternal(x);
       if (idx < 0 || numNodes === 0) {
-        publicAPI.getNanColor(rgb);
```
finetjul, line 392: It would be a convenient signature to also support in `macros.js!getArray()` function

floryst: I certainly think it's a good way to write an array type directly into a destination array. It'll involve rewriting a lot of typescript definitions to include the optional out param...

PR #2887, opened 2023-08-08
Shareable graphics resources

sankhesh:
### Context

Add ability to cache graphics resources at the context level and have mappers re-use them if they are looking at the same data in the same render window. 

### Results

This ability opens up graphics resource sharing across viewports in applications where the same dataset is being rendered by different mappers. An example of this is provided with a QuadView that has three image reslice views and a volume view, all looking at the same 3D vtkImageData. Without this change, each mapper would be uploading the whole dataset to the GPU. 

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/Mapper2D/index.js
+++ b/Sources/Rendering/Core/Mapper2D/index.js
@@ -141,6 +141,23 @@ function vtkMapper2D(publicAPI, model) {
     }
     model.colorBuildString = `${publicAPI.getMTime()}${scalars.getMTime()}${alpha}`;
   };
+
+  publicAPI.getPrimitiveCount = () => {
+    const input = publicAPI.getInputData();
+    const pcount = {
+      points: input.getPoints().getNumberOfValues() / 3,
+      verts:
+        input.getVerts().getNumberOfValues() -
+        input.getVerts().getNumberOfCells(),
+      lines:
+        input.getLines().getNumberOfValues() -
+        2 * input.getLines().getNumberOfCells(),
+      triangles:
+        input.getPolys().getNumberOfValues() -
+        3 * input.getPolys().getNumberOfCells(),
+    };
+    return pcount;
+  };
 }
 
 // ----------------------------------------------------------------------------
--- a/Sources/Rendering/OpenGL/ImageResliceMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageResliceMapper/index.js
@@ -203,6 +203,12 @@ function vtkOpenGLImageResliceMapper(publicAPI, model) {
     if (!scalars) {
       return;
     }
+    if (model._scalars !== scalars) {
+      model._openGLRenderWindow.releaseGraphicsResourcesForObject(
+        model._scalars
+      );
+      model._scalars = scalars;
+    }
 
     const numComp = scalars.getNumberOfComponents();
     let toString = `${image.getMTime()}A${scalars.getMTime()}`;
@@ -221,7 +227,6 @@ function vtkOpenGLImageResliceMapper(publicAPI, model) {
         model.openGLTexture.setOglNorm16Ext(
           model.context.getExtension('EXT_texture_norm16')
         );
-        model.openGLTexture.releaseGraphicsResources(model._openGLRenderWindow);
         model.openGLTexture.resetFormatAndType();
         model.openGLTexture.create3DFilterableFromDataArray(
           dims[0],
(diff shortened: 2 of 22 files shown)


Sources/Rendering/OpenGL/ImageResliceMapper/index.js
```diff
@@ -231,130 +230,174 @@ function vtkOpenGLImageResliceMapper(publicAPI, model) {
...
-    const cfunToString = computeFnToString(
-      ppty,
-      ppty.getRGBTransferFunction,
-      numIComps
-    );
+    const colorTransferFunc = ppty.getRGBTransferFunction();
+    toString = computeFnToString(ppty, colorTransferFunc, numIComps);
+    const cTex =
+      model._openGLRenderWindow.getGraphicsResourceForObject(colorTransferFunc);
+    // const cTex = null;
```
floryst, line 254: Extraneous comment.

floryst (commented):
Exciting work! We will need to update the types for certain classes (e.g. ImageProperty, VolumeProperty, IStatistics, etc.).

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -1344,6 +1391,8 @@ export function extend(publicAPI, model, initialValues = {}) {
   model.bgImage.style.zIndex = '-1';
 
   model._textureResourceIds = new Map();
+  model._graphicsResources = new Map();
+  model._graphicsResourceHash = new Map();
```
jadh4v, line 1395: This is subjective but you probably don't need a separate `Map` for hash strings. Maybe just add that as a part of value in _graphicsResources map.
`_graphicsResources.set( vtkObj, { gObj: resource, hash: hashStr });`

jadh4v (approved):
Some subjective comments, but overall looks good to me. Really cool work!

bruyeret:
Very good work, I hope it will be merged soon!
Do you think it would be possible to factorize the code in the mappers?
I am talking about the modifications to `buildBufferObjects` in the VolumeMapper and ImageResliceMapper

Sources/Rendering/Core/Mapper2D/index.js
```diff
@@ -141,6 +141,23 @@ function vtkMapper2D(publicAPI, model) {
...
+      points: input.getPoints().getNumberOfValues() / 3,
+      verts:
+        input.getVerts().getNumberOfValues() -
+        input.getVerts().getNumberOfCells(),
+      lines:
+        input.getLines().getNumberOfValues() -
+        2 * input.getLines().getNumberOfCells(),
+      triangles:
+        input.getPolys().getNumberOfValues() -
+        3 * input.getPolys().getNumberOfCells(),
```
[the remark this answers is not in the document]
sankhesh: Not necessarily. Polygons are split up into triangles for upload to the GPU. And that's the primitive count that I am looking for here.

Sources/Rendering/Core/Mapper2D/index.js
```diff
@@ -141,6 +141,23 @@ function vtkMapper2D(publicAPI, model) {
     }
     model.colorBuildString = `${publicAPI.getMTime()}${scalars.getMTime()}${alpha}`;
   };
+
+  publicAPI.getPrimitiveCount = () => {
+    const input = publicAPI.getInputData();
+    const pcount = {
+      points: input.getPoints().getNumberOfValues() / 3,
```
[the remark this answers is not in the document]
sankhesh: Yes, this mapper is just for vtkPolyData.

Sources/Rendering/Core/Mapper2D/index.js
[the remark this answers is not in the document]
sankhesh: That is true now but earlier when we were adding extra triangles (for edge rendering, say), the primitive count differed. I'll leave this as is for now.

Sources/Rendering/OpenGL/RenderWindow/index.js
sankhesh: That's a good suggestion. Leaving it as is for now but will modify when I revisit this.

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -1286,6 +1316,59 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
+    if (!vtkObj) {
+      return false;
+    }
+    model._graphicsResources.get(vtkObj)?.releaseGraphicsResources(publicAPI);
+    return (
+      model._graphicsResources.delete(vtkObj) &&
+      model._graphicsResourceHash.delete(vtkObj)
+    );
+  };
+  publicAPI.releaseGraphicsResources = () => {
```
floryst, line 1352: Should this be called in `delete()`? Or is that handled elsewhere?

sankhesh: There is no `delete` at the OpenGL render window level because that is basically deleting the context.
This function should be called by applications that are using the switch data over time and need to free up the GPU allocations for old data.

floryst (approved):
LGTM

PR #2888, opened 2023-08-08
New filter to generate oriented outline of vtkImageData

jadh4v:
### Context

`vtkOutlineFilter` generates outline for any input data using the spatial bounds which is fine for datasets such as `vtkPolyData`. However, it doesn't consider the orientation of a `vtkImageData` or generate the outline oriented based on the direction cosines of the image.

This PR adds a `vtkImageDataOutlineFilter` similar to https://vtk.org/doc/nightly/html/classvtkImageDataOutlineFilter.html which will generate either a polyline outline or triangulated surface for an oriented image outline.

### Results
| generateFaces | generateLines |
|--|--|
| testImageDataOutlineFilter_Faces | testImageDataOutlineFilter_Lines |

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  master
  - **OS**:  Windows 11 Pro, 10.0.22621 Build 22621
  - **Browser**:  Chrome 115.0.5790.171 (Official Build) (64-bit)


Sources/Filters/General/ImageDataOutlineFilter/index.js
```diff
@@ -0,0 +1,149 @@
...
+// ----------------------------------------------------------------------------
+
+function vtkImageDataOutlineFilter(publicAPI, model) {
+  // Set our className
+  model.classHierarchy.push('vtkImageDataOutlineFilter');
+
+  function transformPoints(points, transform) {
+    const tmp = [0, 0, 0];
+    for (let i = 0; i < points.length; i += 3) {
+      const p = points.subarray(i, i + 3);
```
finetjul, line 37: `subarray` is typically a slow operation. 

It's faster to copy points manually

You may also use `vtkTransform.transformPoints` (to improve to support Mat4) or `vtkMatrixBuilder.apply`

Sources/Filters/General/ImageDataOutlineFilter/index.js
```diff
@@ -0,0 +1,149 @@
...
+
+    model.cubeSource.setXLength(spatialExt[1] - spatialExt[0]);
+    model.cubeSource.setYLength(spatialExt[3] - spatialExt[2]);
+    model.cubeSource.setZLength(spatialExt[5] - spatialExt[4]);
+
+    model.cubeSource.setCenter(
+      model.cubeSource.getXLength() / 2.0,
+      model.cubeSource.getYLength() / 2.0,
+      model.cubeSource.getZLength() / 2.0
+    );
```
finetjul, line 76: (optional) you may want to use `vtkBoundingBox` to avoid having to compute those lengths and center manually.

Sources/Filters/General/ImageDataOutlineFilter/index.js
```diff
@@ -0,0 +1,149 @@
...
+
+  // Also make it an algorithm with one input and one output
+  macro.algo(publicAPI, model, 1, 1);
+
+  // Internal persistent objects
+  model.cubeSource = vtkCubeSource.newInstance();
+  model.tmpOut = vtkPolyData.newInstance();
+  model.lineCells = vtkCellArray.newInstance({
+    values: Uint16Array.from(LINE_ARRAY),
+  });
```
finetjul, line 134: should be made protected

Sources/Common/Core/DataArray/index.js
```diff
@@ -425,8 +425,23 @@ function vtkDataArray(publicAPI, model) {
...
 
   publicAPI.deepCopy = (other) => {
+    // Retain current dataType and array reference before shallowCopy call.
+    const currentType = publicAPI.getDataType();
+    const currentArray = model.values;
     publicAPI.shallowCopy(other);
-    publicAPI.setData(other.getData().slice());
+
+    // Avoid array reallocation if size already sufficient
+    // and dataTypes match.
```
sankhesh, line 434: Are you trying to retain the original array pointer here?

jadh4v: Yes. In case the size of arrays is the same or the destination array is large enough, I am trying to avoid an unnecessary new allocation of array. From the point of view of any pipeline (that uses deep copy of DataArrays), and might get updated frequently due to state changes (e.g. mouse drag operation).

Sources/Filters/General/ImageDataOutlineFilter/index.js
jadh4v: Yes. Changed to use `vtkTransform.transformPoints`.

Sources/Filters/General/ImageDataOutlineFilter/index.js
jadh4v: Updated to use `vtkBoundingBox .getLengths()` and `.getCenter()`

Sources/Filters/General/ImageDataOutlineFilter/index.js
jadh4v: Updated. Please re-review to ensure if it was done correctly.

Sources/Filters/General/ImageDataOutlineFilter/index.js
```diff
@@ -0,0 +1,143 @@
...
+      vtkErrorMacro('Unable to fetch spatial extents of input image.');
+      return;
+    }
+
+    model._boundingBox.setBounds(spatialExt);
+    const lengths = model._boundingBox.getLengths();
+    model._cubeSource.setXLength(lengths[0]);
+    model._cubeSource.setYLength(lengths[1]);
+    model._cubeSource.setZLength(lengths[2]);
+    model._cubeSource.setCenter(model._boundingBox.getCenter());
```
finetjul, line 60: vtkCubeSource already has a `setBounds` convenient method. 
Problem is, it does not call "modified" when needed.
It would be a good opportunity to fix it (and here reuse vtkCubeSource.setBounds()).

Sources/Filters/General/ImageDataOutlineFilter/index.js
```diff
@@ -0,0 +1,143 @@
...
+    model._cubeSource.update();
+
+    const out = model._cubeSource.getOutputData();
+    model._tmpOut.getPoints().deepCopy(out.getPoints());
+
+    // Now, transform the cube polydata points in-place
+    // using the image's indexToWorld transformation.
+    model._transform.setMatrix(input.getIndexToWorld());
```
finetjul, line 68: vtkCubeSource already supports bounds "rotations".
Problem is, it supports only X, Y and Z angles.
It would be a good opportunity to extend the rotation to also support a matrix (to apply before the rotations ?)
You wouldn't have to worry here about transforming the points manually. Code would be more factorized.

Sources/Filters/General/ImageDataOutlineFilter/index.js
```diff
@@ -0,0 +1,143 @@
...
+      model._tmpOut.getPoints().getData()
+    );
+
+    // Lastly, generate the necessary cell arrays.
+    if (model.generateFaces) {
+      model._tmpOut.getPolys().deepCopy(out.getPolys());
+    } else {
+      model._tmpOut.getPolys().initialize();
+    }
+    if (model.generateLines) {
```
finetjul, line 80: couldn't this feature be moved to vtkCubeSource ?

Sources/Filters/General/ImageDataOutlineFilter/index.js
finetjul: looks good to me.

Sources/Filters/General/ImageDataOutlineFilter/index.js
jadh4v: I feel like the matrix should be applied the last, as something similar to `setUserMatrix()` function from VTK. Also, the x,y,z rotations wouldn't make sense after the cube has been moved away from the origin.
So basically, the `rotations` and `center` values will provide the initial setup in _model coordinates_, and then the `setMatrix()` option can provide additional option for _world coordinates_ placement.

Sources/Filters/General/ImageDataOutlineFilter/index.js
jadh4v: Can it be considered a _cube_ source still? It would also deviate from what VTK C++ does.

Sources/Filters/Sources/CubeSource/index.js
```diff
@@ -191,47 +218,32 @@ function vtkCubeSource(publicAPI, model) {
...
-    polys[polyIndex++] = 19;
-    polys[polyIndex++] = 17;
-
-    polys[polyIndex++] = 4;
-    polys[polyIndex++] = 20;
-    polys[polyIndex++] = 21;
-    polys[polyIndex++] = 23;
-    polys[polyIndex] = 22;
+    // Apply optional additionally specified matrix transformation
+    vtkMatrixBuilder.buildFromRadian().setMatrix(model.matrix).apply(points);
```
finetjul, line 222: For performance reason, do not call `setMatrix` if `model.matrix` is `undefined` or `null` (should be the default).
(and maybe it should also be the same if `model.rotations` are zeros...

Sources/Filters/General/ImageDataOutlineFilter/index.js
```diff
@@ -0,0 +1,92 @@
...
+  // Forward calls for [set/get]Generate[Faces/Lines] functions to cubeSource:
+  publicAPI.setGenerateFaces = (generateFaces) => {
+    if (model._cubeSource.setGenerateFaces(generateFaces)) {
+      publicAPI.modified();
+    }
+  };
+
+  publicAPI.setGenerateLines = (generateLines) => {
+    if (model._cubeSource.setGenerateLines(generateLines)) {
+      publicAPI.modified();
```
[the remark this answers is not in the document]
jadh4v: Thanks. Please see the suggested changes here: https://github.com/Kitware/vtk-js/pull/2895

Sources/Filters/Sources/CubeSource/index.js
jadh4v: Thanks. Please see the suggested changes here: https://github.com/Kitware/vtk-js/pull/2895

PR #2899, opened 2023-08-22
SegmentedLineRepresentation to connect widget handles 

PaulHax:
Not sure we want to burden the repo with this niche possibly duplicate code, but tossing this out there.

### Context

I wanted to click on a segment of the PolyLineWidget and insert a handle in the middle of the line. 
https://discourse.vtk.org/t/picking-tube-between-points-of-polylinewidget/12154

### Results

Ran with @finetjul suggestion to use a Glyph mapper with cylanders to connect the points of a line.  Can get the index of the cylander from WidgetManager
`const insertHandleIndex = widgetManager.getSelections()[0].getProperties().compositeID + 1;`

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Widgets/Representations/SegmentedLineRepresentation/example/index.js
+++ b/Sources/Widgets/Representations/SegmentedLineRepresentation/example/index.js
@@ -1,0 +1,56 @@
+import '@kitware/vtk.js/favicon';
+
+// Load the rendering pieces we want to use (for both WebGL and WebGPU)
+import '@kitware/vtk.js/Rendering/Profiles/All';
+
+import vtkFullScreenRenderWindow from '@kitware/vtk.js/Rendering/Misc/FullScreenRenderWindow';
+import vtkStateBuilder from '@kitware/vtk.js/Widgets/Core/StateBuilder';
+
+import vtkSegmentedLineRepresentation from '@kitware/vtk.js/Widgets/Representations/SegmentedLineRepresentation';
+
+// ----------------------------------------------------------------------------
+// Standard rendering code setup
+// ----------------------------------------------------------------------------
+
+const fullScreenRenderer = vtkFullScreenRenderWindow.newInstance({
+  background: [0, 0, 0],
+});
+const renderer = fullScreenRenderer.getRenderer();
+const renderWindow = fullScreenRenderer.getRenderWindow();
+
+// -----------------------------------------------------------
+// State
+// -----------------------------------------------------------
+
+const compositeState = vtkStateBuilder
+  .createBuilder()
+  .addDynamicMixinState({
+    labels: ['handles'],
+    mixins: ['origin'],
+    name: 'handle',
+  })
+  .build();
+
+const z = -50;
+const points = [
+  [0, 0, z],
+  [5, 0, z],
+  [5, 10, z],
+];
+points.forEach((point) => {
+  const handle = compositeState.addHandle();
+  handle.setOrigin(point);
+});
+
+// -----------------------------------------------------------
+// Representation
+// -----------------------------------------------------------
+
+const widgetRep = vtkSegmentedLineRepresentation.newInstance({
+  scaleInPixels: false,
+});
+widgetRep.setInputData(compositeState);
+widgetRep.setLabels(['handles']);
+widgetRep.getActors().forEach(renderer.addActor);
+
+renderWindow.render();
--- a/Sources/Widgets/Representations/SegmentedLineRepresentation/example/index.js
+++ b/Sources/Widgets/Representations/SegmentedLineRepresentation/example/index.js
@@ -34,8 +34,9 @@ const compositeState = vtkStateBuilder
 const z = -50;
 const points = [
   [0, 0, z],
-  [5, 0, z],
-  [5, 10, z],
+  [5, 5, z],
+  [3, 10, z],
+  [5, 15, z],
 ];
 points.forEach((point) => {
   const handle = compositeState.addHandle();
@@ -48,6 +49,7 @@ points.forEach((point) => {
 
 const widgetRep = vtkSegmentedLineRepresentation.newInstance({
   scaleInPixels: false,
+  // closePolyLine: true,
 });
 widgetRep.setInputData(compositeState);
 widgetRep.setLabels(['handles']);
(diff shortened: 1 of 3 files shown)


finetjul (commented):
why not derive from vtkGlyphRepresentation ? 

I'm not sure about the SegmentedLineRepresentation. How about LineGlyphRepresentation ?

PaulHax:
`LineGlyphRepresentation` makes sense.  

I did not see how `GlyphRepresentation`'s polydata from state mixins helped me.  I would have needed to override them all and repeat the points iteration in each.  https://github.com/Kitware/vtk-js/pull/2899/files#diff-a7d2b7d96e3c0ce1c14f718829973230f2e3ee06fd6542c28dc6bc39d684475eR115-R139

Did I miss some nice feature GlyphRepresntation would have helped with?

finetjul:
You indeed would have had to overwrite `scale3` and `direction`. I think the `origin` could have stayed as is. The origin of the cylinder glyph would have had to be at the start of the segment instead of its middle.
The overrides could then have been passed as `initialValues` of the `vtkGlyphRepresentation.extend()` constructor. Of course you would have had also to override `hasMixin()` to force it to return `true` even if the states do not hold them.

I do not think it's a problem to repeat the point iteration in each mixin function, it does not change the order of magnitude. Each mixin are small enough to be readable.

I've been thinking that it would be nice if your polyline had "joints" between each segment. This could be done by adding spheres (of diameter the same as the line thickness) at each point handle. You would need however 2 glyphmappers though.

PaulHax:
Ah, setting the center on the cylander glyph means we don't have to compute the center of the segments.  Good tip.

If there was a 1 to 1 state mixin to PolyData array relationship,  GlyphRepresentation seems helpful.  In this case, OriginMixin makes 3 arrays, (position, orientation, scale.)  

Regarding joints, should we keep this simple and ask the using Widgets to use `vtkSphereHandleRepresentation` for them?  That is the original goal for this, to have pickable segments between handles.

finetjul:
> If there was a 1 to 1 state mixin to PolyData array relationship, GlyphRepresentation seems helpful. In this case, OriginMixin makes 3 arrays, (position, orientation, scale.)

Where did you see that OriginMixin makes 3 arrays ?


> Regarding joints, should we keep this simple and ask the using Widgets to use vtkSphereHandleRepresentation for them? That is the original goal for this, to have pickable segments between handles.

Good idea, can you update your example to demonstrate that, that will make the screenshot more appealing.

PaulHax:
Incopreated finetjul's GlyphRepresentation rewrite from here
https://github.com/Kitware/vtk-js/commit/3a9c0442ff819d2d11d6dad71361f6158d7d1f73

Added sphere joints to the example.

Sources/Widgets/Representations/LineGlyphRepresentation/example/index.js
```diff
@@ -0,0 +1,68 @@
...
+// Representation
+// -----------------------------------------------------------
+
+const widgetRep = vtkLineGlyphRepresentation.newInstance({
+  scaleInPixels: false,
+  lineThickness: 0.5, // radius of the cylinder
+});
+widgetRep.setInputData(compositeState);
+widgetRep.setLabels(['handles']);
+widgetRep.getActors().forEach(renderer.addActor);
```
finetjul, line 58: a representation is a prop, therefore you shouldn't have to manually add all the actors of the representation to the renderer. Adding the representation to the renderer should be enough...

Sources/Widgets/Representations/LineGlyphRepresentation/index.js
```diff
@@ -0,0 +1,141 @@
...
+        vtkCylinderSource.newInstance({
+          direction: [1, 0, 0],
+          center: [0.5, 0, 0],
+          capping: false,
+        }),
+      ...initialValues?.pipeline,
+    },
+    applyMixin: {
+      noScale: cylinderScale(publicAPI, model),
+      scale1: cylinderScale(publicAPI, model),
```
finetjul, line 112: I can see some cases where a scale mixin is provided and we want to consider it to scale the cylinders.
Maybe you could add a vtkLineGlyphRepresentation property to "ignore" the scale mixin (e.f. forceLineThickness:bool)

Sources/Widgets/Representations/LineGlyphRepresentation/example/index.js
PaulHax: When i change line 58 to `renderer.addActor(widgetRep);` getting runtime error
```
index.js:339 Uncaught TypeError: prop.getBounds is not a function
    at publicAPI.computeVisiblePropBounds (index.js:339:29)
```
Guess prop is not quite an actor?

finetjul: fair enough.

I guess a representation could be considered as an actor, but there must have been a good reason why it is just a prop (the widget view is however an actor with nested props being representations).

WDYT @floryst ?

Sources/Widgets/Representations/LineGlyphRepresentation/index.js
PaulHax: Added the `forceLineThickness` flag.

finetjul (approved):
LGTM

Documentation/content/examples/index.md
```diff
@@ -68,7 +68,7 @@ This will allow you to see the some live code running in your browser. Just pick
 [![ResliceCursorWidget Example][ResliceCursorWidget]](./ResliceCursorWidget.html "Axial Coronal and Sagittal MPR/Oblique/Reformatted/Resliced/Slab/MIP views")
 [![ShapeWidget Example][ShapeWidget]](./ShapeWidget.html "2D shape widgets with text information")
 [![SphereWidget Example][SphereWidget]](./SphereWidget.html "2D sphere widget controlled with radius")
-[![SplineWidget Example][SplineWidget]](./PaintWidget.html "Widget to draw open or closed (triangularized) sharp/smooth polygon widget")
```
finetjul, line 71: oopsie

Sources/Widgets/Representations/LineGlyphRepresentation/example/index.js
floryst: I guess because representations could be 2D as well. This is probably fine for the examples, as representations are not typically used directly.

Issue #2600, opened 2022-10-12
Update browserslist configuration


FezVrasta:
### High-level description
Right now a lot of code is transpiled, resulting in slower code. One example is the function parameters spread, that is converted into a for loop that allocates a new array, that then needs to be garbage collected.

Could the config be updated to only support modern browsers?

### Environment
- **vtk.js**: 
- **OS**: 
- **Browser**:

finetjul:
VTK.js attempts to be used by the largest user base.

The current `browserslist` setting is restrictive already (i.e. `> 0.5%, last 2 versions, Firefox ESR, not dead`). 

What blocks `browserlist` from supporting [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) is ALL versions of IE (>0.5 but dead?) and Opera Mini(1.07%).

Interestingly, OperaMini and IE < 11 do not support WebGL, VTK.js without WebGL does not make sense. So we could add the `supports webgl` criteria to the browserslist config. That would leave only IE=v11 which is only 0.48% of usage (but already considered dead?).

Can you check if that makes a difference ?

Alternatively, your project can also depend on [vtk.js](https://www.npmjs.com/package/vtk.js) instead of [@kitware/vtk.js](https://www.npmjs.com/package/@kitware/vtk.js) and you implement your own browserslist rule.

PR #2901, opened 2023-08-24
Update the ESM build with newer browserslist db + babel

floryst:
### Context

Updates the browserslist db and bumps babel to output ES6 code for `@kitware/vtk.js`. The `vtk.js` package remains the same.

### Results

ESM build now outputs to ES6 according to the browserslist db. We use the default browserslist browsers.

### Changes
- [ ] Documentation will be updated to reflect this change.

### PR and Code Checklist

### Testing
No testing has been done.

### DIFF
--- a/Documentation/content/docs/intro_vtk_as_es6_dependency.md
+++ b/Documentation/content/docs/intro_vtk_as_es6_dependency.md
@@ -4,7 +4,7 @@ title: Using vtk.js as an ES dependency
 This guide illustrates how to consume vtk.js as an ES dependency.
 
 This document was last updated with the following package versions:
-- `@kitware/vtk.js@v19`
+- `@kitware/vtk.js@v28`
 - `webpack@v5`
 
 ## Starting a new vtk.js project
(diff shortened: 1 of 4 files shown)


jourdain (approved):

finetjul:
That's a breaking change, isn't it ?

floryst:
I've changed this to instead upgrading babel + browserslist db. For some reason I had to upgrade babel in order for it to pick up the newer browserslist db, which ends up targeting ES6.

@jourdain @finetjul I'd like your opinion on this. I'd imagine we may still want to target ES5 browsers. However, it appears that the browserslist `defaults` (which is `> 0.5%, last 2 versions, Firefox ESR, not dead`) no longer includes ES5-only browsers. If we change it to `> 0.2%, last 2 versions, Firefox ESR, not dead`, we get ES5-compatible output. Should we still target ES5, and add documentation on what you should do if you target ES5?

jourdain:
Wow, ES6 is now the mainstream... That is cool.

For vtk.js, what is the cost to keep ES5-compatible output? (increased bundle size, dependency complexity, ??)

finetjul:
I think it's fine to drop ES5, but we need to document how one can build VTK.js to have ES5.

There is that related discussion as well:
https://github.com/Kitware/vtk-js/issues/2600

jourdain:
Can we make the bundled UMD es5 and the ESM one ES6 only?

FezVrasta:
Any update on this?

floryst:
> Any update on this?

I'll revisit this. I need to validate that the UMD package outputs ES5 and add docs.

jourdain (approved):

finetjul:
Lgtm

PR #2902, opened 2023-08-25
perf: use Set to power class hierarchy

FezVrasta:
### Context

The custom ClassHierarchy class is effectively re-implementing the `Set` object. Using `Set` can result in better performance, especially when it doesn't get transpiled. The existing ClassHierarchy when transpiled results in many GC events.

### Results

No GC events and a cleaner interface.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


finetjul:
I guess it is fine to me (except for the broken dashboard). 

Please consider the following discussion for in-depth usage and limitations:
https://github.com/Kitware/vtk-js/pull/1913

FezVrasta:
I can't see what test is failing, they all report a [PASS] status 🤔

finetjul:
That means there is a hang. I just re-started the test.

FezVrasta:
It failed again, it works fine locally, I also tested this change on our app and it looks okay.

floryst:
Test failures: I wonder if chrome 116 specifically is being finicky in github CI.

GC: that will be resolved generally #2901.

Using Set: I am mostly fine with moving to Set. I do consider it a micro-optimization since class hierarchies are never long enough for perf to matter. This will result in a breaking change, since downstream consumers who make their own vtk.js classes will need to switch to using `model.classHierarchy.add`. If you agree, please add such to the commit message.

FezVrasta:
I'm okay in waiting for #2901 to see how the performance is affected, in general I thought it would make to use a Set here as it's what the code is after.

jourdain:
Make sure the `const a = obj.getState()` and `vtk(a)` still works...

FezVrasta:
I tested out the new version and the memory issue is not there anymore, thanks

PR #2903, opened 2023-08-27
2890 add rendering picking screenshots in example gallery

finetjul:
### Context
The example landing page as part of the documentation.

### Results
Added 2 image gallery sections: Rendering and Picking

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


floryst:
Awesome!

One thing I didn't catch from the last example PR: do we need 1080x750 images for the gallery? I think those thumbnails won't ever be viewed at such high resolution.

finetjul:
It may not be useful for the gallery, you're right.
But I'm sure they could be used for other use cases (e.g. blog posts).

I guess the images have a low enough memory footprint to be stored in git.
We could lower the resolution but we would need to store somewhere these mid-size images.

floryst:
My concern is that the size adds up in the repo. I don't really see us ever utilizing these screenshots at a higher resolution than what is shown on the examples page.

I'm mostly good with merging it for now. If we reduce the framerate of the gif, how much space will that shave off?

finetjul:
Sounds good, I resized down all the images and removed some frames to the gif.

Is it worth it if I scale down also the previously added images (or is it too late for git memory storage saving purpose)?

floryst:
We can keep the previously added images for now.

PR #2904, opened 2023-08-28
feat(interactorstyleremotemouse): add option to throttle MouseWheel e…

DevinAPorro:
…vents

Adds a **wheelThrottleDelay** variable that, when set, will throttle MouseWheel events by that amount.

### Context

Provide the option for consumers to throttle MouseWheel events when desired. This is needed in our use case for guaranteeing  all image slices are seen when scrolling in a specific mode (so we want to throttle the number of MouseWheel events that get sent to the server so as to not overwhelm the backend with image requests at a rate that is too fast for the client to display them).  

### Results

image
image

1. In the video below, when the "wheelThrottleDelay" checkbox is checked, we set wheelThrottleDelay to the amount set in the slider (shown in text next to the slider control). In this setup, you will see MouseWheel events that happen at a rate faster than 
wheelThrottleDelay get discarded (with the additional temporary logging I added). 
2. When the "wheelThrottleDelay" checkbox is not checked, all MouseWheel events get sent (identical to the current behavior in vtk.js).
Additional temporary logging added to Sources/Interaction/Style/InteractorStyleRemoteMouse/index.js highlighted:
image
image

https://github.com/Kitware/vtk-js/assets/119877818/90402508-2fc7-474d-97be-75960961ef12

### Changes

1. Public API added: **wheelThrottleDelay** . When set to a value greater than 0, will throttle MouseWheel events by the set amount (in ms). This follows the pattern established in `handleMouseMove`/`throttleDelay`. When set to 0 (or any other falsy value), all MouseWheel events will be sent (the current behavior). wheelThrottleDelay defaults to 0, so there are no changes unless a consumer explicitly sets wheelThrottleDelay to a non-0 value.
4. Private variable added: **wheelLastThrottleTime**. Holds the timestamp for the last MouseWheel event sent, to compare with the next incoming MouseWheel event to determine if the time delta is greater than wheelThrottleDelay (in which case the event will get sent).

### PR and Code Checklist

### Testing

- [X] Tested environment:
  - **vtk.js**:  master (28.8.3)
  - **OS**: Ubuntu 22.04
  - **Browser**: Chrome 116.0.5845.110, FireFox 116.0.3

### DIFF
--- a/Sources/Interaction/Style/InteractorStyleRemoteMouse/index.js
+++ b/Sources/Interaction/Style/InteractorStyleRemoteMouse/index.js
@@ -120,12 +120,23 @@ function vtkInteractorStyleRemoteMouse(publicAPI, model) {
 
   //-------------------------------------------------------------------------
   publicAPI.handleMouseWheel = (callData) => {
-    publicAPI.invokeRemoteWheelEvent({
-      type: 'MouseWheel',
-      ...createRemoteEvent(callData),
-      spinY: callData.spinY,
-    });
-    publicAPI.invokeInteractionEvent(INTERACTION_EVENT);
+    let needToSend = true;
+    if (model.wheelThrottleDelay) {
+      const ts = Date.now();
+      needToSend = model.wheelThrottleDelay < ts - model.wheelLastThrottleTime;
+      if (needToSend) {
+        model.wheelLastThrottleTime = ts;
+      }
+    }
+
+    if (needToSend) {
+      publicAPI.invokeRemoteWheelEvent({
+        type: 'MouseWheel',
+        ...createRemoteEvent(callData),
+        spinY: callData.spinY,
+      });
+      publicAPI.invokeInteractionEvent(INTERACTION_EVENT);
+    }
   };
 
   //-------------------------------------------------------------------------
@@ -283,6 +294,8 @@ const DEFAULT_VALUES = {
   sendMouseMove: false,
   throttleDelay: 33.3, // 33.3 millisecond <=> 30 events/second
   lastThrottleTime: 0,
+  wheelThrottleDelay: 0,
+  wheelLastThrottleTime: 0,
   // remoteEventAddOn: null,
 };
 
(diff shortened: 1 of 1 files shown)


Sources/Interaction/Style/InteractorStyleRemoteMouse/index.js
```diff
@@ -120,12 +120,23 @@ function vtkInteractorStyleRemoteMouse(publicAPI, model) {
...
   //-------------------------------------------------------------------------
   publicAPI.handleMouseWheel = (callData) => {
-    publicAPI.invokeRemoteWheelEvent({
-      type: 'MouseWheel',
-      ...createRemoteEvent(callData),
-      spinY: callData.spinY,
-    });
-    publicAPI.invokeInteractionEvent(INTERACTION_EVENT);
+    let needToSend = true;
+    if (model.wheelThrottleDelay) {
```
finetjul, line 124: Why bother checking `wheelThrottleDelay` ?

If it is 0, then `needToSend = model.whe...` will always be equal to `true`, isn't it ?

DevinAPorro: good point. I added the check just to completely isolate the new code (so the default code path wouldn't really be modified), but I will update. Thanks for the feedback!

DevinAPorro: oh, sorry, looking at the code again, the only benefit to the check is that when wheelThrottleDelay isn't set, Date.now() won't get called (which is slightly expensive)- should I update anyway?

jourdain (approved):
Looks good to me. Thanks for your contribution.

jourdain:
@floryst, do you want to bundle it with others MR?

DevinAPorro:
Thanks @jourdain ! Is there anything else I can/should do at this point? (Sorry, my first time contributing here and just want to make sure I follow the process correctly :) )

floryst:
I'll merge this today!

PR #2909, opened 2023-09-01
Fix logic of computeClosestIJKAxis

bruyeret:
Two mappers with two different slicing modes could have the same ijkMode
For example, two mappers with X and Y slicing mode could both show slice J
Also remove the warning about inaccurate slicing because oblique slicing should be done with the new ImageResliceMapper and the slicing modes XYZ are intentionally used to snap to IJK

Add a function to vtkMath (with tests and docs)

### DIFF
--- a/Sources/Common/Core/Math/index.js
+++ b/Sources/Common/Core/Math/index.js
@@ -2146,6 +2146,56 @@ export function getMajorAxisIndex(vector) {
   return axisIndex;
 }
 
+// Return the closest orthogonal matrix of 1, -1 and 0
+// It works for both column major and row major matrices
+// This function iteratively associate a column with a row by choosing
+// the greatest absolute value from the remaining row and columns
+// For each association, a -1 or a 1 is set in the output, depending on
+// the sign of the value in the original matrix
+export function getSparseOrthogonalMatrix(matrix, n = 3) {
+  // Initialize rows and columns to available indices
+  const rows = new Array(n);
+  const cols = new Array(n);
+  for (let i = 0; i < n; ++i) {
+    rows[i] = i;
+    cols[i] = i;
+  }
+  // No need for the last iteration: i = 0
+  for (let i = n - 1; i > 0; i--) {
+    // Loop invariant:
+    // rows[0:i] and cols[0:i] contain the remaining rows and columns
+    // rows]i:n[ and cols]i:n[ contain the associations found (rows[k] is associated with cols[k])
+    let bestValue = -Infinity;
+    let bestRowI = 0;
+    let bestColI = 0;
+    for (let rowI = 0; rowI <= i; ++rowI) {
+      const row = rows[rowI];
+      for (let colI = 0; colI <= i; ++colI) {
+        const col = cols[colI];
+        const absVal = Math.abs(matrix[row + n * col]);
+        if (absVal > bestValue) {
+          bestValue = absVal;
+          bestRowI = rowI;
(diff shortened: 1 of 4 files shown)


Sources/Common/Core/Math/test/testMath.js
```diff
@@ -754,3 +754,45 @@ test('Test singularValueDecomposition', (t) => {
...
+      matrix: [1, 0, 0, 0, 0, 1, 0, 1, 0],
+      expected: [0, 2, 1],
+    },
+    {
+      matrix: [1, 0, 0, 0, 0, -1, 0, 1, 0],
+      expected: [0, 2, 1],
+    },
+    {
+      matrix: [1, 0, 0, 0, 0, -1, 0, 1, 0],
+      expected: [0, 2, 1],
```
finetjul, line 777: why twice the same test

Sources/Common/Core/Math/test/testMath.js
```diff
@@ -754,3 +754,45 @@ test('Test singularValueDecomposition', (t) => {
...
+    },
+    {
+      matrix: [1, 0, 0, 0, 1, 0, 0, 0, 1],
+      expected: [0, 1, 2],
+    },
+    {
+      matrix: [1, 0, 0, 0, 1, 0, 0, 0, 1],
```
finetjul, line 784: why twice the same test ?

Sources/Common/Core/Math/index.js
```diff
@@ -2146,6 +2146,51 @@ export function getMajorAxisIndex(vector) {
   return axisIndex;
 }
 
+// This function iteratively associate a column with a row by choosing
+// the greatest absolute value from the remaining row and columns
+// Returns an array L such as if the matrix is column major,
+// the row i is associated with column L[i]
+export function getRowColumnAssociations(matrix, n = 3) {
```
finetjul, line 2153: The documentation is not clear to me. 
It seems that you return an order for columns. What is this order ?

bruyeret: The order of the column indices is the order of the rows, this is what I mean by `the row i is associated with column L[i]`
How would you phrase this differently?

bruyeret:
I renamed `getRowColumnAssociations` to `getSparseOrthogonalMatrix` and it now returns an orthogonal matrix of 0, 1 and -1
What do you think @finetjul

Sources/Common/Core/Math/test/testMath.js
```diff
@@ -754,3 +754,47 @@ test('Test singularValueDecomposition', (t) => {
...
+
+test.only('Test getSparseOrthogonalMatrix', (t) => {
+  const successMsg = 'Row/column association is correct';
+  const tests = [
+    {
+      matrix: [
+        0.702, 0.7025, -0.1163, -0.2856, 0.1281, -0.9497, -0.6523, 0.7, 0.2906,
+      ],
+      expected: [0, 1, 0, 0, 0, -1, -1, 0, 0],
+    },
```
finetjul, line 766: can you maybe also add as a test the problem you faced with Kitware Glance ?

finetjul:
LGTM

Sources/Common/Core/Math/test/testMath.js
bruyeret: It is the first test matrix, which use to associate both X and Z to J

finetjul (approved):
LGTM

PR #2914, opened 2023-09-13
fix(interactor): mouseEvent is not available in the worker

sedghi:
### Context

I was trying to consume @kitware/vtk.js inside a webworker which I encountered an error regarding MouseEvent not being available (since it is worker context). This PR will check if the MouseEvent is available then use it

### Results

Added conditional check for mouseEvent

### Changes

replaces 

```js
EMPTY_MOUSE_EVENT = new MouseEvent('');
``` 

with

```js
let EMPTY_MOUSE_EVENT;
if (typeof MouseEvent !== 'undefined') {
  EMPTY_MOUSE_EVENT = new MouseEvent('');
} else {
  // Provide a fallback or handle differently for web worker context
  EMPTY_MOUSE_EVENT = {
    ctrlKey: false,
    altKey: false,
    shiftKey: false,
  };
}

```

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -11,7 +11,17 @@ const { vtkWarningMacro, vtkErrorMacro, normalizeWheel, vtkOnceErrorMacro } =
 // Global methods
 // ----------------------------------------------------------------------------
 
-const EMPTY_MOUSE_EVENT = new MouseEvent('');
+let EMPTY_MOUSE_EVENT;
+if (typeof MouseEvent !== 'undefined') {
+  EMPTY_MOUSE_EVENT = new MouseEvent('');
+} else {
+  // Provide a fallback or handle differently for web worker context
+  EMPTY_MOUSE_EVENT = {
+    ctrlKey: false,
+    altKey: false,
+    shiftKey: false,
+  };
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -11,7 +11,17 @@ const { vtkWarningMacro, vtkErrorMacro, normalizeWheel, vtkOnceErrorMacro } =
...
+let EMPTY_MOUSE_EVENT;
+if (typeof MouseEvent !== 'undefined') {
+  EMPTY_MOUSE_EVENT = new MouseEvent('');
+} else {
+  // Provide a fallback or handle differently for web worker context
+  EMPTY_MOUSE_EVENT = {
+    ctrlKey: false,
+    altKey: false,
+    shiftKey: false,
+  };
```
floryst, line 23: I think we can just set EMPTY_MOUSE_EVENT to this object and do away with the `new MouseEvent` call.

sedghi: sure I can handle that

floryst:
LGTM!

PR #2915, opened 2023-09-13
Fix large scalar values

### LINKED ISSUE #1426: Problem rendering vtkImageData with large scalar values
Using either the `OpenGL/VolumeMapper` (as illustrated by #1425) or also the `OpenGL/ImageMapper` to render `vtkImageData` with large scalar values results in incorrectly rendered images.

I'm not sure precisely when the data range gets too large and the problem starts to happen, but in the (currently failing) testing linked above, the scalar range of the rendered data is roughly `[0, 2.5M]`.  Some other cases where I've seen the issu

As I dug into the problem a little, I discovered that a couple of quick things seem to solve the problem, but aren't good long term solutions in general.  

1. Uploading the scalar data to texture memory as 32 bit values seems to fix the problem (e.g. changing `R16F` to `R32F` https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/RenderWindow/index.js#L556), but causes other tests to fail.  This is also wasteful in most cases we've handled so far, as it doubles the amount of texture memory used for all datasets, compared to what is implemented currently.

2. Forcing the `OpenGL/Texture` module to take the "convert to float" path [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/Texture/index.js#L1081), also resolves this particular issue, but doesn't make sense for most datasets we've encountered so far.

So maybe we could do something along the lines of 1. or 2., above, but only conditionally, depending on the scalar range of the data?

floryst:
### Context
Large scalar values cannot be accurately represented in half float format. When large scalar values are detected, switch to 32-bit floats.

Additionally enables the `OES_texture_float_linear` extension for 32-bit texture filtering.

### Results
- Closes #1426
- Falls back to 32 bit formats if the scalar value is too large
- Enables norm16 extension for ImageMapper

### Changes

- [ ] fixes vtkOpenGLImageMapper to support the norm16 ext
- [ ] fixes to vtkOpenGLTexture to enable the texture_float_linear extension
- [ ] fixes to vtkOpenGLRenderWindow to support 32-bit formats

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: v28.3.3
  - **OS**: macOS
  - **Browser**: Chrome 116, Firefox 117


Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -364,6 +364,20 @@ function vtkOpenGLTexture(publicAPI, model) {
...
     }
 
+    if (
+      [
+        model.context.R32F,
+        model.context.RG32F,
+        model.context.RGB32F,
+        model.context.RGBA32F,
+      ].includes(model.internalFormat) &&
+      !model.context.getExtension('OES_texture_float_linear')
```
floryst, line 374: @sankhesh I noticed that we call getExtension several times in the render path. What if we pre-cached all of the used extensions and looked them up using a cache instead of calling getExtension every time?

sankhesh: Yep, I have that change in the branch with memory leak fixes. I'll clean it up and create a PR soon

sankhesh (approved):
LGTM

floryst:
Rebased and updated the tests to use `tape` rather than `tape-catch`.

PR #2923, opened 2023-09-22
Change style of reslice cursor widget back to as it was in v25

bruyeret:
Remove lighting
Spin cyclinder representations to avoid cropping of the cylinder by the imageReslice

### DIFF
--- a/Sources/Widgets/Representations/LineHandleRepresentation/index.js
+++ b/Sources/Widgets/Representations/LineHandleRepresentation/index.js
@@ -96,6 +96,7 @@ function defaultValues(initialValues) {
     _pipeline: {
       glyph: vtkCylinderSource.newInstance({
         resolution: initialValues.glyphResolution ?? 4,
+        initAngle: initialValues.glyphAngle ?? Math.PI / 4,
         direction: [0, 0, 1],
       }),
     },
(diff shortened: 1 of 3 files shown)


Sources/Widgets/Representations/LineHandleRepresentation/index.js
```diff
@@ -96,6 +96,7 @@ function defaultValues(initialValues) {
     _pipeline: {
       glyph: vtkCylinderSource.newInstance({
         resolution: initialValues.glyphResolution ?? 4,
+        initAngle: initialValues.glyphAngle ?? 0,
```
finetjul, line 99: shouldn't it be PI/4 by default ?

bruyeret: In my opinion it is the responsibility of the reslice cursor to set the right angle for its use
The LineHandleRepresentation doesn't have a good reason to have a rotation by default

finetjul: The good reason is that the default resolution is 4.

If it was not 4, I agree it does not make sense to have an angle != 0.

But in this case it is explicit that we want a rectangular prism that is aligned with orthogonal axes.

No ?

bruyeret: The alignment with the axes is a requirement of the reslice cursor
Is it also a default requirement for the other widgets using this representation?

finetjul: I think it is a good default for ALL the lines, not just RCW.

bruyeret: Ok no problem, it is fixed!

PR #2927, opened 2023-09-29
Fix typed arrays serialization deserialization

Julesdevops:
### Context
The behavior of JSON.stringify for typed arrays is to serialize them as objects with numeric keys. This cause crashes when re-building the object since it would try to fit an object where a typed array is required. This changes makes it so getState replaces typed arrays with plain javascript arrays.

Typical example is `vtkImageData` direction property, which is by default Float64Array
minimal reproduction:
```javascript
import vtk from '@kitware/vtk.js/vtk';
import vtkImageData from '../../Sources/Common/DataModel/ImageData';

const imdt = vtkImageData.newInstance();
const jsonRepr = JSON.stringify(imdt);
const jsonDeserialize = JSON.parse(jsonRepr);

const vtkObj = vtk(jsonDeserialize); // this should crash
```
see also https://github.com/Kitware/glance/issues/480

### Results
makes serialization/deserialization possible and successful

### Changes

This change makes it so we serialize typed arrays as arrays.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: linux
  - **Browser**: chrome, firefox latest

### DIFF
--- a/Sources/macros.js
+++ b/Sources/macros.js
@@ -174,6 +174,10 @@ function safeArrays(model) {
   });
 }
 
+function isTypedArray(value) {
+  return Object.values(TYPED_ARRAYS).some((ctor) => value instanceof ctor);
+}
+
 // ----------------------------------------------------------------------------
 // shallow equals
 // ----------------------------------------------------------------------------
@@ -378,6 +382,8 @@ export function obj(publicAPI = {}, model = {}) {
         jsonArchive[keyName] = jsonArchive[keyName].getState();
       } else if (Array.isArray(jsonArchive[keyName])) {
         jsonArchive[keyName] = jsonArchive[keyName].map(getStateArrayMapFunc);
+      } else if (isTypedArray(jsonArchive[keyName])) {
+        jsonArchive[keyName] = Array.from(jsonArchive[keyName]);
       }
     });
 
--- a/Sources/Testing/testSerialization.js
+++ b/Sources/Testing/testSerialization.js
@@ -3,11 +3,12 @@ import vtk from 'vtk.js/Sources/vtk';
 import macro from 'vtk.js/Sources/macros';
 
 import vtkDataArray from 'vtk.js/Sources/Common/Core/DataArray';
-import vtkPoints from 'vtk.js/Sources/Common/Core/Points';
+import vtkImageData from 'vtk.js/Sources/Common/DataModel/ImageData';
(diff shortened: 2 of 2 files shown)


Julesdevops:
fyi @finetjul @bruyeret

bruyeret (approved):
LGTM 👍

finetjul (commented):
I can think of 2 better (not orthogonal) alternatives:
 - `vtkImageData.indexToWorld` and `worldToIndex` must not be typed array but regular arrays, which are 64 bits anyway.
 - `vtkImageData.indexToWorld` and `worldToIndex` shall not be serialized but must be flagged as protected by adding `_` prefix and recomputed at deserialization.

Julesdevops:
to me those are not alternatives to the fix I'm proposing which is necessary but definitely you have a good point.
probably gonna include those in the scope of this PR

finetjul:
> to me those are not alternatives to the fix I'm proposing which is necessary but definitely you have a good point.
> probably gonna include those in the scope of this PR

I agree, your fix is also necessary.
I have that bad habit to not willing to fix a crash if it can help detect bad design in the first place.

Julesdevops:
we actually uses `Float64Array` a lot for gl-matrix math related stuff. so I wouldn't call it 'bad design' in this case.
if we change that in `vtkImageData` we would need to update everything to use plain old JS arrays to maintain consistency, and I don't think that's worth it/ Let's keep the simple fix for this PR imo.

Sources/macros.js
```diff
@@ -174,6 +174,20 @@ function safeArrays(model) {
   });
 }
 
+function isTypedArray(value) {
+  return (
```
floryst, line 178: If we replace this with `Object.values(TYPED_ARRAYS).some((ctor) => value instanceof ctor)`, we can handle `Big[I|Ui]nt64Array` as well.

floryst:
The image direction matrices are a consequence of using gl-matrix for construction and allocation, since it defaults to Float32Arrays. I partially agree with @finetjul that the index/world transform matrices should be hidden from serialization, but having them is nice for downstream code to not have to always recompute the transforms upon deserialization (e.g. sending over the network).

Julesdevops:
Thanks for review and feedback guys!  
I don't know what's better. I can open an issue to keep track of this discussion. Tbh I just want to get https://github.com/Kitware/glance/issues/480 fixed in this PR

finetjul:
Let's leave it as is.

PR #2928, opened 2023-09-29
Fix calculator for image data

### LINKED ISSUE #2257: vtkCalculator fails with vtkImageData as input
### High-level description
vtkCalculator does not work on vtkImageData because vtkImageData does not have a `getPoints` method.

### Steps to reproduce
1. Use vtkImageData as an input to vtkCalculator.
2. Call `setFormulaSimple`, e.g.

```
calculator.setFormulaSimple(
    FieldDataTypes.POINT,
    ['scalars'],
    'mask',
    value => value > 10 ? 1 : 0
  );
```

### Detailed behavior
`getPoints` is called in two locations in `prepareArrays` that cause problems for the above scenario:

1. [line 128](https://github.com/Kitware/vtk-js/blob/544252718e0e070b842b2da7b8eea37c1ff1aa62/Sources/Filters/General/Calculator/index.js#L125): `arraysIn.push(inData.getPoints());`
(This is due to the coordinate data that gets added in `augmentInputArrays`)

2. [line 194](https://github.com/Kitware/vtk-js/blob/544252718e0e070b842b2da7b8eea37c1ff1aa62/Sources/Filters/General/Calculator/index.js#L194): `(x) => x.getPoints().getNumberOfPoints(),`

### Expected behavior
I was able to get it working for my use case (applying a simple threshold to some image data) by placing checks before the `getPoints` calls:

1. `if (inData.getPoints) arraysIn.push(inData.getPoints());`
2. `(x) => x.getPoints ? x.getPoints().getNumberOfPoints() : x.getNumberOfPoints()`

Not sure if this is the preferred solution though.

### Environment
- **vtk.js**: v22.2.2
- **OS**: Ubuntu-18.04 VM running on Windows 10
- **Browser**: Chrome Version 97.0.4692.99 (Official Build) (64-bit) (Windows)

Julesdevops:
### Context
fix #2257

### Results
we can now use `vtkCalculator` with a `vtkImageData`

### Changes

Do not pass point coordinates to formula function in case of `vtkImageData` input since it relies on the `getPoints` function which does not exist on `vtkImageData`.
Also avoid to call `getPoints` to get number of points, directly use `getNumberOfPoints` on the input dataset.

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: linux
  - **Browser**: firefox 118.0.1, chrome Version 117.0.5938.132 (Official Build) (64-bit)

### DIFF
--- a/Sources/Filters/General/Calculator/test/testCalculator.js
+++ b/Sources/Filters/General/Calculator/test/testCalculator.js
@@ -1,6 +1,7 @@
 import test from 'tape-catch';
 
 import vtkCalculator from 'vtk.js/Sources/Filters/General/Calculator';
+import vtkImageGridSource from 'vtk.js/Sources/Filters/Sources/ImageGridSource';
 import vtkPlaneSource from 'vtk.js/Sources/Filters/Sources/PlaneSource';
 import { AttributeTypes } from 'vtk.js/Sources/Common/DataModel/DataSetAttributes/Constants';
 import { FieldDataTypes } from 'vtk.js/Sources/Common/DataModel/DataSet/Constants';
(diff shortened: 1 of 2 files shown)


Julesdevops:
fyi @bruyeret @finetjul

Sources/Filters/General/Calculator/test/testCalculator.js
```diff
@@ -93,3 +94,17 @@ test('Test vtkCalculator execution', (t) => {
...
+  const source = vtkImageGridSource.newInstance();
+  const filter = vtkCalculator.newInstance();
+
+  filter.setInputConnection(source.getOutputPort());
+  filter.setFormulaSimple(FieldDataTypes.POINT, ['scalars'], 'mask', (value) =>
+    value > 10 ? 1 : 0
+  );
+
+  source.update();
+  filter.update();
```
finetjul, line 108: should you also check if the output of the filter contains a data object with the same attributes as the input data object ?

finetjul (approved):
LGTM

Sources/Filters/General/Calculator/test/testCalculator.js
Julesdevops: definitely. I updated the test

floryst (approved):
LGTM

PR #2933, opened 2023-10-12
Fix ImageReslice direction matrix and quick refactor

bruyeret:
Direction of input image was not correctly supported by the ImageReslice
Direction of output image was not supported at all
Use `getWorldToIndex` and `getIndexToWorld` instead of trying to compute the matrix by hand

Fix the directions and refactor the code to make it clearer (remove `optimizedTransform`, initialize `indexMatrix`)


bruyeret:
I brought back `optimizedTransform` @finetjul 
The refactor only use `getIndexToWorld` and removes the identity check

PR #2935, opened 2023-10-17
Add check for binding a representation in View2DProxy

bruyeret:
The argument of `bindRepresentationToManipulator` should be a proxy
Add a check to ensure that it is indeed a proxy in `addRepresentation`

Also add a test to avoid regression


finetjul (approved):
LGTM but I do not know much of proxies...

Sources/Proxy/Core/View2DProxy/index.js
```diff
@@ -124,7 +124,9 @@ function vtkView2DProxy(publicAPI, model) {
     if (rep.setSlicingMode) {
       rep.setSlicingMode('XYZ'[model.axis]);
     }
-    publicAPI.bindRepresentationToManipulator(rep);
+    if (rep.getProxyId) {
```
jadh4v, line 127: Might be better to add this check inside the `bindRepresentationToManipulator()` call, so that it is handled in other places where it is called, such as line 140.

bruyeret:
I moved the check in `bindRepresentationToManipulator()` as you suggested @jadh4v

PR #2957, opened 2023-11-13
Fix/volume picker

rodrigobasilio2022:
### Context

This PR fixes the Volume Picker in setups where the ray length is large (Cornerstone3D / OHIF). It reimplements the insertectWithLine function of the VTK vtkBox object into vtkBoundingBox, allowing the picker functions to calculate the intersect points (q1 and q2) of the ray into the VolumeActor bounding box and use them to limit the ray

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


Sources/Common/DataModel/BoundingBox/index.d.ts
```diff
@@ -588,6 +588,14 @@ declare class BoundingBox {
    */
   intersects(bounds: Bounds, bBounds: Bounds): boolean;
 
+  /**
+   * Intersect box with line and return the parametric values and points of the two intercepts
+   * @param bounds 
+   * @param p1 
+   * @param p2 
+   */
+  intersectBoundingBoxWithLine(bounds: Bounds, p1: Vector3, p2: Vector3) : object;
```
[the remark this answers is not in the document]
rodrigobasilio2022: Yes, but it returns the parametric values of the intercept points

Sources/Common/DataModel/BoundingBox/index.d.ts
```diff
@@ -588,6 +588,14 @@ declare class BoundingBox {
    */
   intersects(bounds: Bounds, bBounds: Bounds): boolean;
 
+  /**
+   * Intersect box with line and return the parametric values and points of the two intercepts
+   * @param bounds 
+   * @param p1 
+   * @param p2 
+   */
```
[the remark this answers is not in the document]
rodrigobasilio2022: Done

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -198,14 +199,28 @@ function vtkCellPicker(publicAPI, model) {
...
     } else if (mapper.isA('vtkVolumeMapper')) {
-      tMin = publicAPI.intersectVolumeWithLine(
-        p1,
-        p2,
-        clipLine.t1,
-        clipLine.t2,
-        tol,
-        actor
-      );
+      const bbox = vtkBoundingBox.newInstance();
```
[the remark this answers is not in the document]
rodrigobasilio2022: Done

Sources/Common/DataModel/BoundingBox/index.d.ts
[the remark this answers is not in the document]
finetjul: Fair enough, then you need instead to look at [vtkBox.IntersectWithLine](https://gitlab.kitware.com/vtk/vtk/-/blob/master/Common/DataModel/vtkBox.h?ref_type=heads#L103).
Would you mind adding such method in `vtkBox` instead of `vtkBoundingBox` (in order to keep the API as similar to VTK as possible?)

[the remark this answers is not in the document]
rodrigobasilio2022: Done

rodrigobasilio2022:
@finetjul Done the changes. All checks passed

rodrigobasilio2022:
@finetjul Please dont merge. Found a problem here in import

rodrigobasilio2022:
@finetjul How should I call the intersectWithLine in Box object?

finetjul:
> @finetjul How should I call the intersectWithLine in Box object?

Please do something similar to vtkBoundingBox, with a static function that gets exposed publicly

rodrigobasilio2022:
@finetjul Done the changes

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -198,14 +199,33 @@ function vtkCellPicker(publicAPI, model) {
         model.pCoords = pickData.pCoords;
       }
     } else if (mapper.isA('vtkVolumeMapper')) {
-      tMin = publicAPI.intersectVolumeWithLine(
+      // we calculate here the parametric intercept points between the ray and the bounding box, so
+      // if the application defines for some reason a too large ray length (1e6), it restrict the calculation
```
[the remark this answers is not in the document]
rodrigobasilio2022: Very large ray length (2e6) are not problem for parameteric formulas, but in the old implementation of the volume intersect, it was giving 100 steps no matter the size of the ray, of course leading to problems transversing the ray. The current implementation takes the exact size of the ray inside the volume bounding box and use this size to calculate the correct step interval to go voxel by voxel testing the opacity

rodrigobasilio2022:
@finetjul Added the changes requested

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -317,17 +335,20 @@ function vtkCellPicker(publicAPI, model) {
...
+      x2[i] = (q2[i] - origin[i]) / spacing[i];
     }
     const x = [0, 0, 0, 0];
     const xi = [0, 0, 0];
 
     const sliceSize = dims[1] * dims[0];
     const rowSize = dims[0];
-    const nSteps = 100;
+    // here the step is the 1 over the distance between volume index location x1 and x2
+    const step = 1 / Math.sqrt(vtkMath.distance2BetweenPoints(x1, x2));
```
finetjul, line 347: shouldn't step be not smaller than something that correspond to the size of a voxel ?

Sources/Common/DataModel/Box/index.d.ts
```diff
@@ -26,6 +31,16 @@ export interface vtkBox extends vtkObject {
...
 
+    /**
+     * Intersect box with line and return the parametric values and points of the two intercepts
+     * @param bounds 
+     * @param p1 
+     * @param p2 
+     * returns @object IBoxIntersections {t1, t2, x1, x2} object containing the t1, t2 parametric values and
+     * x1, x2 coordinates of the line intercept points in the bounding box
+     */
+    intersectWithLine(p1: Vector3, p2: Vector3) : IBoxIntersections;
```
finetjul, line 42: you said earlier than it can return undefined, so you need to add that possibility

Sources/Common/DataModel/Box/index.d.ts
```diff
@@ -26,6 +31,16 @@ export interface vtkBox extends vtkObject {
      */
     evaluateFunction(x: Vector3): number;
 
+    /**
+     * Intersect box with line and return the parametric values and points of the two intercepts
+     * @param bounds 
+     * @param p1 
+     * @param p2 
+     * returns @object IBoxIntersections {t1, t2, x1, x2} object containing the t1, t2 parametric values and
+     * x1, x2 coordinates of the line intercept points in the bounding box
```
finetjul, line 40: can return undefined

Sources/Rendering/Core/CellPicker/index.js
rodrigobasilio2022: This step is related to the percentage of the size of the distance between x1 and x2. It increments t the parametric value

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -198,14 +199,31 @@ function vtkCellPicker(publicAPI, model) {
...
+            : interceptionObject.t1;
+
+        t2 =
+          interceptionObject.t2 > clipLine.t2
+            ? clipLine.t2
+            : interceptionObject.t2;
+      } else {
+        t1 = clipLine.t1;
+        t2 = clipLine.t2;
+      }
```
finetjul, line 224: if interserctWithLine returned null instead of undefined, then you could simply write:
```
t1 = Math.max(interceptionObject?.t1, clipLine.t1)
t2 = Math.min(interceptionObject?.t2, clipLine.t2)
```

Sources/Rendering/Core/CellPicker/index.js
```diff
@@ -317,17 +335,20 @@ function vtkCellPicker(publicAPI, model) {
...
-    const nSteps = 100;
+    // here the step is the 1 over the distance between volume index location x1 and x2
+    const step = 1 / Math.sqrt(vtkMath.distance2BetweenPoints(x1, x2));
     let insideVolume;
-    for (let t = t1; t < t2; t += 1 / nSteps) {
+    // here we reinterpret the t value as the distance between x1 and x2
+    // When calculating the tMin, we weight t between t1 and t2 values
+    for (let t = 0; t < 1; t += step) {
       // calculate the location of the point
       insideVolume = true;
```
finetjul, line 353: it seems that `insideVolume` will always be equal to `true`. How useful is this variable ?

rodrigobasilio2022: If out of bounds in one dimension it could be false and the code didnt consider the particular  volume index coordinate

Sources/Rendering/Core/CellPicker/index.js
rodrigobasilio2022: I think no because even if interceptionObject is null, interceptionObject?.t1 will be undefined, not null, leading to t1 = undefined

finetjul: Fair enough, then:

```
t1 = interceptionObject?.t1 > clipLine.t1 ? interceptionObject.t1 : clipLine.t1
t2 = interceptionObject?.t2 < clipLine.t2 ? interceptionObject.t2 : clipLine.t2
```

Sources/Rendering/Core/CellPicker/index.js
finetjul: my bad, I didn't read the code afterwards

finetjul (approved):
LGTM

PR #2960, opened 2023-11-21
feat(labelOutline): add per segment label outline thickness

sedghi:
### Context

This pull request will introduce a method to specify individual segment outline thicknesses for the label map outline.

### Results

It will look like this with different outline thickness for each segment.

CleanShot 2023-11-21 at 16 18 41

### Changes

The signature of the `setLabelOutlineThickness` has been changed to support both a single number (old API) and a new way of handling it (via an array).

```js
setLabelOutlineThickness(2) => will set 2px for all segments
setLabelOutlineThickness([2]) => will set 2px for all segments
setLabelOutlineThickness([1,3]) => will set 1px for segment index 1  and 3 px for segment index 2 
```

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: Latest
  - **OS**: MacOS 13.5.2 (22G91)
  - **Browser**:  Chrome Version 119.0.6045.159 (Official Build) (arm64)

### DIFF
--- a/Sources/Rendering/Core/VolumeProperty/index.d.ts
+++ b/Sources/Rendering/Core/VolumeProperty/index.d.ts
@@ -204,11 +204,20 @@ export interface vtkVolumeProperty extends vtkObject {
 	 */
 	setIndependentComponents(independentComponents: boolean): boolean;
 
-  /**
-   *
-   * @param {Number | Number[]} labelOutlineThickness
-  */
-  setLabelOutlineThickness(labelOutlineThickness: number | number[]): boolean;
+	/**
+	 * It will set the label outline thickness for the labelmaps. It can accept
+	 * a single number or an array of numbers. If a single number is provided,
+	 * it will be used for all the components. If an array is provided, it indicates
+	 * the thickness for each segment index. For instance if you have a labelmap
+	 * with 3 segments (0: background 1: live 2: tumor), you can set the thickness
+	 * to [2,4] to have a thicker outline for the tumor (thickness 4). It should be 
+	 * noted that if the number of segments is greater than the number of thickness
+	 * the first thickness will be used for the remaining segments.
+	 * 
+	 * 
+	 * @param {Number | Number[]} labelOutlineThickness
+	 */
+	setLabelOutlineThickness(labelOutlineThickness: number | number[]): boolean;
 
 
 	/**
--- a/Sources/Rendering/Core/VolumeProperty/index.d.ts
+++ b/Sources/Rendering/Core/VolumeProperty/index.d.ts
@@ -207,13 +207,12 @@ export interface vtkVolumeProperty extends vtkObject {
 	/**
 	 * It will set the label outline thickness for the labelmaps. It can accept
 	 * a single number or an array of numbers. If a single number is provided,
-	 * it will be used for all the components. If an array is provided, it indicates
+	 * it will be used for all the segments. If an array is provided, it indicates
 	 * the thickness for each segment index. For instance if you have a labelmap
-	 * with 3 segments (0: background 1: live 2: tumor), you can set the thickness
+	 * with 3 segments (0: background 1: liver 2: tumor), you can set the thickness
 	 * to [2,4] to have a thicker outline for the tumor (thickness 4). It should be 
-	 * noted that if the number of segments is greater than the number of thickness
-	 * the first thickness will be used for the remaining segments.
-	 * 
+	 * noted that the thickness is in pixel and also the first array value will 
+	 * control the default thickness for all the segments if not specified.
 	 * 
 	 * @param {Number | Number[]} labelOutlineThickness
 	 */
--- a/Sources/Rendering/Core/VolumeProperty/index.d.ts
+++ b/Sources/Rendering/Core/VolumeProperty/index.d.ts
@@ -204,11 +204,12 @@ export interface vtkVolumeProperty extends vtkObject {
 	 */
 	setIndependentComponents(independentComponents: boolean): boolean;
 
-	/**
-	 *
-	 * @param {Number} labelOutlineThickness 
-	 */
-	setLabelOutlineThickness(labelOutlineThickness: number): boolean;
+  /**
+   *
+   * @param {Number | Number[]} labelOutlineThickness
+  */
+  setLabelOutlineThickness(labelOutlineThickness: number | number[]): boolean;
+
 
 	/**
 	 *
--- a/Sources/Rendering/Core/VolumeProperty/index.d.ts
+++ b/Sources/Rendering/Core/VolumeProperty/index.d.ts
@@ -212,7 +212,7 @@ export interface vtkVolumeProperty extends vtkObject {
 	 * with 3 segments (0: background 1: liver 2: tumor), you can set the thickness
 	 * to [2,4] to have a thicker outline for the tumor (thickness 4). It should be 
 	 * noted that the thickness is in pixel and also the first array value will 
-	 * control the default thickness for all the segments if not specified.
+	 * control the default thickness for all labels when 0 or not specified.
 	 * 
 	 * @param {Number | Number[]} labelOutlineThickness
 	 */
--- a/Sources/Rendering/Core/VolumeProperty/index.js
+++ b/Sources/Rendering/Core/VolumeProperty/index.js
@@ -203,6 +203,17 @@ function vtkVolumeProperty(publicAPI, model) {
   publicAPI.getInterpolationTypeAsString = () =>
     macro.enumToString(InterpolationType, model.interpolationType);
 
+  publicAPI.setLabelOutlineThickness = (thickness) => {
+    // handle the old API which was a single number for all segments
+    if (typeof thickness === 'number') {
+      model.labelOutlineThickness = [thickness];
+    } else {
+      model.labelOutlineThickness = thickness;
+    }
+
+    publicAPI.modified();
+  };
+
   const sets = [
     'useGradientOpacity',
     'scalarOpacityUnitDistance',
@@ -252,7 +263,7 @@ const DEFAULT_VALUES = {
   specular: 0.2,
   specularPower: 10.0,
   useLabelOutline: false,
-  labelOutlineThickness: 1,
+  labelOutlineThickness: [1],
   labelOutlineOpacity: 1.0,
 };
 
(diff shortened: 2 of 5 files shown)


Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
```diff
@@ -43,7 +43,7 @@ varying vec3 vertexVCVSOutput;
 //VTK::ImageLabelOutlineOn
 
 #ifdef vtkImageLabelOutlineOn
-uniform int outlineThickness;
+uniform int outlineThickness[256];
```
sedghi, line 46: I'm not sure if this is correct or not, but I believe we need to have a fixed size for uniform arrays. I think 256 segments might be enough, but maybe we should target a higher number?

finetjul:
> `setLabelOutlineThickness([1,3]) => will set 1px for segment index 1 (value 1) and 3 px for segment index 3 (value 3)`

and 3px for segment index 3, right ?

sedghi:
> > `setLabelOutlineThickness([1,3]) => will set 1px for segment index 1 (value 1) and 3 px for segment index 3 (value 3)`
> 
> and 3px for segment index 2, right ?

correct, typo

Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl
finetjul: Maybe it can be uint8 ? 
1024 should be high enough without taking too much space...

Sources/Rendering/Core/VolumeProperty/index.d.ts
```diff
@@ -204,11 +204,12 @@ export interface vtkVolumeProperty extends vtkObject {
...
 
-	/**
-	 *
-	 * @param {Number} labelOutlineThickness 
-	 */
-	setLabelOutlineThickness(labelOutlineThickness: number): boolean;
+  /**
+   *
+   * @param {Number | Number[]} labelOutlineThickness
+  */
```
finetjul, line 210: wrong indent

sedghi:
@finetjul The value is referring to the segment value, sorry for the awful example and confusion

Sources/Rendering/Core/VolumeProperty/index.js
```diff
@@ -203,6 +203,17 @@ function vtkVolumeProperty(publicAPI, model) {
   publicAPI.getInterpolationTypeAsString = () =>
     macro.enumToString(InterpolationType, model.interpolationType);
 
+  publicAPI.setLabelOutlineThickness = (thickness) => {
+    // handle the old API which was a single number for all segments
+    if (typeof thickness === 'number') {
+      model.labelOutlineThickness = [thickness];
+    } else {
+      model.labelOutlineThickness = thickness;
```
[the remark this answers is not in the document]
sedghi: Yeah this part I was not sure about. I'll sure the setGetArray and see if it works

sedghi:
@finetjul I updated the code. I couldn't set the length to 1024 for the thickness array. I was getting this error. It seems like I can only put up to 983 for some reason, but nothing more. Do you know what is happening?

CleanShot 2023-11-24 at 08 14 59

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -1071,7 +1071,9 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
       const labelOutlineOpacity = actor.getProperty().getLabelOutlineOpacity();
 
       // Create an array of size 256 and make sure it is padded with zeros
```
finetjul, line 1073: Create a large enough array and make sure it is padded with zeros

Sources/Rendering/Core/VolumeProperty/index.d.ts
```diff
@@ -204,11 +204,20 @@ export interface vtkVolumeProperty extends vtkObject {
...
-  setLabelOutlineThickness(labelOutlineThickness: number | number[]): boolean;
+	/**
+	 * It will set the label outline thickness for the labelmaps. It can accept
+	 * a single number or an array of numbers. If a single number is provided,
+	 * it will be used for all the components. If an array is provided, it indicates
+	 * the thickness for each segment index. For instance if you have a labelmap
+	 * with 3 segments (0: background 1: live 2: tumor), you can set the thickness
+	 * to [2,4] to have a thicker outline for the tumor (thickness 4). It should be 
+	 * noted that if the number of segments is greater than the number of thickness
+	 * the first thickness will be used for the remaining segments.
```
finetjul, line 215: how about having the first array value control the "default thickness" ?

Your condition `if the number of segments is greater than the number of thickness the first thickness will be used for the remaining segments` would otherwise be an issue if you want to "emphasize" the first label (because it is the current label) without impacting the other labels that rely on the default value.

I also wonder if the label of value 0 is ALWAYS background hence not visible. Maybe there could be usefulness to have an actual representation (therefore thickness) for the background value (label=0)

Sources/Rendering/Core/VolumeProperty/index.d.ts
```diff
@@ -204,11 +204,20 @@ export interface vtkVolumeProperty extends vtkObject {
...
-   *
-   * @param {Number | Number[]} labelOutlineThickness
-  */
-  setLabelOutlineThickness(labelOutlineThickness: number | number[]): boolean;
+	/**
+	 * It will set the label outline thickness for the labelmaps. It can accept
+	 * a single number or an array of numbers. If a single number is provided,
+	 * it will be used for all the components. If an array is provided, it indicates
+	 * the thickness for each segment index. For instance if you have a labelmap
+	 * with 3 segments (0: background 1: live 2: tumor), you can set the thickness
```
finetjul, line 212: 1: live -> 1: liver

finetjul:
> @finetjul I updated the code. I couldn't set the length to 1024 for the thickness array. I was getting this error. It seems like I can only put up to 983 for some reason, but nothing more. Do you know what is happening?
> 
> CleanShot 2023-11-24 at 08 14 59

I guess it is because large arrays are not meant to be passed as uniforms, but instead as textures. @bruyeret ?

bruyeret:
> > @finetjul I updated the code. I couldn't set the length to 1024 for the thickness array. I was getting this error. It seems like I can only put up to 983 for some reason, but nothing more. Do you know what is happening?
> > CleanShot 2023-11-24 at 08 14 59
> 
> I guess it is because large arrays are not meant to be passed as uniforms, but instead as textures. @bruyeret ?

Yes there is a limit to the number of locations per program: `GL_MAX_UNIFORM_LOCATIONS`, and arrays take as many locations as there are elements.

Using textures is a good idea 👍

sedghi:
@bruyeret But adding a texture will reduce the number of textures that we can use in the shader no?

sedghi:
Thanks I will update the PR with texture approach

Sources/Rendering/Core/VolumeProperty/index.d.ts
sedghi: I updated the doc. 
I'm not entirely certain about the background thickness, as it can be a bit challenging around the edges, I suppose. I am not modifying the outline thickness behaviour for background in this PR, so perhaps we could consider waiting for that specific scenario and create a separate PR for adjusting the background's outline thickness?

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -1584,6 +1593,77 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
+        model._openGLRenderWindow.setGraphicsResourceForObject(
+          labelOutlineThicknessArray,
+          model.labelOutlineThicknessTexture,
+          model.labelOutlineThicknessTextureString
+        );
+      }
+    } else {
+      model.labelOutlineThicknessTexture = lTex.vtkObj;
+      model.labelOutlineThicknessTextureString = lTex.hash;
+    }
```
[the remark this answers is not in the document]
sedghi: you mean just the else or the whole if-else statement?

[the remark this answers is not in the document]
finetjul: the whole if-else statement. no ?

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -1067,10 +1074,10 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
       const labelOutlineThickness = actor
         .getProperty()
         .getLabelOutlineThickness();
+      const maxThickness = Math.max(...labelOutlineThickness);
```
finetjul, line 1077: use reduce here as well

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -1682,6 +1693,80 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
 
     model.VBOBuildTime.modified();
   };
+
+  publicAPI.updateLabelOutlineThicknessTexture = (actor) => {
```
finetjul, line 1697: technically it is not an `actor` but a `volume`

finetjul (approved):
LGTM.
@bruyeret WDYT ?

bruyeret (approved):
Looks good to me!
One non-blocking thing: why normalizing the array?
A limitation of label outline thickness to 255px doesn't seem bad to me

sedghi:
@bruyeret You are right, I removed it

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -1682,6 +1687,75 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
...
     model.VBOBuildTime.modified();
   };
+
+  publicAPI.updateLabelOutlineThicknessTexture = (volume) => {
+    const labelOutlineThicknessArray = volume
+      .getProperty()
+      .getLabelOutlineThickness();
+
+    if (!labelOutlineThicknessArray) {
+      return;
```
finetjul, line 1697: This means that if you reset an existing array, it won't be removed from the GPU.
Shouldn't it be released from GPU memory?

What behavior do you expect in the shader ?

sedghi: it is not about reset I think, it is either it is defined or not. I tried changing it and it worked fine

https://github.com/Kitware/vtk-js/assets/7490180/4dc6d881-2a83-4073-bf94-2292fb78aaea

finetjul: When you start you have an array, what happens if you call: `setLabelOutlineThicknessArray(null)` ?

sedghi: removed that unnecessary check and let the toString handle it

finetjul (approved):
LGTM

sedghi:
@finetjul Can we merge this if approved please?

PR #2965, opened 2023-12-12
fix(TubeFilter): do not drop output arrays

floryst:
### Context

Output cell/point arrays were being dropped when they shouldn't be. Every update invocation creates a new polydata array, so there are no old arrays to be dropped.

Related: https://discourse.vtk.org/t/tubefilter-broken-for-celldata/12863

### Results

Cell and point data arrays should be properly copied to the output polydata.

### Changes

- [ ] Fix TubeFilter not copying cell/pointdata properly

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


sankhesh (approved):
LGTM

PR #2966, opened 2023-12-14
feat(rendering): add labelmap outline thickness for image mappers

sedghi:
### Context

We have previously introduced the [volume outline](https://kitware.github.io/vtk-js/examples/VolumeOutline.html). Moreover, I have[ recently implemented the feature](https://github.com/Kitware/vtk-js/pull/2960) that allows defining per segment outline thickness. This new pull request aims to accomplish a similar task for imageMappers. Specifically, it will: 
1) Enable the display of outlines for the outline of a segmentation in image mapper.
2) Provide the ability to define per segment outline thickness in pixels.

### Results

With this new addition, we can now apply segmentations on imageMapper with varying border thickness.

image

To test run `npm run example ImageLabelOutline`

### Changes

- relevant changes to the image mapper and shader code
- New ImageLabelOutline example
- New testing 

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  latest
  - **OS**:  macOS 13.6.3 (22G436)
  - **Browser**:  Version 120.0.6099.109 (Official Build) (arm64)


Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -37,6 +37,18 @@ function computeFnToString(property, pwfun, numberOfComponents) {
...
+function splitStringOnEnter(inputString) {
+  // Split the input string into an array of lines based on "Enter" (newline) characters
+  const lines = inputString.split('\n');
+
+  // Remove any leading or trailing whitespace from each line and filter out empty lines
+  const trimmedLines = lines
+    .map((line) => line.trim())
+    .filter((line) => line.length > 0);
+
+  return trimmedLines;
```
finetjul, line 49: while this is nice to read, using split, map and filter is slow.
I think the following would be way faster:
```
const trimmedLines = [];

for (const line of inputString.split('\n')) {
  const trimmedLine = line.trim();

  if (trimmedLine.length > 0) {
    trimmedLines.push(trimmedLine);
  }
}

return trimmedLines;
```

Sources/Rendering/Core/ImageProperty/index.js
```diff
@@ -169,8 +172,12 @@ export function extend(publicAPI, model, initialValues = {}) {
     'diffuse',
     'opacity',
     'useLookupTableScalarRange',
+    'useLabelOutline',
+    'labelOutlineOpacity',
   ]);
 
+  macro.setGetArray(publicAPI, model, ['labelOutlineThickness']);
```
finetjul, line 179: Please update documentation accordingly

Sources/Rendering/OpenGL/ImageMapper/index.js
sedghi: When I follow your recommendation, I encounter this issue. Perhaps you could consider turning off the eslint rule if you prefer this approach.

CleanShot 2023-12-21 at 09 07 58

Sources/Rendering/Core/ImageProperty/index.js
sedghi: Done

Sources/Rendering/OpenGL/ImageMapper/index.js
finetjul: My bad, then let's do it simpler :

```
const lines = inputString.split('\n');
const trimmedLines = [];
for (let i = 0; i < lines.length; ++i) {
  const trimmedLine = lines[i].trim();
  if (trimmedLine.length > 0) {
    trimmedLines.push(trimmedLine);
  }
}
return trimmedLines;
```

sedghi: Done

PR #2967, opened 2023-12-14
fix(macro): fix shift + mousewheel on mac by using spinX values when …

### LINKED ISSUE #2834: [Bug] - Unable to zoom when holding shift when scrolling with MouseWheel
### Bug description

Zoom interactor does not work with `scrollEnabled: true` and `shift: true` when rolling a mousewheel.  All other variants work in terms of modifier keys.  Specifically shift + a mouse wheel interaction.  Effects both zoom and zoomToMouse.

### Steps to reproduce

https://kitware.github.io/vtk-js/examples/InteractorStyleManipulator.html

image

Go to the above demo, set the right most scroll column to as seen in the screenshot.  You'll see when shift + mousewheel scroll is applied, nothing happens.  It correctly zooms for the other modifier key combinations ctrl + mousewheel and alt + mousewheel.

**Please note this works when 2-finger scrolling with say a touchpad (macbook).  It will not work when rolling a mousewheel.  All other modifier keys with mousewheel work: `ctrl` `alt` etc.**

### Detailed Behavior

_No response_

### Expected Behavior

Zoom occurs

### Environment

Whatever is latest deployed on https://kitware.github.io/vtk-js/examples/InteractorStyleManipulator.html

DevinAPorro:
…spinY is 0

Shift + mousewheel on mac is used for "horizontal scrolling". Per comment in #2834, set spinY and pixelY in normalizeWheel to spinX/pixelX when y values are 0

re #2834

### Context

Fix #2834 . Shift + mousewheel on mac doesn't work.

### Results

Shift + mousewheel on mac now works because spinX and pixelX are set to the spinY and pixelY values when spinX/pixelX are 0 or undefined. Before change Shift + mousewheel on mac does not scroll. After fix, scrolling works correctly.

### Changes

1. Updated macros.js `normalizeWheel` function per this [comment](https://github.com/Kitware/vtk-js/issues/2834#issuecomment-1584661716) in #2834.

### PR and Code Checklist

### Testing

- [X] Tested environment:
  - **vtk.js**: master (28.10.0)
  - **OS**: macOS 14 Sonoma
  - **Browser**: Chrome 120.0.6099.109

### DIFF
--- a/Sources/Testing/testMacro.js
+++ b/Sources/Testing/testMacro.js
@@ -650,11 +650,51 @@ test('Macro methods keystore tests', (t) => {
   t.end();
 });
 
+test('Macro methods normalizeWheel tests', (t) => {
+  const wheelEvent = { wheelDeltaX: 120 };
+
+  let normalizeWheelReturn = macro.normalizeWheel(wheelEvent);
+  t.equal(
+    normalizeWheelReturn.spinY,
+    -1,
+    'spinY should be set when only wheelDeltaX is defined'
+  );
+  t.equal(
+    normalizeWheelReturn.spinY,
+    normalizeWheelReturn.spinX,
+    'spinY should equal spinX when only wheelDeltaX is defined'
+  );
+  t.equal(
+    normalizeWheelReturn.pixelY,
+    -10,
+    'pixelY should be set when only wheelDeltaX is defined'
+  );
+  t.equal(
+    normalizeWheelReturn.pixelY,
+    normalizeWheelReturn.pixelX,
+    'pixelY should equal pixelX when only wheelDeltaX is defined'
+  );
+
+  wheelEvent.wheelDeltaY = 240;
+  normalizeWheelReturn = macro.normalizeWheel(wheelEvent);
+  t.equal(
+    normalizeWheelReturn.spinY,
+    -2,
+    'spinY should be set using wheelDeltaY when defined'
+  );
+  t.equal(
+    normalizeWheelReturn.pixelY,
+    -20,
+    'pixelY should be set using wheelDeltaY when defined'
+  );
+
+  t.end();
+});
+
 test('Macro debounce can be cancelled', (t) => {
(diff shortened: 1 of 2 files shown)


DevinAPorro:
I know this change seems too simple...but I think this is all there is to it (it does fix Shift + mousewheel scrolling on the mac), but please let me know if there are additional changes that should be made. I was mainly going off [this comment](https://github.com/Kitware/vtk-js/issues/2834#issuecomment-1584661716) in the bug, but I didn't want to remove the `spinX` / `pixelX` properties entirely in case someone uses them in the future.

finetjul:
My only concern is "if one day" an interactor needs to support spinX AND spinY. You wouldn't be able to get the actual value of both spins.
But maybe that "if one day" won't show up so we should keep it simple like you suggest.
@floryst  WDYT ?

DevinAPorro:
> My only concern is "if one day" an interactor needs to support spinX AND spinY. You wouldn't be able to get the actual value of both spins. But maybe that "if one day" won't show up so we should keep it simple like you suggest. @floryst WDYT ?

I see what you mean, but since the suggestion was to merge the spinX and spinY values, I thought it sounded like there were no plans to use spinY in the near future? The only thing I can think of offhand is to add a flag or something similar to indicate when spinX is set to the spinY value, although I'm probably missing something obvious :) . If anyone ends up needing spinY (and doesn't want it "mirrored" to spinX), a flag (or some other way of handling) should be easy to add...but maybe we add it if/when it's needed? Happy to make changes as well though, just let me know. Thanks!

finetjul:
@DevinAPorro let's leave it like that. It can be modified/improved in the future. 
Can you please create a test to make sure the next person that works on that does not break the use case you fixed with this PR.

DevinAPorro:
@finetjul I added a simple unit test to cover the functionality I modified. Please let me know if it is sufficient, or if you would like me to add additional tests.

PR #2976, opened 2024-01-04
Add typescript definitions

bruyeret:
Main addition: ImageReslice typescript header


finetjul (approved):

bruyeret:
This merge request did not trigger a release
From the logs of [this action](https://github.com/Kitware/vtk-js/actions/runs/7415010723/job/20177270450):
```There are no relevant changes, so no new version is released.```
I think that it is because my commit commit message starts with `docs`
How can I trigger a release anyway @floryst?

floryst:
`docs()` commit messages don't trigger a release. `fix()` is probably best for ts definition changes that need to be released.

The easiest solution is to wait for a merge that contains a release-able commit. I'm not familiar with manually cutting a release, though I suppose it could be possible.

Issue #2978, opened 2024-01-05
[Bug] Likely memory leak when using vtkHttpDataSetReader


dominik-werner-casra:
### Bug description

When using vtk.js in a SPA where the displayed CT image data is changed without refreshing the page, the browser's memory usage continuously increases until parts of the page stop working.
As a workaround we have removed the usage of ```vtkHttpDataSetReader.setUrl()``` from our code and replaced it with our own implementation which omits the problem. However, our own method is not as feature rich as ```vtkHttpDataSetReader```.

### Steps to reproduce

To reproduce this problem, we have created a very simple nodejs application with two different pages:
- **http://localhost:8080/example_with_issue.html**: This uses the ```vtkHttpDataSetReader.setUrl()``` method and causes the memory to be filled while the sequence runs.
- **http://localhost:8080/example_with_workaround**: This uses the custom method to load data and allows the browser to dispose old data.

To run the examples extract the zip archive and in the folder run:
```bash
node server.js
```

The server uses samples from **kitware.github.io** that are served in a way that the client cannot cache them in order to simulate many different CT images.
[vtk_memory_leak_demo.zip](https://github.com/Kitware/vtk-js/files/13842815/vtk_memory_leak_demo.zip)

### Detailed Behavior

When using vtk.js in a SPA where the displayed CT image data is changed without refreshing the page, the browser's memory usage continuously increases until parts of the page stop working, usually with an error similar to this:
```
RangeError: Array buffer allocation failed
    at new ArrayBuffer (<anonymous>)
    at new Uint16Array (<anonymous>)
    at r (<anonymous>:2:1727148)
    at Vk.e.create3DFromRaw (<anonymous>:2:1740847)
    at Vk.e.create3DFilterableFromRaw (<anonymous>:2:1741812)
    at Nz.e.buildBufferObjects (<anonymous>:2:2117370)
    at Nz.e.updateBufferObjects (<anonymous>:2:2115011)
    at Nz.e.renderPieceStart (<anonymous>:2:2112251)
    at Nz.e.renderPiece (<anonymous>:2:2114756)
    at Nz.e.volumePass (<anonymous>:2:2093181)
```
As a workaround we have removed the usage of ```vtkHttpDataSetReader.setUrl()``` from our code and replaced it with our own implementation which omits the problem:
```javascript
const NEW_ARRAY = {
    'Uint8Array': arr => arr,
[...]

PR #2979, opened 2024-01-05
fix: Added limit to array cache

dominik-werner-casra:
### Context
This PR addresses the issue described in #2978 where using ```vtkHttpDataSetReader.setUrl()``` multiple times without refreshing the page would cause the browser to consume more and more memory until the application crashes. Usually, the console would display an error similar to this:
```
RangeError: Array buffer allocation failed
    at new ArrayBuffer (<anonymous>)
    at new Uint16Array (<anonymous>)
    at r (<anonymous>:2:1727148)
    at Vk.e.create3DFromRaw (<anonymous>:2:1740847)
    at Vk.e.create3DFilterableFromRaw (<anonymous>:2:1741812)
    at Nz.e.buildBufferObjects (<anonymous>:2:2117370)
    at Nz.e.updateBufferObjects (<anonymous>:2:2115011)
    at Nz.e.renderPieceStart (<anonymous>:2:2112251)
    at Nz.e.renderPiece (<anonymous>:2:2114756)
    at Nz.e.volumePass (<anonymous>:2:2093181)
```

### Results
The changes in this PR will add a more sophisticated caching method that allows to define a limit for the cache size. Once the limit is reached, the cached array data that was accessed the furthest in the past will be deleted until the cache size no longer exceeds the configured limit. Additionally, caching can be disabled altogether for instance in a scenario where the same data is unlikely to be used multiple times.

### Changes
New methods that where added include the following:
```typescript
/**
* Gets if arrays are stored in cache.
*/
getCacheArrays(): boolean;

/**
* Sets if arrays are stpred in cache.
*/
setCacheArrays(value: boolean): void;

/**
* Gets the maximum size cached arrays are allowed to occupy.
* Size is given in MiB.
* If cache size is exceeded, the arrays that where not accessed
* the longest are removed.
*/
getMaxCacheSize(): number;

/**
* Sets the maximum size cached arrays are allowed to occupy.
* Size is given in MiB.
* If cache size is exceeded, the arrays that where not accessed
* the longest are removed.
*/
setMaxCacheSize(value: numnber): void;
```
The ```fetchArray()``` function was altered to:
- Store promises in the ```cachedPromises``` variable instead of ```cachedArrays```
- Remove all cached promises that where resolved
- Store new arrays in ```cachedArrays``` if caching is enabled
- Keep track of the access times of cache entries
- Remove cache entries if cache exceeds configured max size

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 29.3.2
  - **OS**: Windows 10
  - **Browser**: Chrome 120.0.6099.200


Sources/IO/Core/HttpDataSetReader/index.d.ts
```diff
@@ -148,6 +148,33 @@ export interface vtkHttpDataSetReader extends vtkHttpDataSetReaderBase {
...
+	 */
+	getMaxCacheSize(): number;
+
+	/**
+	 * Sets the maximum size cached arrays are allowed to occupy.
+	 * Size is given in MiB.
+	 * If cache size is exceeded, the arrays that where not accessed
+	 * the longest are removed.
+	 */
+	setMaxCacheSize(value: numnber): void;
```
finetjul, line 176: Maybe we could have:
 `maxCacheSize = undefined` => unlimited cache size
 `maxCacheSize = none` => no cache usage

That way you do not need to create a separate `cacheArrays` variable with accessors.

finetjul (commented):
That looks good, thanks for catching and fixing that tricky bug.

Can you please create a unit test as well (to make sure there is no regression in the future) ? You could do returned array reference comparison.

(when developping, you can do `t.only()` to speedup the debug).

dominik-werner-casra:
> That looks good, thanks for catching and fixing that tricky bug.
> 
> Can you please create a unit test as well (to make sure there is no regression in the future) ? You could do returned array reference comparison.
> 
> (when developping, you can do `t.only()` to speedup the debug).

Thank you very much for the quick review and the feedback! I am looking into it right now.

Sources/IO/Core/HttpDataSetReader/index.js
```diff
@@ -145,26 +147,59 @@ function vtkHttpDataSetReader(publicAPI, model) {
   // Internal method to fetch Array
   function fetchArray(array, options = {}) {
     const arrayId = `${array.ref.id}|${array.vtkClass}`;
-    if (!cachedArrays[arrayId]) {
+
+    // Check if array is present in cache
+    if (cachedArrays[arrayId]) {
```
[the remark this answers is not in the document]
dominik-werner-casra: You are right, to make it consistent I should create a code path that completely omits caching.

Sources/IO/Core/HttpDataSetReader/index.d.ts
dominik-werner-casra: I see your point but there is one question about ```maxCacheSize = none```. As far as I know ```none``` is not a built-in value in Javascript therefore, how should I add it? One option would be to allow strings to be set:
```javascript
setMaxCacheSize('none')  // disable caching
```
However, this means that we allow different datatypes for one property (nullable/undefinable numbers and string). I know that this is possible and probably fine in Javascript but as a developer more used to languages like C# or C++ this is a bit icky. How about we rename the ```cacheArrays``` property to ```arrayCachingEnabled```?

dominik-werner-casra: I have implemented my idea and I think using ```arrayCachingEnabled``` makes it more obvious what the intention is than the old name or using ```setMaxCacheSize('none'). Interested what your opinion is @finetjul.

finetjul: By `none` I meant `null`

dominik-werner-casra: Makes sense :)
But even with that in mind, I would still argue to use a separate property to control this. I would say the behavior
```
null -> caching disabled
undefined -> unlimited cache size
```
Is not that intuitive for people unfamiliar with the implementation.

finetjul: Fair enough. I'm fine with `setArrayCachingEnabled`.
But what behavior will you have if maxCacheSize is undefined/null ? caching will be disabled ?

finetjul:
> Another option could be:
> 
> * cachesize = 0 => no cache
> * cachesize = -1 => infinity

That's what I suggested here: https://github.com/Kitware/vtk-js/pull/2979#discussion_r1444839469

@jourdain if you think that it is explicit enough and it follows VTK.js coding style, then let's stick to a unique cacheSize variable instead of 2. I prefer that option, but @dominik-werner-casra  feared that it was too hackish.

Sources/IO/Core/HttpDataSetReader/index.js
```diff
@@ -145,26 +147,66 @@ function vtkHttpDataSetReader(publicAPI, model) {
...
+          while (cacheSize > cacheSizeLimit && sortedArrayCache.length > 0) {
+            const [oldId, entry] = sortedArrayCache.pop();
+            delete cachedArrays[oldId];
+            cacheSize -= entry.array.values.byteLength;
+          }
+
+          // Edge case: If the new entry is bigger than the cache limit
+          if (cacheSize > cacheSizeLimit) {
+            delete cachedArrays[arrayId];
+          }
```
floryst, line 196: You can drop this edge case code handling with some code rearrangements.

```
// Add array to cache
...

// sort cache. Last element is the new array.
```

Additionally, I think we should have a warning if the new entry is bigger than the cache limit.

```
if (!cachedArrays[arrayId]) vtkWarningMacro('Cache size is too small for the dataset')
```

Sources/IO/Core/HttpDataSetReader/test/testHttpDataSetReader.js
```diff
@@ -0,0 +1,222 @@
+import test from 'tape-catch';
+
+import vtkHttpDataSetReader from '..';
+import MockDataAccessHelper from '../../DataAccessHelper/MockDataAccessHelper';
```
floryst, line 4: I recommend using the full import path `vtk.js/Sources/IO/Core/HttpDatasetReader` for consistency.

dominik-werner-casra:
> > Another option could be:
> > 
> > * cachesize = 0 => no cache
> > * cachesize = -1 => infinity
> 
> That's what I suggested here: [#2979 (comment)](https://github.com/Kitware/vtk-js/pull/2979#discussion_r1444839469)
> 
> @jourdain if you think that it is explicit enough and it follows VTK.js coding style, then let's stick to a unique cacheSize variable instead of 2. I prefer that option, but @dominik-werner-casra feared that it was too hackish.

I actually think that using ```0``` and ```-1``` changes the situation a bit. Also if this conforms better with the vtk.js coding style I am happy to change it. Additionally, I would use ```undefined``` and ```null``` similar to ```0``` to disable the cache, this way you could write a check for caching like this:

```javascript
if (reader.getMaxCacheSize()) {
  // cache enabled
}
else {
  // cache disabled
}
```
What are your thoughts @finetjul and @jourdain?

finetjul:
> What are your thoughts @finetjul and @jourdain?

That's good with me.

Sources/IO/Core/HttpDataSetReader/test/testHttpDataSetReader.js
dominik-werner-casra: I changed the paths in my most recent commit.

Sources/IO/Core/HttpDataSetReader/index.js
dominik-werner-casra: Changes implemented as requested.

Sources/IO/Core/HttpDataSetReader/index.d.ts
```diff
@@ -148,6 +148,40 @@ export interface vtkHttpDataSetReader extends vtkHttpDataSetReaderBase {
 	 */
 	getUrl(): string;
 
+	/**
+	 * Gets if arrays are stored in cache.
+	 */
+	getArrayCachingEnabled(): boolean;
+
+	/**
+	 * Sets if arrays are stpred in cache.
```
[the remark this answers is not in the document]
dominik-werner-casra: Property was removed, so typo no longer exists.

Sources/IO/Core/HttpDataSetReader/index.d.ts
dominik-werner-casra: The property has been removed and the following has been implemented:
```
-1 -> unlimited cache
0 -> cache disabled
null -> cache disabled
undefined -> cache disabled
```

dominik-werner-casra:
@finetjul all requests have been implemented, ready for the next round.

Sources/IO/Core/HttpDataSetReader/index.js
```diff
@@ -145,26 +147,71 @@ function vtkHttpDataSetReader(publicAPI, model) {
...
+          const sortedArrayCache = Object.entries(cachedArrays).sort((a, b) =>
+            Math.sign(b[1].lastAccess - a[1].lastAccess)
+          );
+
+          // Check cache size
+          const cacheSizeLimit = model.maxCacheSize * MiB;
+          let cacheSize = Object.values(cachedArrays).reduce(
+            (accSize, entry) => accSize + entry.array.values.byteLength,
+            0
+          );
```
finetjul, line 189: sort and reduce are not very fast, but I guess it is fine to leave it as is. I presume it is not on a critical path.

finetjul (commented):
LGTM

Sources/IO/Core/HttpDataSetReader/index.js
dominik-werner-casra: I have tested it with 7000+ entries by letting the example I have attached to #2978 run for a while and there do not seem to be any issues. I leave it for now.

Sources/IO/Core/HttpDataSetReader/index.js
```diff
@@ -145,26 +147,71 @@ function vtkHttpDataSetReader(publicAPI, model) {
...
+          // Return here if caching is disabled
+          if (!model.maxCacheSize) {
+            return newArray;
+          }
+
+          // Add array to cache
+          cachedArrays[arrayId] = { array: newArray, lastAccess: new Date() };
+
+          // If maxCacheSize is set to -1 the cache is unlimited
+          if (model.maxCacheSize === -1) {
```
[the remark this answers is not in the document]
dominik-werner-casra: Makes sense, the new commit changes this to use your suggestion 👍

dominik-werner-casra:
I am very sad to say that my version of the **HttpDataSetReader** does not work properly anymore as soon as the cache is actually used. I realized this when I investigated the failing tests (first I thought the failing tests had nothing to do with my changes). As it turns out, separating ```cachedPromises``` and ```cachedArrays``` poses problems due to the way the code around the whole thing is implemented.
I am working on a new implementation that changes less of the overall structure to limit the side effects. This means that promieses and arrays are stored in the same global variable as before but it is now called differently:
```
cachedArrays -> cachedArraysAndPromises
```
I really don't like this very much but I have not found a better solution without rewriting large parts of the component.

dominik-werner-casra:
@finetjul I have pushed the new implementation. Locally, all tests are passing now and the lint rules seem to be happy as well.

finetjul (approved):
LGTM

dominik-werner-casra:
@finetjul Thank you for once again reviewing my changes. It seems that the tests are passing now on Firefox but strangely not on Chrome within the GitHub actions environment. I have tried to run the tests with Chrome locally (Windows 10, Chrome 120.0.0.0 x64) and it worked fine:
```
Browser: Chrome 120.0.0.0 (Windows 10)
        Success:        3078
        Skipped:        0
        Failed: 0
        Total:  3078
10 01 2024 09:11:07.698:INFO [tape-html-reporter]: report written to file Utilities/TestResults/Test-Report.html
```
Is there a timeout that needs to be configured? Or could there be an issue with my changes?
Thanks for the support!

floryst:
I'll re-run CI. That's a CI-only transient issue that we haven't been able to fully fix.

dominik-werner-casra:
@floryst The issue seems to persist. Do you want to re-run or is there something else we can do?

dominik-werner-casra:
It does not seem to get any better. Is someone looking into this? Would it make sense to remove the Chrome runner until it is clear what causes the issue, so the changes can be merged before it has any conflicts with the master branch? 

I still have a bit of time right now to work on this and provide support. Since I do this on behalf of our company, it will become more difficult looking into this the longer the PR takes to merge.

dominik-werner-casra:
@floryst How do we proceed?

floryst:
I've pushed a change to your branch that hopefully resolves the testing issue.

floryst:
So there is some funkiness going on with the tests. A snippet is provided below. For some reason, CI doesn't receive any more data after the 11th call. This was observed twice in CI with concurrency set to 1.
```
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] Cache was disabled through setting maxCacheSize to "null"'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test01 (not cached)] Method "fetchJSON()" is called 7 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test01 (not cached)] Method "fetchArray()" is called 7 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test01 (not cached)] Cache contains expected elements: '
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test02 (not cached)] Method "fetchJSON()" is called 8 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test02 (not cached)] Method "fetchArray()" is called 8 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test02 (not cached)] Cache contains expected elements: '
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test01 (not cached)] Method "fetchJSON()" is called 9 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test01 (not cached)] Method "fetchArray()" is called 9 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test01 (not cached)] Cache contains expected elements: '
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test03 (not cached)] Method "fetchJSON()" is called 10 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test03 (not cached)] Method "fetchArray()" is called 10 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test03 (not cached)] Cache contains expected elements: '
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test02 (not cached)] Method "fetchJSON()" is called 11 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test02 (not cached)] Method "fetchArray()" is called 11 time(s).'
Chrome 120.0.0.0 (Linux x86_64) INFO: '	[PASS] [http://mockData/test02 (not cached)] Cache contains expected elements: '
15 01 2024 20:19:24.081:WARN [Chrome 120.0.0.0 (Linux x86_64)]: Disconnected (0 times) , because no message in 30000 ms.
Chrome 120.0.0.0 (Linux x86_64) ERROR
  Disconnected , because no message in 30000 ms.
```

dominik-werner-casra:
Do you thing there might be something wrong with the tests I added? It is hard to understand since I cannot reproduce this in my environment. Is there anything you want me to try or change?

dominik-werner-casra:
I have created a Docker image to debug this on a more similar system to the one used in the CI checks. However, in my case the (using only Chrome), the browser always looses connection after a while (also completing many tests, successfully) and after 3 retry attempts fails. Importantly, this also happens when I remove all the tests I add and also on the master branch.
At this point I have no more ideas what I can do on our side. This fix is very important because without it, vtk.js is not usable in single page applications. It seems the fix is working and we are now stuck waiting for the pipeline issue to be resolved. What are the next steps?
Below you can find the Dockerfile I used for testing:
```Dockerfile
FROM ubuntu:20.04

# install curl
RUN DEBIAN_FRONTEND=noninteractive apt-get update
RUN DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y curl

# install nodejs
RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y nodejs

# install xvfb
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y xvfb

# install chrome for headless testing
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y wget
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y fonts-liberation xdg-utils libxss1 libappindicator1 libindicator7
RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y alsa-topology-conf alsa-ucm-conf at-spi2-core dbus-user-session dconf-gsettings-backend dconf-service dmsetup gir1.2-glib-2.0 glib-networking glib-networking-common glib-networking-services gsettings-desktop-schemas libargon2-1 libasound2 libasound2-data libatk-bridge2.0-0 libatspi2.0-0 libcap2 libcolord2 libcryptsetup12 libdconf1 libdevmapper1.02.1 libepoxy0 libgbm1 libgirepository-1.0-1 libgtk-3-0 libgtk-3-bin libgtk-3-common libip4tc2 libjson-c4 libjson-glib-1.0-0 libjson-glib-1.0-common libkmod2 liblcms2-2 libnspr4 libnss-systemd libnss3 libpam-systemd libproxy1v5 librest-0.7-0 libsoup-gnome2.4-1 libsoup2.4-1 libu2f-udev libwayland-cursor0 libwayland-egl1 libwayland-server0 libxkbcommon0 networkd-dispatcher python3-dbus python3-gi systemd systemd-sysv systemd-timesyncd udev
RUN dpkg -i google-chrome*.deb
RUN apt-get update
RUN wget https://chromedriver.storage.googleapis.com/2.22/chromedriver_linux64.zip
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y unzip
RUN unzip chromedriver_linux64.zip
RUN cp ./chromedriver /usr/bin/
RUN chmod ugo+rx /usr/bin/chromedriver
RUN DEBIAN_FRONTEND=noninteractive apt-get -y install libxpm4 libxrender1 libgtk2.0-0 libnss3 libgconf-2-4
RUN DEBIAN_FRONTEND=noninteractive apt-get -y install xorg xvfb gtk2-engines-pixbuf
RUN DEBIAN_FRONTEND=noninteractive apt-get -y install dbus-x11 xfonts-base xfonts-100dpi xfonts-75dpi xfonts-cyrillic xfonts-scalable
RUN DEBIAN_FRONTEND=noninteractive apt-get -y install imagemagick x11-apps

# Set user and group
ARG user=appuser
ARG group=appuser
ARG uid=1000
ARG gid=1000
RUN groupadd -g ${gid} ${group}
RUN useradd -u ${uid} -g ${group} -s /bin/sh -m ${user} # <--- the '-m' create a user home directory

# get repo
COPY ./vtk-js /repo/vtk-js

# install node modules
WORKDIR /repo/vtk-js
RUN npm install
RUN chmod -R o+w ./*

USER appuser

ENTRYPOINT [ "xvfb-run", "--auto-servernum", "npm", "run", "test", "--", "--browsers", "Chrome" ]
```

daker:
Maybe try launching chrome with --disable-dev-shm-usage ?

dominik-werner-casra:
Hello @daker thank you for your suggestion. I tried to supply this option by using:
```Dockerfile
ENTRYPOINT [ "xvfb-run", "--auto-servernum", "npm", "run", "test", "--", "--browsers", "Chrome", "--disable-dev-shm-usage"]
```
Sadly, the result is still the same: some tests run and after a while it gets stuck. I am not sure if the **xvfb-run** command passes these arguments to Chrome though. At this point I am also not sure how efficient it is to do this with my custom Docker image since this probably still does not capture the GitHub environment. Also my time is limited since I am working on other projects as well and hoping to wrap things up here.

daker:
the `--disable-dev-shm-usage` needs to be added to the flags here https://github.com/Kitware/vtk-js/blob/master/karma.conf.js#L89

dominik-werner-casra:
Thank you for the fast reply! I have tried again with the change in **karma.conf.js** but the result is still the same.

Sources/IO/Core/HttpDataSetReader/test/testHttpDataSetReader.js
```diff
@@ -0,0 +1,238 @@
...
+        },
+        (err) => {
+          console.error(err);
+        }
+      );
+    };
+
+    // start test sequence with first test
+    processNextTestCase();
+  });
```
floryst, line 32: This can be simplified to:
```suggestion
  let promise = Promise.resolve();
  testCases.forEach((testCase) => {
    promise = promise.then(() => new Promise(testCase));
  });
  promise.catch((err) => {
    console.error(err);
  });
  return promise;
```

Sources/IO/Core/HttpDataSetReader/test/testHttpDataSetReader.js
```diff
@@ -0,0 +1,238 @@
...
+          }
+        }
+        resolve(v);
+      },
+      (err) => reject(err)
+    );
+  };
+}
+
+test('Caching capabilities of vtkHttpDataSetReader with unlimited cache.', (t) => {
```
floryst, line 106:
```suggestion
test('Caching capabilities of vtkHttpDataSetReader with unlimited cache.', async (t) => {
```

Sources/IO/Core/HttpDataSetReader/test/testHttpDataSetReader.js
```diff
@@ -0,0 +1,238 @@
+import test from 'tape-catch';
```
floryst, line 1:
```suggestion
import test from 'tape';
```

Sources/IO/Core/HttpDataSetReader/test/testHttpDataSetReader.js
```diff
@@ -0,0 +1,238 @@
...
+    // load array for the second time and it is still not cached
+    fetchArrayTest(t, reader, callTracker, 'http://mockData/test03', false, []),
+  ];
+
+  // verify that cache can be disable by setting maxCacheSize to "0", "null" and "undefined"
+  runTests([
+    ...createDisabledCacheTests(0),
+    ...createDisabledCacheTests(null),
+    ...createDisabledCacheTests(undefined),
+  ]).then(t.end);
```
floryst, line 237:
```suggestion
  await runTests([
    ...createDisabledCacheTests(0),
    ...createDisabledCacheTests(null),
    ...createDisabledCacheTests(undefined),
  ]);
  t.end();
```

Sources/IO/Core/HttpDataSetReader/test/testHttpDataSetReader.js
floryst: This is more of a readability change than a functional change.

floryst (commented):
I suspect the underlying issue is that an unhandled promise rejection is causing tape to hang, since t.end() never gets called.

I've provided some code suggestions that should allow you to identify the root cause of the issue.

dominik-werner-casra:
Thank you @floryst for your suggestions. I had to add some missing ```async``` and ```await``` keywords and also added some timeouts to make sure the promises don't get stuck. However, the second browser always gets stuck now on my machine. Since I am not a Javascript developer and am not familiar with the frameworks used here I am really at the end of my ideas.

dominik-werner-casra:
Any idea how we are going to proceed?

floryst:
It appears that your changes have caused some tests utilizing vtkHttpDataSetReader to start failing, resulting in test timeout. This appears to only occur in the github environment, so some digging is needed.

What I can do is identify those tests using the dataset reader, update them to properly return a promise, and then have you rebase onto those changes so we can see why

dominik-werner-casra:
I would appreciate that. I have some things to do at the moment but I am hoping to look into this again this week.

floryst:
#3001 should cover the relevant tests that are timing out for your PR.

dominik-werner-casra:
@floryst I finally had a few minutes to spare and synchronized with the current **master**. Now the both browsers run through all the tests. However, the second browser (Firefox if I test locally) terminates with 23 failed tests

dominik-werner-casra:
A colleague and I finally found the issue! The ```Sources\IO\Core\DataAccessHelper\MockDataAccessHelper.js``` was generating test data that became too big during the tests. That seems to be the reason why it was always the second browser that failed (chrome on GitHub and Firefox for me locally).

dominik-werner-casra:
Any idea why the ```commitlint``` check failed? I put the ```fix:``` in front of every commit message.

floryst:
Check the commits tab. The github-merged commits don't conform (They start with "Update ...."). Rebasing to rename or squash those commits is needed.

floryst:
Whoops, I clicked "close and comment" instead of "comment".

dominik-werner-casra:
It seems to be related to these updates:
image

image

What do I have to do here? I never use squash or rebase with git.

floryst:
I went ahead and squashed those commits for you.

dominik-werner-casra:
Awesome! It finally worked. Thanks for all the help!

daker:
@dominik-werner-casra can the mockdata files be move to the test folder ? i believe they will be shipped with build files

dominik-werner-casra:
Let me know if I need to do something before we can merge.

PR #2980, opened 2024-01-06
feat(reslicecursor): expose a convenient getPlaneExtremities function

finetjul:
This is useful to control the center using sliders.

### Context
It is not obvious how to control the reslice cursor widget with sliders.

### Results
This PR demonstrates how to control the reslice cursor widget with sliders and expose a convient function.

### Changes
- [x] ResliceCursorWidget.getPlaneExtremities() has been created

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows
  - **Browser**: Chrome

### DIFF
--- a/Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
+++ b/Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
@@ -1,5 +1,7 @@
 import macro from 'vtk.js/Sources/macros';
 import vtkAbstractWidgetFactory from 'vtk.js/Sources/Widgets/Core/AbstractWidgetFactory';
+import vtkBoundingBox from 'vtk.js/Sources/Common/DataModel/BoundingBox';
+import vtkBox from 'vtk.js/Sources/Common/DataModel/Box';
 import vtkPlane from 'vtk.js/Sources/Common/DataModel/Plane';
 import vtkPlaneSource from 'vtk.js/Sources/Filters/Sources/PlaneSource';
 import vtkPlaneManipulator from 'vtk.js/Sources/Widgets/Manipulators/PlaneManipulator';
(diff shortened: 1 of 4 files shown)


Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -647,6 +643,29 @@ function vtkResliceCursorWidget(publicAPI, model) {
...
+    const p1 = vtkMath.multiplyAccumulate(
+      publicAPI.getWidgetState().getCenter(),
+      dirProj,
+      -10000,
+      []
+    );
+    const p2 = vtkMath.multiplyAccumulate(
+      publicAPI.getWidgetState().getCenter(),
+      dirProj,
+      10000,
```
floryst, line 658: These seem like "sufficiently large" constants for most realistic data, but still has edge case behavior for images with very large bounds. p1 and p2 could end up inside the bounds, which means instersectionPoints is empty (presuming intersectWithLine does segment-plane intersection).

finetjul: should now be fixed

Sources/Widgets/Widgets3D/ResliceCursorWidget/index.js
```diff
@@ -647,6 +644,32 @@ function vtkResliceCursorWidget(publicAPI, model) {
...
+      []
+    );
+    const p2 = vtkMath.multiplyAccumulate(
+      publicAPI.getWidgetState().getCenter(),
+      dirProj,
+      length,
+      []
+    );
+    // FIXME: support oriented bounds
+    const intersectionPoints = vtkBox.intersectWithLine(
```
bruyeret, line 666: An idea of solution:
From the image direction, we already have the 3 normals of the 6 planes of the image oriented bounding box.
Now let's say we have for each plane `i` the equation `X.N_i = a_i` (the equation of a plane, we get `a_i` using indexToWorld)
And the equation of the line `X = b * N`
Then the intersection of the plane `i` and the line verifies: `(b * N). N_i = a_i` so `b = a / (N.N_i)` when `N.N_i` is not 0.
We get 6 different values for `b` and only keep the biggest negative and the smallest positive (if there is a negative and a positive value).
From `b`, we get the point by doing `X = b * N`.

bruyeret (approved):

PR #2992, opened 2024-01-19
fix(handleWheel): mouse scrolling behavior for out-of-focus windows

jadh4v:
# Brief Description
When rebinding events with a different container, make sure all timeouts and state variables from previous container are cleared.

This PR is part of a react-vtk-js fix for out-of-focus wheel events (https://github.com/Kitware/react-vtk-js/pull/125)

### Changes

### PR and Code Checklist

### Testing

- [x] ~~This change adds or fixes unit tests~~ 
- [x] Tested environment:
  - **vtk.js**: latest master 
  - **OS**: Windows 11, macOS Sonoma 
  - **Browser**: Chrome 120

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -92,6 +92,9 @@ function vtkRenderWindowInteractor(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkRenderWindowInteractor');
 
+  // Capture "parentClass" api for internal use
+  const superClass = { ...publicAPI };
+
   // Initialize list of requesters
   const animationRequesters = new Set();
 
@@ -215,8 +218,12 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     return event.pointerType || '';
   }
 
-  publicAPI.bindEvents = (container) => {
-    model.container = container;
+  const _bindEvents = () => {
+    if (model.container === null) {
+      return;
+    }
+
+    const { container } = model;
     container.addEventListener('contextmenu', preventDefault);
     container.addEventListener('wheel', publicAPI.handleWheel);
     container.addEventListener('DOMMouseScroll', publicAPI.handleWheel);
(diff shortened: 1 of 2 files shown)


Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -217,6 +217,13 @@ function vtkRenderWindowInteractor(publicAPI, model) {
 
   publicAPI.bindEvents = (container) => {
     model.container = container;
+    // Clear any previous timeouts and state variables that control mouse / touchpad behavior.
+    clearTimeout(model.moveTimeoutID);
+    clearTimeout(model.wheelTimeoutID);
+    model.moveTimeoutID = 0;
+    model.wheelTimeoutID = 0;
+    wheelCoefficient = 1.0;
```
finetjul, line 225: Shouldn't you simply call unbindEvents on `model.container` ?

And move those clear timeouts in `unbindEvents` ?

jadh4v: That makes sense. The `RenderWindowInteractor` assumes that it is bound to a single container at any given point. Adding an explicit unbind will make it more obvious, and the right thing to do.

jadh4v: @finetjul  updated the branch

finetjul: I also think that the (un)binding should be done by `setContainer()`

floryst: One thought regarding `setContainer`: we should be able to add it without breaking bind/unbindEvents. Just need to be sure that setContainer operates similarly to bind/unbindEvents, and that if someone calls setContainer and then bindEvents, we don't get two listeners registered.

jadh4v: What behavior do we want for `setContainer(null)` ?

finetjul: > What behavior do we want for setContainer(null) ?

`setContainer(null)` it should simply unbind the current container. No binding.

> `we should be able to add it without breaking bind/unbindEvents`. 

@floryst are you saying we should still support (backward compatibility) the case were users do not call `setContainer()` but only bind/unbind ?

finetjul (commented):
While that works, I find the logic quite convoluted.
How about the following: 
```
unbindEvents() {
  ...
  // Kept for backward compatibility purpose
  publicAPI.setContainer(null);
}
setContainer(newContainer) {
  if (newContainer === model.container) {
     return false;
  }
  unbindEvents();
  model.container = newContainer;
  bindEvents();
  return true;
}

// container parameter is for backward compatibility purpose
bindEvents(container = model.container){
  if (model.container != container) {
    setContainer(container)
    return;
  }
  ...
}
...
publicAPI.delete(
  ...
  setContainer(null);
}
```
And to not forget about `modified()` and `_onContainerChanged()` (see `macro.js::findSetter()`,
```
setContainer(newContainer) {
  if (newContainer === model.container) {
     return false;
  }
  unbindEvents();
  const res = superclass.setContainer(newContainer);
  bindEvents(model.container);
  return res;
}
...
macro.setGet(...[container]);
```

jadh4v:
Oh yes, thanks for reminding about `onChanged` and `modified`.

I don't agree on the new call tree you suggested. It is more complex to my eyes. Specially since `setContainer()` calls both `bindEvents` and `unbindEvents` and then `bind\unbindEvents` again call `setContainer`. It makes the code hard to read since there is a possible circular call which can only be disentangled through reasoning about the values of arguments/variables.

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -1126,6 +1154,11 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     model.currentRenderer = r;
   };
 
+  publicAPI.setContainer = (container) => {
+    publicAPI.unbindEvents();
+    return publicAPI.bindEvents(container);
+  };
```
finetjul, line 1160: calling `setContainer(null)` will not call `model.container = null`

Sources/Rendering/Core/RenderWindowInteractor/index.js
floryst: That's my preference, yes. Current applications use bind/unbind, so setContainer should exist as an alternative. We should document a preference for setContainer in bind/unbind docs. We could mark bind/unbind as deprecated and remove them in the future.

jadh4v: Pushed another update based on both of your suggestions.

finetjul (approved):
LGTM

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -1126,8 +1172,19 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     model.currentRenderer = r;
   };
 
+  publicAPI.setContainer = (container) => {
+    if (container === undefined) {
+      vtkErrorMacro('Container cannot be undefined.');
+    }
```
floryst, line 1178: Did you mean to return here or just show a warning message? Another option is to slightly relax this condition by removing this if statement and running `setContainer(container ?? null)` below to always get null.

floryst (commented):
One last change needed: the `getContainer()` type needs to be updated to be `getContainer(): Nullable<HTMLElement>`. You can also update the `setContainer` type to use Nullable for consistency.

floryst:
LGTM

PR #2995, opened 2024-01-22
Drop tape catch

floryst:
### Context
Drops tape-catch in favor of just using tape.

Probably related to failing tests in https://github.com/Kitware/vtk-js/pull/2979.

This change means unhandled exceptions will terminate the test runner rather than resulting in a failed test. Since there are no tests that rely on unhandled exceptions to fail the test, this change doesn't affect anything.

In exchange, we get better handling of unhandled rejected promises. Tests can now be written with async/await and tape will fail a test if a test rejects with an error.

### Results
- tape-catch dropped in favor of tape
### Changes

- [x] updated docs

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: tip (at the time of writing)
  - **OS**: macOS
  - **Browser**: Chrome + Firefox


finetjul (approved):
I like that idea of simplification.

PR #2997, opened 2024-01-23
fix(PaintWidget): blocked event w/o intersection

floryst:
### Context
Events were being aborted even when the manipulator returns no world coordinates.

### Results
The paint widget does not abort events during non-intersecting actions.

### PR and Code Checklist

### Testing

- [ ] Tested environment: not tested


finetjul (commented):
LGTM

PR #2999, opened 2024-01-24
fix(PaintWidget): worldCoords can be null

floryst:
### Context
Using the paint widget with a manipulator that can return null.

### Results

Paint widget no longer errors when worldCoords is null.

### Changes

- [x] PaintWidget updated

### PR and Code Checklist

### Testing

- [x] Tested locally


floryst:
I noticed a bunch of other widgets don't expect worldCoords to be null. I may lump those changes into this PR.

finetjul (approved):
LGTM

PR #3005, opened 2024-01-31
docs(ItkWasmVolumeExample): update to use @itk-wasm/image-io

thewtex:
ITK-Wasm image-io is now distributed in its own package, @itk-wasm/image-io.

It is available as an ESM bundle. We load this with a dynamic import.

readImageArrayBuffer is now readImage. The `webWorker` argument is not required. Data is passed as object with a `data` Uint8Array and path string.


finetjul:
LGTM

daker (approved):
LGTM

floryst (approved):

PR #3008, opened 2024-02-01
Add typescript specific to widget behavior

bruyeret:
Use Symbols in the type of extend function as typescript can't differentiate two `extend` functions.


floryst (approved):

daker (approved):
LGTM

PR #3009, opened 2024-02-01
Add a CSS helper for CSS filters

bruyeret:
Create utility functions to manipulate equivalent of filters in vtk.js.
Based on the W3C specification.

### DIFF
--- a/Sources/Common/Core/Math/index.js
+++ b/Sources/Common/Core/Math/index.js
@@ -609,7 +609,7 @@ export function multiply3x3_mat3(a_3x3, b_3x3, out_3x3) {
   }
 }
 
-export function multiplyMatrix(a, b, rowA, colA, rowB, colB, out_rowXcol) {
+export function multiplyMatrix(a, b, rowA, colA, rowB, colB, outRowAColB) {
   // we need colA == rowB
   if (colA !== rowB) {
     vtkErrorMacro('Number of columns of A must match number of rows of B.');
@@ -623,10 +623,10 @@ export function multiplyMatrix(a, b, rowA, colA, rowB, colB, out_rowXcol) {
   for (let i = 0; i < rowA; i++) {
     // output col
     for (let j = 0; j < colB; j++) {
-      out_rowXcol[i * colB + j] = 0;
+      outRowAColB[i * colB + j] = 0;
       // sum for this point
       for (let k = 0; k < colA; k++) {
-        out_rowXcol[i * colB + j] += copyA[i * colA + k] * copyB[j + colB * k];
+        outRowAColB[i * colB + j] += copyA[i * colA + k] * copyB[j + colB * k];
       }
     }
   }
--- a/Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
+++ b/Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
@@ -1,0 +1,91 @@
+/**
+ * A helper file to transform RGBA points using CSS filters equivalent
+ * The equivalents of CSS filters using SVG filters can be found here:
+ * https://www.w3.org/TR/filter-effects-1/#ShorthandEquivalents
+ * For each SVG filter, you can look for the maths behind it on the same page:
+ * https://www.w3.org/TR/filter-effects-1/#FilterPrimitivesOverview
+ *
+ * For example, the saturate filter equivalent is here:
+ * https://www.w3.org/TR/filter-effects-1/#saturateEquivalent
+ * And the maths behind the feColorMatrix of type saturate is here:
+ * https://www.w3.org/TR/filter-effects-1/#ref-for-attr-valuedef-type-saturate
+ *
+ * The transforms are done using matrices of size 5 by 5. They are row major
+ * as in vtkMath. The vectors representing the RGBA points uses
+ * [R, G, B, A, 1] vectors, with each channel between 0 and 1.
+ */
+
+import { identity, multiplyMatrix } from 'vtk.js/Sources/Common/Core/Math';
+
+export const luminanceWeights = [0.213, 0.715, 0.072];
+
+export function getNewFilter() {
+  return new Array(25);
+}
+
+export function getIdentityFilter(outFilter = getNewFilter()) {
+  return identity(5, outFilter);
+}
+
+export function combineFilters(
+  baseFilter,
+  newFilter,
+  outFilter = getNewFilter()
+) {
+  multiplyMatrix(newFilter, baseFilter, 5, 5, 5, 5, outFilter);
+  return outFilter;
+}
+
+export function applyFilter(filter, r, g, b, a = 1) {
+  const vec = [r, g, b, a, 1];
+  multiplyMatrix(filter, vec, 5, 5, 5, 1, vec);
+  return vec.slice(0, 4);
+}
+
+export function getLinearFilter(slope, intercept, outFilter = getNewFilter()) {
+  getIdentityFilter(outFilter);
+  for (let row = 0; row < 3; ++row) {
+    outFilter[row * 5 + row] = slope;
+    outFilter[row * 5 + 4] = intercept;
+  }
+  return outFilter;
+}
+
+// https://www.w3.org/TR/filter-effects-1/#contrastEquivalent
+// https://www.w3.org/TR/filter-effects-1/#attr-valuedef-type-linear
+export function getContrastFilter(contrast, outFilter = getNewFilter()) {
+  const slope = contrast;
+  const intercept = -(0.5 * contrast) + 0.5;
+  return getLinearFilter(slope, intercept, outFilter);
+}
+
+// https://www.w3.org/TR/filter-effects-1/#saturateEquivalent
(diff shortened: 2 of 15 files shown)


Sources/Common/Core/Math/index.js
```diff
@@ -609,24 +609,31 @@ export function multiply3x3_mat3(a_3x3, b_3x3, out_3x3) {
...
   // we need colA == rowB
   if (colA !== rowB) {
     vtkErrorMacro('Number of columns of A must match number of rows of B.');
   }
 
   // If a or b is used to store the result, copying them is required
-  const copyA = [...a];
-  const copyB = [...b];
+  const copyA =
+    a === out_rowA_colB || (a.buffer && a.buffer === out_rowA_colB.buffer)
```
finetjul, line 620: why `a === out_rowA_colB` is not enough ?

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
```diff
@@ -0,0 +1,91 @@
...
+ * The transforms are done using matrices of size 5 by 5. They are row major
+ * as in vtkMath. The vectors representing the RGBA points uses
+ * [R, G, B, A, 1] vectors, with each channel between 0 and 1.
+ */
+
+import { identity, multiplyMatrix } from 'vtk.js/Sources/Common/Core/Math';
+
+export const luminanceWeights = [0.213, 0.715, 0.072];
+
+export function getNewFilter() {
```
finetjul, line 22: rename to `createCSSFilter`?

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
```diff
@@ -0,0 +1,91 @@
...
+  return new Array(25);
+}
+
+export function getIdentityFilter(outFilter = getNewFilter()) {
```
finetjul, line 26: isn't `getNewFilter()` always called even when an `outFilter` parameter is passed?
If so, then `return identity(5, outFilter || getNewFilter());` might be more appropriate

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
```diff
@@ -0,0 +1,91 @@
...
+  return outFilter;
+}
+
+export function applyFilter(filter, r, g, b, a = 1) {
+  const vec = [r, g, b, a, 1];
+  multiplyMatrix(filter, vec, 5, 5, 5, 1, vec);
+  return vec.slice(0, 4);
+}
+
+export function getLinearFilter(slope, intercept, outFilter = getNewFilter()) {
```
finetjul, line 45: `createLinearFilter`

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
```diff
@@ -0,0 +1,91 @@
...
+    outFilter[row * 5 + 4] = intercept;
+  }
+  return outFilter;
+}
+
+// https://www.w3.org/TR/filter-effects-1/#contrastEquivalent
+// https://www.w3.org/TR/filter-effects-1/#attr-valuedef-type-linear
+export function getContrastFilter(contrast, outFilter = getNewFilter()) {
+  const slope = contrast;
+  const intercept = -(0.5 * contrast) + 0.5;
```
finetjul, line 58: `0.5 * (1 - contrast)`

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
```diff
@@ -0,0 +1,91 @@
...
+  for (let row = 0; row < 3; ++row) {
+    outFilter[row * 5 + row] = slope;
+    outFilter[row * 5 + 4] = intercept;
+  }
+  return outFilter;
+}
+
+// https://www.w3.org/TR/filter-effects-1/#contrastEquivalent
+// https://www.w3.org/TR/filter-effects-1/#attr-valuedef-type-linear
+export function getContrastFilter(contrast, outFilter = getNewFilter()) {
```
finetjul, line 56: createContrastFilter

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
```diff
@@ -0,0 +1,91 @@
...
+  const slope = contrast;
+  const intercept = -(0.5 * contrast) + 0.5;
+  return getLinearFilter(slope, intercept, outFilter);
+}
+
+// https://www.w3.org/TR/filter-effects-1/#saturateEquivalent
+// https://www.w3.org/TR/filter-effects-1/#ref-for-attr-valuedef-type-saturate
+export function getSaturateFilter(saturate, outFilter = getNewFilter()) {
```
finetjul, line 64: `createSaturateFilter`

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
```diff
@@ -0,0 +1,91 @@
...
+    for (let row = 0; row < 3; ++row) {
+      outFilter[row * 5 + col] = row === col ? diagonalValue : nonDiagonalValue;
+    }
+  }
+  return outFilter;
+}
+
+// https://www.w3.org/TR/filter-effects-1/#brightnessEquivalent
+// https://www.w3.org/TR/filter-effects-1/#attr-valuedef-type-linear
+export function getBrightnessFilter(brightness, outFilter = getNewFilter()) {
```
finetjul, line 79: `createBrightnessFilter`

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
```diff
@@ -0,0 +1,91 @@
...
+  const slope = brightness;
+  const intercept = 0;
+  return getLinearFilter(slope, intercept, outFilter);
+}
+
+// https://www.w3.org/TR/filter-effects-1/#invertEquivalent
+// https://www.w3.org/TR/filter-effects-1/#attr-valuedef-type-table
+export function getInvertFilter(invert, outFilter = getNewFilter()) {
```
finetjul, line 87: `createInvertFilter`

finetjul (commented):
Please add a test or an example for the CSS filters functions.
Otherwise it's hard to see how it can be reused.

Sources/Common/Core/Math/index.js
bruyeret: Because you can create two typed arrays that are different but use the same buffer:
```js
const x = new Float32Array(1);
const y = new Float32Array(x.buffer);

x[0] = 42;
console.log(x); // [42]
y[0] = 3.14;
console.log(x); // [3.14]

console.log(x === y); // false
console.log(x.buffer === y.buffer); // true
```

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
bruyeret: No, it is called only if no parameter is given
Here is a code snippet to test it:
```js
function foo() {
  console.log("foo");
  return 42;
}

function bar(x = foo()) {
  return x + 10;
}

bar(); // prints: foo
bar(2); // nothing is printed
```

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
bruyeret: I know it is equivalent, I wrote it like this only to stick to what is written in the specs:
https://www.w3.org/TR/filter-effects-1/#contrastEquivalent

Sources/Common/Core/Math/index.js
finetjul: Fair enough, but if you are sharing the same buffer in 2 different arrays and passing those arrays to `multiplyMatrix`, you seem ok to take the risk that `a` is not copied (if it is correctly documented).

Or said differently, here is an example where `a` and `b` do not have to be copied that could make sense:
```
const inputsOutputs = new Float32Array(10000);
...
multiplyMatrix(inputsOutputs.subArray(0), inputsOutputs.subArray(9), 3,3, inputsOutputs.subArray(18));
```

And if it is an issue for the caller, then he can still do the `copy` him/herself before calling `multiplyMatrix`. Preventing the copy is not possible with your implementation

Sources/Rendering/Core/ColorTransferFunction/CssFilters.js
finetjul: fair enough

Sources/Common/Core/Math/index.js
bruyeret: Yes, it is a limitation of this implementation, it sometimes makes copies that are not needed.
I just wanted to avoid some copies but keep the old API.
I could you `byteOffset` in case of TypedArray, but isn't it overengineering?

finetjul: I suggest to keep the test light: `const copyA = a === out_rowA_colB ? [...a] : a;`

bruyeret: I reverted and created a new issue

finetjul (approved):
LGTM

PR #3020, opened 2024-02-23
Add vtkpicker pick3d point public api

bourdaisj:
### Context
vtkPicker currently cannot pick using 2 points in world coordinates. Add such an API.

### Results
Picking in world coordinates.

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


bourdaisj:
@rodrigobasilio2022 if you want to take a look.

Sources/Rendering/Core/Picker/index.js
```diff
@@ -89,9 +89,171 @@ function vtkPicker(publicAPI, model) {
...
+        p1Mapper[1] /= p1Mapper[3];
+        p1Mapper[2] /= p1Mapper[3];
+
+        p2Mapper[0] /= p2Mapper[3];
+        p2Mapper[1] /= p2Mapper[3];
+        p2Mapper[2] /= p2Mapper[3];
+
+        for (let i = 0; i < 3; i++) {
+          ray[i] = p2Mapper[i] - p1Mapper[i];
+        }
```
finetjul, line 171: `vtkMath.subtract` (or `vec3.subtract`) ?

Sources/Rendering/Core/Picker/index.js
```diff
@@ -89,9 +89,171 @@ function vtkPicker(publicAPI, model) {
...
+
+        //  Have the ray endpoints in mapper space, now need to compare this
+        //  with the mapper bounds to see whether intersection is possible.
+        //
+        //  Get the bounding box of the modeller.  Note that the tolerance is
+        //  added to the bounding box to make sure things on the edge of the
+        //  bounding box are picked correctly.
+        const bounds = mapper
+          ? vtkBoundingBox.inflate(mapper.getBounds(), tolerance)
+          : [0, 0, 0, 0, 0, 0];
```
finetjul, line 181: `[0,0,0,0,0,0]` is a valid bounding box. There is a constant (or convenient function ) that creates uninvalid/uninitialized bounds

Sources/Rendering/Core/Picker/index.js
```diff
@@ -89,9 +89,171 @@ function vtkPicker(publicAPI, model) {
...
+            p[1] = (1.0 - t[0]) * p1World[1] + t[0] * p2World[1];
+            p[2] = (1.0 - t[0]) * p1World[2] + t[0] * p2World[2];
+
+            let actorId = -1;
+            for (let i = 0; i < model.actors.length; i++) {
+              if (model.actors[i] === prop) {
+                actorId = i;
+                break;
+              }
+            }
```
finetjul, line 206: `model.actors.findIndex` ?

Sources/Rendering/Core/Picker/index.js
```diff
@@ -89,9 +89,171 @@ function vtkPicker(publicAPI, model) {
...
+      const mapper = prop.getMapper();
+      pickable =
+        prop.getNestedPickable() &&
+        prop.getNestedVisibility() &&
+        prop.getProperty().getOpacity() > 0.0;
+
+      //  If actor can be picked, get its composite matrix, invert it, and
+      //  use the inverted matrix to transform the ray points into mapper
+      //  coordinates.
+      if (pickable) {
```
finetjul, line 131: instead of having a long indentation, simply return early here.

Sources/Rendering/Core/Picker/index.js
```diff
@@ -122,46 +284,33 @@ function vtkPicker(publicAPI, model) {
...
     model.selectionPoint[1] = selectionY;
     model.selectionPoint[2] = selectionZ;
 
-    if (!renderer) {
-      vtkErrorMacro('Picker::Pick Must specify renderer');
-      return;
-    }
+    let displayCoords = [];
+    const p1World = new Float64Array(4);
+    const p2World = new Float64Array(4);
```
finetjul, line 305: why mix Float64array and regular array ?

Sources/Rendering/Core/Picker/index.js
```diff
@@ -237,134 +389,36 @@ function vtkPicker(publicAPI, model) {
...
 
-        if (mapper) {
-          bbox.setBounds(mapper.getBounds());
-          bbox.inflate(tolerance);
-        } else {
-          bbox.reset();
-        }
+    const view = renderer.getRenderWindow().getViews()[0];
+    const dims = view.getViewportSize(renderer);
+    const aspect = dims[0] / dims[1];
```
finetjul, line 416: won't work if `dims[1]` is 0 (see recent fix by Forrest)

Sources/Rendering/Core/Picker/test/testPicker.js
```diff
@@ -0,0 +1,146 @@
...
+
+  const selectionPoint = [-1, 0, 0, 1.0];
+  const focalPoint = [2, 0, 0, 1.0];
+  picker.pick3DPoint(selectionPoint, focalPoint, renderer);
+
+  const pickedActors = picker.getActors();
+  t.equal(pickedActors.length, 2);
+
+  gc.releaseResources();
+});
```
finetjul, line 139: no `t.end()` ?

Sources/Rendering/Core/Picker/test/testPicker.js
```diff
@@ -0,0 +1,146 @@
...
+
+    let pickedActors = picker.getActors();
+    t.equal(pickedActors[0], coneActor);
+
+    coneActor.setPickable(false);
+    picker.pick3DPoint(selectionPoint, focalPoint, renderer);
+    pickedActors = picker.getActors();
+    t.equal(pickedActors.length, 0);
+
+    gc.releaseResources();
```
finetjul, line 100: no `t.end()` ?

Sources/Rendering/Core/Picker/test/testPicker.js
```diff
@@ -0,0 +1,146 @@
...
+  const picker = vtkPicker.newInstance();
+
+  const selectionPoint = [-1, 0, 0, 1.0];
+  const focalPoint = [2, 0, 0, 1.0];
+  picker.pick3DPoint(selectionPoint, focalPoint, renderer);
+
+  const pickedActors = picker.getActors();
+  t.equal(pickedActors[0], coneActor);
+
+  gc.releaseResources();
```
finetjul, line 67: no `t.end()`?

Sources/Rendering/Core/Picker/index.js
```diff
@@ -89,9 +89,171 @@ function vtkPicker(publicAPI, model) {
...
+    const t = [];
+
+    const ray = [];
+    let pickable;
+    const hitPosition = [];
+
+    if (!renderer) {
+      vtkErrorMacro('You must specify a renderer.');
+      return;
+    }
```
finetjul, line 112: `renderer` already checked in the calling methods. You must decide whose responsibility is it to test for renderer validity, but do not do it more than needed

Sources/Rendering/Core/Picker/index.js
```diff
@@ -89,9 +89,171 @@ function vtkPicker(publicAPI, model) {
...
+          t[0] = publicAPI.intersectWithLine(
+            p1Mapper,
+            p2Mapper,
+            tolerance * 0.333 * (scale[0] + scale[1] + scale[2]),
+            prop,
+            mapper
+          );
+
+          if (t[0] < Number.MAX_VALUE) {
+            const p = [];
```
finetjul, line 195: `scale` is predefined before the for loop. Why not do the same for `p` ?

Sources/Rendering/Core/Picker/index.js
floryst: > `[0,0,0,0,0,0]` is a valid bounding box. There is a constant (or convenient function ) that creates uninvalid/uninitialized bounds

The symbol is [INIT_BOUNDS](https://github.com/Kitware/vtk-js/blob/master/Sources/Common/DataModel/BoundingBox/index.js#L5-L12)

Sources/Rendering/Core/Picker/test/testPicker.js
floryst: `gc.releaseResources()` calls `t.end()`. Not obvious from the call, but that's what the function does. The context comes from `testUtils.createGarbageCollector(t)` above.

Sources/Rendering/Core/Picker/index.js
```diff
@@ -89,9 +89,171 @@ function vtkPicker(publicAPI, model) {
...
+    tempArray.forEach((obj) => {
+      model.pickedPositions.push(obj.pickedPosition);
+      model.actors.push(obj.actor);
+    });
+  }
+
   // Intersect data with specified ray.
   // Project the center point of the mapper onto the ray and determine its parametric value
-  publicAPI.intersectWithLine = (p1, p2, tol, mapper) => {
+  publicAPI.intersectWithLine = (p1, p2, tolerance, prop, mapper) => {
```
floryst, line 256: This changes the public API for the vtkPicker, which should be reflected in the vtkPicker types.

Sources/Rendering/Core/Picker/index.js
bourdaisj: To be honest, I didn't write those lines myself, only moved it around. It probably has a good reasons to use `Float64Array`. I'm going to use `Float64Array` everywhere since it makes it pre-allocates the memory and makes it clear how many items it contains.

finetjul: fair enough

bourdaisj: we compute `p1World` and `p2World`. on the other hand, we get `displayCoords` from another API

Sources/Rendering/Core/Picker/index.js
bourdaisj: I assume you mean `model.actors.indexOf`

Sources/Rendering/Core/Picker/index.js
bourdaisj: Yep, will do. 
I need to update it to comply with `CellPicker.intersectWithLine`.

Sources/Rendering/Core/Picker/index.js
bourdaisj: I assume some variables are defined outside of the closure to avoid allocating/destroying every time the function is called by `forEach`?
Is this really a big game changer for performance? What's the policy/best practices that should be applied regarding this in VTK.js?
Naively I would say either we define everything in the closure, either everything outside.

finetjul: There is no rule in VTK.js.
I think it should be declared outside a `for` loop in `time sensitive` functions.
And I agree with you that it should be all or none. Not a mix of it.
Because the picking is `time sensitive` I therefore suggest to do pre-allocation everywhere.

Sources/Rendering/Core/Picker/index.js
```diff
@@ -89,19 +89,144 @@ function vtkPicker(publicAPI, model) {
...
+    //  transformed to actors coordinates!  Reduces overall computation!!!).
+    //  Note that only vtkProp3D's can be picked by vtkPicker.
+    const props = model.pickFromList ? model.pickList : renderer.getActors();
+
+    props.forEach((prop) => {
+      const mapper = prop.getMapper();
+      const pickable =
+        prop.getNestedPickable() &&
+        prop.getNestedVisibility() &&
+        prop.getProperty().getOpacity() > 0.0;
```
finetjul, line 118: are you sure there always a property in a prop ?
In case of image actors, I think `getProperty()` returns `null`, but `getImageProperty()` does not. Isn't it ?

bourdaisj: There should always be a property.  
`getProperty` is defined in both `vtkActor` and `vtkVolume`, which to my knowledge are the only two concrete implementation of `vtkProp3D`.  
Also, it is specified that the function always return `vtkProperty` (for `vtkActor`) or `vtkVolumeProperty` (for `vtkVolume`)
that being said, it is a problem that `getProperty` is not defined in `vtkProp3D` itself, so in theory other subclasses could be created and not implement `getProperty`.  
Lastly, this line is a problem for the call to `getOpacity`, which does not exists on a `vtkVolumeProperty`.  I'm going to look for a fix.

Sources/Rendering/Core/Picker/index.js
bourdaisj: is the `for...of` syntax allowed in VTK.js?

Sources/Rendering/Core/Picker/index.js
bourdaisj: see recent commits

bourdaisj:
@finetjul @floryst re-requesting review

floryst:
LGTM

floryst:
> can we merge someday? I'm not really in a hurry right now but it could change in the next weeks. @finetjul @floryst

Since this is also a breaking change, we can merge it with #3033, which is also a breaking change.

Sources/Rendering/Core/Picker/index.js
```diff
@@ -89,19 +89,148 @@ function vtkPicker(publicAPI, model) {
...
+      // Note that the tolerance is added to the bounding box to make sure things on the edge of the
+      // bounding box are picked correctly.
+      const bounds = mapper
+        ? vtkBoundingBox.inflate(mapper.getBounds(), tolerance)
+        : [...vtkBoundingBox.INIT_BOUNDS];
+
+      if (vtkBoundingBox.intersectBox(bounds, p1Mapper, ray, hitPosition, [])) {
+        mat4.getScaling(transformScale, model.transformMatrix);
+
+        // TODO: better naming?
```
finetjul, line 160: to remove

Sources/Rendering/Core/PointPicker/index.d.ts
```diff
@@ -35,19 +37,21 @@ export interface vtkPointPicker extends vtkPicker {
 	 * 
 	 * @param {Vector3} p1 
 	 * @param {Vector3} p2 
-	 * @param {Number} tol 
+	 * @param {Number} tolerance 
+	 * @param {vtkProp3D} actor
 	 * @param {vtkMapper} mapper 
 	 */
-	intersectWithLine(p1: Vector3, p2: Vector3, tol: number, mapper: vtkMapper): number;
+	intersectWithLine(p1: Vector3, p2: Vector3, tolerance: number, actor: Nullable<vtkProp3D>, mapper: vtkMapper): number;
```
finetjul, line 44: rename `actor` into `prop` to be consistent with `vtkPicker` ?

finetjul (approved):
LGTM

floryst:
A follow-up question regarding the breaking change nature of this PR: should `intersectWithLine` and friends be public API methods? It might be better to have them as protected methods (e.g. `model.intersectWithLine = () => ...`). @finetjul @jourdain

finetjul:
> A follow-up question regarding the breaking change nature of this PR: should `intersectWithLine` and friends be public API methods? It might be better to have them as protected methods (e.g. `model.intersectWithLine = () => ...`). @finetjul @jourdain

They are public in VTK C++

finetjul:
> They look protected in the nightly docs: https://vtk.org/doc/nightly/html/classvtkPointPicker.html#a10445a7288ee6baf61371a64a3e69087

my bad, I checked too fast. 
Then yes, protecting them makes sense.

bourdaisj:
@daker @floryst What is the rule for declaration typing of protected methods?

floryst:
> @daker @floryst What is the rule for declaration typing of protected methods?

There's no need to declare protected methods, since they're not exposed on the public API.

PR #3021, opened 2024-02-23
fix(OpenGL/Framebuffer): size can be null

floryst:
### Context
If the OpenGL render window has [0,0] size, it's possible for the associated framebuffer to not be correctly initialized ([see these lines](https://github.com/Kitware/vtk-js/blob/b9988635d1beed28a1b13f7c5e1ac8b668b77e6c/Sources/Rendering/OpenGL/ForwardPass/index.js#L60-L68)). The resulting error is opaque (errors in `renderPieceStart`) and is not indicative of the root cause.

### Results
If the framebuffer is not yet initialized, return null as the size. This matches some existing usages of `framebuffer.getSize()` that checks for null-ness.

### Changes

- [x] vtkFramebuffer.getSize() now can return null
- [x] Updated usages of framebuffer.getSize() to handle the null case

### PR and Code Checklist

### Testing

### DIFF
--- a/Sources/Rendering/OpenGL/Framebuffer/index.js
+++ b/Sources/Rendering/OpenGL/Framebuffer/index.js
@@ -225,12 +225,8 @@ function vtkFramebuffer(publicAPI, model) {
   };
 
   publicAPI.getSize = () => {
-    const size = [0, 0];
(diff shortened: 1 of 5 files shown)


finetjul (approved):
LGTM

PR #3022, opened 2024-02-23
fix(WidgetManager): call modified on widget remove

floryst:
### Context
The widget manager does not trigger onModified when removing widgets.

### Results
The widget manager will begin triggering onModified when removing widgets.

### Changes

- [x] widget manager changes to trigger onModified on widget removal

### PR and Code Checklist

### Testing


floryst:
Actually, this is not necessary. Removing widgets already calls onModified, and I was observing different behavior. This change actually triggers onModified twice, so I'm closing this PR.

PR #3026, opened 2024-02-29
fix(imagemapper): change slice interaction to use window size instead…

sankhesh:
… of viewport

This will ensure that different sized viewports in the same window will exhibit the same behavior.

### PR and Code Checklist


floryst:
LGTM

PR #3028, opened 2024-03-05
Add MIP to ImageCPRMapper

bruyeret:
Add MIP in CPR with Maximum, Minimum and Average modes.
See the example for a demonstration of how to use MIP.
image

### DIFF
--- a/Sources/Rendering/Core/ImageCPRMapper/Constants.js
+++ b/Sources/Rendering/Core/ImageCPRMapper/Constants.js
@@ -1,0 +1,9 @@
+export const ProjectionMode = {
+  MAX: 0,
+  MIN: 1,
+  AVERAGE: 2,
+};
+
+export default {
+  ProjectionMode,
+};
--- a/Sources/Rendering/Core/ImageCPRMapper/example/index.js
+++ b/Sources/Rendering/Core/ImageCPRMapper/example/index.js
@@ -6,6 +6,7 @@ import '@kitware/vtk.js/Rendering/Profiles/All';
 // Force the loading of HttpDataAccessHelper to support gzip decompression
 import '@kitware/vtk.js/IO/Core/DataAccessHelper/HttpDataAccessHelper';
 
+import { ProjectionMode } from '@kitware/vtk.js/Rendering/Core/ImageCPRMapper/Constants';
 import { radiansFromDegrees } from 'vtk.js/Sources/Common/Core/Math';
 import { updateState } from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers';
 import { vec3, mat3, mat4 } from 'gl-matrix';
@@ -44,8 +45,12 @@ const renderWindow = fullScreenRenderer.getRenderWindow();
 
 fullScreenRenderer.addController(controlPanel);
 const angleEl = document.getElementById('angle');
+const animateEl = document.getElementById('animate');
 const centerlineEl = document.getElementById('centerline');
 const modeEl = document.getElementById('mode');
+const projectionModeEl = document.getElementById('projectionMode');
+const projectionThicknessEl = document.getElementById('projectionThickness');
+const projectionSamplesEl = document.getElementById('projectionSamples');
 
 const interactor = renderWindow.getInteractor();
 interactor.setInteractorStyle(vtkInteractorStyleImage.newInstance());
@@ -354,6 +359,20 @@ angleEl.addEventListener('input', () =>
   setAngleFromSlider(radiansFromDegrees(Number.parseFloat(angleEl.value, 10)))
 );
 
+let animationId;
+animateEl.addEventListener('change', () => {
+  if (animateEl.checked) {
+    animationId = setInterval(() => {
+      const currentAngle = radiansFromDegrees(
+        Number.parseFloat(angleEl.value, 10)
+      );
+      setAngleFromSlider(currentAngle + 0.1);
+    }, 60);
+  } else {
+    clearInterval(animationId);
+  }
+});
+
 function useStraightenedMode() {
   mapper.useStraightenedMode();
   updateDistanceAndDirection();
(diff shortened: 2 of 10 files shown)


Sources/Rendering/Core/ImageCPRMapper/Constants.js
```diff
@@ -0,0 +1,9 @@
+export const MipMode = {
```
finetjul, line 1: Average (i.e. AverageIP) is not a Mip (Maximum Intensity Projection).
Therefore I suggest to rename MipMode into BlendMode, SlabMode or ProjectionMode.

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -157,6 +158,59 @@ export interface vtkImageCPRMapper extends vtkAbstractMapper3D {
...
+	 * Even if the name of the mode is MIP (which implies a maximum intensity projection),
+	 * other modes are available such as MinIP and Average Intensity Projection
+	 */
+	getMipMode(): MipMode;
+
+	/**
+	 * @see getMipMode
+	 * @param mipMode
+	 */
+	setMipMode(mipMode: MipMode): boolean;
```
finetjul, line 212: Conventions say that acronyms must not be camel-cased. it is MIP or mip, but never Mip

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -157,6 +158,59 @@ export interface vtkImageCPRMapper extends vtkAbstractMapper3D {
...
+	 * Using an odd number is advised
+	 * If this number is even, the center of the slab will not be sampled
+	 */
+	getMipNumberOfSamples(): number;
+
+	/**
+	 * @see getMipNumberOfSamples
+	 * @param mipNumberOfSamples
+	 */
+	setMipNumberOfSamples(mipNumberOfSamples: number): boolean;
```
finetjul, line 199: setSlabNumberOfSamples

Sources/Rendering/Core/ImageCPRMapper/index.d.ts
```diff
@@ -157,6 +158,59 @@ export interface vtkImageCPRMapper extends vtkAbstractMapper3D {
...
+	 * @see getMipNumberOfSamples
+	 * @see getMipMode
+	 */
+	getUseMip(): boolean;
+
+	/**
+	 * @see getUseMip
+	 * @param useMip
+	 */
+	setUseMip(useMip: boolean): boolean;
```
finetjul, line 174: couldn't it be automatically assumed by the number of samples ? (if 1 it means it uses mip)

Sources/Rendering/Core/ImageCPRMapper/index.js
```diff
@@ -103,7 +104,10 @@ function vtkImageCPRMapper(publicAPI, model) {
       : orientationDataArray.getNumberOfComponents();
     switch (numComps) {
       case 16:
-        convert = mat4.getRotation;
+        convert = (outQuat, inMat) => {
+          mat4.getRotation(outQuat, inMat);
+          quat.normalize(outQuat, outQuat);
+        };
```
finetjul, line 110: shouln't this go into another commit (same PR) ?

Sources/Rendering/Core/ImageCPRMapper/Constants.ts
daker, line 1: Constants.d.ts

Sources/Rendering/Core/ImageCPRMapper/example/index.js
```diff
@@ -29,6 +29,7 @@ import widgetBehavior from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget
 import controlPanel from './controller.html';
 import aortaJSON from './aorta_centerline.json';
 import spineJSON from './spine_centerline.json';
+import { ProjectionMode } from '../Constants';
```
daker, line 32: `import { ProjectionMode } from 'vtk.js/Sources/Rendering/Core/ImageCPRMapper/Constants';`

daker (changes requested):

Sources/Rendering/Core/ImageCPRMapper/example/index.js
bruyeret: Thank you for the review!
Is there a convention to always use absolute import?
I prefer using relative imports when the relationship between the two files is not likely to change in case of a refactor
For example:
```
|-- Utils/
|   |-- index.js
|-- Classes/
|   |-- MyClass/
|   |   |-- index.js
```
Here, from `Classes/MyClass/index.js`, I import `Utils/index.js` instead of `../../Utils/index.js` because the position of the `Utils` folder is unlikely to change, but `MyClass` could be for example renamed or moved

What do you think?

daker: Displaying relative will not help developers, the code of the examples is displayed on the [website ](https://kitware.github.io/vtk-js/examples/ImageCPRMapper.html) as a "starter" using the recommended version of vtk.js which is ESM, we can do relative import but will not to add another step to rewrite the import into ESM version import

*.js -> UMD version vtk.js/Sources
example/*.js -> ESM version @kitware/vtk.js
*.d.ts -> Relative import [rewritten ](https://github.com/Kitware/vtk-js/blob/master/rollup.config.js#L180-L207) into ESM at build

@floryst can give more insights

bruyeret: Oh ok yes it makes a lot of sense for the example, I didn't think about it, thanks!

finetjul (approved):
LGTM

Sources/Rendering/Core/ImageCPRMapper/example/index.js
```diff
@@ -6,6 +6,7 @@ import '@kitware/vtk.js/Rendering/Profiles/All';
 // Force the loading of HttpDataAccessHelper to support gzip decompression
 import '@kitware/vtk.js/IO/Core/DataAccessHelper/HttpDataAccessHelper';
 
+import { ProjectionMode } from '@kitware/vtk.js/Rendering/Core/ImageCPRMapper/Constants';
 import { radiansFromDegrees } from 'vtk.js/Sources/Common/Core/Math';
 import { updateState } from 'vtk.js/Sources/Widgets/Widgets3D/ResliceCursorWidget/helpers';
```
daker, line 11: can you also change all the import to `@kitware/vtk.js` ?

daker (approved):
LGTM

PR #3029, opened 2024-03-06
Add custom color mix functions for volume mapper fragment shader

bruyeret:
The custom shader code is used to merge blend components
Presets can be used, I added an `ADDITING` preset that is useful when using medical images with labels
Also added a test with 2 baselines


Sources/Rendering/Core/VolumeMapper/test/testColorMix.js
```diff
@@ -0,0 +1,207 @@
...
+          y <= 0.7 * dims[1] &&
+          z >= 0.3 * dims[2] &&
+          z <= 0.7 * dims[2];
+        newData[iTuple * nComp + 1] = isInMask ? 1 : 0;
+      }
+    }
+  }
+  array.setData(newData);
+
+  const maskCtfun = vtkColorTransferFunction.newInstance();
```
finetjul, line 97: why not `registerResources` ?

bruyeret:
Just need to add a preset for `COLORIZING` to be ready

Sources/Rendering/Core/VolumeProperty/index.js
```diff
@@ -203,6 +203,30 @@ function vtkVolumeProperty(publicAPI, model) {
...
+          float opacitySum = opacity0 + opacity1;
+          if (opacitySum == 0.0) {
+            return vec4(0.0);
+          }
+          #if vtkLightComplexity > 0
+            applyLighting(tColor0, normal0);
+            applyLighting(tColor1, normal1);
+          #endif
+          vec3 mixedColor = mix(tColor0, tColor1, opacity1 / opacitySum);
+          return vec4(mixedColor, opacitySum);`;
```
finetjul, line 224: this WebGL code should not be in vtkVolumeProperty. Instead, it should go in the `OpenGL/VolumeMapper` file.

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -44,6 +44,14 @@ function arrayEquals(a, b) {
   return true;
 }
 
+// TODO: Do we want this in some shared utility? Shouldwe just use lodash.isEqual
+function deepEqual(a, b) {
+  return (
+    typeof a === typeof b &&
+    (typeof a !== 'object' || arrayEquals(Object.entries(a), Object.entries(b)))
+  );
+}
```
finetjul, line 53: why not `import DeepEqual from 'fast-deep-equal';`?

Sources/Rendering/Core/VolumeProperty/Constants.d.ts
```diff
@@ -9,8 +9,13 @@ export declare enum OpacityMode {
 	PROPORTIONAL = 1,
 }
 
+export declare enum ColorMixPreset {
+	ADDITING = 0,
```
sankhesh, line 13: Calling this `ADDITIVE` would be more appropriate and more in-line with the volume mapper blend modes.

Sources/Rendering/Core/VolumeProperty/index.js
```diff
@@ -203,6 +203,30 @@ function vtkVolumeProperty(publicAPI, model) {
...
+          float opacity1 = goFactor.y * pwfValue1;
+          float opacitySum = opacity0 + opacity1;
+          if (opacitySum == 0.0) {
+            return vec4(0.0);
+          }
+          #if vtkLightComplexity > 0
+            applyLighting(tColor0, normal0);
+            applyLighting(tColor1, normal1);
+          #endif
+          vec3 mixedColor = mix(tColor0, tColor1, opacity1 / opacitySum);
```
sankhesh, line 223: Maybe I'm not understanding the big picture here.

How is this different from the current code where independent components are alpha-blended?
Secondly, this is not doing an additive blend, is it?

bruyeret: If we have alphas `a0` and `a1`, and colors `c0` and `c1`, then:
With alpha blending:
```
color = (c0 * a0 + (1 - a0) * c1 * a1 ) / alpha
alpha = a0 + (1 - a0) * a1
```
With additive mode:
```
color = (c0 * a0 + c1 * a1) / alpha
alpha = a0 + a1
```

Also how to create alpha blending with independent components in the current shader?

Do you suggest a different name instead or additive mode ?

Sources/Rendering/Core/VolumeProperty/index.js
```diff
@@ -203,6 +203,16 @@ function vtkVolumeProperty(publicAPI, model) {
   publicAPI.getInterpolationTypeAsString = () =>
     macro.enumToString(InterpolationType, model.interpolationType);
 
+  publicAPI.getColorMixCode = (apiSpecificPresets) => {
```
finetjul, line 206: I don't think we want such type of "API code" function here.
I suggest VolumeProperty only store a preset value.
For custom code (which will be rarely used), I think this can be done with the "shader replacement" mechanisnm. So you could have 3 presets: "ADDITING/ADDITIVE", "COLORIZING" and "CUSTOM". The custom preset simply adds a "comment" that can be replaced using the shader replacement.
WDYT ?

finetjul (approved):
LGTM

finetjul:
@sankhesh I had to merge it, but feel free to add comments if needed. @bruyeret can fix them in a following PR.

PR #3031, opened 2024-03-15
refactor(AxesActor): ability to invert axis

rodrigobasilio2022:
### Context

This PR aims to enable the ability to invert axis in AxesActor. I creates an example, named AxesActor, that one can test the functionality.

### Changes

### PR and Code Checklist

### Testing

- [] Tested environment:
  - **vtk.js**:  latest
  - **OS**:  Windows 11
  - **Browser**:  Chrome

### DIFF
--- a/Examples/Geometry/AxesActor/index.js
+++ b/Examples/Geometry/AxesActor/index.js
@@ -1,0 +1,80 @@
+import '@kitware/vtk.js/favicon';
+
+// Load the rendering pieces we want to use (for both WebGL and WebGPU)
+import '@kitware/vtk.js/Rendering/Profiles/Geometry';
+
+import macro from '@kitware/vtk.js/macros';
+import vtkAxesActor from '@kitware/vtk.js/Rendering/Core/AxesActor';
+import vtkFullScreenRenderWindow from '@kitware/vtk.js/Rendering/Misc/FullScreenRenderWindow';
+
+import controlPanel from './controlPanel.html';
+
+console.warn(
+  'Click on index.ts to open source code for this example --------->'
+);
+
+// ----------------------------------------------------------------------------
+// Standard rendering code setup
+// ----------------------------------------------------------------------------
+
+const fullScreenRenderer = vtkFullScreenRenderWindow.newInstance({
+  background: [0.2, 0.3, 0.4],
+});
+const renderer = fullScreenRenderer.getRenderer();
+const renderWindow = fullScreenRenderer.getRenderWindow();
+
+const axesActor = vtkAxesActor.newInstance();
+renderer.addActor(axesActor);
+
+renderer.resetCamera();
+renderWindow.render();
+
+// ----------------------------------------------------------------------------
+// UI control handling
+// ----------------------------------------------------------------------------
+
+fullScreenRenderer.addController(controlPanel);
+
+function updateRendering() {
+  axesActor.update();
+  renderer.resetCameraClippingRange();
+  renderWindow.render();
+}
+
+document.querySelector('.recenter').addEventListener('change', (e) => {
+  const config = axesActor.getConfig();
+  config.recenter = !!e.target.checked;
+  axesActor.setConfig(config);
+  updateRendering();
+});
+
+document.querySelector('.xAxisInvert').addEventListener('change', (e) => {
+  const config = axesActor.getXConfig();
+  config.invert = !!e.target.checked;
+  axesActor.setXConfig(config);
+  updateRendering();
+});
+
+document.querySelector('.yAxisInvert').addEventListener('change', (e) => {
+  const config = axesActor.getYConfig();
+  config.invert = !!e.target.checked;
+  axesActor.setYConfig(config);
+  updateRendering();
+});
+
+document.querySelector('.zAxisInvert').addEventListener('change', (e) => {
+  const config = axesActor.getZConfig();
+  config.invert = !!e.target.checked;
+  axesActor.setZConfig(config);
+  updateRendering();
+});
+
+// -----------------------------------------------------------
+// Make some variables global so that you can inspect and
+// modify objects in your browser's developer console:
+// -----------------------------------------------------------
+
+global.setLoggerFunction = macro.setLoggerFunction;
+global.axesActor = axesActor;
+global.renderer = renderer;
+global.renderWindow = renderWindow;
--- a/Sources/Rendering/Core/AxesActor/index.js
+++ b/Sources/Rendering/Core/AxesActor/index.js
@@ -21,10 +21,14 @@ function centerDataSet(ds) {
     .apply(ds.getPoints().getData());
 }
 
-function shiftDataset(ds, axis) {
+function shiftDataset(ds, axis, invert = false) {
   const bounds = ds.getPoints().getBounds();
   const center = [0, 0, 0];
-  center[axis] = -bounds[axis * 2];
+  if (invert) {
+    center[axis] = -bounds[axis * 2 + 1];
+  } else {
+    center[axis] = -bounds[axis * 2];
+  }
   vtkMatrixBuilder
     .buildFromDegree()
     .translate(...center)
@@ -65,36 +69,48 @@ function vtkAxesActor(publicAPI, model) {
   publicAPI.setMapper(_mapper);
 
   publicAPI.update = () => {
+    let currentConfig = {
+      ...model.config,
+      ...model.xConfig,
+    };
+
     const xAxis = vtkArrowSource
-      .newInstance({ direction: [1, 0, 0], ...model.config })
+      .newInstance({ direction: [1, 0, 0], ...currentConfig })
       .getOutputData();
     if (model.config.recenter) {
       centerDataSet(xAxis);
     } else {
-      shiftDataset(xAxis, 0);
+      shiftDataset(xAxis, 0, currentConfig.invert);
     }
-    addColor(xAxis, ...model.xAxisColor);
+    addColor(xAxis, ...currentConfig.color);
 
+    currentConfig = {
+      ...model.config,
+      ...model.yConfig,
+    };
     const yAxis = vtkArrowSource
-      .newInstance({ direction: [0, 1, 0], ...model.config })
+      .newInstance({ direction: [0, 1, 0], ...currentConfig })
       .getOutputData();
     if (model.config.recenter) {
       centerDataSet(yAxis);
     } else {
-      shiftDataset(yAxis, 1);
+      shiftDataset(yAxis, 1, currentConfig.invert);
     }
+    addColor(yAxis, ...currentConfig.color);
 
-    addColor(yAxis, ...model.yAxisColor);
-
+    currentConfig = {
+      ...model.config,
+      ...model.zConfig,
+    };
     const zAxis = vtkArrowSource
-      .newInstance({ direction: [0, 0, 1], ...model.config })
+      .newInstance({ direction: [0, 0, 1], ...currentConfig })
       .getOutputData();
     if (model.config.recenter) {
       centerDataSet(zAxis);
     } else {
-      shiftDataset(zAxis, 2);
+      shiftDataset(zAxis, 2, currentConfig.invert);
     }
-    addColor(zAxis, ...model.zAxisColor);
+    addColor(zAxis, ...currentConfig.color);
 
     const source = vtkAppendPolyData.newInstance();
     source.setInputData(xAxis);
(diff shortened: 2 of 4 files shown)


Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -35,7 +39,7 @@ function shiftDataset(ds, axis) {
 
 function addColor(ds, r, g, b) {
   const size = ds.getPoints().getData().length;
-  const rgbArray = new Uint8ClampedArray(size);
+  const rgbArray = new Uint8Array(size);
```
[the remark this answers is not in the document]
rodrigobasilio2022: The clamp array was affecting the colors for some reason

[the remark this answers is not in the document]
finetjul: this should be part of a separate PR. (would need to be investigated a bit more...)

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -65,34 +69,50 @@ function vtkAxesActor(publicAPI, model) {
...
+      direction = [-1, 0, 0];
+    } else {
+      direction = [1, 0, 0];
+    }
     const xAxis = vtkArrowSource
-      .newInstance({ direction: [1, 0, 0], ...model.config })
+      .newInstance({ direction, ...model.config })
       .getOutputData();
     if (model.config.recenter) {
       centerDataSet(xAxis);
```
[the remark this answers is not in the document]
rodrigobasilio2022: I need the inversion parameter in the shiftDataset because just changing the direction could end up in a axis being dislocated, detached from the others. This not happens with the centerDataset because of the centering operation

Sources/Rendering/Core/AxesActor/index.js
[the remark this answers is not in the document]
rodrigobasilio2022: Ok. I undo the change

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -169,7 +191,12 @@ export function extend(publicAPI, model, initialValues = {}) {
   // Inheritance
   vtkActor.extend(publicAPI, model, initialValues);
 
-  macro.setGet(publicAPI, model, ['config']);
+  macro.setGet(publicAPI, model, [
+    'config',
+    'xAxisInvert',
+    'yAxisInvert',
+    'zAxisInvert',
+  ]);
```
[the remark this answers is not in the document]
rodrigobasilio2022: Done it

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -154,11 +174,13 @@ export const DEFAULT_VALUES = {
     tipLength: 0.2,
     shaftResolution: 60,
     shaftRadius: 0.03,
-    invert: false,
```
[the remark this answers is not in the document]
rodrigobasilio2022: Done the separate config for each axis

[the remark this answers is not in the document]
rodrigobasilio2022: I agree with separated config for each axis and propably the place to put the label in the future...

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -65,36 +69,43 @@ function vtkAxesActor(publicAPI, model) {
   publicAPI.setMapper(_mapper);
 
   publicAPI.update = () => {
+    let direction;
+    let config = model.xConfig || model.config;
+    direction = [config.invert ? -1 : 1, 0, 0];
     const xAxis = vtkArrowSource
-      .newInstance({ direction: [1, 0, 0], ...model.config })
+      .newInstance({ direction, ...model.config })
```
finetjul, line 76: I meant something as such:
```
const xAxis = vtkArrowSource.newInstance({ direction: [1, 0, 0], ...model.config, ...model.xConfig })
```

Sources/Rendering/Core/AxesActor/index.js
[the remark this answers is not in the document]
finetjul: You must use `defaultValue()` instead of `DEFAULT_VALUE` (find examples in VTK.js code base).
Otherwise your configs will be shared between ALL instances.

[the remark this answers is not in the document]
finetjul: You changed the API of the class. Hence it is no longer a `refactor` commit, it is a BREAKING_CHANGE commit and requires a BUMP of VTK.js major version.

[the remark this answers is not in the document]
rodrigobasilio2022: Sorry, i've incorrectly interpret your review

rodrigobasilio2022:
@finetjul Sorry I interpret incorrectly your messages. I rewrite the code, undo some changes and made as expect, I hope. One thing I didnt agree, but understand, is to leave the incorrect color assignment to other PR. 
I think we should create a new PR just to undo the UInt8 ->UInt8Clamped change to fix arrow color bug and leave the proper debugging for this issue to a future PR.

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -146,30 +186,47 @@ function vtkAxesActor(publicAPI, model) {
...
+    },
+    xConfig: {},
+    yConfig: {},
+    zConfig: {},
+    xAxisColor: [255, 0, 0],
+    yAxisColor: [255, 255, 0],
+    zAxisColor: [0, 128, 0],
+    xAxisInvert: false,
+    yAxisInvert: false,
+    zAxisInvert: false,
```
finetjul, line 208: I do not think those xyzAxisInvert are needed anymore.

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -146,30 +186,47 @@ function vtkAxesActor(publicAPI, model) {
...
+    ...initialValues,
```
finetjul, line 209: I think it should be more like:
```
return {
  xConfig: {
    color: [255, 0, 0],
    invert: false,
    ...initialValues.?xConfig
  },
  yConfig: {
    color: [255, 255, 0],
    invert: false,
    ...initialValues.?yConfig
  },
  zConfig: {
    color: [255, 0, 0],
    invert: false,
    ...initialValues.?zConfig
  },
  config: {
     recenter: true,
     [...]
     ...initialValues.?config
   }
}
```

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -117,24 +133,48 @@ function vtkAxesActor(publicAPI, model) {
...
 
-  publicAPI.setXAxisColor = (c) => {
-    if (setXAxisColor(c)) {
+  publicAPI.setXConfig = (c) => {
+    if (setXConfig(c)) {
+      _debouncedUpdate();
+      return true;
+    }
+    return false;
+  };
```
finetjul, line 142: there is now better than overriding setters:
```
model._onXConfigChanged = _debouncedUpdate;
model._onYConfigChanged = _debouncedUpdate;
model._onZConfigChanged = _debouncedUpdate;
```
they will be automatically called when `setXConfig()` is change the model value (see `macros.js:findSetter:getSetter::setter`)

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -65,34 +69,38 @@ function vtkAxesActor(publicAPI, model) {
...
     if (model.config.recenter) {
       centerDataSet(yAxis);
     } else {
-      shiftDataset(yAxis, 1);
+      shiftDataset(yAxis, 1, model.yAxisInvert);
     }
 
     addColor(yAxis, ...model.yAxisColor);
 
+    direction = [0, 0, model.zAxisInvert ? -1 : 1];
```
finetjul, line 96: if `vtkArrowSource` got `invert: true` (because `model.zConfig` contains `invert: true`, then you do not need to flip the `direction` anymore.

finetjul (changes requested):

Examples/Geometry/AxesActor/index.js
```diff
@@ -0,0 +1,80 @@
+import '@kitware/vtk.js/favicon';
+
+// Load the rendering pieces we want to use (for both WebGL and WebGPU)
+import '@kitware/vtk.js/Rendering/Profiles/Geometry';
+
+import macro from '@kitware/vtk.js/macros';
+import vtkAxesActor from 'vtk.js/Sources/Rendering/Core/AxesActor';
```
finetjul, line 7: @kitware/vtk.js...

Examples/Geometry/AxesActor/index.js
```diff
@@ -0,0 +1,80 @@
...
+console.warn(
+  'Click on index.ts to open source code for this example --------->'
+);
+
+// ----------------------------------------------------------------------------
+// Standard rendering code setup
+// ----------------------------------------------------------------------------
+
+const fullScreenRenderer = vtkFullScreenRenderWindow.newInstance({
+  background: [0.9, 0.9, 0.9],
```
finetjul, line 21: only 2 examples have this background color. Can you please choose another color that is more popular in the code base (0, 0.1, 0.3, 0.5 or 1).

(Eventually we should change the other 2 instances as well...)

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -107,76 +123,88 @@ function vtkAxesActor(publicAPI, model) {
...
+
+  const { setConfig } = publicAPI;
+
+  publicAPI.setConfig = (c) => {
+    if (setConfig(c)) {
       _debouncedUpdate();
       return true;
     }
     return false;
   };
```
finetjul, line 158: I think if you do `model._onConfigChanged = _debouncedUpdate();`, you don't have to overwrite `setConfig()`, don't you ?

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -107,76 +123,88 @@ function vtkAxesActor(publicAPI, model) {
...
-  const { setConfig, setXAxisColor, setYAxisColor, setZAxisColor } = publicAPI;
-
-  publicAPI.setConfig = (c) => {
-    if (setConfig(c)) {
-      _debouncedUpdate();
-      return true;
-    }
-    return false;
+  publicAPI.setXAxisColor = (color) => {
+    model.xConfig.color = color;
```
finetjul, line 127:
```
publicAPI.setXAxisColor = (color) =>
  publicAPI.setXConfig({...publicAPI.getXConfig(), color});
```

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -107,76 +123,88 @@ function vtkAxesActor(publicAPI, model) {
...
-    return false;
+  publicAPI.setZAxisColor = (color) => {
+    model.zConfig.color = color;
+    _debouncedUpdate();
   };
 
-  publicAPI.setZAxisColor = (c) => {
-    if (setZAxisColor(c)) {
+  publicAPI.getXAxisColor = (color) =>
+    model.xConfig?.color || model.config?.color;
```
finetjul, line 142:
```
publicAPI.getXAxisColor = () =>
    model.getXConfig().color;
```

finetjul (changes requested):

finetjul (changes requested):
LGTM, you need to update `index.d.ts` accordingly though.
(and actually, some  `set...From()` and `get...ByReference()` methods that were automatically generated by `macros.setGetArray()`

rodrigobasilio2022:
@finetjul Changed the index.d.ts

finetjul:
> @finetjul Changed the index.d.ts

Thank you, but you did not add `setXConfig()` and co into index.d.ts

rodrigobasilio2022:
I thought the macros.setGetArray() would handle them... I didthe npm run build and upload...

rodrigobasilio2022:
@finet Done

Sources/Rendering/Core/AxesActor/index.js
```diff
@@ -107,76 +123,70 @@ function vtkAxesActor(publicAPI, model) {
...
+      ...initialValues?.zConfig,
+    },
+  };
+}
 
 // ----------------------------------------------------------------------------
 
 export function extend(publicAPI, model, initialValues = {}) {
-  Object.assign(model, DEFAULT_VALUES, initialValues);
+  Object.assign(model, defaultValues(initialValues));
```
finetjul, line 184: this line is not needed, if you already pass `defaultValues()` in the `extend()` function below

rodrigobasilio2022:
@finetjul I think thisi s it...

finetjul (approved):

finetjul:
One last comment (thanks for your patience): you need to rephrase your commit message. It is not a refactor:
`refactor(AxesActor): ability to invert axis`
=>
`feat(AxesActor): add option to flip axes individually` 
or something like that

rodrigobasilio2022:
@finetjul Done

PR #3033, opened 2024-03-18
fix(reslicecursorwidget): fix camera reset on rotation after a camera…

finetjul:
… translation

BREAKING CHANGE: ResliceCursorWidget.interactionEvent has been modified

### Context
When the user translates a camera (shift+left drag), and then rotate the reslice cursor widget in another view, the camera was reset. This is because the focal point offsets were not "stored" after the camera translation. The rotation was therefore resetting the camera translation. 

### Results
A solution would have been to observe the interactor style each time the camera is manipulated. Instead, we prefer to save the focal point offsets before any reslice cursor widget interaction happens, namely when startInteraction is triggered. 

### Changes
This makes computeFocalPointOffset parameter in the reslice cursor widget interactionEvent obsolete. It has therefore been removed along with the associated canUpdateFocalPoint. This is not the behavior's responsibility to tell the event consumers what they must do.

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows
  - **Browser**: Chrome


Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -405,7 +431,7 @@ reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
                 renderer: obj.renderer,
                 resetFocalPoint: false,
                 keepFocalPointPosition,
-                computeFocalPointOffset,
+                computeFocalPointOffset: !keepFocalPointPosition,
```
bruyeret, line 434: It looks like `keepFocalPointPosition` is always `!computeFocalPointOffset`, does it make sens to keep both variables?
It would make sense to say that if we don't keep focal position, then we always need to compute focal point offset

bruyeret (approved):

bruyeret:
> @bruyeret could I have broken the CPR widget ?

I tested and it doesn't look broken, I don't think that the change could affect it

Sources/Widgets/Widgets3D/ResliceCursorWidget/example/index.js
```diff
@@ -45,7 +49,10 @@ const viewColors = [
 ];
 
 const viewAttributes = [];
+window.va = viewAttributes;
 const widget = vtkResliceCursorWidget.newInstance();
+window.widget = widget;
+window.splineWidget = vtkSplineWidget.newInstance();
```
floryst, line 55: Just checking to make sure you didn't mean to commit these debugging lines.

floryst:
LGTM

floryst:
Whenever this is ready to merge, we should merge it with #3020 to cut a single major release, instead of two.

PR #3038, opened 2024-03-22
Fix projection slab thickness unit in ImageCPRMapper

bruyeret:
Documentation says for the projection slab thickness:

>  Thickness of the projection slab in image coordinates (NOT in voxels)
> Usually in millimeters if the spacing of the input image is set from a DICOM

But the implementation used voxels as unit which doesn't make a lot of sense as it can vary along axes

This PR fixes the unit of this value and the example accordingly.

### DIFF
--- a/Sources/Rendering/Core/ImageCPRMapper/example/index.js
+++ b/Sources/Rendering/Core/ImageCPRMapper/example/index.js
@@ -421,8 +421,15 @@ projectionModeEl.addEventListener('input', (ev) => {
 });
 
 projectionThicknessEl.addEventListener('input', (ev) => {
-  const thickness = Number.parseFloat(projectionThicknessEl.value, 10);
-  mapper.setProjectionSlabThickness(thickness);
+  const thicknessRatio = Number.parseFloat(projectionThicknessEl.value, 10);
+  const image = mapper.getInputData();
+  if (image) {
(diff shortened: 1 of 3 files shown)


Sources/Rendering/Core/ImageCPRMapper/example/index.js
```diff
@@ -421,8 +421,15 @@ projectionModeEl.addEventListener('input', (ev) => {
...
-  const thickness = Number.parseFloat(projectionThicknessEl.value, 10);
-  mapper.setProjectionSlabThickness(thickness);
+  const thicknessRatio = Number.parseFloat(projectionThicknessEl.value, 10);
+  const image = mapper.getInputData();
+  if (image) {
+    const spacing = image.getSpacing();
+    const dimensions = image.getDimensions();
+    const diagonal = vec3.len(vec3.mul([], spacing, dimensions));
+    const thickness = diagonal * thicknessRatio;
+    mapper.setProjectionSlabThickness(thickness);
```
finetjul, line 431: shouldn't you update the documentation to explain the expected unit

bruyeret: The expected unit was right but the implementation was wrong
I fixed the implementation to make it coherent with the documentation

PR #3039, opened 2024-03-22
Fix vtkAxesActor types

LeviEyal:
### Context
The type definition of the following methods of vtkAxesActor were modified to support the current required parameters:

- setXAxisColor
- setYAxisColor
- setZAxisColor

##### The correct usage of the methods:
```ts
axes.setZAxisColor([255, 0, 0])
```

But the type definition required to supply r,g,b as three arguments, instead of inside an array

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


finetjul:
Thanks for the catch.
Because the recent change is 1 commit before the new major release, let's "pretend" it was a breaking change hence part of the new major release.
Would you mind updating the section `From 29.x to 30` of BREAKING_CHANGES.md accordingly ?

LeviEyal:
> Thanks for the catch. Because the recent change is 1 commit before the new major release, let's "pretend" it was a breaking change hence part of the new major release. Would you mind updating the section `From 29.x to 30` of BREAKING_CHANGES.md accordingly ?

Done

daker (approved):
LGTM

finetjul (approved):

PR #3045, opened 2024-03-25
feat(WidgetManager): customize cursor styles

floryst:
### Context
https://discourse.vtk.org/t/cannot-change-default-cursor-or-reslicecursorwidget/13609/3

The widget manager doesn't support customizing the default cursor styles.

### Results

The widget manager now exposes an API to configure the default + hover cursor styles.

### Changes

- `vtkWidgetManager.setCursorStyles()` API added.

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -151,7 +151,12 @@ function vtkWidgetManager(publicAPI, model) {
     }
 
     // Default cursor behavior
-    model._apiSpecificRenderWindow.setCursor(widget ? 'pointer' : 'default');
+    const cursorStyles = publicAPI.getCursorStyles();
+    const style = widget ? 'hover' : 'default';
+    const cursor = cursorStyles[style];
+    if (cursor) {
+      model._apiSpecificRenderWindow.setCursor(cursor);
+    }
 
     model.activeWidget = null;
     let wantRender = false;
@@ -461,7 +466,7 @@ function vtkWidgetManager(publicAPI, model) {
 // Object factory
 // ----------------------------------------------------------------------------
 
-const DEFAULT_VALUES = {
+const defaultValues = (initialValues = {}) => ({
   // _camera: null,
   // _selector: null,
   // _currentUpdateSelectionCallID: null,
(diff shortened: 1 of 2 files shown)


Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -476,6 +481,10 @@ const DEFAULT_VALUES = {
   previousSelectedData: null,
   widgetInFocus: null,
   captureOn: CaptureOn.MOUSE_MOVE,
+  cursorStyles: {
+    default: 'default',
+    hover: 'pointer',
+  },
```
finetjul, line 487: you must use the `defaultValues()` function notation if you have an object as value, otherwise it will be shared between ALL the widget manager instances.

finetjul (changes requested):

Sources/Widgets/Core/WidgetManager/index.js
floryst: Thanks for catching that, fixed.

Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -476,16 +481,25 @@ const DEFAULT_VALUES = {
...
   captureOn: CaptureOn.MOUSE_MOVE,
-};
+  ...initialValues,
+
+  cursorStyles: initialValues.cursorStyles
+    ? { ...initialValues.cursorStyles }
+    : {
+        default: 'default',
+        hover: 'pointer',
+      },
```
finetjul, line 491:
```
cursorStyles: {
 default: 'default',
 hover: 'pointer',
 ...initialValues.cursorStyles
}
```

finetjul (approved):

Sources/Widgets/Core/WidgetManager/index.js
floryst: I'm doing a ternary here because I've defined `setCursorStyles` to be all-or-nothing. If the user manages the styles externally, then they would want to call `manager.setCursorStyles({})`.

finetjul: fair enough

PR #3058, opened 2024-04-11
fix(ImageOutline): allow zero outline by hiding it

sedghi:
### Context
There seems to be a left out debugging statement (by me), in the older PR which prevents the zero outline value
One should be able to turn off the outline for one segment by setting it to 0. 

`labelMap.actor.getProperty().setLabelOutlineThickness([0, 3]);`

Right now it makes it blue (meant for debugging)

CleanShot 2024-04-11 at 14 25 38@2x

After 

CleanShot 2024-04-11 at 14 31 02@2x

### Results
You can now set the outline to 0 and the outline for that segment will be off

### Changes
Removed the debugging if statement

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: Latest
  - **OS**: macOS Ventura
  - **Browser**: Latest Chrome

### DIFF
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -351,11 +351,6 @@ function vtkOpenGLImageMapper(publicAPI, model) {
                     return;
                   }
 
-                  if (actualThickness == 0) {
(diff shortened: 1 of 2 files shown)


finetjul (approved):
LGTM

PR #3060, opened 2024-04-16
Fix ImageResliceMapper and improve example

### LINKED ISSUE #3059: [Bug] Issues in PR(Share GPU ressources across render windows)
### Bug description

About PR(https://github.com/Kitware/vtk-js/pull/3051), It is a great solution for share GPU memory in vtk.js . I tried and  find two issues. 
1. renderer.getRenderWindow().getViews() return [] for some classes. If i use pick() in Picker.js, it will throw error "Uncaught TypeError: Cannot read properties of undefined (reading 'getViewportSize')" , I think 'renderWindow.addView(renderWindowView)' is missing after 'mainRenderWindowView.addMissingNode(renderWindow)' called

2. ‘ImageResliceMapper’should also be changed same as  'VolumeMapper', beacause it's also use GPU to render MPR images.

### Steps to reproduce

1. use  ‘ImageResliceMapper’ to create  overy two MPR windows
2. check is the GPU memory(texture) shared between two MPR windows
3. call 'renderer.getRenderWindow().getViews()' , it will return empty array

### Detailed Behavior

_No response_

### Expected Behavior

1.  GPU memroy should be shared
2.  'renderer.getRenderWindow().getViews()'  should not be empty

### Environment

- vtk.js version: 30.4.0
- Browsers: chrome
- OS: windows11

bruyeret:
Fixes the issues mentionned in #3059

The picking has been tested and should work fine.

I kept it in a stash but it is maybe better to share it in a branch:
I modified the HW selector test to show that picking still works in the children render windows [in my fork](https://github.com/bruyeret/vtk-js/tree/picking-multi-render-window).

Close #3059

### DIFF
--- a/Examples/Rendering/ManyRenderWindows/index.js
+++ b/Examples/Rendering/ManyRenderWindows/index.js
@@ -14,6 +14,12 @@ import vtkRenderWindowInteractor from '@kitware/vtk.js/Rendering/Core/RenderWind
 import vtkVolume from '@kitware/vtk.js/Rendering/Core/Volume';
 import vtkVolumeMapper from '@kitware/vtk.js/Rendering/Core/VolumeMapper';
 
+import vtkImageSlice from '@kitware/vtk.js/Rendering/Core/ImageSlice';
+import vtkImageResliceMapper from '@kitware/vtk.js/Rendering/Core/ImageResliceMapper';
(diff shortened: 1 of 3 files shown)


finetjul (approved):
LGTM

PR #3061, opened 2024-04-21
fix(colortransferfunction): add access to set/get Scale

antonskrebkov:
In this case issue was created in vtk discourse

Discourse post thread: https://discourse.vtk.org/t/set-desired-scale-in-vtkcolortransferfunction/13736

Results:
Now you can access the Scale variable.

Changes:
Simply added access to changing the value of the Scale variable by the user with getScale() and setScale() in vtkColorTransferFunction, which was not there for some reason.

### DIFF
--- a/Sources/Rendering/Core/ColorTransferFunction/index.js
+++ b/Sources/Rendering/Core/ColorTransferFunction/index.js
@@ -1324,9 +1324,10 @@ export function extend(publicAPI, model, initialValues = {}) {
   macro.setGet(publicAPI, model, [
     'useAboveRangeColor',
     'useBelowRangeColor',
-    'colorSpace',
     'discretize',
     'numberOfValues',
+    { type: 'enum', name: 'colorSpace', enum: ColorSpace },
+    { type: 'enum', name: 'scale', enum: Scale },
   ]);
 
   macro.setArray(


finetjul:
LGTM.

I suggest you rename your commit though: 
 - fix(colortransferfunction) -> fix(colortransferfunction)
 - added access for changing Scale using set -> add access to set/get Scale

antonskrebkov:
> LGTM.
> 
> I suggest you rename your commit though:
> 
> * fix(colortransferfunction) -> fix(colortransferfunction)
> * added access for changing Scale using set -> add access to set/get Scale

Thanks, the commit name has been corrected

finetjul (approved):
LGTM

antonskrebkov:
@finetjul , may I ask you when these changes will be included to the library?

finetjul:
Now :-) (sorry, I skipped it)

antonskrebkov:
@finetjul , it looks like something went wrong during the merge: https://github.com/Kitware/vtk-js/commit/f1dc4a2cd5865637db8bea6c20633e060f03f2d2. If so, what could be the reason for failing the check?

finetjul:
maybe the release failed, but the next released worked. So it is available as of [v30.5.0](https://github.com/Kitware/vtk-js/releases/tag/v30.5.0)

antonskrebkov:
> maybe the release failed, but the next released worked. So it is available as of [v30.5.0](https://github.com/Kitware/vtk-js/releases/tag/v30.5.0)

Good, but I was wondering this because apparently neither setColorSpace nor setScale are working properly before and after merge. I mean, they set the values and with getColorSpace and getScale you can see that the index has changed, but I don’t notice such changes as there are in Paraview. I also wonder why colorTransferFunction has a usingLogScale method, which always returns false?

image

antonskrebkov:
> you might want to place breakpoints in [vtkColorTransferFunction.getTable()](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ColorTransferFunction/index.js#L481) to see if the log scale is considered.

You are right, the logScale is considered in getTable, but I just realised that Scale and ColorSpace are only considered in getTable method, and in my basic pipeline there is no need to get the table (I think so). My code looks something like this:

```
    const points = presetData[0].Points
    const colorRgbPoints = colorData[0].RGBPoints
    
    const pp = stretchColorMap(colorRgbPoints, points)
    ctfun.removeAllPoints()
    
    if (isLogScaleEnabled) {
      ctfun.setScale(Scale.LOG10)
    } else {
      ctfun.setScale(Scale.LINEAR)
    }
    
    for (let i = 0; i < pp.length; i += 4) {
      const x = pp[i]
      const r = pp[i + 1]
      const g = pp[i + 2]
      const b = pp[i + 3]
      ctfun.addRGBPoint(x, r, g, b)
    }

    actor.getProperty().setRGBTransferFunction(0, ctfun)
    
```

where pp is:
image

So the question is how to properly get table in my case?
image

What are xStart and xEnd (indices or values)?
What is table and is my pp array suitable for table parameter?
Is size it is just the length of the array? 

And if I can get a table using this method, how should I apply it to the actor?

The reason I ask this is because I can't find an example of use of this method, and the documentation is unfortunately a bit taciturn :(

finetjul:
I just added an example [here](https://github.com/Kitware/vtk-js/pull/3072).
Make sure your min range is not 0 so that log scaling is considered.
You can debug until `usingLogScale` is not false.

antonskrebkov:
> I just added an example [here](https://github.com/Kitware/vtk-js/pull/3072). Make sure your min range is not 0 so that log scaling is considered. You can debug until `usingLogScale` is not false.

Thank you very much for your example. I indeed see changes when switching the **Log** checkbox.

However, your example uses `vtkMapper`, which has a `get/setLookupTable` method, while `vtkVolumeMapper` (used in my case) lacks it. So I have to use `actor.getProperty().setRGBTransferFunction(0, ctfun)` (where actor is `vtkVolume`) to apply the colors to my image. Could this be a consequence of the fact that nothing changes for me when I change the `scale` and `colorSpace`? And are there any ways out of this situation?

finetjul:
I'm sorry, I do not have time to further investigate the issue. 
Please consider getting a support contract for tailored support.

PR #3063, opened 2024-04-24
fix(outlineThickness): allow 0 thickness per segment index

sedghi:
### Context

Incorrect logic prevents each segment from having a 0 outline. The previous logic defaulted to the first segment when the thickness was 0. The accurate approach should be to default to the first segment when it is undefined, as it should be permissible to set any segment's thickness to 0.

### Results

`labelMap.actor.getProperty().setLabelOutlineThickness([3, 0]);` (segment 1 (red)-> 3 , segment 2 (green) -> 0)

**Previously** 

CleanShot 2024-04-23 at 22 48 31@2x

**After**

CleanShot 2024-04-23 at 22 47 55@2x

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -1376,7 +1376,7 @@ function vtkOpenGLImageMapper(publicAPI, model) {
       // Assuming labelOutlineThicknessArray contains the thickness for each segment
       for (let i = 0; i < lWidth; ++i) {
         // Retrieve the thickness value for the current segment index.
-        // If the value is undefined, use the first element's value as a default, otherwise use the value (event if 0)
+        // If the value is undefined, use the first element's value as a default, otherwise use the value (even if 0)
         const thickness =
           typeof labelOutlineThicknessArray[i] !== 'undefined'
             ? labelOutlineThicknessArray[i]
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -1376,9 +1376,11 @@ function vtkOpenGLImageMapper(publicAPI, model) {
       // Assuming labelOutlineThicknessArray contains the thickness for each segment
       for (let i = 0; i < lWidth; ++i) {
         // Retrieve the thickness value for the current segment index.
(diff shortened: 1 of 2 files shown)


sedghi:
I'm not sure why the test on firefox fails, but for chrome passes

sedghi:
@finetjul sorry for the mention, but are you around to review this quick fix?

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -1376,9 +1376,11 @@ function vtkOpenGLImageMapper(publicAPI, model) {
       // Assuming labelOutlineThicknessArray contains the thickness for each segment
       for (let i = 0; i < lWidth; ++i) {
         // Retrieve the thickness value for the current segment index.
-        // If the value is undefined, null, or 0, use the first element's value as a default.
+        // If the value is undefined, use the first element's value as a default, otherwise use the value (event if 0)
```
finetjul, line 1379: (event if 0) -> (even if 0)

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -1782,9 +1782,12 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
       // Assuming labelOutlineThicknessArray contains the thickness for each segment
       for (let i = 0; i < lWidth; ++i) {
         // Retrieve the thickness value for the current segment index.
-        // If the value is undefined, null, or 0, use the first element's value as a default.
+        // If the value is undefined, use the first element's value as a default, otherwise use the value (event if 0)
```
finetjul, line 1785: (event if 0) -> (even if 0)

finetjul (approved):
LGTM, minor typo though

Sources/Rendering/OpenGL/ImageMapper/index.js
sedghi: Thanks updated both

sedghi:
@finetjul Can we merge this? We're blocked from releasing until this is resolved unfortunatelly

finetjul:
> @finetjul Can we merge this? We're blocked from releasing until this is resolved unfortunatelly

Sure thing. (I was on vacation last week).

PR #3064, opened 2024-04-28
fix(ImageResliceMapper): Axis orthogonality, Image size and texture coordinates

yarous224:
`const w2io = mat3.fromValues(image?.getDirection());` 

`mat3.fromValues`needs expanded array, the old version would output a matrix full of NaN here.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


yarous224:
### Image Size
The polygons generated by ImageResliceMapper do not take into account the size of the pixels themselves.

This is a 512*512 spacing 0.782 image. It length is 400.384mm.
#### Before Change:
bb54c760cd380cfb238e11fbe86a197

#### After Change:
55aae1f4723109b03716f057b2a90da

yarous224:
### Texture coordinates
ImageResliceMapper does not account for the 0.5 offset of the texture matrix when calculating the matrix from world coordinates to texture coordinates, resulting in behavior that is inconsistent with other software (e.g., Radiant).

Before Change
image
After Change
image
Radiant
image

floryst:
@sankhesh can you take a look?

Sources/Rendering/OpenGL/ImageResliceMapper/index.js
```diff
@@ -1217,7 +1221,14 @@ function vtkOpenGLImageResliceMapper(publicAPI, model) {
           );
           const otherAxes = [(orthoAxis + 1) % 3, (orthoAxis + 2) % 3].sort();
           const dim = image.getDimensions();
-          const ext = [0, dim[0] - 1, 0, dim[1] - 1, 0, dim[2] - 1];
+          const ext = [
```
sankhesh, line 1224: Please use `image.getSpatialExtent` here.

Sources/Rendering/OpenGL/ImageResliceMapper/index.js
```diff
@@ -1133,17 +1138,16 @@ function vtkOpenGLImageResliceMapper(publicAPI, model) {
       resGeomString = resGeomString.concat(`Plane${slicePlane.getMTime()}`);
       if (image) {
         resGeomString = resGeomString.concat(`Image${image.getMTime()}`);
+        // Check to see if we can bypass oblique slicing related bounds computation
```
sankhesh, line 1141: I'm not sure I follow the reason for this change. Firstly, this function will not be called unless there is an input(https://github.com/Kitware/vtk-js/blob/a8b7a5cf99c546f28d92bad21ac71a3404c1bff8/Sources/Rendering/OpenGL/ImageResliceMapper/index.js#L121-L126). Secondly, even if execution gets here without an input, an xy plane should still be detected as orthonormal.

sankhesh (changes requested):

daker:
@yarous224 i believe your commits have been merged in the other PR, can you close this one ?

floryst:
Based on the conversations, it appears that this PR has been superseded in #3093. Please re-open with a comment if this is not the case.

PR #3065, opened 2024-04-30
docs(offlineviewer): update style to avoid scrollbars when embeded as iframe

ChristosT:
### Context
This allows to avoid creating scrollbars when using the offline viewer with trame-vtk in the context of jupyter notebooks.

cc: @jourdain 

### Changes

### PR and Code Checklist

-  [x] [semantic-release](https://github.com/semantic-release/semantic-release) commit messages
-  [x] Run `npm run reformat` to have correctly formatted code

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Examples/Applications/OfflineLocalView/viewer.module.css
+++ b/Examples/Applications/OfflineLocalView/viewer.module.css
@@ -49,3 +49,11 @@
   border-radius: 10px;
   user-select: none;
 }
+
+body {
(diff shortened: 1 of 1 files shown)


jourdain:
Can you make it a `docs(...): ...` commit so we don't need to create a new release for that.

PR #3070, opened 2024-05-15
Add getter and setter for resetCameraOnFirstRender in view proxy

bruyeret:
Add `resetCameraOnFirstRender` in the setGet macro

It was possible to set it when creating the view proxy (defaults to true), but it is reset to `true` when rendering with an empty representation list which is sometimes undesirable


finetjul (approved):
LGTM

PR #3071, opened 2024-05-17
fix(cropper): refactor handle normal calculation

rodrigobasilio2022:
### Context

This PR aims to fix a bug in cropper when dealing with medical volumes in orientations other than axial or transversal. The current handleNormal calculation doesn't account for the slice orientation. For example in a axial series, the slice direction is the world z axis, but in a sagittal series the slice direction is the world x axis, as indicated in the below figure, extracted from (https://vtkusers.public.kitware.narkive.com/Nzzv62Jr/correct-anatomical-orientation-of-volumes-from-niftii-and-dicom)

image

So this PR propose a different calculation of the handle normal, based on the calculation of the normalized vector between the world center and the handle position. I believe it is a simpler solution that allows the proper functionality even when in the future the cropper widget could be rotated

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Widgets/Widgets3D/ImageCroppingWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/ImageCroppingWidget/behavior.js
@@ -3,8 +3,8 @@ import macro from 'vtk.js/Sources/macros';
 import {
   AXES,
   transformVec3,
-  rotateVec3,
   handleTypeFromName,
+  calculateDirection,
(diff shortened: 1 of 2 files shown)


floryst (approved):

rodrigobasilio2022:
@finetjul

PR #3073, opened 2024-05-21
feat(Widgets): smooth drag behavior

floryst:
### Context
Dragging widget handles has a jerky behavior, as the handles jump to the mouse position.

### Results

The manipulators now return a new property, `worldDelta`, which is the 3D vector delta between the current and the previous call to `handleEvent`.

### Changes

- [x] Widgets have been updated to use `handleEvent().worldDelta` for smoother dragging

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Widgets/Manipulators/AbstractManipulator/index.d.ts
+++ b/Sources/Widgets/Manipulators/AbstractManipulator/index.d.ts
@@ -49,11 +49,37 @@ export interface vtkAbstractManipulator extends vtkObject {
     setUseCameraNormal(useCameraNormal: boolean): boolean;
 
     /**
-     * 
+     * Processes a vtkRenderWindowInteractor event into 3D world positional info.
+     *
+     * Returns an object containing:
+     * - worldCoords: a 3D coordinate corresponding to the 2D event.
+     * - worldDelta: a 3D position delta between the current and the previous call to handleEvent.
+     * - worldDirection: a 3D directional vector. Only on select manipulators.
+     *
+     * worldCoords can be null if the pointer event enters an invalid manipulator region. For example,
+     * the PickerManipulator returns null when the pointer event is off of the picked geometry.
+     *
+     * worldDelta only makes sense between two calls of `handleEvent`. In a queue of `handleEvent` calls,
+     * the i-th call returns the delta between the i-th worldCoords and the (i-1)-th worldCoords.
+     * Thus, calling `handleEvent` is necessary for maintaining a valid worldDelta even when the return
+     * value is ignored.
+     *
+     * There are three cases where worldDelta needs to handle null events:
+     * 1. the first call to `handleEvent`, since there is no previously cached event position.
+     *    worldDelta is set to [0, 0, 0].
+     * 2. if the current `handleEvent` call returns a null worldCoords. worldDelta is set to [0, 0, 0].
+     * 3. if the previous `handleEvent` call returned a null worldCoords. In this case, worldDelta is the
+     *     delta between the current worldCoords and the previous non-null worldCoords, referring to the
+     *     previous 2 cases when applicable.
+     *
      * @param callData 
      * @param glRenderWindow 
      */
-    handleEvent(callData: any, glRenderWindow: vtkOpenGLRenderWindow): { worldCoords: Nullable<Vector3>, worldDirection?: Matrix3x3 };
+    handleEvent(callData: any, glRenderWindow: vtkOpenGLRenderWindow): {
+      worldCoords: Nullable<Vector3>,
+      worldDelta: Vector3,
+      worldDirection?: Matrix3x3,
+    };
 
     /* ------------------------------------------------------------------- */
 
--- a/Sources/Widgets/Manipulators/AbstractManipulator/index.js
+++ b/Sources/Widgets/Manipulators/AbstractManipulator/index.js
@@ -1,4 +1,5 @@
 import macro from 'vtk.js/Sources/macros';
+import { subtract } from 'vtk.js/Sources/Common/Core/Math';
 
 // ----------------------------------------------------------------------------
 // vtkAbstractManipulator methods
@@ -7,6 +8,8 @@ import macro from 'vtk.js/Sources/macros';
 function vtkAbstractManipulator(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkAbstractManipulator');
+
+  model._prevWorldCoords = [];
 
   publicAPI.getOrigin = (callData) => {
     if (model.userOrigin) return model.userOrigin;
@@ -26,6 +29,27 @@ function vtkAbstractManipulator(publicAPI, model) {
     if (model.handleNormal) return model.handleNormal;
     if (model.widgetNormal) return model.widgetNormal;
     return [0, 0, 1];
+  };
+
+  model._computeDeltaFromPrevCoords = (curWorldCoords) => {
+    if (!model._prevWorldCoords?.length || !curWorldCoords?.length)
+      return [0, 0, 0];
+    return subtract(curWorldCoords, model._prevWorldCoords, []);
+  };
+
+  model._addWorldDeltas = (manipulatorResults) => {
+    const { worldCoords: curWorldCoords } = manipulatorResults;
+    const worldDelta = model._computeDeltaFromPrevCoords(curWorldCoords);
+    if (curWorldCoords) model._prevWorldCoords = curWorldCoords;
+
+    const deltas = {
+      worldDelta,
+    };
+
+    return {
+      ...manipulatorResults,
+      ...deltas,
+    };
   };
 }
 
--- a/Sources/Widgets/Widgets3D/AngleWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/AngleWidget/behavior.js
@@ -1,4 +1,5 @@
 import macro from 'vtk.js/Sources/macros';
+import { add } from 'vtk.js/Sources/Common/Core/Math';
 import vtkPointPicker from 'vtk.js/Sources/Rendering/Core/PointPicker';
 
 const MAX_POINTS = 3;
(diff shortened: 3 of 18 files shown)


Sources/Widgets/Core/DragHelper/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+function vtkWidgetDragHelper(publicAPI, model) {
+  model.classHierarchy.push('vtkWidgetDragHelper');
+
+  publicAPI.calcDelta = (dragWorldCoords) => {
+    if (dragWorldCoords?.length !== 3 || !model.lastOrigin) return [0, 0, 0];
+    return subtract(dragWorldCoords, model.lastOrigin, [0, 0, 0]);
+  };
+
+  publicAPI.calcPosition = (position, dragWorldCoords) => {
```
finetjul, line 12: you could provide an optional 3rd element that will contain the result array. (for performance reasons)

Sources/Widgets/Core/DragHelper/index.d.ts
```diff
@@ -0,0 +1,61 @@
...
+  getLastOrigin(): Nullable<Vector3>;
+
+  /**
+   * Calculates the delta from the last origin to the given coordinate.
+   * @param {Nullable<number[]>} dragWorldCoords the next drag interaction coordinate, if any
+   */
+  calcDelta(dragWorldCoords: Nullable<number[]>): Vector3;
+
+  /**
+   * Calculates the new position from a starting position given a drag interaction coordinate.
```
finetjul, line 27: Can you give more explanations in this documentation ? It's not really clear what it does.

Sources/Widgets/Core/DragHelper/index.d.ts
```diff
@@ -0,0 +1,61 @@
...
+   * @param {Nullable<Vector3>} startPosition the start position
```
finetjul, line 28: Please details what "startPosition" is:
 - last known origin of the widget?
 - first origin of the widget when mouse was first clicked
 - ...

Sources/Widgets/Core/DragHelper/index.d.ts
```diff
@@ -0,0 +1,61 @@
...
+   * @param {Vector3} dragWorldCoords the drag event world coordinate
+   */
+  calcPosition(
+    startPosition: Nullable<Vector3>,
+    dragWorldCoords: Vector3
+  ): Vector3;
+}
+
+/**
+ * Method used to decorate a given object (publicAPI+model) with vtkPlane characteristics.
```
finetjul, line 38: incorrect doc

Sources/Widgets/Core/DragHelper/index.js
```diff
@@ -0,0 +1,44 @@
...
+    if (!position) return dragWorldCoords;
+    const delta = publicAPI.calcDelta(dragWorldCoords);
+    return add(position, delta, [0, 0, 0]);
+  };
+
+  const origSetLastOrigin = publicAPI.setLastOrigin;
+
+  publicAPI.setLastOrigin = (origin) => {
+    // allow setting null as the last origin
+    if (origin && origin.length !== 3) return false;
```
finetjul, line 23: shouldn't it trigger a vtkErrorMacro ?

Sources/Widgets/Core/DragHelper/index.js
```diff
@@ -0,0 +1,44 @@
...
+    return origSetLastOrigin(origin);
+  };
+}
+
+const defaultValues = (initialValues) => ({
+  lastOrigin: null,
```
finetjul, line 29: maybe rename into lastHandleOrigin or lastWidgetOrigin ?

Personal preference: I prefer "currentOrigin" vs "lastOrigin". For some reason working with a "last" information carries a notion of "obsolete" information.

finetjul (commented):
Some ideas:
 - find a way to move the dragHelper into the "manipulator" in order to 1) simplify the overall usage for "widgets developers" 2) group all "coordinate manipulation" into one class 3) simplify custom "drag" behaviors by changing/configuring the manipulator.
 - separate the "lastOrigin" from the "firstOriginClicked". It could be useful for some manipulators to know use that firstOriginClicked position.
 - I think our naming convention is a bit confusing. There are 3 words used interchangeably: "position", "coords" and "origin"

floryst:
I've decided to try something different in the 2nd commit: the manipulator now exposes a `worldDelta` property, which is the 3D vector difference between the previously handled event and the currently handled event. So far I've only done the [LineWidget](https://github.com/Kitware/vtk-js/pull/3073/files#diff-7fb5f9ba6a5b3f56d275bd25d13d359dd54322db636cd9beabfbd421a71b251a), [AngleWidget](https://github.com/Kitware/vtk-js/pull/3073/files#diff-841c9bbf0787ef517ac395498609b9961732c1855204fb980e3d8f205cef3b03), and [ImplicitPlaneWidget](https://github.com/Kitware/vtk-js/pull/3073/files#diff-77b18c2899d081bb11d53ee63049bbb8865a39f7731be80289f685813800de8a).

Generally speaking, the worldDelta should be sufficient for most dragging functionality. In constrained scenarios (e.g. see the ImplicitPlaneWidget code), an extra state variable is needed per constrained state.

I think this is sufficient for now. I'm not yet sure if we need to have the manipulators understand the startDrag/onDrag/endDrag lifecycle just yet. If you think this is sufficient, I'll change all of the widgets to use worldDelta.

floryst:
The second commit drops the drag helper in favor of exposing a `worldDelta` property on the manipulator's handleEvent() return value.

This doesn't work well with complex-constrained manipulations, e.g. the SeedWidget + PickerManipulator. In those scenarios, it's not the user manipulation that needs to be constrained, but rather the seed actor's position.

floryst:
@finetjul please take a look again!

Sources/Widgets/Manipulators/AbstractManipulator/index.d.ts
```diff
@@ -49,11 +49,20 @@ export interface vtkAbstractManipulator extends vtkObject {
     setUseCameraNormal(useCameraNormal: boolean): boolean;
 
     /**
+     * Processes a vtkRenderWindowInteractor event into 3D world positional info.
      * 
+     * Returns an object containing:
+     * - worldCoords: a 3D coordinate corresponding to the 2D event. Can be null.
+     * - worldDelta: a 3D position delta between the current and the previous call to handleEvent.
```
finetjul, line 56: please explain what is returned the first time it is called.

Sources/Widgets/Manipulators/AbstractManipulator/index.js
```diff
@@ -27,6 +30,27 @@ function vtkAbstractManipulator(publicAPI, model) {
     if (model.widgetNormal) return model.widgetNormal;
     return [0, 0, 1];
   };
+
+  model._computeDeltaFromPrevCoords = (curWorldCoords) => {
+    if (!model._prevWorldCoords?.length || !curWorldCoords?.length)
+      return [0, 0, 0];
```
finetjul, line 36: if _prevWorldCoords is empty/null, shouldn't it return curWorldCoords?
if curWorldCoords is empty/null, shouldn't it return _prevWorldCoords ?

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
```diff
@@ -85,20 +87,24 @@ export default function widgetBehavior(publicAPI, model) {
...
-        return macro.EVENT_ABORT;
+      if (worldCoords.length && model.activeState.setOrigin) {
+        if (model.activeState === model.widgetState.getMoveHandle()) {
+          model.activeState.setOrigin(worldCoords);
+          publicAPI.invokeInteractionEvent();
+          return macro.EVENT_ABORT;
+        }
+        if (model._isDragging) {
+          model.activeState.setOrigin(
+            add(model.activeState.getOrigin(), worldDelta, [])
```
finetjul, line 103: please explain in comments why here you use `worldDelta`, but not if `activeState === model.widgetState.getModeHandle)`. It is not obvious to me.

Sources/Widgets/Manipulators/AbstractManipulator/index.d.ts
finetjul: please explain that sometimes it is important to call handleEvent "just" to have worldDelta in correct state for next call.

Sources/Widgets/Widgets3D/PolyLineWidget/behavior.js
```diff
@@ -37,7 +38,15 @@ export default function widgetBehavior(publicAPI, model) {
         model._isDragging) &&
       model.activeState.setOrigin // e.g. the line is pickable but not draggable
     ) {
-      model.activeState.setOrigin(worldCoords);
+      if (model._isDragging) {
+        // only use the drag helper if we are dragging a handle,
+        // not moving the move handle
```
finetjul, line 43: I think this documentation could go in `handleEvent`, to know when to rely on worldDelta, and when not to rely on worldDelta. 
It is not obvious why dragging a handle requires a drag helper (which is an obsolete name ?)

Sources/Widgets/Widgets3D/ShapeWidget/behavior.js
```diff
@@ -498,11 +498,19 @@ export default function widgetBehavior(publicAPI, model) {
       }
     } else if (model._isDragging) {
       if (model.activeState === model.point1Handle) {
-        model.point1Handle.setOrigin(worldCoords);
-        model.point1 = worldCoords;
+        model.point1 = vtkMath.add(
+          model.point1Handle.getOrigin(),
+          worldDelta,
+          []
```
finetjul, line 504: `vtkMath.add(model.point1Handle.getOrigin(), worldDelta, model.point1)` to avoid array instanciation...

Sources/Widgets/Manipulators/AbstractManipulator/index.js
floryst: I think that if there's no previous world coordinate (e.g. first call), then returning [0,0,0] makes sense, since there's no movement. I can also see the case for returning `curWorldCoords` only in the case where `curWorldCoords` is not null, since it's interpreted as a delta from the origin, but that has a built-in assumption on the first call.

If `curWorldCoords` is null, then we shouldn't return the previous world coordinate. That would invalidate the concept of a delta entirely.

Sources/Widgets/Widgets3D/PolyLineWidget/behavior.js
floryst: This is an obsolete comment that needs to be removed.

Sources/Widgets/Widgets3D/AngleWidget/behavior.js
floryst: The move handle can have a null origin, so `add(moveHandle.getOrigin(), worldDelta, [])` crashes. I'll add a comment about it.

floryst:
I've pushed a new commit (which will be squashed prior to merging):
- updated worldDelta docs
- simplified worldDelta handling. The logic should now be more explicit about when we call `setOrigin(worldCoords)` (when the current origin is null) versus when we call `setOrigin(curOrigin + worldDelta)` (current origin is not null).

finetjul (approved):
LGTM

PR #3077, opened 2024-06-04
VolumeMapper & OpenGLTexture: support updating image regions

floryst:
### Context

When updating a 3D texture, the entire texture is uploaded to the GPU. If a volume is incrementally updated (e.g. slice-by-slice), then there is a lot of unchanged data being uploaded at every update.

This changeset lets the VolumeMapper specify image regions that have updated. Internally, this tells vtkOpenGLTexture to only update those regions via `texSubImage3D`.

This change can also be applied to other mappers that use the 3D textures, e.g. the ImageResliceMapper.

### Results

The volume mapper now has the following methods: `setUpdatedExtents(extents)` and `getUpdatedExtents()`. This property defines which extents need to be updated.

```js
const extent = [20, 49, 20, 49, 20, 49];
volumeMapper.setUpdatedExtents([...volumeMapper.getUpdatedExtents(), extent]);
```

This property acts as a "request" to update a set of extents. It is cleared after every successful render call.

### Changes

- [x] vtkVolumeMapper and vtkImageResliceMapper: `setUpdatedExtents()` and `getUpdatedExtents()`
- [x] vtkOpenGLTexture: support updated extents in `create3D*` functions

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


floryst:
@sankhesh @finetjul it would be good to do a review of the API, since this is an initial draft at a working API.Some initial thoughts:
- Maybe add a `addRegionsToUpdate` method
- Should the regions be user-cleared or cleared after every render call

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -611,14 +614,100 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+    const [sx, sy, sz] = size;
+    const sxy = sx * sy;
+    const startOffset = start[2] * sxy + start[1] * sx + start[0];
+
+    let writeOffset = outOffset;
+    for (let zi = 0; zi < sz; zi++) {
+      const zOffset = startOffset + zi * sxy;
+      for (let yi = 0; yi < sy; yi++) {
+        const readOffset = zOffset + yi * sx;
+        const chunk = data.subarray(readOffset, readOffset + sx);
```
finetjul, line 640: did you check the performances ? 
`subarray()` is somehow costly, I found out that replacing it with a `for` loop  is sometimes faster.

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -611,14 +614,100 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+        outArray.set(chunk, writeOffset);
+        writeOffset += chunk.length;
+      }
+    }
+  }
+
+  //----------------------------------------------------------------------------
+
+  /**
+   * /omputes the region's size.
```
finetjul, line 650: Computes

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -627,11 +716,17 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+          if (updateRegions) {
+            pixData.push(
+              new Float32Array(readRegions(data[idx], volumeRegions))
+            );
+          } else {
+            const dataArrayToCopy =
+              data[idx].length > pixCount
+                ? data[idx].subarray(0, pixCount)
+                : data[idx];
+            pixData.push(new Float32Array(dataArrayToCopy));
```
finetjul, line 728: You can move the effect of `constructor(subarray())`  into `constructor(data, offset, length)`, you avoid the creation of a sub array

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1391,55 +1500,114 @@ function vtkOpenGLTexture(publicAPI, model) {
...
     const is3DArray = true;
-    const pixData = updateArrayDataType(dataType, dataArray, is3DArray);
+    const pixData = updateArrayDataType(
+      dataType,
+      dataArray,
+      is3DArray,
+      // if we need to reallocate the entire texture, then process the entire volume
+      rebuildEntireTexture ? [] : updatedRegions
+    );
+    // TODO(floryst): updatedRegions is currently incompatible with webgl1, since there's no region scaling
```
finetjul, line 1529: should probably written in the index.d.ts documentation...

Sources/Rendering/Core/VolumeMapper/index.d.ts
```diff
@@ -1,8 +1,16 @@
...
 import vtkAbstractMapper3D, { IAbstractMapper3DInitialValues } from "../AbstractMapper3D";
 import { BlendMode, FilterMode } from "./Constants";
 
+/**
+ * Represents a 3D region of a volume.
+ */
+export interface VolumeRegion {
+  start: Vector3;
+  size: Vector3;
+}
```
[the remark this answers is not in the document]
sankhesh: I'd prefer `UpdateExtent` (over regions or bounds) to be in sync with the `vtkAlgorithm` terminology.

Sources/Rendering/Core/VolumeMapper/index.d.ts
```diff
@@ -278,6 +286,25 @@ export interface vtkVolumeMapper extends vtkAbstractMapper3D {
...
+   * Tells the mapper to only update the specified regions.
+   * 
+   * If there are zero regions, the mapper updates the entire volume texture.
+   * Otherwise, the mapper will only update the texture by the specified regions
+   * during the next render call.
+   * 
+   * This array is cleared after a successful render.
+   * @param regions 
+   */
+  setRegionsToUpdate(regions: VolumeRegion[]): boolean;
```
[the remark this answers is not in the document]
floryst: I'm not sure what indent you are referring to here. It's all aligned formatting-wise.

Sources/Rendering/OpenGL/Texture/index.js
floryst: I have not, but I do recall subarray in v8 being much slower than usual. Best be safe if that's still an issue.

floryst:
I've pushed a new commit:
- rename "region" to "extent" and use the existing Extent type
- minor perf improvements to the extent handling (not measured)

Sources/Rendering/Core/VolumeMapper/index.d.ts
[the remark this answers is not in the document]
finetjul: you have only 2 spaces for indent (not 3?).
Other functions in this files are indented with tabs.

Sources/Rendering/OpenGL/Texture/index.js
finetjul: fair enough

Sources/Rendering/Core/VolumeMapper/index.d.ts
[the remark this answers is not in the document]
sankhesh: > Other functions in this files are indented with tabs.

Yeah, why is that? Shouldn't we be following vtk coding conventions and stick to 2 spaces instead of tabs?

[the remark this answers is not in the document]
floryst: Here it is: https://github.com/Kitware/vtk-js/pull/3078

finetjul (approved):
LGTM

floryst:
This is now ready for any final reviews @sankhesh. All commits will be squashed prior to merging.

A summary of the recent commits:
- Add updatedExtents support for the vtkImageResliceMapper
- Add a test for vtkOpenGLVolumeMapper and fix an updatedExtents-related bug in vtkOpenGLTexture

Sources/Rendering/OpenGL/ImageResliceMapper/index.js
```diff
@@ -220,24 +220,39 @@ function vtkOpenGLImageResliceMapper(publicAPI, model) {
...
+
+    if (reBuildTex || hasUpdatedExtents) {
+      // If hasUpdatedExtents, then the texture is partially updated
+      const updatedExtents = [...model.renderable.getUpdatedExtents()];
+      // clear the array to acknowledge the update.
+      model.renderable.setUpdatedExtents([]);
+
+      // Build the image scalar texture
+      const dims = image.getDimensions();
       model.openGLTexture.create3DFilterableFromDataArray(
```
sedghi, line 249: We should consider transitioning from positional arguments to object-based arguments. It would make it easier to track which arguments have default values and which ones need to be added. It wouldn't be a breaking change since this are not really public APIs that people use IMO

floryst: This is a good idea to consider. Probably best to wrap up such a change with the next major version increment anyways to be safe from a public API standpoint (even if it's mostly used internally).

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1393,55 +1520,114 @@ function vtkOpenGLTexture(publicAPI, model) {
...
-        model.height,
-        model.depth,
-        0,
-        model.format,
-        model.openGLDataType,
-        scaledData[0]
-      );
+
+      model._paramCache = {
+        tex3dAllocated: true,
```
bruyeret, line 1596: What is the point of `tex3dAllocated`, as it is always true?

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1393,55 +1520,114 @@ function vtkOpenGLTexture(publicAPI, model) {
     model._openGLRenderWindow.activateTexture(publicAPI);
     publicAPI.createTexture();
     publicAPI.bind();
+
+    const hasUpdatedExtents = updatedExtents.length > 0;
+    const paramCache = model._paramCache;
+
+    // It's possible for the texture parameters to change while
+    // streaming, so check for such a change.
+    const rebuildEntireTexture =
```
bruyeret, line 1529: You can create a `newParamCache` object and call `DeepEqual` between this `newParamCache` and `model._paramCache`. Then, if they are different, you just have to do `model._paramCache = newParamCache;`

sankhesh (approved):
LGTM

floryst:
I'll merge this once #3221 goes in, since I need to rebase and handle conflicts anyways.

sedghi:
@floryst Could we perhaps release a minor version with the installation fix before merging this larger change?

https://github.com/Kitware/vtk-js/issues/3231
https://github.com/Kitware/vtk-js/issues/3226

floryst:
Oh I saw your comment a little too late, sorry. I'll get a fix out for that.

PR #3081, opened 2024-06-08
Add vtk threshold point

finetjul:
### Context
Add vtkThresholdPoints filter.

### Results
A new filter with an example are added.

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows
  - **Browser**: Chrome


Sources/Filters/Core/ThresholdPoints/index.js
```diff
@@ -0,0 +1,251 @@
...
+  // Set our classname
+  model.classHierarchy.push('vtkThresholdPoints');
+
+  publicAPI.requestData = (inData, outData) => {
+    const input = inData[0];
+    const output = vtkPolyData.newInstance();
+    outData[0] = output;
+
+    if (model.criterias.length === 0) {
+      output.shallowCopy(input);
```
finetjul, line 58: @floryst @jourdain I was surprised to have to shallow copy. 
If I simply do outputData[0] = input, then the shader does not refresh.
This happens when the threshold filter has previously generated a new ouput polydata and now it simply returns an old polydata. There must be some kind of time stamp issue. I'm not sure if that should be fixed instead in the mapper. WDYT ?

floryst: It's probably b/c `input.getMTime()` is older than `output.getMTime()` at that point, so there's probably a `shouldUpdate` flag set to false in the vtk algorithm macro. `outputData[0] = input` would probably work if `input.modified()` was called as well, but that's probably undesirable, so shallowCopy seems fine to me.

floryst:
Are you planning on adding typescript definitions as well?

finetjul:
> Are you planning on adding typescript definitions as well?

I was not planning to because I do not use typescript. Shall we make it mandatory for new contributions ?

floryst:
> I was not planning to because I do not use typescript. Shall we make it mandatory for new contributions ?

It's a good idea to do so nowadays, especially since vtk.js is ingested by typescript projects.

floryst (approved):

Issue #2858, opened 2023-06-26
[Bug]  vtkInteractiveOrientationWidget does not work since 28.0.0


zbouslikhin:
### Bug description

vtkInteractiveOrientationWidget stopped working since v28.0.0 update onwards. Pressing the interactiveWidget does not trigger onOrientationChange.
The InteractiveOrientationWidget example also does not seem to work: https://kitware.github.io/vtk-js/examples/InteractiveOrientationWidget.html

### Steps to reproduce

See: https://codesandbox.io/s/playground-93onj8?file=/src/Scene/SceneManager.tsx:5303-5334

### Expected Behavior

Run onOrientationChange when clicking the widget

### Environment

- vtk.js version: 28.0.0
- Browsers: Firefox 114.0.2 
- OS: Windows 10 Pro Version 21H2

floryst:
Thanks for the report! We'll look into this. Do you know if it worked in the latest v27 release?

zbouslikhin:
> Thanks for the report! We'll look into this. Do you know if it worked in the latest v27 release?

Yes, it does work in the latest v27 release

jourdain:
Could be one of
- WidgetManager: Support multiple renderers in the same render window ([4def3ed](https://github.com/Kitware/vtk-js/commit/4def3ed1a49b20ead9a2f15f70f4d941a15ea45c))
- BehaviorManager: Allow decorated RenderWindow to work as well ([1024f07](https://github.com/Kitware/vtk-js/commit/1024f0719970fdcca63b552f1c2aaa6b728768b4))

judebourlay:
tried it and indeed it seems broken due to https://github.com/Kitware/vtk-js/commit/4def3ed1a49b20ead9a2f15f70f4d941a15ea45c

bourdaisj:
Closing as completed  
see https://github.com/Kitware/vtk-js/pull/3082

PR #3082, opened 2024-06-10
Fix interactive orientation widget

bourdaisj:
### Context
See https://github.com/Kitware/vtk-js/issues/2858

### Results
A solution is provided to make widget works as intended.

### Changes

The provided solution expose a new API on InteractorStyle to force an explicit renderer instead of using the event poked renderer.
The second change is to make the OrientationMarkerWidget self renderer interactive.
Third fix is in the example itself, it consists of calling `setParentRenderer` on the OrientationMarkerWidget instance with the renderer that should be used for camera synchronization.

### PR and Code Checklist

### Testing
- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Interaction/Style/InteractorStyleImage/index.js
+++ b/Sources/Interaction/Style/InteractorStyleImage/index.js
@@ -15,7 +15,7 @@ function vtkInteractorStyleImage(publicAPI, model) {
   publicAPI.superHandleMouseMove = publicAPI.handleMouseMove;
   publicAPI.handleMouseMove = (callData) => {
     const pos = callData.position;
-    const renderer = callData.pokedRenderer;
+    const renderer = model.getRenderer(callData);
 
     switch (model.state) {
       case States.IS_WINDOW_LEVEL:
@@ -101,7 +101,7 @@ function vtkInteractorStyleImage(publicAPI, model) {
 
   //--------------------------------------------------------------------------
   publicAPI.handleMouseWheel = (callData) => {
-    const camera = callData.pokedRenderer.getActiveCamera();
+    const camera = model.getRenderer(callData).getActiveCamera();
 
     let distance = camera.getDistance();
     distance += callData.spinY;
@@ -115,7 +115,8 @@ function vtkInteractorStyleImage(publicAPI, model) {
       distance = range[1];
     }
     camera.setDistance(distance);
-    const props = callData.pokedRenderer
+    const props = model
+      .getRenderer(callData)
       .getViewProps()
       .filter((prop) => prop.isA('vtkImageSlice'));
     props.forEach((prop) => {
--- a/Sources/Interaction/Widgets/OrientationMarkerWidget/index.d.ts
+++ b/Sources/Interaction/Widgets/OrientationMarkerWidget/index.d.ts
@@ -145,6 +145,13 @@ export interface vtkOrientationMarkerWidget extends vtkObject {
    * Updates the orientation widget viewport size.
    */
   updateViewport(): void;
+
+  /**
+   * An instance of this class will spawn its own renderer, by default non interactive.
+   * This behavior is configurable through the interactiveRenderer property when initializing the instance.
+   * @returns true if the renderer was created as interactive, false otherwise.
+   */
+  getInteractiveRenderer(): boolean;
 }
 
 /**
--- a/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
+++ b/Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
@@ -148,7 +148,7 @@ function vtkOrientationMarkerWidget(publicAPI, model) {
       }
       // Highest number is foreground
       selfRenderer.setLayer(renderWindow.getNumberOfLayers() - 1);
-      selfRenderer.setInteractive(false);
+      selfRenderer.setInteractive(model.interactiveRenderer);
 
       selfRenderer.addViewProp(model.actor);
       model.actor.setVisibility(true);
@@ -281,6 +281,7 @@ export const DEFAULT_VALUES = {
   minPixelSize: 50,
   maxPixelSize: 200,
   parentRenderer: null,
+  interactiveRenderer: false,
 };
 
 // ----------------------------------------------------------------------------
@@ -291,7 +292,12 @@ export function extend(publicAPI, model, initialValues = {}) {
   // Build VTK API
   macro.obj(publicAPI, model);
 
-  macro.get(publicAPI, model, ['enabled', 'viewportCorner', 'viewportSize']);
+  macro.get(publicAPI, model, [
+    'enabled',
+    'viewportCorner',
+    'viewportSize',
+    'interactiveRenderer',
+  ]);
 
   // NOTE: setting these while the widget is enabled will
   // not update the widget.
--- a/Sources/Rendering/Core/InteractorStyle/index.d.ts
+++ b/Sources/Rendering/Core/InteractorStyle/index.d.ts
@@ -1,5 +1,7 @@
 import { EventHandler, vtkSubscription } from '../../../interfaces';
+import { Nullable } from '../../../types';
 import vtkInteractorObserver from '../InteractorObserver';
+import vtkRenderer from '../Renderer';
 
 export interface vtkInteractorStyle extends vtkInteractorObserver {
   /**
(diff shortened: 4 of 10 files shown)


bourdaisj:
@finetjul can we move `InteractorStyle.getRenderer` to protected or private? I don't see the need for it being in to the public api.

Sources/Rendering/Core/InteractorStyle/index.d.ts
```diff
@@ -210,6 +210,19 @@ export interface vtkInteractorStyle extends vtkInteractorObserver {
...
    */
   handleKeyPress(callData: unknown): void;
+
+  /**
+   * Explicitly defines a renderer to be used for event handling.
+   * If not set, the pokedRenderer of the event will be used.
+   *
+   * @param {vtkRenderer} renderer
+   */
+  setForcedRenderer(renderer: vtkRenderer): void;
```
floryst, line 220: What about calling it "focusedRenderer"? I don't have a strong naming pref, but "forcedRenderer" seems a little opaque to me.

In addition, we should allow setting null here to be able to turn off this functionality.

bourdaisj: I agree. `focusedRenderer` sounds clearer

bourdaisj: the choice to use `forcedRenderer` comes from `vtkRenderWindowInteractor`  [here](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/RenderWindowInteractor/index.js#L1190)

Sources/Interaction/Style/InteractorStyleImage/index.js
```diff
@@ -115,7 +115,8 @@ function vtkInteractorStyleImage(publicAPI, model) {
       distance = range[1];
     }
     camera.setDistance(distance);
-    const props = callData.pokedRenderer
+    const props = publicAPI
+      .getRenderer(callData)
```
finetjul, line 119: model.getRenderer

Sources/Rendering/Core/InteractorStyle/index.d.ts
```diff
@@ -210,6 +212,19 @@ export interface vtkInteractorStyle extends vtkInteractorObserver {
...
    */
   handleKeyPress(callData: unknown): void;
+
+  /**
+   * Explicitly defines a renderer to be used for event handling.
+   * If never called or called with null, the pokedRenderer of the event will be used.
+   *
+   * @param {Nullable<vtkRenderer>} renderer
+   */
+  setFocusedRenderer(renderer: Nullable<vtkRenderer>): void;
```
finetjul, line 222: a setter does not return a void but a boolean

Sources/Interaction/Widgets/OrientationMarkerWidget/index.js
```diff
@@ -148,7 +148,7 @@ function vtkOrientationMarkerWidget(publicAPI, model) {
       }
       // Highest number is foreground
       selfRenderer.setLayer(renderWindow.getNumberOfLayers() - 1);
-      selfRenderer.setInteractive(false);
+      selfRenderer.setInteractive(true);
```
finetjul, line 151: This specific widget does not require an interactive renderer.

Sources/Interaction/Widgets/OrientationMarkerWidget/index.d.ts
```diff
@@ -147,6 +147,13 @@ export interface vtkOrientationMarkerWidget extends vtkObject {
 	 * Updates the orientation widget viewport size.
 	 */
 	updateViewport(): void;
+
+	/**
+	 * An instance of this class will spawn its own renderer, by default non interactive.
+	 * This behavior is configurable through the spawnInteractiveRenderer property when initializing the instance.
+	 * @returns true if the renderer was created as interactive, false otherwise.
+	 */
+	getSpawnInteractiveRenderer(): boolean;
```
finetjul, line 156: you also need to add it to `IOrientationMarkerWidgetInitialValues`

finetjul: Rename into `interactiveRenderer` or `interactive`?

finetjul: I thought you would also spawn `vtkInteractiveOrientationWidget` within `OrientationMarkerWidget` when `interactive` is true...

bourdaisj:
will squash before merging

Sources/Widgets/Widgets3D/InteractiveOrientationWidget/example/index.js
```diff
@@ -5,27 +5,15 @@ import '@kitware/vtk.js/Rendering/Profiles/Geometry';
...
-  const axis = [0, 0, 0];
-  const idx = Math.abs(vec3[idxA]) > Math.abs(vec3[idxB]) ? idxA : idxB;
-  const value = vec3[idx] > 0 ? 1 : -1;
-  axis[idx] = value;
-  return axis;
-}
+import {
+  createInteractiveOrientationMarkerWidget,
+  alignCameraOnViewWidgetOrientationChange,
+} from '../helpers';
```
finetjul, line 16: examples should not use relative imports (need to be copy-pasted)

Sources/Widgets/Widgets3D/InteractiveOrientationWidget/helpers.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+    if (direction[1]) {
+      camera.setViewUp(majorAxis(viewUp, 0, 2));
+    }
+    if (direction[2]) {
+      camera.setViewUp(majorAxis(viewUp, 0, 1));
+    }
+
+    orientationMarkerWidget.updateMarkerOrientation();
+    widgetManager.enablePicking();
```
finetjul, line 98: I'm surprised you have to call `enablePicking()` each time

Sources/Widgets/Widgets3D/InteractiveOrientationWidget/helpers.js
```diff
@@ -0,0 +1,101 @@
...
+  return axis;
+}
+
+export function createOrientationMarkerWidget(interactor, mainRenderer) {
+  const axes = vtkAxesActor.newInstance();
+
+  const orientationWidget = vtkOrientationMarkerWidget.newInstance({
+    actor: axes,
+    interactor,
+    spawnInteractiveRenderer: true,
```
finetjul, line 20: spawnInteractiveRenderer -> interactiveRenderer

Sources/Widgets/Widgets3D/InteractiveOrientationWidget/helpers.js
```diff
@@ -0,0 +1,101 @@
...
+  });
+  orientationWidget.setEnabled(true);
+  orientationWidget.setViewportCorner(
+    vtkOrientationMarkerWidget.Corners.BOTTOM_LEFT
+  );
+  orientationWidget.setViewportSize(0.1);
+  orientationWidget.setMinPixelSize(100);
+  orientationWidget.setMaxPixelSize(300);
+
+  orientationWidget.setParentRenderer(mainRenderer);
```
finetjul, line 30: NIP: pass viewportSize, minPixelSize, maxPixelSize and parentRenderer in `newInstance()`

Sources/Widgets/Widgets3D/InteractiveOrientationWidget/helpers.js
bourdaisj: just copy/pasted the existed boilerplate tbh. no idea what that would be required but if its there I guess there's a good reason

bourdaisj:
@daker Could you please take a look at the TS stuff?

daker (approved):
LGTM

Sources/Widgets/Widgets3D/InteractiveOrientationWidget/helpers.js
floryst: I think it might be better to call `widgetManager.renderWidgets()`. This will do the same thing as `enablePicking()`, except not do a capture pass if the user has picking disabled.

finetjul (approved):
LGTM

Sources/Widgets/Widgets3D/InteractiveOrientationWidget/helpers.js
bourdaisj: I do not see any difference between not calling `widgetManager.enablePicking` nor `widgetManager.renderWidgets`, or calling one or the other

finetjul: let's remove it then

floryst (approved):

PR #3085, opened 2024-06-14
Remove node deletion from "removeNode"

bruyeret:
If this function deletes a node, it should be called "deleteNode" not "removeNode"
This caused an error in the example `ManyRenderWindows`


finetjul (approved):
LGTM

PR #3086, opened 2024-06-18
Picking translucent

### LINKED ISSUE #2800: [Bug] HardwareSelector incorrect attributeID/propID with opacity < 1
### High-level description

Picking using HardwareSelector in the presence of opacity can return incorrect attributeID and propID, which can correspond to the furthest away face/object. A hack is proposed in the forum post linked below, where opacity is set to 1 before the `hardwareSelector.getSourceDataAsync` call, and then reset immediately after

### Steps to reproduce

https://jsfiddle.net/rmart/g8jxwkdz/1/ to reproduce, 

https://jsfiddle.net/rmart/g8jxwkdz/12/ implements the "hack"

### Detailed Current Behavior

Logged here https://discourse.vtk.org/t/hardware-selector-with-transparency-returning-wrong-attribute-id-prop-id/11270

> Hello, I am looking for some help to make hardware selector return correct attributeID and propID in the presence of opacity.
>
> Very basically, I am trying to select faces of a polydata using the hardware selector, but am finding that when actors have opacity < 1, the attributeID and propID returned (often, but not always) refer to the furthest away objects instead of the one I am clicking on (i.e., my pick appears to go through to the other side… as if the order was reversed…). This applies to faces in a single object (i.e., I pick faces on the other side of the transparent object), as well as when making a pick when another transparent object is behind (faces from the other object are selected). The really odd thing is that the xyz position is correct, its just the attributeID and propID that point to the wrong thing.
>
> The behavior is demonstrated in this screengrab, and in the js-fiddle that follows:
>
> picking-hw-opacity
> https://jsfiddle.net/rmart/g8jxwkdz/1/
>
>  am wondering if there is something clearly wrong being done in this example? Most of the logic is borrowed from the Hardware Selector example on the vtk-js docs.
>
> Thanks in advance!

Using the hack in a renderWindow with lots of complex actors, with transparency, results in the HardwareSelector picking the right objects and getting the right attribute ID's and is much more performant than falling back on the raycast-pickers.

### Environment

- vtk.js version: latest, https://unpkg.com/vtk.js@27.3.0/vtk.js
- Browsers: Chrome, firefox
- OS: macos, windows, linux

sankhesh:
PropID of the last rendered prop was returned by the hardware selector when picking translucent actors overlaid on each other.
### Context

fix #2800 
### Results

Corrent prop ids being returned.
### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 11
  - **Browser**: Chrome

### DIFF
--- a/Sources/Rendering/OpenGL/Actor/index.js
+++ b/Sources/Rendering/OpenGL/Actor/index.js
@@ -152,7 +152,10 @@ function vtkOpenGLActor(publicAPI, model) {
   // Renders myself
   publicAPI.translucentPass = (prepass, renderPass) => {
     if (prepass) {
-      model.context.depthMask(false);
+      model.context.depthMask(
+        model._openGLRenderer.getSelector() &&
+          model.renderable.getNestedPickable()
+      );
       publicAPI.activateTextures();
     } else if (model.activeTextures) {
       for (let index = 0; index < model.activeTextures.length; index++) {
--- a/Sources/Rendering/Core/HardwareSelector/test/testHardwareSelector.js
+++ b/Sources/Rendering/Core/HardwareSelector/test/testHardwareSelector.js
@@ -8,7 +8,13 @@ import vtkActor from 'vtk.js/Sources/Rendering/Core/Actor';
 import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
 import vtkPlaneSource from 'vtk.js/Sources/Filters/Sources/PlaneSource';
 import vtkSphereSource from 'vtk.js/Sources/Filters/Sources/SphereSource';
-import { FieldAssociations } from 'vtk.js/Sources/Common/DataModel/DataSet/Constants';
+import vtkConeSource from 'vtk.js/Sources/Filters/Sources/ConeSource';
+import vtkCalculator from 'vtk.js/Sources/Filters/General/Calculator';
+import { AttributeTypes } from 'vtk.js/Sources/Common/DataModel/DataSetAttributes/Constants';
+import {
+  FieldDataTypes,
+  FieldAssociations,
+} from 'vtk.js/Sources/Common/DataModel/DataSet/Constants';
 
 test('Test HardwareSelector', (tapeContext) => {
   const gc = testUtils.createGarbageCollector(tapeContext);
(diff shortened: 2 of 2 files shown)


Sources/Rendering/OpenGL/Actor/index.js
```diff
@@ -152,7 +152,10 @@ function vtkOpenGLActor(publicAPI, model) {
   // Renders myself
   publicAPI.translucentPass = (prepass, renderPass) => {
     if (prepass) {
-      model.context.depthMask(false);
+      model.context.depthMask(
+        model._openGLRenderer.getSelector() &&
```
bruyeret, line 156: We may want a configurable threshold on opacity, that enables picking for translucent actors
For example, we don't want to pick when the opacity is 0
This threshold can be ignored here as we want to completely disable rendering of translucent actors if they don't reach the threshold (if we only disable depth mask, the actors will be rendered anyway), so the threshold test could be when doing the actual rendering

package-lock.json
daker, line 1: Please remove the package-lock.json changes

daker (changes requested):

bruyeret (commented):
The CI has caught the `test.only` usage that should be removed
Thanks for adding this feature and a test case 👍

package-lock.json
bruyeret: It was intentional, and has been done in a separate commit:
> `chore(package-lock): fixed node package issues with npm audit fix`
Why is it bad?

Sources/Rendering/OpenGL/Actor/index.js
sankhesh: @bruyeret Sorry for not getting back to you sooner. We don't need a configurable opacity threshold for picking simply because if the actor's opacity is 0, it will not be rendered (even for selection). If users desire to ignore certain actors from selection, that can be done via `PickableOff`. WDYT?

package-lock.json
sankhesh: @daker I removed the package-lock changes and will move them to a separate PR.

Sources/Rendering/OpenGL/Actor/index.js
bruyeret: Yes that's ok, it works 👍

bruyeret (approved):

PR #3087, opened 2024-06-18
fix(ImageProperty): color + opacity funcs can be null

floryst:
### Context
`imageProperty.setRGBTransferFunction(index, null)` and `imageProperty.setScalarOpacity(index, null)` should be allowed in the types.

### Results
Types are fixed to allow null functions.

### Changes

### PR and Code Checklist

### Testing
N/A


daker (approved):
LGTM

PR #3092, opened 2024-06-25
GPU resources garbage collection

bruyeret:
Add garbage collection of shared graphics resources.
The render window now owns the shared resources.
Owners of graphics resources are responsible to register and unregister their usage of graphics resources.
Remove sharing of `scalars` from the cpu image reslice mapper, as it owns the scalars and doesn't share them as the volume mapper and the gpu image reslice mapper do.
Users don't have to do anything - except using a parent RenderWindow - to use this feature.

### DIFF
--- a/Sources/Rendering/OpenGL/ImageMapper/index.js
+++ b/Sources/Rendering/OpenGL/ImageMapper/index.js
@@ -89,7 +89,7 @@ function vtkOpenGLImageMapper(publicAPI, model) {
         !oldOglRenderWindow.isDeleted() &&
         oldOglRenderWindow !== model._openGLRenderWindow
       ) {
-        // We unregister ourself when the render window changes
+        // Unregister the mapper when the render window changes
         unregisterGraphicsResources(oldOglRenderWindow);
       }
       model.context = model._openGLRenderWindow.getContext();
@@ -1016,14 +1016,16 @@ function vtkOpenGLImageMapper(publicAPI, model) {
           model.colorTexture,
           cfunToString
         );
-        model._openGLRenderWindow.unregisterGraphicsResourceUser(
-          model._colorTransferFunc,
-          publicAPI
-        );
-        model._openGLRenderWindow.registerGraphicsResourceUser(
-          colorTransferFunc,
-          publicAPI
-        );
+        if (colorTransferFunc !== model._colorTransferFunc) {
+          model._openGLRenderWindow.registerGraphicsResourceUser(
+            colorTransferFunc,
+            publicAPI
+          );
+          model._openGLRenderWindow.unregisterGraphicsResourceUser(
+            model._colorTransferFunc,
+            publicAPI
+          );
+        }
         model._colorTransferFunc = colorTransferFunc;
       }
     } else {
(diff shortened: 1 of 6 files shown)


Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -929,19 +950,15 @@ function vtkOpenGLImageMapper(publicAPI, model) {
       model._openGLRenderWindow.getGraphicsResourceForObject(colorTransferFunc);
 
     const reBuildC =
-      !cTex?.vtkObj?.getHandle() ||
-      cTex?.hash !== cfunToString ||
-      model.colorTextureString !== cfunToString;
+      !cTex?.oglObject?.getHandle() || cTex?.hash !== cfunToString;
     if (reBuildC) {
+      model.colorTexture = vtkOpenGLTexture.newInstance({
```
finetjul, line 955: fyi, a `newInstance()` call is not cheap in VTK.js. 
That's why we were trying to "reuse" them.
I'm not sure why you got rid of that.

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -67,9 +80,17 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
+      const oldOglRenderWindow = model._openGLRenderWindow;
       model._openGLRenderWindow = model._openGLRenderer.getLastAncestorOfType(
         'vtkOpenGLRenderWindow'
       );
+      if (
+        oldOglRenderWindow &&
+        !oldOglRenderWindow.isDeleted() &&
+        oldOglRenderWindow !== model._openGLRenderWindow
+      ) {
+        unregisterGraphicsResources(oldOglRenderWindow);
```
finetjul, line 92: I think it deserves some in-code comments.

Sources/Rendering/OpenGL/RenderWindow/index.js
```diff
@@ -1204,49 +1205,66 @@ function vtkOpenGLRenderWindow(publicAPI, model) {
...
+  };
+
   publicAPI.getGraphicsMemoryInfo = () => {
     let memUsed = 0;
-    model._graphicsResources.forEach((gObj, vtkObj) => {
-      memUsed += gObj.getAllocatedGPUMemoryInBytes();
+    model._graphicsResources.forEach(({ oglObject }) => {
+      memUsed += oglObject.getAllocatedGPUMemoryInBytes();
     });
     return memUsed;
```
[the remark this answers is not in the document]
bruyeret: There is no `reduce()` on objects of class `Map`
I just changed a name in existing code

Sources/Rendering/OpenGL/ImageMapper/index.js
bruyeret: I removed it because the render window is now the owner of the resource
Only the owner can free the texture

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -89,6 +89,7 @@ function vtkOpenGLImageMapper(publicAPI, model) {
         !oldOglRenderWindow.isDeleted() &&
         oldOglRenderWindow !== model._openGLRenderWindow
       ) {
+        // We unregister ourself when the render window changes
```
finetjul, line 92: Unregister the mapper when the render window changes.

(we tend to avoid using "we", "I", "ourself" in comments).

sankhesh (approved):
LGTM

bruyeret:
I fixed the potential issue that we talked about @finetjul

finetjul (approved):
LGTM

PR #3093, opened 2024-07-08
Fix reslice mapper

bruyeret:
I fixed 4 bugs and added a test case in the image reslice mapper.

The changes are small, and the commits are self explanatory:
- [Fix orthoSlicing detection](https://github.com/Kitware/vtk-js/commit/07180d3db922e76ee96c7289d883bf31b9612d31): The w2io matrix was always full of NaN because of a bad usage of mat3.fromValues
- [Use spatial extent from image data itself](https://github.com/Kitware/vtk-js/commit/263c0b57f2f75cfa945590bd63b44e5450e3eb26)
- [Use default scaling when there is no vbo scaling](https://github.com/Kitware/vtk-js/commit/37c06769fc72e49cd6512540771494cad7903ab3) 
- [Fix offset of WCTCMatrix](https://github.com/Kitware/vtk-js/commit/f72a1519bf0b4ba108c67cc973a7f06df2a0d615): The `indexToWorld` matrix transforms an integer index coordinate into the center of a voxel.
 The WCTCMatrix transforms world coordinates to texture coordinates.
 Therefore we need to add  a 0.5 voxel offset (see `getSpatialExtent`) to this matrix.
 We need this offset because the voxel region is all around the voxel coordinate instead of after it.
 This is similar to what is done in ITK, see this link:
 https://itk.org/ITKSoftwareGuide/html/Book1/ITKSoftwareGuide-Book1ch4.html#x45-540004.1.4

The added test uses a green and red background with a very low tolerance to avoid any regression.


sankhesh:
@bruyeret Could you please attribute the author from #3064 via either `git commit --author` or a `Co-Authored-by:` trailer in the relevant commits?

The changes look good to me otherwise.

Sources/Rendering/Core/ImageResliceMapper/test/testImageResliceMapperBorderRendering.js
```diff
@@ -0,0 +1,131 @@
...
+  ofun.addPoint(maxScalar, 1);
+  ppty.setPiecewiseFunction(ofun);
+
+  ppty.setInterpolationTypeToNearest();
+  actor.setProperty(ppty);
+
+  const imageData = gc.registerResource(vtkImageData.newInstance());
+
+  imageData.setOrigin(0, 0, 0);
+  imageData.setSpacing(1, 1, 1);
```
finetjul, line 82: it would be great to have a non 1 spacing (in case the 0.5 offset is spacing-related)

Sources/Rendering/Core/ImageResliceMapper/test/testImageResliceMapperSlabTypes.png
finetjul, line 1: Not sure why we now get black lines on top and bottom images.

Sources/Rendering/OpenGL/ImageResliceMapper/index.js
```diff
@@ -599,7 +599,10 @@ function vtkOpenGLImageResliceMapper(publicAPI, model) {
       }
 
       if (program.isUniformUsed('vboScaling')) {
-        program.setUniform3fv('vboScaling', cellBO.getCABO().getCoordScale());
+        program.setUniform3fv(
+          'vboScaling',
+          cellBO.getCABO().getCoordScale() ?? [1, 1, 1]
```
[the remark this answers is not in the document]
bruyeret: No, there is no documentation for this function, it is an auto generated getter and the default value is null

Sources/Rendering/Core/ImageResliceMapper/test/testImageResliceMapperSlabTypes.png
bruyeret: This dark line doesn't grow in screen space when zooming
image
image

This is likely caused by floating point imprecisions

bruyeret:
> @bruyeret Could you please attribute the author from #3064 via either `git commit --author` or a `Co-Authored-by:` trailer in the relevant commits?

Thanks, I didn't notice this PR, I coo authored all the `fix` commits 👍

PR #3094, opened 2024-07-08
Fix ogl render window typing

bruyeret:
It didn't inherit the view node and inherited vtkAlgorithm even if it is not one


daker (approved):
LGTM

PR #3095, opened 2024-07-11
Use 3D textures when needed for cell data colors

bruyeret:
When coloring by point data, with interpolateScalarsBeforeMapping set to true, we can't use a 3D texture because we interpolate texture coordinates.

When coloring by cell data, there will be no texture coordinate interpolation as all the points of a cell have the same texture coordinates. Therefore we can use 3D texture to be able to use more values. We need to be able to interpolate between two neighboring texels, this is why we have to use a zigzag pattern for the color texture.

The theoretical limit for the number of voxels is 2048**3 = 2**33. We could also use the 3D texture when interpolateScalarsBeforeMapping is false, but this is not done in this PR.


sankhesh (approved):
LGTM

Do we have a test for cell scalars with texture map?

bruyeret:
> Do we have a test for cell scalars with texture map?
I don't think so, but I can add one

Sources/Common/Core/LookupTable/index.js
```diff
@@ -61,7 +61,8 @@ function vtkLookupTable(publicAPI, model) {
   publicAPI.usingLogScale = () => false;
 
   //----------------------------------------------------------------------------
-  publicAPI.getNumberOfAvailableColors = () => model.table.length / 4;
+  // Remove special colors (min, max, NaN) from available colors
```
finetjul, line 64: Number does not include special colors (`min`, `max` and `NaN`).

Sources/Rendering/Core/Mapper/index.d.ts
```diff
@@ -291,21 +275,10 @@ export interface vtkMapper extends vtkAbstractMapper3D {
...
-   * @param {Number} scalarValue
-   * @param {Number} rangeMin
-   * @param {Number} invRangeWidth
-   */
-  scalarToTextureCoordinate(
-    scalarValue: number,
-    rangeMin: number,
-    invRangeWidth: number
-  ): IScalarToTextureCoordinate;
+  mapScalarsToTexture(scalars: any, cellFlag: boolean, alpha: number): void;
```
finetjul, line 281: looks like it is a breaking change. you might want to flag that commit as such (and update BREAKING_CHANGES.md)

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -368,26 +567,20 @@ function vtkMapper(publicAPI, model) {
...
-        scalars,
-        scalarComponent,
-        range,
-        numberOfColorsInRange,
-        useLogScale
-      );
-      model.colorCoordinatesString = colorCoordinatesString;
-    }
+    // Create new coordinates if necessary, this function uses cache if possible.
+    // We can't use a zigzag pattern with point data, as interpolation of texture coordinates will be wrong
```
finetjul, line 571: We can't use [...] -> A zigzag pattern can't be used [...]

finetjul (commented):
I agree, a non-regression test would be good...

Sources/Common/Core/LookupTable/index.js
bruyeret: The table length does include these special colors, because in `forceBuild`, the length of the table is defined as such:
```js
const maxIndex = model.numberOfColors - 1;
// ...
model.table.length = 4 * maxIndex + 16;
```
So this function simply returns the `model.numberOfColors`. I could eventually simplify it by doing:
```js
publicAPI.getNumberOfAvailableColors =  publicAPI.getNumberOfColors;
```

And the special colors should not be included in `getNumberOfAvailableColors` because we add them by hand in `mapScalarsToTexture`. Also in vtkColorTransferFunction, it basically returns `model.nodes.length` which does not include NaN colors.

Sources/Rendering/Core/Mapper/index.d.ts
bruyeret: Yes, I was wondering if we really need to have this function in the publicAPI, we could simply make it private.
It would be a breaking change too though.
What do you think?

bruyeret:
I rebased on master and added 3 commits to add a regression test, a fix two minor issues

Sources/Common/Core/LookupTable/index.js
finetjul: All I meant was to rephrase the "comment" because mentionning "Remove" seems to say that there is a "side-effect" of calling this get function.

Sources/Rendering/Core/Mapper/index.d.ts
finetjul: Maybe not private but protected. I do not have a strong opinion on this. 
Please flag your commit as BREAKING_CHANGE though...

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -185,117 +441,7 @@ function vtkMapper(publicAPI, model) {
...
-        outputV[outputIdx++] = outputs.texCoordS;
-        outputV[outputIdx++] = outputs.texCoordT;
-        inputIdx += numComps;
-      }
-    }
-    return output;
-  };
-
-  publicAPI.mapScalarsToTexture = (scalars, alpha) => {
+  publicAPI._mapScalarsToTexture = (scalars, cellFlag, alpha) => {
```
finetjul, line 444: this is not removing it from the public API, to make the method protected, you need to add it to "model" instead of "publicAPI"

bruyeret:
I fixed the protected function 👍 
I also added a small commit for fixing the shared GPU resources when picking

bruyeret:
I moved the fix to this PR: #3098
This PR contains a breaking change and should be merged after 3098

floryst:
@bruyeret I would like to merge 9b2970796ee322cd6f2db11699ecc492e2a07e27 separately in a different PR. We need it as a higher priority fix.

bruyeret:
> @bruyeret I would like to merge [9b29707](https://github.com/Kitware/vtk-js/commit/9b2970796ee322cd6f2db11699ecc492e2a07e27) separately in a different PR. We need it as a higher priority fix.

I moved the commit in PR #3105

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -27,6 +27,262 @@ function notImplemented(method) {
...
+  // Caching using the "arguments" special object (because it is a pure function)
+  const argStrings = new Array(arguments.length);
+  for (let argIndex = 0; argIndex < arguments.length; ++argIndex) {
+    // eslint-disable-next-line prefer-rest-params
+    const arg = arguments[argIndex];
+    argStrings[argIndex] = arg.getMTime?.() ?? arg;
+  }
+  const cacheString = argStrings.join('/');
+  const cachedResult = colorTextureCoordinatesCache.get(cacheString);
+  if (cachedResult) {
```
Antti-Palola, line 160: _Actually_ the cache can never be hit again as `MTimes` can't match, this should just store a single previous key and value or change the cache key logic.

Antti-Palola (changes requested):

Sources/Rendering/Core/Mapper/index.js
bruyeret: The input array are the scalar values. So if you don't change the cell data itself, but you change the lookup table, it will use the cached coordinates.

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -325,38 +472,90 @@ function vtkMapper(publicAPI, model) {
...
+      const numberOfAvailableColors =
+        model.lookupTable.getNumberOfAvailableColors();
+
+      // Maximum dimensions and number of colors in range
+      const maxTextureWidthForCells = 2048;
+      const maxColorsInRangeForCells = maxTextureWidthForCells ** 3 - 3; // 3D but keep a color for min, max and NaN
+      const maxTextureWidthForPoints = 4096;
+      const maxColorsInRangeForPoints = maxTextureWidthForPoints - 2; // 1D but keep a color for min and max (NaN is in a different row)
+      // Minimum number of colors excluding minColor, maxColor and NaNColor
+      const minColorsInRange = 2;
```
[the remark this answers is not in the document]
bruyeret: Why should points have a minimum of 3 colors ?
The reason for using 2 is that we divide somewhere by `(numberOfNonSpecialColors - 1)`, so it should be greater than 1.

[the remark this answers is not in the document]
Antti-Palola: the comment is then the misleading part as it talks about 3 colors but value is set to 2.

Sources/Rendering/Core/Mapper/index.js
Antti-Palola: Ok, I mixed this input to be the lookup. But it still stands that if any of the values change you can never get back the old values returned by `getMTime` so only a single entry should be cached?

Sources/Rendering/Core/Mapper/index.js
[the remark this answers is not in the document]
bruyeret: I changed the comment 👍

Sources/Rendering/Core/Mapper/index.js
```diff
@@ -27,6 +27,262 @@ function notImplemented(method) {
...
+) {
+  // Caching using the "arguments" special object (because it is a pure function)
+  const argStrings = new Array(arguments.length);
+  for (let argIndex = 0; argIndex < arguments.length; ++argIndex) {
+    // eslint-disable-next-line prefer-rest-params
+    const arg = arguments[argIndex];
+    argStrings[argIndex] = arg.getMTime?.() ?? arg;
+  }
+  const cacheString = argStrings.join('/');
+  const cachedResult = colorTextureCoordinatesCache.get(cacheString);
```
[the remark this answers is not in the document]
bruyeret: I made a cache that only stores one texture coordinate per input data array and that uses a weak map, so that the texture coordinates array is freed when the input array is not referenced anywhere

Sources/Rendering/Core/Mapper/index.js
bruyeret: In theory you could use a single input data array with different lookup tables, so you could need several cache entries for a single input data array.
In practice this almost never happens so I changed the cache to only cache one result per entry, as you suggested 👍

Antti-Palola (approved):
Took this PR for a test run and it seems to be working well. Now we're just hitting the walls of GPU picker overflow on really big models 😄

Antti-Palola:
Is there a schedule for merging this?

PR #3098, opened 2024-07-17
Fix rendering when picking with hardware selector in a shared render window

bruyeret:
When two render windows A and B share the same context using `addRenderWindow` (see ManyRenderWindows example), this is how a render window does its rendering:
- Render in the shared context
- Copy the content of the shared context to the render window canvas

The following issue happened when using the hardware selector with a shared context when A renders, B renders and A uses the hardware selector:
- `A` renders in the shared context and the context of the shared context is copied into `A`'s canvas
- `B` renders in the shared context and the context of the shared context is copied into `B`'s canvas
- `A` renders in the shared context for the hardware selector, which does not actually render something, so the content of the context remains what `B` just put there and then the content of the shared context is copied into `A`'s canvas

The issue is that `A`'s canvas now contains what B just rendered.

To fix the issue, disable copy of the shared context content when a renderer has a selector


finetjul (approved):
LGTM

PR #3099, opened 2024-07-18
Fix resolve coincident topology api

floryst:
### Context
- `mapper.set({ relativeCoincidentTopologyLineOffsetParameters: { factor, offset } })` does not work because the coincident topology setters do not support single-object parameters. This is (mostly) inconsistent with how we expect setters to work.
- Refactored the coincident topology helper types to reduce repetition and inconsistencies with type definitions across mappers.

### Results

- improved types and coincident topology API

### Changes

### PR and Code Checklist

### Testing
N/A

### DIFF
--- a/Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.d.ts
+++ b/Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.d.ts
@@ -3,6 +3,11 @@ export interface ICoincidentTopology {
   offset: number;
 }
 
+export enum Resolve {
+  Off,
+  PolygonOffset,
+}
+
 export interface StaticCoincidentTopologyMethods {
   /**
    *
@@ -88,12 +93,12 @@ export interface StaticCoincidentTopologyMethods {
    *
    * @param mode
    */
-  setResolveCoincidentTopology(mode?: number): boolean;
+  setResolveCoincidentTopology(mode?: Resolve): boolean;
 
   /**
    *
    */
-  getResolveCoincidentTopology(): number;
+  getResolveCoincidentTopology(): Resolve;
 
   /**
    *
@@ -117,10 +122,7 @@ export interface StaticCoincidentTopologyMethods {
 }
 
 export default interface CoincidentTopologyHelper
-  extends Omit<
-    StaticCoincidentTopologyMethods,
-    'setResolveCoincidentTopology' | 'getResolveCoincidentTopology'
-  > {
+  extends StaticCoincidentTopologyMethods {
   /**
    *
    * @param {ICoincidentTopology} params
@@ -174,17 +176,6 @@ export default interface CoincidentTopologyHelper
     factor: number,
     offset: number
   ): boolean;
-
-  /**
-   *
-   * @param resolveCoincidentTopology
-   */
-  setResolveCoincidentTopology(resolveCoincidentTopology: boolean): boolean;
-
-  /**
-   *
-   */
-  getResolveCoincidentTopology(): boolean;
 
   /**
    *
--- a/Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.d.ts
+++ b/Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.d.ts
@@ -1,0 +1,203 @@
+export interface ICoincidentTopology {
+  factor: number;
+  offset: number;
+}
+
+export interface StaticCoincidentTopologyMethods {
+  /**
+   *
+   * @param {ICoincidentTopology} params
+   */
+  setResolveCoincidentTopologyPolygonOffsetParameters(
+    params: ICoincidentTopology
+  ): boolean;
+
+  /**
+   *
+   * @param {ICoincidentTopology} params
+   */
+  setResolveCoincidentTopologyLineOffsetParameters(
+    params: ICoincidentTopology
+  ): boolean;
+
+  /**
+   *
+   * @param {ICoincidentTopology} params
+   */
+  setResolveCoincidentTopologyPointOffsetParameters(
+    params: ICoincidentTopology
+  ): boolean;
+
+  /**
+   *
+   * @param {Number} factor
+   * @param {Number} offset
+   */
+  setResolveCoincidentTopologyPolygonOffsetParameters(
+    factor: number,
+    offset: number
+  ): boolean;
+
+  /**
+   *
+   * @param {Number} factor
+   * @param {Number} offset
+   */
+  setResolveCoincidentTopologyLineOffsetParameters(
+    factor: number,
+    offset: number
+  ): boolean;
+
+  /**
+   *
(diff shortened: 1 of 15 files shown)


Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
```diff
@@ -53,6 +69,8 @@ function implementCoincidentTopologyMethods(publicAPI, model) {
     publicAPI[methodName] = staticOffsetAPI[methodName];
   });
 
+  macro.setGet(publicAPI, model, ['resolveCoincidentTopology']);
+
```
floryst, line 73: This is an interesting change. The static methods declare `vtkMapper.setResolveCoincidentToplogy(mode = 0)`, while the instance methods are used as a boolean value (e.g. `mapper.setResolveCoincidentTopology(true)`). Should we change expectations of the latter to match the former?

Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
```diff
@@ -5,8 +5,26 @@ import macro from 'vtk.js/Sources/macros';
...
+        newOffset = offset;
+      }
+
+      const changed =
+        model[item.key]?.factor !== newFactor ||
+        model[item.key]?.offset !== newOffset;
+
+      model[item.key] = { factor: newFactor, offset: newOffset };
+      return changed;
     };
```
finetjul, line 28: It looks like it would be easier to do that:
```
macro.set(publicAPI, model, [{ type: 'object', name: item.method}]);
```
This would only support calls such as: `setCoincidentTopologyParameters({factor:1, offset:1});` (already the case in `WidgetRepresentation`)

I guess we could also extend  `vtkMacros.findSetter()` so that we would do: 
```
macro.set(publicAPI, model, [{ type: 'object', name: item.method, params:['factor', 'offset']}]);
```

and that would allow `setCoincidentTopologyParameters(1, 1);` in addition to  `setCoincidentTopologyParameters({factor:1, offset:1});`

Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
finetjul: I think `mapper.setResolveCoincidentTopology(true)` should not be allowed.

sankhesh: I agree with @finetjul. The setter should just be setting the type of resolution.

Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
floryst: Ah, thanks for reminding me of that setter format. I agree that having a `params` field would be useful, and I'm satisfied with the way you've structured it so I'll add that in.

floryst:
New changes:
- `CoincidentTopologyHelper` exposes a `Resolve` object: `Resolve.Off` and `Resolve.PolygonOffset` are the two enumerations available.
- `get/setResolveCoincidentTopology()` uses the above `Resolve` enum.
- `macro.setGet` on object fields now supports a `params: [prop1, ...]` config, which enables invoking the setter with object fields as arguments, e.g. `setter(prop1, prop2, ...)`.
- `objectSetterMap` is exported from macros.js, but is marked for internal use only (used in CoincidentTopologyHelper.js) and is not exposed in the typescript typings.

finetjul:
There is a breaking-change PR-ready if you want/need to do breaking change in this PR: https://github.com/Kitware/vtk-js/pull/3095

floryst:
Is there a timeline for #3095?

bruyeret:
No, we will merge it as soon as possible

finetjul:
> Is there a timeline for #3095?

we would like to merge it asap.

Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
```diff
@@ -1,13 +1,22 @@
...
 
+export { Resolve } from 'vtk.js/Sources/Rendering/Core/Mapper/Static';
+
 function addCoincidentTopologyMethods(publicAPI, model, nameList) {
   nameList.forEach((item) => {
     publicAPI[`get${item.method}`] = () => model[item.key];
-    publicAPI[`set${item.method}`] = (factor, offset) => {
-      model[item.key] = { factor, offset };
-    };
+    publicAPI[`set${item.method}`] = macro.objectSetterMap.object(
```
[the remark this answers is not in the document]
floryst: `item.method` differs significantly from `item.key`. For instance, the `item.method` could be `ResolveCoincidentTopology${key}OffsetParameters`, while `item.key` would be `Polygon`. `macro.setGet` doesn't support custom getter names based on the key.

Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.d.ts
```diff
@@ -0,0 +1,194 @@
...
+   *
+   * @param {Number} value
+   */
+  setResolveCoincidentTopologyPolygonOffsetFaces(value: number): boolean;
+
+  /**
+   *
+   * @param mode
+   */
+  setResolveCoincidentTopology(mode?: number): boolean;
```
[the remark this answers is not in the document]
floryst: Yep,  thanks for catching that.

Sources/Testing/testMacro.js
```diff
@@ -451,6 +459,20 @@ test('Macro methods object tests', (t) => {
     'Getter shall return a copy'
   );
 
+  myTestClass.setMyParameterizedObjectProp({ foo: 12, bar: 89 });
```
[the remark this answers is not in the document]
floryst: I don't see a reason to further test setting the entire object, since that functionality should be the same as if we didn't pass in a `params: [...]` option. I will add another prop to the object just to demonstrate that you can set an object with keys outside of the `params` list.

Sources/Testing/testMacro.js
```diff
@@ -451,6 +459,20 @@ test('Macro methods object tests', (t) => {
...
+  t.deepEqual(
+    myTestClass.getMyParameterizedObjectProp(),
+    { foo: 12, bar: 89 },
+    'Setter changes on a different object'
+  );
+
+  myTestClass.setMyParameterizedObjectProp(16, 52);
```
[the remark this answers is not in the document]
floryst: `(16)` is ambiguous. If there is only 1 argument then I treat that as setting the property directly. Deciding what to do based on the argument type is not a good idea, because what if the property's value is itself an object?

I'll add a test for 3 arguments.

floryst:
Updates
* fixed `setResolveCoincidentTopology`'s signature to use the `Resolve` enum
* added two more tests to testMacro

finetjul (approved):
LGTM

Issue #3062, opened 2024-04-22
[Bug] Memory leak in WidgetState


Antti-Palola:
### Bug description

Instantiating for example a plane widget and properly removing the widget and its manager with delete leaks memory.

### Steps to reproduce

* Instantiate a widget
* Delete widget
* Widget state leaks memory

### Detailed Behavior

Loaded widgets three times, notice the groups of 3 distinct size, which are all different different states when loading the same view three times.

![Screenshot 2024-04-22 at 09 54 02](https://github.com/Kitware/vtk-js/assets/100124171/fe2a12c7-b3fe-4203-ab5e-528c82e5557d)

Cause would seem to be https://github.com/Kitware/vtk-js/blob/21534d1515f234c5034f9ec1c1e423575379a14d/Sources/Widgets/Core/WidgetState/index.js#L57 that pushes a callback in `unbindState`

### Expected Behavior

No memory leaks

### Environment

- vtk.js version: main of 22.4.2024 ([21534d1](https://github.com/Kitware/vtk-js/commit/21534d1515f234c5034f9ec1c1e423575379a14d))
- Browsers: Chrome
- OS: Mac OS 14.3.

Antti-Palola:
Could someone take a look at this? Shouldn't the unbind state remove the callback rather than add it?

floryst:
I'll take a look at this.

floryst:
This appears to be a leak only in vtkLineWidget. The logic you linked shouldn't leak, since it unsubscribes from all nested states, removes the nested state, then re-subscribes to the remaining nested states. In effect, removing the associated subscription with the removed nested state.

PR #3100, opened 2024-07-19
fix(LineWidget): onModified memory leak

floryst:
### Context

#3062 reported a memory leak with vtkLineWidget. It remains to be seen if widgets generally leak memory.

### Results

- Line widget no longer leaks memory

### Changes

### PR and Code Checklist

### Testing
N/A

### DIFF
--- a/Sources/Widgets/Widgets3D/LineWidget/index.js
+++ b/Sources/Widgets/Widgets3D/LineWidget/index.js
@@ -160,7 +160,7 @@ function vtkLineWidget(publicAPI, model) {
     model.widgetState.getMoveHandle().setOrigin(center);
   });
 
-  model.widgetState.getPositionOnLine().onModified(() => {
+  let linePosSub = model.widgetState.getPositionOnLine().onModified(() => {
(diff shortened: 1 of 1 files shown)


finetjul (approved):
Could a unit test detect memory leaks on widgets automatically?

floryst:
There's `performance.memory`, but it's deprecated, doesn't have firefox support, and is likely a granular measure to avoid fingerprinting. The other approach would be to use a WeakMap, but that would require being able to reliably trigger GC, which we can't.

finetjul:
@floryst could we inject code in `macro.obj()` or `delete()` to keep track of created/deleted objects ?

floryst:
Yes, that can help track leaked objects, but it would be noisy b/c there's a lot of vtk.js objects that stick around for the duration of a scene. However, I don't know if it'll help with objects we don't directly create (such as nested widget states), since we'd want to use WeakMap to avoid holding a strong reference to tracked objects. Did you have some specific mechanism in mind?

PR #3101, opened 2024-07-20
fix the judgement of model.useLookupTableScalarRange

YenseanHub:
### Context
when I use mapper, set the parameter "useLookupTableScalarRange" for "true". It seems the lookupTable did not work, So I set it "false"  ,the lookupTable will work. 
I look up the source of vtkMapper, the original version is "if (!model.useLookupTableScalarRange)"
It is more reasonable is the judgement is " if (model.useLookupTableScalarRange)" ,Right?
### Results
useLookupTableScalarRange work as expected

### Changes
delete the "!"

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  latest master
  - **OS**:  Windows 10
  - **Browser**:   Chrome 126

### DIFF
--- a/Sources/Rendering/Core/Mapper/index.js
+++ b/Sources/Rendering/Core/Mapper/index.js
@@ -156,7 +156,7 @@ function vtkMapper(publicAPI, model) {
     const toString = `${publicAPI.getMTime()}${scalars.getMTime()}${alpha}`;
     if (model.colorBuildString === toString) return;
 
-    if (!model.useLookupTableScalarRange) {
+    if (model.useLookupTableScalarRange) {
       publicAPI
         .getLookupTable()
         .setRange(model.scalarRange[0], model.scalarRange[1]);


bruyeret (changes requested):
From the ts header of the getter for `useLookupTableScalarRange`:
```ts
  /**
   * Control whether the mapper sets the lookuptable range based on its
   * own ScalarRange, or whether it will use the LookupTable ScalarRange
   * regardless of it’s own setting. By default the Mapper is allowed to set
   * the LookupTable range, but users who are sharing LookupTables between
   * mappers/actors will probably wish to force the mapper to use the
   * LookupTable unchanged.
   *
   * @param {Boolean} useLookupTableScalarRange
   * @default false
   */
  setUseLookupTableScalarRange(useLookupTableScalarRange: boolean): boolean;
```

So the check that you see here is perfectly fine, it can be read:
"If we don't use the lookup table scalar range, then override it"

(Also the same check has been copy pasted in the `Glyph3DMapper` and the `Mapper2D`)

bruyeret:
I will close this PR but if you still have an issue, you can open an issue and tag me 😉

PR #3104, opened 2024-07-26
fix(OBJReader): Add support for negative indices

amin-kchaou:
### Context
According to the Wavefront OBJ spec ([here](https://paulbourke.net/dataformats/obj/) and [here](https://www.martinreddy.net/gfx/3d/OBJ.spec)), indices specified in the face elements (the `f` lines) may be negative. In that case, they refer to a vertex data counting up the list (e.g. -1 is the vertex data element immediately before the current line, -2 is the one before that, and so on...).

The current OBJReader implementation doesn't take into account the fact that these indices can be negative, leading to incorrect data lookup and overall an incorrect polydata.

image
  

### Results
Here's an example textured cube that contains negative indices for 2 faces:
- Before:
Screenshot from 2024-07-26 15-10-32

- After: 
Screenshot from 2024-07-26 15-14-03

### Changes
-  Updated `faceMap` function in `OBJReader` to take into account negative indices. 

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Linux Ubuntu 24-04
  - **Browser**: Chrome 126.0.6478.126
- Test cube files: [link](https://filebin.net/n6ybiv3u4soo9wux)

### DIFF
--- a/Sources/IO/Misc/OBJReader/index.js
+++ b/Sources/IO/Misc/OBJReader/index.js
@@ -37,9 +37,20 @@ function begin(splitMode) {
 
 function faceMap(str) {
   const idxs = str.split('/').map((i) => Number(i));
-  const vertexIdx = idxs[0] - 1;
-  const textCoordIdx = idxs[1] ? idxs[1] - 1 : vertexIdx;
-  const vertexNormal = idxs[2] ? idxs[2] - 1 : vertexIdx;
+  const vertexIdx = idxs[0] > 0 ? idxs[0] - 1 : data.v.length / 3 + idxs[0];
+  let textCoordIdx;
+  if (idxs[1]) {
+    textCoordIdx = idxs[1] > 0 ? idxs[1] - 1 : data.vt.length / 2 + idxs[1];
+  } else {
+    textCoordIdx = vertexIdx;
+  }
+
+  let vertexNormal;
+  if (idxs[2]) {
+    vertexNormal = idxs[2] > 0 ? idxs[2] - 1 : data.vn.length / 3 + idxs[2];
+  } else {
+    vertexNormal = vertexIdx;
+  }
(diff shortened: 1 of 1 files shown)


Sources/IO/Misc/OBJReader/index.js
```diff
@@ -37,9 +37,20 @@ function begin(splitMode) {
...
+  } else {
+    textCoordIdx = vertexIdx;
+  }
+
+  let vertexNormal;
+  if (idxs[2]) {
+    vertexNormal = idxs[2] > 0 ? idxs[2] - 1 : data.vn.length / 3 + idxs[2];
+  } else {
+    vertexNormal = vertexIdx;
+  }
```
finetjul, line 53: how about:

```
let vertexIdx = idxs[0] - 1;
vertexIdx = vertexIdx <= 0 ? vertexIdx + 1 + data.v.length / 3 : vertexIdx;
let textCoordIdx = idxs[1] ? idxs[1] - 1 : vertexIdx;
textCoordIdx= textCoordIdx<= 0 ? textCoordIdx + 1 + data.vt.length / 2 : textCoordIdx;
let vertexNormal = idxs[2] ? idxs[2] - 1 : vertexIdx;
vertexNormal = vertexNormal <= 0 ? vertexNormal + 1 + data.vt.length / 2 : vertexNormal;
```

finetjul (commented):
You might want to create a unit test that enforces non regression for negative indices.

finetjul (approved):
LGTM

amin-kchaou:
@finetjul It seems that the test for hardware acceleration barely fails, but I don't think my PR affects it. Should I bump the limit a bit?

floryst:
Those tests are flaky, so re-running usually fixes it. LGTM!

PR #3107, opened 2024-07-31
Update index.d.ts

luchen666:
In the norm method in vtkMath, its second parameter should be non-required. However, in ts, an error reporting situation occurs every time.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Common/Core/Math/index.d.ts
+++ b/Sources/Common/Core/Math/index.d.ts
@@ -311,7 +311,7 @@ export function cross(x: Vector3, y: Vector3, out: Vector3): Vector3;
  * @param {Number[]} x
  * @param {Number} n
  */
-export function norm(x: number[], n: number): number;
+export function norm(x: number[], n: number = 3): number;
 
 /**
  * Normalize in place. Returns norm.


finetjul (changes requested):
Please respect the [semantic release](https://github.com/semantic-release/semantic-release) naming scheme of VTK.js.

You can use `npm run commit` to format your commit messages.

Sources/Common/Core/Math/index.d.ts
```diff
@@ -311,7 +311,7 @@ export function cross(x: Vector3, y: Vector3, out: Vector3): Vector3;
  * @param {Number[]} x
  * @param {Number} n
  */
-export function norm(x: number[], n: number): number;
+export function norm(x: number[], n: number = 3): number;
```
daker, line 314:
```typescript
export function norm(x: number[], n?: number): number;
```

daker (changes requested):

PR #3113, opened 2024-08-07
Add GPU resource sharing for the CPR image mapper

bruyeret:
Add resource sharing for the vtkImageCPRMapper.
Factorize the computation of hashes to avoid deoptimization of resource sharing.


xiefanhua:
How to add blood vessel mapper

bruyeret:
You can see [this example](https://kitware.github.io/vtk-js/examples/ImageCPRMapper.html)

PR #3115, opened 2024-08-20
Fix wrong center computation from bounding box

bruyeret:
Two examples did the same mistake (there were a copy paste).
They should add the coordinates instead of subtract.
Use `vtkBoundingBox.getCenter` instead.


finetjul (approved):
LGTM

Issue #3030, opened 2024-03-12
[Bug] Handling of clamped arrays?


abustany:
### Bug description

When creating data arrays from Uint8ClampedArray, VTK infers the texture type to be a float texture, not a UCHAR. This is (I suppose) because VtkDataTypes.UNSIGNED_CHAR is "Uint8Array", so the data type does not get properly detected when creating the texture.

### Steps to reproduce

1. Create a DataArray from a Uint8ClampedArray, with some random pixel data
2. Display this data on screen (using an ImageMapper etc etc)

### Detailed Behavior

getDefaultTextureInternalFormat in Rendering/OpenGL/RenderWindow.js doesn't match the data type to be uchar, and falls down to the default float case.

### Expected Behavior

The OpenGL texture uses the right data type (UCHAR), that does not require the `OES_texture_float_linear extension`.

### Environment

- vtk.js version: 29.7.0
- Browsers: Latest firefox+chrome
- OS: Linux

finetjul:
@sedghi  @bruyeret @sankhesh

longuto:
How to correct this issue? When using volume rendering, it appears on the Android mobile browser,

longuto:
How to correct this issue? When using volume rendering, it appears on the Android mobile browser.
![Dingtalk_20240315164407](https://github.com/Kitware/vtk-js/assets/21220020/814a2569-b0dc-4c4e-a0bf-e0074e10cf0b)

longuto:
@abustany Can you patch him up.  My mistake seems to be the same as yours, but it cannot be resolved.

longuto:
He's really strange. If my browser doesn't support the OES texture_ float_line and EXT texture_ orm16 extensions, I have to make the changes in the following way to see the rendered Actor.

The version of vtkjs is 29.7.0
![Dingtalk_20240408154803](https://github.com/Kitware/vtk-js/assets/21220020/ade3ae99-ce98-4f7a-8fa9-ec568298a175)

longuto:
He's really strange. If my browser doesn't support the OES_texture_ float_linear and EXT_texture_norm16 extensions, I have to make the changes in the following way to see the rendered Actor.
The version of vtkjs is 29.7.0
![Dingtalk_20240408154803](https://github.com/Kitware/vtk-js/assets/21220020/abad2665-d545-4276-a479-3771836fce98)

daker:
Already fixed

PR #3117, opened 2024-08-28
fix: force half float in texture if it's exact and OES_texture_float_linear is not available

slak44:
### Context
The origin of the problem is this bug in OHIF, a consumer of vtk: https://github.com/OHIF/Viewers/issues/4286
For certain mobile Android devices, instead of rendering a texture, the canvas was solid gray/black. It happens regardless of OS version/browser, but eventually I narrowed it down to devices that use Mali GPUs (which is likely because they don't support OES_texture_float_linear).

Since it was working everywhere previously and I had a good version of OHIF, I bisected until this [first bad commit](https://github.com/OHIF/Viewers/commit/21e8a2bd0b7cc72f90a31e472d285d761be15d30), which among other things, bumps the version of vtk.js from `29.3.0` to `29.7.0`. After checking all the vtk versions in that range, I found that `29.5.0` was the last good one, and `29.5.1` was the first bad one with gray/black canvases. This issue has been present on master ever since.

Luckily (or unluckily) this was a patch release that only included 2 commits. Of them, the relevant one is [this](https://github.com/Kitware/vtk-js/commit/4eb4e630fb0ea33e2a1e6939270159ea5d0dc5eb).

The commit changed the texture format to sometimes use *32F versions, and added this warning: 
image
The problem with using full floats is in webgl2 they aren't texture filterable without OES_texture_float_linear, while half floats are always filterable even without the extension. Using full floats triggers the warning/black textures.

I assume all of these are the same problem and would be fixed by this change:
- https://github.com/Kitware/vtk-js/issues/3030
- https://github.com/cornerstonejs/cornerstone3D/issues/1152
- https://github.com/OHIF/Viewers/issues/4286

### Results
With this change, canvases are no longer black in OHIF and render the images again everywhere.

### Changes
This change forces half floats in cases where:
1. Full floats would have been used instead
2. The extension for texture filtering is not supported in the gl context
3. The data can be stored accurately in half floats

I don't know if there's any way to solve this if the data doesn't fit in halfs, maybe with another option similar to preferSizeOverAccuracy.

### PR and Code Checklist

### Testing
I only reproduced this on Androids with Mali GPUs. I would assume it happens on any device that supports webgl2 but not OES_texture_float_linear. 
If you have such a device, you can see the problem here (a gray square is rendered, instead of an image of a brain CT): https://viewer.ohif.org/viewer?StudyInstanceUIDs=2.16.840.1.114362.1.11972228.22789312658.616067305.306.2

For testing the changes, the easiest way I know is to clone, compile and run [OHIF](https://github.com/OHIF/Viewers), and configure resolutions in its package.json to use these changes in vtk:
```
  "resolutions": {
    "@kitware/vtk.js": "link:../vtk-js/dist/esm",
``` 
Then open this: http://localhost:3000/viewer?StudyInstanceUIDs=2.16.840.1.114362.1.11972228.22789312658.616067305.306.2

With the changes enabled you should see the brain CT, without the changes you should not.

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -408,7 +408,7 @@ function vtkOpenGLTexture(publicAPI, model) {
     return result;
   };
 
-  publicAPI.useHalfFloat = () => model.enableHalfFloat && model.canUseHalfFloat;
+  publicAPI.useHalfFloat = () => model.canUseHalfFloat;
 
   //----------------------------------------------------------------------------
   publicAPI.setInternalFormat = (iFormat) => {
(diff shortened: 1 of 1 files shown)


Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -408,7 +408,7 @@ function vtkOpenGLTexture(publicAPI, model) {
     return result;
   };
 
-  publicAPI.useHalfFloat = () => model.enableHalfFloat && model.canUseHalfFloat;
```
floryst, line 411: I think `model.enableHalfFloat` should be `model.enableUseHalfFloat`.

sankhesh (approved):
LGTM

PR #3122, opened 2024-09-17
feat: Add WebXR to default Rendering export

bruyeret:
WebXR wasn't exposed in the file sent by CDNs
See [this post](https://discourse.vtk.org/t/vtkxrrenderwindow-is-web-served-vtk-js-service/14554) in the discourse


finetjul (changes requested):
Aren't you missing an index.js file in WebXR ?

bruyeret:
Yes I am working on it, but I don't know how to expose the `Constants.js` file
Do you know how should I proceed?
It looks like it is usually not exposed but it is surprising

finetjul (approved):
LGTM

PR #3128, opened 2024-09-27
feat(ImageSlice): add forceOpacity and forceTransparent

PaulHax:
To maintain consistent render effect with stacked image slices no matter the opacity of the slices.

### Context

When the prop's transparency crosses from .999 to 1, it gets slotted in the Opaque render pass and can change its render order: https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ImageSlice/index.js#L24-L26

When multiple slices are layered changing the tranparency can inavertently change how they are blended.  Would be nice to keep a slice in a consistent render pass despite the opacity.  Then can avoid this workaround: https://github.com/Kitware/VolView/pull/654/files#diff-3b32a6ae4684ef94faf56b59ea525d8c8adccc792d96cb56613eec2aedf6faf6R20

### Results

From an added test: before with 2 actors where one is translucent
```
  actorBelow.getProperty().setOpacity(0.9);
  actorAbove.getProperty().setOpacity(1);
```
image

After with adding `actorAbove.setForceTranslucent(true)`
image

### Changes

### PR and Code Checklist

### Testing


finetjul (approved):
LGTM

finetjul:
You may want to forceOpaque when you want to render something translucent but writing in the depth buffer.
You may want to forceTransparent when you want to render something opaque without writing in the depth buffer.

sedghi:
Thanks @finetjul 

I guess the default (without setting these force properties) is to render something translucent without a depth buffer and render something opaque with writing to the depth buffer?

What are the main reasons that one might or might not want to write to the depth buffer?

finetjul:
> I guess the default (without setting these force properties) is to render something translucent without a depth buffer and render something opaque with writing to the depth buffer?

Yes, this is the default.

> What are the main reasons that one might or might not want to write to the depth buffer?

Want to write in depth buffer: if you don't want other actors to render behind the actor
Don't want to write in depth buffer: if you want other actors to render behind/in front of the actor

I do not have a special use case in mind (besides Paul's), but I remember needing those in VTK C++. That's why Paul just "brought" those "forceXXX" functions from VTK-C++ to VTK.js

sedghi:
Amazing, thanks

PR #3131, opened 2024-09-30
vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters errors

PaulHax:
### Context

When calling the static mapper `vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters`, there is an error because the "class" does not have a publicAPI  with a `modified()` function.

### Results

Does not error now when calling `vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters`.  

Some other tests are failing after adding the test in this PR.  Seems that calling `setResolveCoincidentTopologyPolygonOffsetParameters` with the initial results of `getResolveCoincidentTopologyPolygonOffsetParameters` does not completely undo the effect?
image

### Changes

The error fix involved adding a noOp `modified` function.  Alternatives could be to check if `publicAPI` has `modified` in macros.js or upgrade the Mapper static object with macros.js' `function obj(publicAPI = {}, model = {})` "constructor."   

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
+++ b/Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
@@ -29,7 +29,10 @@ const staticOffsetModel = {
   Line: { factor: 1, offset: -1 },
   Point: { factor: 0, offset: -2 },
 };
-const staticOffsetAPI = {};
+const noOp = () => undefined;
+const staticOffsetAPI = {
+  modified: noOp,
+};
 
 addCoincidentTopologyMethods(
   staticOffsetAPI,
--- a/Sources/Rendering/Core/Mapper/test/testCoincidentTopologyHelper.js
+++ b/Sources/Rendering/Core/Mapper/test/testCoincidentTopologyHelper.js
@@ -1,0 +1,38 @@
+import test from 'tape';
+import testUtils from 'vtk.js/Sources/Testing/testUtils';
+
+import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
+
+test('Test calling CoincidentTopologyHelper static functions', async (t) => {
+  const gc = testUtils.createGarbageCollector(t);
+
+  // set back to starting/default values to avoid side effects for other tests
+  const startingParameters =
+    vtkMapper.getResolveCoincidentTopologyPolygonOffsetParameters();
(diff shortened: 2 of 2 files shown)


PaulHax:
I guess the open questions/todos on this is issue PR:

- [ ] Do we want to keep supporting the `vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters` function?  We are using it in VolView: https://github.com/Kitware/VolView/blob/main/src/main.js#L34
- [ ] Is this "fix" by adding the noOp `modified` function acceptable?
- [ ] How to test `vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters` without affecting other tests?

finetjul:
> Do we want to keep supporting the vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters function? We are using it in VolView: https://github.com/Kitware/VolView/blob/main/src/main.js#L34

I think so, there is also a Static ResolveCoincidentTopology in VTK C++.

finetjul (commented):
Shouldn't you expose a "static" MTime ? so that it can be considered when deciding to building/updating the shader ?

PaulHax:
> Shouldn't you expose a "static" MTime ? so that it can be considered when deciding to building/updating the shader ?

I think that makes sense as the global/static factor/offset are added with instance values as a uniform for the mappers shaders.  How to do that?  Upgrade the static object to be a full `macros.obj`, then each mapper registers a `onModified` for the static object and updates its own `MTime` when the static object changes?

>  How to test vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters without affecting other tests?

Do you have any insight on that?

finetjul:
> I think that makes sense as the global/static factor/offset are added with instance values as a uniform for the mappers shaders. How to do that? Upgrade the static object to be a full macros.obj, then each mapper registers a onModified for the static object and updates its own MTime when the static object changes?

Yes, that makes sense

finetjul:
>>How to test vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters without affecting other tests?
> Do you have any insight on that?

I'm not sure what the issue is TBH

PaulHax:
superseded by https://github.com/Kitware/vtk-js/pull/3142

PaulHax:
Can't reopen this PR so opening #3143

PR #3133, opened 2024-10-01
Multi volume

bruyeret:
This PR adds support for multi volume.
It is not intended to be merged, as the shader takes too much time to compile on some systems (e.g. Windows + NVidia + Chrome).
It gives an overview of how it can be done one wants to add multi volume support to vtk.js.

See `Examples/Volume/VolumeMapperLightAndShadow/index.js` for an example that uses multiple volumes.

See #3132 for the PR that is based on this PR, that adds multi component


finetjul:
It is to be noted that https://github.com/Kitware/vtk-js/pull/3132 has various fixes that are not in 3133.

PR #3143, opened 2024-10-07
Fix static vtkMapper coincident topology functions 

PaulHax:
When calling the static mapper `vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters`, there is an error because the "class" does not have a publicAPI  with a `modified()` function.  `publicAPI.modifed()` is called by macros.js object setter macro.

### Results

Does not error now when calling `vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters`.  

### Changes

The error fix involved adding a noOp `modified` function.  Alternatives could be to check if `publicAPI` has `modified` in macros.js or upgrade the Mapper static object with macros.js' `function obj(publicAPI = {}, model = {})` "constructor."   

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
+++ b/Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
@@ -29,7 +29,10 @@ const staticOffsetModel = {
   Line: { factor: 1, offset: -1 },
   Point: { factor: 0, offset: -2 },
 };
-const staticOffsetAPI = {};
+const noOp = () => undefined;
+const staticOffsetAPI = {
+  modified: noOp,
+};
 
 addCoincidentTopologyMethods(
   staticOffsetAPI,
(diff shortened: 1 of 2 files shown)


Sources/Rendering/Core/Mapper/CoincidentTopologyHelper.js
```diff
@@ -29,7 +29,10 @@ const staticOffsetModel = {
   Line: { factor: 1, offset: -1 },
   Point: { factor: 0, offset: -2 },
 };
-const staticOffsetAPI = {};
+const noOp = () => undefined;
+const staticOffsetAPI = {
+  modified: noOp,
```
finetjul, line 34: why bother setting the key value to `modified` if it will be ignored  line 65 ?

PaulHax: > When calling the static mapper vtkMapper.setResolveCoincidentTopologyPolygonOffsetParameters, there is an error because the "class" does not have a publicAPI with a modified() function.

finetjul: gotcha, you might want to add a "comment" in the code to make it less confusing.

PaulHax: Added comment.

finetjul (approved):
LGTM

PR #3173, opened 2024-11-09
PointPicker: PointId would be overwritten with multiple actors.

yarous224:
### Context

When using multiple actors with PointPicker, the pointID is overwritten by the latter actor, regardless of whether it finds a point. I referred to VTK C++ to modify this part.

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


floryst (approved):
LGTM based on the C++ implementation.

PR #3191, opened 2024-12-15
fix(IFCImporter): fix normals

### LINKED ISSUE #3189: [Bug] Wrong normals in IFC ?
### Bug description

Depending on the camera position, the IFC scene example gets all black

image
image

### Steps to reproduce

open the [IFC example](https://kitware.github.io/vtk-js/examples/IFCImporter.html) and rotate the camera

### Detailed Behavior

Maybe there is a problem with the normals ? the lights ?
Alternatively, should the scene example file be revised and add more "ambient" light ?

### Expected Behavior

I would expect a more lit scene

### Environment

- vtk.js version: master
- Browsers: Chrome
- OS: Windows

daker:
### Context
fix #3189

### Results

https://github.com/user-attachments/assets/6db63ed2-ff76-4a48-862b-fa9c4d173bd5

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/IO/Geometry/IFCImporter/index.js
+++ b/Sources/IO/Geometry/IFCImporter/index.js
@@ -14,6 +14,7 @@ import vtkPolyData from 'vtk.js/Sources/Common/DataModel/PolyData';
 import vtkCellArray from 'vtk.js/Sources/Common/Core/CellArray';
 import vtkAppendPolyData from 'vtk.js/Sources/Filters/General/AppendPolyData';
 import vtkMatrixBuilder from 'vtk.js/Sources/Common/Core/MatrixBuilder';
+import { mat3 } from 'gl-matrix';
 
 const { vtkErrorMacro } = macro;
 
@@ -94,11 +95,16 @@ function vtkIFCImporter(publicAPI, model) {
         .buildFromRadian()
         .setMatrix(userMatrix);
 
+      const normalMatrix = vtkMatrixBuilder
+        .buildFromRadian()
+        .multiply3x3(mat3.fromMat4(mat3.create(), userMatrix));
+
       for (let i = 0; i < vertices.length; i += 6) {
         const point = [vertices[i], vertices[i + 1], vertices[i + 2]];
         const normal = [vertices[i + 3], vertices[i + 4], vertices[i + 5]];
 
-        transformMatrix.apply(point).apply(normal);
+        transformMatrix.apply(point);
+        normalMatrix.apply(normal);
 
         pointValues[i / 2] = point[0];
         pointValues[i / 2 + 1] = point[1];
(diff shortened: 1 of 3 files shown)


Sources/IO/Geometry/IFCImporter/index.js
```diff
@@ -244,32 +245,40 @@ function vtkIFCImporter(publicAPI, model) {
...
-      let apd = vtkAppendPolyData.newInstance();
+      const oapd = vtkAppendPolyData.newInstance();
 
       opaqueMeshes.forEach((mesh) => {
         const pd = createColoredPolyDataFromIFCMesh(mesh);
-        apd.addInputData(pd);
+        oapd.addInputData(pd);
       });
 
+      const opdn = vtkPolyDataNormals.newInstance();
```
finetjul, line 255: is this overwritting existing normals ? If so, it might not be ideal. 
vtkPolyDataNormals does not perform well in case of sharp edges.

daker: this is only for "merged geometries" which create two actors one to group opaque meshes and one transparent to group transparent meshes, the default option will create actors with their correct normals.

finetjul: Would you say that the AppendPolyData is "dropping" the normals of the input polydatas ?

daker: By looking at the code, it doesn't seems to be the case

finetjul: vtkAppendPolyData seems to preserve normals if they exist in ALL the input polydatas.

So why do you need to recompute normals here ? Is is because some polydata are missing normals ?

daker: i think it should be fixed

Sources/IO/Geometry/IFCImporter/index.js
```diff
@@ -94,11 +94,34 @@ function vtkIFCImporter(publicAPI, model) {
...
+        userMatrix[10],
+        0,
+        0,
+        0,
+        0,
+        1,
+      ];
+      const rotationMatrix = vtkMatrixBuilder
+        .buildFromRadian()
+        .setMatrix(rotMatrix);
```
finetjul, line 117: Could the following work ?
`const rotationMatrix = vtkMatrixBuilder.buildFromRadian().multiply3x3(mat3.fromMat4(userMatrix))`

daker: It works with a small adjustment

finetjul (approved):
LGTM

PR #3193, opened 2024-12-15
fix(GLTFImporter): fix GLTF Draco normals

### LINKED ISSUE #3192: [Bug] GLTF Draco wrong normals
### Bug description

Wrong Normals with Draco format

https://github.com/user-attachments/assets/782ae8de-6d7e-489d-a892-8eb6483e233b

### Steps to reproduce

Open https://kitware.github.io/vtk-js/examples/GLTFImporter/index.html?model=CesiumMan&flavor=glTF-Draco&scene=0&viewAPI=WebGPU

### Detailed Behavior

_No response_

### Expected Behavior

Draco format renders correctly.

### Environment

- vtk.js version:
- Browsers:
- OS:

daker:
### Context

fix #3192

### Results

https://github.com/user-attachments/assets/742a0bfd-aae1-4aee-b2bc-92ebb4312eef

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/IO/Geometry/GLTFImporter/Extensions.js
+++ b/Sources/IO/Geometry/GLTFImporter/Extensions.js
@@ -86,7 +86,7 @@ export function handleKHRLightsPunctual(extension, transformMatrix, model) {
  *
  * @param {object} extension - The KHR_draco_mesh_compression extension object.
  */
-export async function handleKHRDracoMeshCompression(extension) {
+export function handleKHRDracoMeshCompression(extension) {
   const reader = vtkDracoReader.newInstance();
   reader.parse(extension.bufferView);
   return reader.getOutputData();
--- a/Sources/IO/Geometry/DracoReader/example/index.js
+++ b/Sources/IO/Geometry/DracoReader/example/index.js
@@ -49,16 +49,13 @@ function update() {
 // Dynamic script loading from CDN
 // ----------------------------------------------------------------------------
 
-// Prevent error when draco try to set the export on module
-window.module = {};
-
 // Add new script tag with draco CDN
 vtkResourceLoader
-  .loadScript('https://unpkg.com/draco3d@1.3.4/draco_decoder_nodejs.js')
-  .then(() => {
+  .loadScript('https://unpkg.com/draco3d@1.5.7/draco_decoder_nodejs.js')
+  .then(async () => {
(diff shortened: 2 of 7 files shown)


Sources/IO/Geometry/GLTFImporter/Extensions.js
```diff
@@ -89,7 +90,10 @@ export function handleKHRLightsPunctual(extension, transformMatrix, model) {
 export async function handleKHRDracoMeshCompression(extension) {
   const reader = vtkDracoReader.newInstance();
   reader.parse(extension.bufferView);
-  return reader.getOutputData();
+  const pdn = vtkPolyDataNormals.newInstance();
```
finetjul, line 93: shouldn't you NOT compute polydatanormals if the file already contains normals ?

daker: Yes, i saw that the vtkDracoReader is already setting the normals but i am not sure why the result is not correct, i will see if i can find why

daker: If i disable the [setNormals ](https://github.com/Kitware/vtk-js/blob/89758edd084dc59dd85908afc2e49ba9d1dc4230/Sources/IO/Geometry/DracoReader/index.js#L174) in vtkDracoReader the rendering is the expected one and the normals array from GLTF and GLTF-Draco are the same size but not same values

finetjul: You might want to use the GlyphMapper to visualize the normals created by Draco. Maybe that would give a clue about what is wrong...

daker:
@finetjul I have upgraded the draco lib to the latest version and i generate the normals if they are missing.

finetjul (approved):
LGTM.
@floryst could it be considered as breaking change ?

daker:
@finetjul the only breaking change is the switch to promise/async setDracoDecoder imposed by the library.

floryst (approved):
I'm good to keep this as a non-breaking change. I'm partly making this call due to draco3d not designating it as a breaking change (assuming they follow semantic versioning), as they introduce it in the 1.3.6->1.4.0 release.

PR #3197, opened 2024-12-25
fix(CanvasView): allow setting canvas element

daker:
### Context
Allow setting the canvas element for `RemoteView`

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Misc/RemoteView/index.js
+++ b/Sources/Rendering/Misc/RemoteView/index.js
@@ -17,7 +17,9 @@ function vtkRemoteView(publicAPI, model) {
   model.classHierarchy.push('vtkRemoteView');
 
   // Constructor
-  model.canvasView = vtkCanvasView.newInstance();
+  model.canvasView = vtkCanvasView.newInstance({
+    canvas: model.canvasElement,
+  });
   model.interactorStyle = vtkInteractorStyleRemoteMouse.newInstance();
 
   model.interactor = vtkRenderWindowInteractor.newInstance();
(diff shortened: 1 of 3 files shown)


floryst (approved):
FYI @jourdain

Sources/Rendering/Misc/RemoteView/index.js
```diff
@@ -225,6 +228,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'rpcMouseEvent',
     'rpcGestureEvent',
     'rpcWheelEvent',
+    'canvasElement',
```
jourdain, line 231: The `set` won't be that useful since canvas vs canvasElement are 2 different names.

daker: can you explain more ? in line 20 i am using the following

```js
model.canvasView = vtkCanvasView.newInstance({
    canvas: model.canvasElement,
});
```

jourdain: Yes but what I mean is by calling `obj.setCanvasElement(canvas)` won't update the `vtkCanvasView` canvas. So instead of putting __canvasElement__ inside a __setGet__ macro, I would be fine to only put it inside a __get__ macro.

jourdain (approved):

Sources/Rendering/Misc/RemoteView/index.js
```diff
@@ -219,6 +222,7 @@ export function extend(publicAPI, model, initialValues = {}) {
     'interactiveRatio',
     'stillQuality',
     'stillRatio',
+    'canvasElement',
```
finetjul, line 225: if no `canvasElement` is specified at construction time (i.e. as a parameter to `newInstance()`), `getCanvasElement()` will always return null. Shouldn't you update `model.canvasElement` when you create the canvas ?

jourdain: indeed that could be a good addon

jourdain (approved):

PR #3204, opened 2025-01-10
fix(WidgetManager): cross-renderer widget usage

### LINKED ISSUE #3144: [Bug] InteractiveOrientationWidget stuck in active state
### Bug description

InteractiveOrientationWidget stops the main view from working anymore. Can be fixed by clicking on the widget again.

### Steps to reproduce

Go to https://kitware.github.io/vtk-js/examples/InteractiveOrientationWidget.html

* Start a quick mouse drag from orientation widget and end the drag outside the widget in the canvas area.
* OrientationMarkerWidget triggers orientation change but keeps active state
* Try to rotate model
* Model is stuck

### Detailed Behavior

This often happens if you want to spin your model around and start the drag too close to the widget.

https://github.com/Kitware/vtk-js/blob/4691eead12c3291ca786028b8f03e84c8dcf0bfb/Sources/Widgets/Widgets3D/InteractiveOrientationWidget/behavior.js#L32 `getActive` returns true and thus the mouse event is aborted.

### Expected Behavior

The widget never gets stuck in active mode.

The same kind of behavior can be caused by starting a normal model rotate and ending the drag outside browser: model will stay in rotation mode. Could this be improved by ending LeftButtonPress event on "drags outside"? I guess this event is somehow virtually created from pointerup/pointerdown events.

### Environment

- vtk.js version: latest
- Browsers: any
- OS: any

floryst:
### Context

Widgets should not remain in an active state if interaction occurs outside of the containing renderer.

An alternative to deactivating all widgets in the widget manager in such a scenario is to filter out interaction events at the widget level that are not targeting the widget's renderer. Such an alternative has the benefit of allowing widgets to remain active for future in-renderer interactions.

I'm still figuring out if this change might break any existing expectations.

Addresses #3144. 

### Results

- Fixes #3144, and probably any other widget that operates in a overlapping renderer
### Changes

- [x] Fixes to vtkWidgetManager
### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -183,6 +183,17 @@ function vtkWidgetManager(publicAPI, model) {
     }
   }
 
+  const deactivateAllWidgets = () => {
+    let wantRender = false;
+    for (let i = 0; i < model.widgets.length; i++) {
+      const w = model.widgets[i];
+      wantRender ||= !!w.getActiveState();
+      w.deactivateAllHandles();
+    }
+
+    if (wantRender) model._interactor.render();
+  };
+
   const handleEvent = async (callData, fromTouchEvent = false) => {
     if (
       !model.isAnimating &&
@@ -192,6 +203,8 @@ function vtkWidgetManager(publicAPI, model) {
       const callID = Symbol('UpdateSelection');
       model._currentUpdateSelectionCallID = callID;
       await updateSelection(callData, fromTouchEvent, callID);
+    } else {
+      deactivateAllWidgets();
     }
   };
 


jord1e:
Hi @floryst, this seems to work great

I am getting some weird behaviour with the widget disappearing and my screen zooming out in some situations, but I am 99.99% sure that it is because of my own flawed self-implemented zooming logic.

In general, clicking the widget works as expected, and I'm no longer getting the freezes on Firefox/Windows

Thank you so much, look forward to this being merged if no other issues are encountered

floryst:
@finetjul what do you think about this approach vs the one I mentioned in the PR description?

lahdemaki:
Thanks for the suggested fix! Any update on accepting this or some other fix to https://github.com/Kitware/vtk-js/issues/3144 @finetjul ?

finetjul:
@lahdemaki we actually discussed it recently with @floryst. We discussed a slightly different approach. @floryst have you had a chance to work on the approach ?

floryst:
Update: I have a slightly different approach in the https://github.com/Kitware/vtk-js/tree/add-renderer-change-event branch, which adds a RendererChange event. Whenever a widget detects rendererChange, it deactivates itself.

This approach doesn't work well with the vtkPaintWidget. In the scenario with two renderers, if the mouse leaves the scene the paint widget deactivates, requiring users to re-activate the widget when re-entering the scene.

@finetjul I think a better approach would be to have each of a widget's event handlers filter only on events targeting the associated view widget's renderer. What do you think?

floryst:
I'll pick this back up this week. I'm opting to instead filter events rather than the current implementation, because disabling widgets when changing renderer focus is not ideal.

Antti-Palola:
@floryst did you get to check this out?

PR #3207, opened 2025-01-22
fix(URLExtract): use builtin parser

floryst:
### Context
Certain URLs are incorrectly parsed, e.g. URL parameters that have equal signs (`=`) as part of their value. (Yes, those symbols should be percent-encoded, but that's not always the case.)

See: https://github.com/Kitware/VolView/issues/688

### Results
Use URLSearchParams to parse the URL instead of string splits.

### Changes

- [x] vtkURLExtract uses URLSearchParams

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Common/Core/URLExtract/index.js
+++ b/Sources/Common/Core/URLExtract/index.js
@@ -39,13 +39,8 @@ function extractURLParameters(
 ) {
   const summary = {};
   const convert = castToNativeType ? toNativeType : identity;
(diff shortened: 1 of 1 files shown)


daker (approved):
LGTM

finetjul (approved):
LGTM

PR #3210, opened 2025-02-04
refactor(glyph3dmapper): Typescript definition

sankhesh:
The mapper expects array name strings for the scale and orientation arrays. This change ensures that the typescript definition reflects the same.

BREAKING CHANGE: scaleArray and orientationArray changed API from `number[]` to `string`

### Context

The Glyph3DMapper had inconsistent API typescript definitions for scale and orientation array parameters.

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/Glyph3DMapper/index.d.ts
+++ b/Sources/Rendering/Core/Glyph3DMapper/index.d.ts
@@ -12,17 +12,65 @@ interface IPrimitiveCount {
 export interface IGlyph3DMapperInitialValues extends IMapperInitialValues {
   orient?: boolean;
   orientationMode?: OrientationModes;
-  orientationArray?: number[];
+  orientationArray?: string;
   scaling?: boolean;
   scaleFactor?: number;
   scaleMode?: ScaleModes;
-  scaleArray?: number[];
+  scaleArray?: string;
   matrixArray?: number[];
   normalArray?: number[];
   colorArray?: number[];
 }
 
 export interface vtkGlyph3DMapper extends vtkMapper {
+  /**
+   * Get the bounds for this mapper as [xmin, xmax, ymin, ymax,zmin, zmax].
+   * @return {Bounds} The bounds for the mapper.
+   */
+  getBounds(): Bounds;
+
+  /**
+   *
+   */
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/Glyph3DMapper/index.d.ts
```diff
@@ -12,85 +12,101 @@ interface IPrimitiveCount {
...
-  orientationArray?: number[];
+  orientationArray?: string;
   scaling?: boolean;
   scaleFactor?: number;
   scaleMode?: ScaleModes;
-  scaleArray?: number[];
+  scaleArray?: string
   matrixArray?: number[];
   normalArray?: number[];
   colorArray?: number[];
```
finetjul, line 22: why not change the other array names ?

finetjul (commented):
This does not seem to be a BREAKING_CHANGE to me. 
It's just a "documentation" fix.

Sources/Rendering/Core/Glyph3DMapper/index.d.ts
sankhesh: Sorry, I didn't understand your question. Did you mean changing the type for `matrixArray`, etc.? Those are not names but data arrays, hence `number[]`.

sankhesh:
> This does not seem to be a BREAKING_CHANGE to me.
> It's just a "documentation" fix.

Awesome. I wasn't sure about this either. I figured people using typescript will see this as an API change.

Sources/Rendering/Core/Glyph3DMapper/index.d.ts
finetjul: ohh, I see. I was wrongly assuming that all arrays would behave the same.

my bad.

floryst (approved):
Agreed, this is just fixing the incorrect docs.

floryst (changes requested):
That was supposed to be a "Request Changes" for dropping "BREAKING CHANGES" from the commit message

floryst (approved):

PR #3217, opened 2025-02-14
feat(stlreader): add removeDuplicateVertices method to vtkSTLReader

jreyero12:
### Context
Edges could not be hidden when displaying a stl

### Results
vs

### Changes

- [x] TypeScript definitions were updated to match those changes

- [x] Tested environment:
  - **vtk.js**: 32.9.1
  - **OS**: Ubuntu 22.04.3 LTS
  - **Browser**: chrome 112.0.5615.165

### DIFF
--- a/Sources/IO/Geometry/STLReader/index.d.ts
+++ b/Sources/IO/Geometry/STLReader/index.d.ts
@@ -94,6 +94,12 @@ export interface vtkSTLReader extends vtkSTLReaderBase {
    * @param {ISTLReaderOptions} [option] The STL reader options.
    */
   setUrl(url: string, option?: ISTLReaderOptions): Promise<string | any>;
+
+  /**
+   *
+   * @param offset
+   */
+  removeDuplicateVertices(offset: number = 5): void;
 }
 
 /**
--- a/Sources/IO/Geometry/STLReader/index.js
+++ b/Sources/IO/Geometry/STLReader/index.js
@@ -331,6 +331,60 @@ function vtkSTLReader(publicAPI, model) {
   publicAPI.requestData = (inData, outData) => {
     publicAPI.parse(model.parseData);
   };
+
+  publicAPI.removeDuplicateVertices = (offset = 5) => {
+    if (!model.output || !model.output[0]) {
+      console.warn('Load polydata first.');
+      return;
+    }
+    const polydata = model.output[0];
+
+    const vertices = polydata.getPoints().getData();
+    const faces = polydata.getPolys().getData();
+
+    if (!vertices || !faces) {
+      console.warn('No valid polydata.');
+      return;
+    }
+
+    const vMap = new Map();
+    const vIndexMap = new Map();
+    let vInc = 0;
+    for (let i = 0; i < vertices.length; i += 3) {
+      const k1 = (vertices[i] * 10 ** offset).toFixed(0);
+      const k2 = (vertices[i + 1] * 10 ** offset).toFixed(0);
+      const k3 = (vertices[i + 2] * 10 ** offset).toFixed(0);
+      const key = `${k1},${k2},${k3}`;
+      if (vMap.get(key) !== undefined) {
+        vIndexMap.set(i / 3, vMap.get(key));
+      } else {
+        vIndexMap.set(i / 3, vInc);
+        vMap.set(key, vInc);
+        vInc++;
+      }
+    }
+
+    const outVerts = new Float32Array(vMap.size * 3);
+    vMap.keys().forEach((k) => {
+      const j = vMap.get(k) * 3;
+      [outVerts[j], outVerts[j + 1], outVerts[j + 2]] = k
+        .split(',')
+        .map((e) => +e * 10 ** -offset);
+    });
+
+    const outFaces = new Int32Array(faces);
+    for (let i = 0; i < faces.length; i += 4) {
+      outFaces[i] = 3;
+      outFaces[i + 1] = vIndexMap.get(faces[i + 1]);
+      outFaces[i + 2] = vIndexMap.get(faces[i + 2]);
+      outFaces[i + 3] = vIndexMap.get(faces[i + 3]);
+    }
+
+    polydata.getPoints().setData(outVerts);
+    polydata.getPolys().setData(outFaces);
+
+    publicAPI.modified();
+  };
 }
 
 // ----------------------------------------------------------------------------
--- a/Sources/IO/Geometry/STLReader/example/index.js
+++ b/Sources/IO/Geometry/STLReader/example/index.js
@@ -7,7 +7,7 @@ import vtkActor from '@kitware/vtk.js/Rendering/Core/Actor';
 import vtkFullScreenRenderWindow from '@kitware/vtk.js/Rendering/Misc/FullScreenRenderWindow';
 import vtkMapper from '@kitware/vtk.js/Rendering/Core/Mapper';
 import vtkSTLReader from '@kitware/vtk.js/IO/Geometry/STLReader';
-
+import vtkPolyDataNormals from '@kitware/vtk.js/Filters/Core/PolyDataNormals';
 // ----------------------------------------------------------------------------
 // Example code
 // ----------------------------------------------------------------------------
@@ -17,7 +17,9 @@ const mapper = vtkMapper.newInstance({ scalarVisibility: false });
 const actor = vtkActor.newInstance();
 
 actor.setMapper(mapper);
-mapper.setInputConnection(reader.getOutputPort());
+const normals = vtkPolyDataNormals.newInstance();
+normals.setInputConnection(reader.getOutputPort());
+mapper.setInputConnection(normals.getOutputPort());
 
 // ----------------------------------------------------------------------------
 
@@ -54,6 +56,7 @@ function handleFile(event) {
     const fileReader = new FileReader();
     fileReader.onload = function onLoad(e) {
       reader.parseAsArrayBuffer(fileReader.result);
+      reader.removeDuplicateVertices(5);
       update();
     };
     fileReader.readAsArrayBuffer(files[0]);


Sources/IO/Geometry/STLReader/index.d.ts
```diff
@@ -94,6 +94,12 @@ export interface vtkSTLReader extends vtkSTLReaderBase {
    * @param {ISTLReaderOptions} [option] The STL reader options.
    */
   setUrl(url: string, option?: ISTLReaderOptions): Promise<string | any>;
+
+  /**
+   *
+   * @param offset
+   */
+  removeDuplicateVertices(offset: number = 5): void;
```
finetjul, line 102: Please make it an option so that, the function is being called 
 automatically after reading the STL.

finetjul: Maybe `offset` could be changed into `tolerance` or `precision`

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -331,6 +331,60 @@ function vtkSTLReader(publicAPI, model) {
...
+        vIndexMap.set(i / 3, vMap.get(key));
+      } else {
+        vIndexMap.set(i / 3, vInc);
+        vMap.set(key, vInc);
+        vInc++;
+      }
+    }
+
+    const outVerts = new Float32Array(vMap.size * 3);
+    vMap.keys().forEach((k) => {
```
finetjul, line 368: forEach is usually a bit slow, we prefer regular `for` loops.

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -331,6 +331,60 @@ function vtkSTLReader(publicAPI, model) {
...
+      const j = vMap.get(k) * 3;
+      [outVerts[j], outVerts[j + 1], outVerts[j + 2]] = k
+        .split(',')
+        .map((e) => +e * 10 ** -offset);
+    });
+
+    const outFaces = new Int32Array(faces);
+    for (let i = 0; i < faces.length; i += 4) {
+      outFaces[i] = 3;
```
finetjul, line 377: are we sure we always have 3 points per face ? no more ?

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -331,6 +331,60 @@ function vtkSTLReader(publicAPI, model) {
...
+      outFaces[i + 1] = vIndexMap.get(faces[i + 1]);
+      outFaces[i + 2] = vIndexMap.get(faces[i + 2]);
+      outFaces[i + 3] = vIndexMap.get(faces[i + 3]);
+    }
+
+    polydata.getPoints().setData(outVerts);
+    polydata.getPolys().setData(outFaces);
```
finetjul, line 384: how about lines ? and vertices ?

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -331,6 +331,60 @@ function vtkSTLReader(publicAPI, model) {
...
+
+    publicAPI.modified();
```
finetjul, line 386: if you don't detect duplicated, maybe you shouldn't call modified()

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -331,6 +331,60 @@ function vtkSTLReader(publicAPI, model) {
...
   };
+
+  publicAPI.removeDuplicateVertices = (offset = 5) => {
+    if (!model.output || !model.output[0]) {
+      console.warn('Load polydata first.');
+      return;
+    }
+    const polydata = model.output[0];
+
+    const vertices = polydata.getPoints().getData();
```
finetjul, line 342: a `vertex` is a "cell" in the VTK world, it's better to call it `points` or `coords`

finetjul (changes requested):

Sources/IO/Geometry/STLReader/index.js
jreyero12: Yes, an stl can only have 3 points per face.
It is a triangular mesh and can only be triangular.

Sources/IO/Geometry/STLReader/index.d.ts
jreyero12: I don't know how to make it be called at the beginning 
Where can I see an example ?
Do I have to add a parameter in DEFAULT_VALUES ?

Sources/IO/Geometry/STLReader/index.js
jreyero12: I don't know what you want me to do here

jourdain: stl only has triangles, no lines or vertices.

Sources/IO/Geometry/STLReader/index.d.ts
jourdain: I think the idea is that VTK do not have methods that modify its output. Usually we have methods to define a state and mark the filter modified() so when it re-execute, it can compute its new output.

In other words, you should have set/get for removeDuplicateVertices(true/false), tolerance(float) or you could combine both into just tolerance (set/getMergeVerticesTolerance()). And if tolerance is negative or 0 you do not remove duplicate points in the request data method.

finetjul: You can look at `automaticPlaneGeneration` in `TextureMapToPlane`

Sources/IO/Geometry/STLReader/index.d.ts
```diff
@@ -76,6 +81,12 @@ export interface vtkSTLReader extends vtkSTLReaderBase {
    */
   requestData(inData: any, outData: any): void;
 
+  /**
+   *
+   * @param tolerance
+   */
+  removeDuplicateVertices(tolerance: number = 5): void;
```
finetjul, line 88: I think this function should not be made public.

Sources/IO/Geometry/STLReader/index.d.ts
```diff
@@ -96,10 +107,10 @@ export interface vtkSTLReader extends vtkSTLReaderBase {
   setUrl(url: string, option?: ISTLReaderOptions): Promise<string | any>;
 
   /**
-   *
-   * @param offset
+   * Turn on/off automatic removeDuplicateVertices
```
finetjul, line 110: You might want to give a bit more details:
```
After reading the STL file, if `tolerance` is >= 0, then points with the same coordinates at 10 power tolerance are merged.
For a smooth rendering, you might want to compute normals with vtkPolyDataNormals.
```

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -289,6 +289,10 @@ function vtkSTLReader(publicAPI, model) {
 
     // Add new output
     model.output[0] = polydata;
+
+    if (model.removeDuplicateVertices > 0) {
```
finetjul, line 293: I guess 0 could be a valid value. 
how about comparing with -1 instead ?

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -326,37 +330,43 @@ function vtkSTLReader(publicAPI, model) {
...
+      publicAPI.removeDuplicateVertices(model.removeDuplicateVertices);
+    }
   };
 
   publicAPI.requestData = (inData, outData) => {
     publicAPI.parse(model.parseData);
   };
 
-  publicAPI.removeDuplicateVertices = (offset = 5) => {
+  publicAPI.removeDuplicateVertices = (tolerance = 5) => {
```
finetjul, line 343: can you please move it outside of `publicAPI` ?
(like `triangulatePolygon` in `ContourTriangulator`)

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -365,12 +375,15 @@ function vtkSTLReader(publicAPI, model) {
     }
 
     const outVerts = new Float32Array(vMap.size * 3);
-    vMap.keys().forEach((k) => {
+    const keys = Array.from(vMap.keys());
+    for (let i = 0; i < keys.length; i++) {
```
finetjul, line 379: I suppose `for (const key of vMap.keys()) {` works as well and avoids a copy.

jreyero12: eslint eslint does not allow "iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations."

Sources/IO/Geometry/STLReader/index.js
jreyero12: Wouldn't it be better to leave it inside the function vtkSTLReader function as fetchData so that it can access model and publicAPI from removeDuplicateVertices.
Like this: 
```
function vtkSTLReader(publicAPI, model) {
	function removeDuplicateVertices(tolerance = 5) {
		model.output
		publicAPI.modified()
	}
}
```

finetjul: Yes, that works as well.

Sources/IO/Geometry/STLReader/index.js
finetjul: Ok, let's leave it as is then. No worries.

jreyero12:
Is there anything left to do ?

jourdain (approved):

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -341,6 +407,7 @@ const DEFAULT_VALUES = {
   // baseURL: null,
   // dataAccessHelper: null,
   // url: null,
+  removeDuplicateVertices: 0,
```
finetjul, line 410: please disable the duplicates removal by default (`// removeDuplicateVertices: null` or `removeDuplicateVertices: -1`).
That would be a breaking change otherwise.

finetjul (changes requested):

Sources/IO/Geometry/STLReader/index.js
jreyero12: Ready

finetjul (approved):
LGTM

PR #3220, opened 2025-02-18
fix(webxr): remove controllers rays when stopping XR

munozco:
### Context

When using `vtkWebXRRenderWindowHelper` with `drawControllersRay=true`, rays are drawn to represent the controllers
but they are not removed when stopping the XR session. This leads to the camera not being centered correctly.

Here's a comparison of before starting the XR session and after stopping it:
before
after

### Results

Thanks to this small fix, the rays' actors are removed when exiting the XR session, and the camera can be re-centered 

### Changes

When `vtkWebXRRenderWindowHelper.stopXR()` is called, every ray's actor is removed so they are not visible anymore.
  - No changes in definitions

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: latest master ([413f07a](https://github.com/Kitware/vtk-js/commit/413f07a1cd4fe0166e3e5564a4e1899260dc01b1))
  - **OS**: Windows 11 WSL2 Debian 12
  - **Browser**: Chromium 133.0.6943.98

### DIFF
--- a/Sources/Rendering/WebXR/RenderWindowHelper/index.js
+++ b/Sources/Rendering/WebXR/RenderWindowHelper/index.js
@@ -173,6 +173,15 @@ function vtkWebXRRenderWindowHelper(publicAPI, model) {
       const gl = model.renderWindow.get3DContext();
       gl.bindFramebuffer(gl.FRAMEBUFFER, null);
 
+      // Remove controllers ray
+      const ren = model.renderWindow.getRenderable().getRenderers()[0];
+      model.xrSession.inputSources.forEach((inputSource) => {
+        if (model.inputSourceToRay[inputSource.handedness]) {
+          ren.removeActor(model.inputSourceToRay[inputSource.handedness].actor);
+          model.inputSourceToRay[inputSource.handedness].visible = false;
+        }
+      });
+
       await model.xrSession.end().catch((error) => {
         if (!(error instanceof DOMException)) {
           throw error;


munozco:
@bourdaisj @finetjul please review

finetjul (approved):
LGTM

Sources/Rendering/WebXR/RenderWindowHelper/index.js
```diff
@@ -173,6 +173,15 @@ function vtkWebXRRenderWindowHelper(publicAPI, model) {
       const gl = model.renderWindow.get3DContext();
       gl.bindFramebuffer(gl.FRAMEBUFFER, null);
 
+      // Remove controllers ray
+      const ren = model.renderWindow.getRenderable().getRenderers()[0];
+      model.xrSession.inputSources.forEach((inputSource) => {
+        if (model.inputSourceToRay[inputSource.handedness]) {
+          ren.removeActor(model.inputSourceToRay[inputSource.handedness].actor);
```
bourdaisj, line 180: why not destroying the pipeline instead of only removing/hiding actor?

floryst:
I think this was mistakenly closed from the other issue. Re-opening.

Sources/Rendering/WebXR/RenderWindowHelper/index.js
munozco: If we want to call `startXR()` again, we would have to call `createRay()` again. [This line](https://github.com/munozco/vtk-js/blob/5f688d88ea2d0df1f291f70c0c876b2eb6f7ba15/Sources/Rendering/WebXR/RenderWindowHelper/index.js#L235) checks if the pipeline already exists, so I think this is fine if we keep the same pipeline for the rays between xr sessions.

PR #3221, opened 2025-02-18
feat: improve performance of ImageMapper.buildBufferObjects

jmannau:
### Context
This fix addresses this issue in the Cornerstone library https://github.com/cornerstonejs/cornerstone3D/issues/1756

When rendering cine-loop dicom files, each frame is rendered in a loop and playback performance is poor on lower powered devices.  When a performance trace is completed, `computeFastRange` is one of the largest total time in the cineloop playback.

On each frame, `getRange` is being called on an instance of `vtkDataArray` created in the `ImageMapper` causing the cumulative processing time.

This PR fixes this issue by updating `ImageMapper.buildBufferObjects` so that `create2DFilterableForRaw` can accept the range from the source `imageScalars`, meaning the range is preset, removing the need to calculate it where it is a direct copy.

### Results

As described in this comment https://github.com/cornerstonejs/cornerstone3D/issues/1756#issuecomment-2661911052 this PR fixes the issue, significantly reducing the cumulative computation time.

### Changes

There are 2 main changes:
1. Updating OpenGL `vtkOpenGLTexture.create2DFilterableFromRaw` to accept an optional `ranges` parameter. If included, this parameter is password to the new `vtkDtaArray` instance that is created.
2. Updating OpenGL `ImageMapper.buildBufferObjects` so that if the `(ijkMode === SlicingMode.K || ijkMode === SlicingMode.NONE)` and `sliceOffet === 0` then the range from the source `imageScalars` is passed to `create2DFilterableFromRaw`.  

In the context of Cornerstone, this means that the range for each frame only needs to be calculated once on first render and then on subsequent renders, the range can be preset reducing computation time.

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

### Funding
This contribution is funded by [Pliny](https://pliny.com.au).


finetjul (commented):
I do not see any change in the code. 
Only style change.

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -1237,6 +1237,14 @@ function vtkOpenGLImageMapper(publicAPI, model) {
       const spatialExt = image.getSpatialExtent();
       const basicScalars = imgScalars.getData();
       let scalars = null;
+      /**
+       * if available, used to store the range of the scalars array. If
+       * available, pre-setting the scalras range can prevent it being
```
finetjul, line 1242: if available is repeated.
scalras range -> scalar range

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -1310,6 +1318,24 @@ function vtkOpenGLImageMapper(publicAPI, model) {
         ptsArray[9] = spatialExt[1];
         ptsArray[10] = spatialExt[3];
         ptsArray[11] = sliceDepth;
+        if (sliceOffset === 0) {
+          // If the sliceOffset is 0, the scalars array is a view of the
+          // original scalars array. As a micro-optimization, we can get the
+          // range from the original scalars array and set the range to the
+          // scalars array.  This means there is no need to re-calculate the
+          // ranges for the scalars array.
```
finetjul, line 1326: Isn't `imgScalars.getRange()` the range of the whole 3D volume, while `scalars` contains only a 2D slice ?

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -1310,6 +1318,24 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
+          ranges = [];
+          for (let i = 0; i < numComp; i++) {
+            const [min, max] = imgScalars.getRange(i);
```
finetjul, line 1329: maybe we could expose a convenient `getRanges()` function

jmannau: Would you like me to include a `getRanges()` api in `DataArray` as part of this API or a separate?

Sources/Rendering/OpenGL/ImageMapper/index.js
jmannau: I can't comment on this sorry.

finetjul: the problem is that it might be incorrect to use for a single slice the range computed over the whole 3D volume.

jmannau: I am far from an expert so feel free to suggest an improvement.  From my testing with cornerstone, my understanding is that `if (ijkMode === SlicingMode.K || ijkMode === SlicingMode.NONE)` AND `sliceOffset===0` then `scalars` is a direct copy of imgScalars. Is it possible to further narrow this check to guarantee that we can use the range? perhaps ` && sliceSize === basicScalars.length`?

finetjul: sliceSize is width*height, it is not width*height*depth. So it is not a is a direct copy of imgScalars, it is simply a direct subset of imgScalars but it IS NOT imgScalars. It is way smaller. So they may have a different "range".

However, for performance reasons, it can be argued that to use the "range of the volume" in place of the "range of a slice of the volume". For sure, we know that the range of a slice is at least within the range of its "volume". In a way, it does not make much sense that each slice gets a different range when we scroll along a volume. Therefore, we could decide that for `ImageMapper`, the range of the slice is never computed and the range of the volume is always used instead. That wouldn't be specific to the `ijkMode` value. It is just always the case.

Sources/Rendering/OpenGL/ImageMapper/index.js
finetjul: yes, it can be done as part of this PR.

Sources/Rendering/OpenGL/Texture/index.d.ts
```diff
@@ -244,7 +244,8 @@ export interface vtkOpenGLTexture extends vtkViewNode {
    * @param numComps The number of components in the texture.
    * @param dataType The data type of the texture.
    * @param data The raw data for the texture.
-   * @param preferSizeOverAccuracy Whether to prefer texture size over accuracy.
+   * @param [preferSizeOverAccuracy=false] Whether to prefer texture size over accuracy. Defaults to false.
+   * @param [ranges] The ranges of the data (optional).
```
finetjul, line 248: The precomputed ranges of the data (optional). Provided to prevent computation of the data ranges.

Sources/Rendering/OpenGL/ImageMapper/index.js
jmannau: I have refactored this and created a `getRanges` function in DataArray public api.

Sources/Rendering/OpenGL/ImageMapper/index.js
jmannau: I would appreciate your input to advise the best option here.  

1. I can update `if (sliceOffset === 0) {` to `if (sliceOffset === 0 && scalars.length === basicScalars.length) {` ensuring the range is strictly valid
2. Leave the broader `if (sliceOffset === 0) {` as it is currently, justified as you have in your comment above.

Sources/Rendering/OpenGL/Texture/index.d.ts
jmannau: Updated.

Sources/Rendering/OpenGL/ImageMapper/index.js
finetjul: I would set the range with no condition. No matter what `ijkMode` nor `sliceOffset` worth.

jmannau: That is updated.

floryst:
@jmannau since you're modifying `create2DFilterableFromRaw` positional parameters, I plan on implementing #3222 so that your changes will be adding named parameters instead. I'll update this thread when it's implemented.

Sources/Common/Core/DataArray/index.js
```diff
@@ -281,6 +281,22 @@ function vtkDataArray(publicAPI, model) {
     return model.rangeTuple;
   };
 
+  publicAPI.getRanges = () => {
```
finetjul, line 284: Look at `getRange()`: for multi-components, model.ranges has the size of numberOfComponents + 1 (where the last item is the range across all components):
```
...
if (model.numberOfComponents > 1) {
  const [min, max] = publicAPI.getRange(-1);
  ...
  ranges.push(range);
}
return ranges;
};`

File: Sources/Common/Core/DataArray/index.d.ts [hunk 6]
```diff
@@ -83,6 +83,11 @@ export interface vtkDataArray extends vtkObject {
    */
   setRange(rangeValue: vtkRange, componentIndex: number): Range;
 
+  /**
+   *
```
finetjul, line 87: please add documentation

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -1314,6 +1314,21 @@ function vtkOpenGLImageMapper(publicAPI, model) {
...
 
+      /**
+       *
+       * Fetch the ranges of the source volume, `imgScalars`, and use them when
+       * creating the texture. Whilst the pre-calculated ranges may not be
+       * strictly correct for the slice, it is guarateed to be within the
+       * source volume's range.
+       *
+       * There is a significant performance improvement by pre-setting the range
+       * of the scalars array particaulrly when scrolling through the source
```
finetjul, line 1325: particularly

Sources/Common/Core/DataArray/index.js
```diff
@@ -281,6 +281,22 @@ function vtkDataArray(publicAPI, model) {
...
+      const range = {
+        min,
+        max,
+        component: i,
+      };
+      ranges.push(range);
+    }
+    return ranges;
+  };
+
```
finetjul, line 299: please use `getRanges()` in `vtkDataArray.deepCopy()`

Sources/Common/Core/DataArray/index.js
finetjul: I think an option to not enforce range computation might be useful because calling `getRanges()` might end up being more costly than needed (e.g. let's say you are only using with the range of the first component (or the overall range) of a multi-component array, calling `getRanges()` would end up computing the range of all the components):
```
publicAPI.getRanges => (computeRanges=true) {
  if (!computeRanges) {
     return DeepCopy(model.ranges);
  }
  ...
}
```

jmannau: I can add this. I see that `structuredClone` is used elsewhere.  Should I use `structuredClone` (https://caniuse.com/?search=structuredClone) or is there a preferred method for `deepCopy`?

finetjul: `structuredClone` is fine by me.

Sources/Common/Core/DataArray/index.js
```diff
@@ -281,6 +281,37 @@ function vtkDataArray(publicAPI, model) {
...
+    }
+    // where the number of components is greater than 1, the last element in
+    // the range array is the min,max magnitude of the entire dataset.
+    if (model.numberOfComponents > 1) {
+      /** @type {import('../../../interfaces').vtkRange} */
+      const [min, max] = publicAPI.getRange(-1);
+      const range = {
+        min,
+        max,
+        component: -1,
```
finetjul, line 308: you do not need `component`

Sources/Common/Core/DataArray/index.js
```diff
@@ -281,6 +281,37 @@ function vtkDataArray(publicAPI, model) {
...
+    }
+    /** @type {import('../../../interfaces').vtkRange[]} */
+    const ranges = [];
+    for (let i = 0; i < model.numberOfComponents; i++) {
+      const [min, max] = publicAPI.getRange(i);
+      /** @type {import('../../../interfaces').vtkRange} */
+      const range = {
+        min,
+        max,
+        component: i,
```
finetjul, line 296: you do not need `component`

Sources/Common/Core/DataArray/test/testDataArray.js
```diff
@@ -202,6 +235,47 @@ test('Test vtkDataArray getRange function with multi-channel data.', (t) => {
   t.end();
 });
 
+test('Test vtkDataArray getRanges function with multi-channel data.', (t) => {
+  // create a data array with 3 channel data.
+  const newArray = new Uint16Array(256 * 3);
```
finetjul, line 240: there is no need to make that array that large. 
If you make it up tot 3*10, the magnitude will be a round number and you wouldn't need the convenient `compareFloat` function

jmannau: I can't say I understand how this magnitude is calculated. I just tried an array of 3 * 10, but the maximum of the magnitude was 33.67491648096547.  I tried with other small ranges (3*2 - the max is 3.741657386773941) and wasn't able to get the max to be a whole number.

finetjul: indeed, that won't work with 3 channels, it will work with 4 channels though.

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1368,7 +1368,8 @@ function vtkOpenGLTexture(publicAPI, model) {
     numberOfComponents,
     dataType,
     values,
-    preferSizeOverAccuracy = false
+    preferSizeOverAccuracy = false,
+    ranges = undefined
```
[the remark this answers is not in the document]
finetjul: @jmannau would you mind also add the `ranges` parameter to the 3D functions as well ?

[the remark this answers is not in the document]
jmannau: All done

floryst:
The lint check is failing because the body line is too long:

```
⧗   input: fix(DataArray): update getRanges

add computeRanges arg. If set to false will clone the existing range. If true, will calculate the ranges. Added the magnitude ranges for multi-component ranges
✖   body's lines must not be longer than 100 characters [body-max-line-length]

✖   found 1 problems, 0 warnings
```

jmannau:
I have updated the commit message

jmannau:
Updated and thank you for your review and patience!

finetjul (approved):
LGTM

(thanks for your patience)

Sources/Common/Core/DataArray/index.d.ts
```diff
@@ -83,6 +83,19 @@ export interface vtkDataArray extends vtkObject {
    */
   setRange(rangeValue: vtkRange, componentIndex: number): Range;
 
+  /**
+   * Returns an array of the ranges for each component of the DataArray.
+   * Defaults to computing all the ranges if they aren't already computed.
+   *
+   * Passing `getRanges(false)` will return  a clone of the ranges that have
+   * already been computed. This is useful when you want to avoid recomputing
+   * the ranges, which can be expensive.
```
floryst, line 92: Please add a note about the last element in the returned ranges being the min,max magnitude of the entire dataset. This can be the same as your comment in publicAPI.getRanges().

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -1314,6 +1314,21 @@ function vtkOpenGLImageMapper(publicAPI, model) {
         vtkErrorMacro('Reformat slicing not yet supported.');
       }
 
+      /**
+       *
+       * Fetch the ranges of the source volume, `imgScalars`, and use them when
+       * creating the texture. Whilst the pre-calculated ranges may not be
+       * strictly correct for the slice, it is guarateed to be within the
```
floryst, line 1321: `guarateed` -> `guaranteed`

floryst (commented):
Just two minor things and we will be good to merge!

finetjul (approved):
LGTM

PR #3223, opened 2025-02-25
test: fix async tests

floryst:
### Context

I ran into a case where calls to `t.end()` in async tests broke unrelated tests by being called in a future execution.

Tape tests should not call t.end() in async tests. The GC object is no longer instantiated with the tape context when returning a Promise.

Some of the tests are supposed to be WebGL-only.

There are a lot of changes, but most should be straightforward.

### Results

Tests should all still pass. This PR ensures that there are no delayed `t.end()` calls.

### Changes

- [ ] Tests have been updated

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Documentation/content/docs/develop_test.md
+++ b/Documentation/content/docs/develop_test.md
@@ -45,15 +45,15 @@ test('Validate vtkMyClass properties', (t) => {
 ```js ClassName/test/testRendering.js
 import test from 'tape';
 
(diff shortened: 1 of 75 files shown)


jourdain (approved):

finetjul (approved):
LGTM

PR #3224, opened 2025-02-25
refactor(OpenGL/Texture): use named parameters

### LINKED ISSUE #3222: [Refactor] vtkOpenGLTexture create* methods to use named arguments
### Motivation

Some of the `vtkOpenGLTexture.create*` methods are starting to accrue many parameters (see #3221 and #3077 changes for parameter additions). Example methods:
- [create2DFromRaw](https://github.com/Kitware/vtk-js/blob/722db8a5e9c5e52190f416622c90d07b8a8327f5/Sources/Rendering/OpenGL/Texture/index.js#L853)
- [create2DFilterableFromDataArray](https://github.com/Kitware/vtk-js/blob/722db8a5e9c5e52190f416622c90d07b8a8327f5/Sources/Rendering/OpenGL/Texture/index.js#L1384)
- [create3DFilterableFromDataArray](https://github.com/Kitware/vtk-js/blob/722db8a5e9c5e52190f416622c90d07b8a8327f5/Sources/Rendering/OpenGL/Texture/index.js#L1615)

I would like to propose changing these methods to use named parameters over positional parameters. (See [this comment for more context](https://github.com/Kitware/vtk-js/pull/3077#discussion_r1702070508).)

This issue is for tracking as well as soliciting opinions over implementation approaches and whether this is something to move forward on.

Tagging @finetjul @jourdain @sedghi 

### All named parameters

One approach is to move completely to named parameters:

```js
publicAPI.create3DFilterableFromDataArray = ({ width, height, depth, dataArray, preferSizeOverAccuracy = false } = {}) => { ... }
```

This will name every parameter, but also means that leaving out a required parameter might not be as obvious (which can also be an issue with very long positional params, but I don't consider that a fruitful point of discussion for this suggestion).

My vote is for this option, but there is another option to consider.

### Optional named parameters

We can instead make only the optional parameters part of an "options" record:

```js
publicAPI.create3DFilterableFromDataArray = (width, height, depth, dataArray, { preferSizeOverAccuracy = false } = {}) => { ... }
```

This makes sense for optional params, but we still have the issue with many positionals (e.g. is it width/height or height/width in `create3DFilterableFromDataArray(100, 100, ...)`?)

### Backwards compatibility

* These are mostly used internally, so cutting a new major release is a soft requirement.
* If we moved these from arrow functions to regular functions, we could detect when `arguments.length > 1` and unpack the arguments accordingly.

floryst:
### Context

The create* methods have enough parameters to warrant using named parameters.

Closes #3222 

### Results

- vtkOpenGLTexture `create*` methods now use named parameters

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -1,6 +1,8 @@
 ## From 32.x to 33
 
 - **vtkMapper**: many properties have moved to `vtkVolumeProperty`. The full list of changed methods is: `getAnisotropy`, `getComputeNormalFromOpacity`, `getFilterMode`, `getFilterModeAsString`, `getGlobalIlluminationReach`, `getIpScalarRange`, `getIpScalarRangeByReference`, `getLAOKernelRadius`, `getLAOKernelSize`, `getLocalAmbientOcclusion`, `getPreferSizeOverAccuracy`, `getVolumetricScatteringBlending`, `setAnisotropy`, `setAverageIPScalarRange`, `setComputeNormalFromOpacity`, `setFilterMode`, `setFilterModeToNormalized`, `setFilterModeToOff`, `setFilterModeToRaw`, `setGlobalIlluminationReach`, `setIpScalarRange`, `setIpScalarRangeFrom`, `setLAOKernelRadius`, `setLAOKernelSize`, `setLocalAmbientOcclusion`, `setPreferSizeOverAccuracy`, `setVolumetricScatteringBlending`.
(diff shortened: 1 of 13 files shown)


floryst:
FYI @sedghi

sedghi:
Looks great!

sankhesh (approved):
LGTM

PR #3230, opened 2025-03-11
fix(AbstractPicker): model.pickFromList type is boolean

PaulHax:
Init pickFromList to false not 0.

Fix TypeScript defs.

### DIFF
--- a/Sources/Rendering/Core/AbstractPicker/index.d.ts
+++ b/Sources/Rendering/Core/AbstractPicker/index.d.ts
@@ -1,6 +1,6 @@
 import { vtkObject } from '../../../interfaces';
 import { Vector3 } from '../../../types';
-import vtkActor from '../Actor';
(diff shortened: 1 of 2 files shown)


finetjul:
I guess it can be seen as a breaking change for typescript users. @floryst

PaulHax:
I'm mostly interested in relaxing the `picker.addPickList` type from `vtkActor` to `vtkProp3D`.  Happy to back out the pickFromList boolean cleanup as it seems its common in vtk.js code to use 0 and 1 as booleans.

floryst:
I don't see this as a breaking change, because the only valid values are 0 and 1, and so changing the types from number to boolean is more of a fix to properly annotate the intended type.

I'm good with relaxing the vtkActor type to vtkProp3D.

floryst (approved):

PR #3237, opened 2025-04-04
feat(mappers): set opacity/color/label tex widths

floryst:
### Context

When the opacity and/or color transfer functions have a range width larger than 1024, the uniform texture sampler undersamples the functions, leading to incorrect colors and opacities in the resultant image.

### Results

Users can now configure the sampling texture widths to avoid undersampling functions with large ranges.

### Changes

- [ ] ImageMapper, VolumeMapper, ImageCPRMapper, and ImageResliceMapper now support customizable texture widths for opacity, color, and label outline (when applicable) textures

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


finetjul (approved):
LGTM

It would be nice if that number could be "automatic" (<=0 means it takes a reasonable the LUT range).

You might want to mention in the comments that it should be max 4096

dereklukacs:
Shouldn't the max width be `model.context.getParameter(model.context.MAX_TEXTURE_SIZE)` which is hardware dependent? 

I can get 2^14 working well on my machine (M4 Pro Max) but only 2^13 working on CI.

finetjul:
> Shouldn't the max width be `model.context.getParameter(model.context.MAX_TEXTURE_SIZE)` which is hardware dependent?
> 
> I can get 2^14 working well on my machine (M4 Pro Max) but only 2^13 working on CI.

https://web3dsurvey.com/webgl/parameters/MAX_TEXTURE_SIZE

sankhesh (approved):
LGTM

floryst:
New changes:
- added docs on reasonable and max texture sizes
- Widths <= 0 will now be interpreted as "use max texture size"

finetjul (approved):
LGTM

PR #3239, opened 2025-04-14
fix(renderwindowinteractor): observe KeyPress events on RW container …

### LINKED ISSUE #1856: RWI keypress event listener
RenderWindowInteractor (RWI) [observes](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/RenderWindowInteractor/index.js#L235-L243) key events (keydown, keypress, keyup) on `body`.

For plain HTML app, it makes sense: If a key down event is not handled by any DOM element, then the key press event gets to the VTK view. That's handy.

However, React handles events [AFTER](https://drek4537l1klr.cloudfront.net/mardan/Figures/06fig04_alt.jpg) native event listeners (addEventListener).

Therefore, `onKeyDown` callback gets called after the RWI. Calling `stopPropagate()` within the callback is therefore useless and will not prevent RWI from processing the event because it is already processed.

What do you suggest ?
 1) do native event listener within React (e.g. `addEventListener()` in `componentDidMount()`)
 2) Make RWI stops listening `<body>` but `<container>` instead ?
 3) Delegate event listening to the framework (React, Vue...)

My understanding is that it is the same issue with Vue. I have not looked nor tried, but few weeks ago, I observed something similar in PVGlance. If you enable "the First Person mode" and use keydown/keyup to change the camera zoom factor, it also moves the camera forward in the view.

finetjul:
…instead of document

Observing on document produces many false positives. For example, a key press event was for an <input> element but RWI was also processing it.

BREAKING CHANGE: This may break key press event current behavior. Use tabIndex=0 on RW containers to fix it.

fix #1856

### Context
Read the description on #1856

### Results
This time, it enforces the events are observed on the RW and not on the whole document.
This is important when there are multiple RW in a document.

### Changes

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows
  - **Browser**: Chrome


finetjul:
image

jourdain:
I defer to Forrest on that one

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -237,15 +237,16 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     });
     container.addEventListener('pointerup', publicAPI.handlePointerUp);
     container.addEventListener('pointercancel', publicAPI.handlePointerCancel);
-    document.addEventListener('keypress', publicAPI.handleKeyPress);
-    document.addEventListener('keydown', publicAPI.handleKeyDown);
-    document.addEventListener('keyup', publicAPI.handleKeyUp);
+    container.addEventListener('keypress', publicAPI.handleKeyPress);
+    container.addEventListener('keydown', publicAPI.handleKeyDown);
+    container.addEventListener('keyup', publicAPI.handleKeyUp);
```
floryst, line 242: It might be better to listen for keyup on the document in the case that the focus leaves the container while a key is still down.

floryst:
Since this is a breaking change, let's merge this into the beta branch. I need to merge my "named parameters" branch into beta as well. That way, we have one major change as opposed to two.

Sources/Rendering/Core/RenderWindowInteractor/index.js
finetjul: I understand your concern, I find your suggestion a bit "fragile" and prone to further errors. 
Can we wait to have a concrete example when there such case happen to see how to best fix it ?

finetjul:
> Since this is a breaking change, let's merge this into the beta branch. I need to merge my "named parameters" branch into beta as well. That way, we have one major change as opposed to two.

I'm fine with merging it into the beta branch for now, but can the final merge into master happen soon-ish (e.g. in no more than 1 week)?

Sources/Rendering/Core/RenderWindowInteractor/index.js
floryst: I'm not sure what issues would arise, unless there is a "keyup" operation that isn't cleaning up from a "keydown" event, or the "keyup" listener is not a no-op for viewports that are not focused. I just see the one case where someone expects start/end semantics with keydown/keyup, as there is no concept of "keyleave".

Realistically, I don't think it affects our codebases at the moment. I've always preferred handling keypresses at the application level rather than at the vtk.js level. I'm fine keeping it as-is if there are tangible issues with listening for keyup on the document.

floryst:
> > Since this is a breaking change, let's merge this into the beta branch. I need to merge my "named parameters" branch into beta as well. That way, we have one major change as opposed to two.
> 
> I'm fine with merging it into the beta branch for now, but can the final merge into master happen soon-ish (e.g. in no more than 1 week)?

Sure. I think we're good to go with the existing changes in beta (the volume mapper changes). I'll check to see if it merges cleanly.

PR #3241, opened 2025-04-22
feat(RenderWindowInteractor): commit mouse movement data in pointerlock

Jo-Byr:
Fix pending TODO in MouseCameraTrackballFirstPersonManipulator to move pointerLock mouse movement logic to vtkRenderWindowInteractor

Modify RenderWindowInteractor to return mouse movement data in pointerLock mode
Freeze InteractorStyleManipulator manipulator during pointer
Lock Update MouseCameraTrackballFirstPersonManipulator

### DIFF
--- a/Sources/Interaction/Manipulators/MouseCameraTrackballFirstPersonManipulator/index.js
+++ b/Sources/Interaction/Manipulators/MouseCameraTrackballFirstPersonManipulator/index.js
@@ -15,73 +15,30 @@ function vtkMouseCameraTrackballFirstPersonManipulator(publicAPI, model) {
   const internal = {
     interactor: null,
     renderer: null,
-    previousPosition: null,
   };
 
   //--------------------------------------------------------------------------
 
   publicAPI.onButtonDown = (interactor, renderer, position) => {
-    internal.previousPosition = position;
-
     if (model.usePointerLock && !interactor.isPointerLocked()) {
       Object.assign(internal, { interactor, renderer });
       interactor.requestPointerLock();
-      publicAPI.startPointerLockInteraction();
     }
-  };
-
-  //--------------------------------------------------------------------------
-
-  publicAPI.startPointerLockInteraction = () => {
-    const { interactor } = internal;
-
-    // TODO: at some point, this should perhaps be done in
-    // RenderWindowInteractor instead of here.
-    // We need to hook into mousemove directly for two reasons:
-    // 1. We need to keep receiving mouse move events after the mouse button
-    //    is released. This is currently not possible with
-    //    vtkInteractorStyleManipulator.
-    // 2. Since the mouse is stationary in pointer lock mode, we need the
-    //    event.movementX and event.movementY info, which are not currently
-    //    passed via interactor.onMouseMove.
-    document.addEventListener('mousemove', publicAPI.onPointerLockMove);
-
-    let subscription = null;
-    const endInteraction = () => {
-      document.removeEventListener('mousemove', publicAPI.onPointerLockMove);
-      subscription.unsubscribe();
-    };
-    subscription = interactor.onEndPointerLock(endInteraction);
-  };
-
-  //--------------------------------------------------------------------------
-
-  publicAPI.onPointerLockMove = (e) => {
-    const sensitivity = model.sensitivity;
-    const yaw = -1 * e.movementX * sensitivity;
-    const pitch = -1 * e.movementY * sensitivity;
-
-    publicAPI.moveCamera(yaw, pitch);
   };
 
   //--------------------------------------------------------------------------
 
   publicAPI.onMouseMove = (interactor, renderer, position) => {
-    // This is currently only being called for non pointer lock mode
     if (!position) {
       return;
     }
 
-    const { previousPosition } = internal;
-
     const sensitivity = model.sensitivity;
-    const yaw = (previousPosition.x - position.x) * sensitivity;
-    const pitch = (position.y - previousPosition.y) * sensitivity;
+    const yaw = -position.movementX * sensitivity;
+    const pitch = -position.movementY * sensitivity;
 
     Object.assign(internal, { interactor, renderer });
     publicAPI.moveCamera(yaw, pitch);
-
-    internal.previousPosition = position;
   };
 
   //--------------------------------------------------------------------------
--- a/Sources/Interaction/Style/InteractorStyleManipulator/index.js
+++ b/Sources/Interaction/Style/InteractorStyleManipulator/index.js
@@ -1,4 +1,5 @@
 import macro from 'vtk.js/Sources/macros';
+import { MouseButton } from 'vtk.js/Sources/Rendering/Core/RenderWindowInteractor/Constants';
 import vtkInteractorStyle from 'vtk.js/Sources/Rendering/Core/InteractorStyle';
 
 const { vtkDebugMacro } = macro;
@@ -268,19 +269,19 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
   //-------------------------------------------------------------------------
   publicAPI.handleLeftButtonPress = (callData) => {
     model.previousPosition = callData.position;
-    publicAPI.onButtonDown(1, callData);
+    publicAPI.onButtonDown(MouseButton.LeftButton, callData);
   };
 
   //-------------------------------------------------------------------------
   publicAPI.handleMiddleButtonPress = (callData) => {
     model.previousPosition = callData.position;
-    publicAPI.onButtonDown(2, callData);
+    publicAPI.onButtonDown(MouseButton.MiddleButton, callData);
   };
 
   //-------------------------------------------------------------------------
   publicAPI.handleRightButtonPress = (callData) => {
     model.previousPosition = callData.position;
-    publicAPI.onButtonDown(3, callData);
+    publicAPI.onButtonDown(MouseButton.RightButton, callData);
   };
 
   //-------------------------------------------------------------------------
@@ -397,17 +398,17 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
 
   //-------------------------------------------------------------------------
   publicAPI.handleLeftButtonRelease = () => {
-    publicAPI.onButtonUp(1);
+    publicAPI.onButtonUp(MouseButton.LeftButton);
   };
 
   //-------------------------------------------------------------------------
   publicAPI.handleMiddleButtonRelease = () => {
-    publicAPI.onButtonUp(2);
+    publicAPI.onButtonUp(MouseButton.MiddleButton);
   };
 
   //-------------------------------------------------------------------------
   publicAPI.handleRightButtonRelease = () => {
-    publicAPI.onButtonUp(3);
+    publicAPI.onButtonUp(MouseButton.RightButton);
   };
 
   //-------------------------------------------------------------------------
@@ -421,10 +422,16 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
     ) {
       model.currentManipulator.onButtonUp(model._interactor);
       model.currentManipulator.endInteraction();
-      model.currentManipulator = null;
-      model._interactor.cancelAnimation(publicAPI.onButtonDown);
+      if (!model._interactor.isPointerLocked()) {
+        model.currentManipulator = null;
+      }
       publicAPI.invokeEndInteractionEvent(END_INTERACTION_EVENT);
     }
+  };
+
+  //-------------------------------------------------------------------------
+  publicAPI.handleEndPointerLock = () => {
+    model.currentManipulator = null;
   };
 
   //-------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -12,6 +12,10 @@ const START_INTERACTION_EVENT = { type: 'StartInteractionEvent' };
 const INTERACTION_EVENT = { type: 'InteractionEvent' };
 const END_INTERACTION_EVENT = { type: 'EndInteractionEvent' };
 
+const LEFT_MOUSE_BUTTON = 1;
+const MIDDLE_MOUSE_BUTTON = 2;
+const RIGHT_MOUSE_BUTTON = 3;
```
finetjul, line 17: to move in a Constants.js file

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -680,6 +688,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
       ...getModifierKeysFor(event),
       position: getScreenEventPositionFor(event),
       deviceType: getDeviceTypeFor(event),
+      movement: getScreenMovementFor(event),
```
finetjul, line 691: isn't there a way to merge `position` with `movement`  ?

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -427,6 +432,15 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
     }
   };
 
+  //-------------------------------------------------------------------------
+  publicAPI.ignoreButtonUp = (button) =>
```
finetjul, line 436: you need to document why you want to ignore the button up

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -427,6 +432,15 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
...
+    button === LEFT_MOUSE_BUTTON && model._interactor.isPointerLocked();
+
+  //-------------------------------------------------------------------------
+  publicAPI.handleEndPointerLock = () => {
+    model.currentManipulator = null;
```
finetjul, line 441: shouldn't there be an `endInteraction()` triggered as well as `cancelAnimation` ...?

Sources/Interaction/Manipulators/MouseCameraTrackballFirstPersonManipulator/index.js
```diff
@@ -26,57 +26,20 @@ function vtkMouseCameraTrackballFirstPersonManipulator(publicAPI, model) {
...
-    const pitch = -1 * e.movementY * sensitivity;
-
-    publicAPI.moveCamera(yaw, pitch);
-  };
-
-  //--------------------------------------------------------------------------
-
-  publicAPI.onMouseMove = (interactor, renderer, position) => {
+  publicAPI.onMouseMove = (interactor, renderer, position, movement) => {
     // This is currently only being called for non pointer lock mode
```
finetjul, line 35: is this comment still valid ?

Sources/Interaction/Manipulators/MouseCameraTrackballFirstPersonManipulator/index.js
```diff
@@ -26,57 +26,20 @@ function vtkMouseCameraTrackballFirstPersonManipulator(publicAPI, model) {
...
-    if (!position) {
+    if (!movement || !interactor.isPointerLocked()) {
       return;
     }
 
-    const { previousPosition } = internal;
-
     const sensitivity = model.sensitivity;
-    const yaw = (previousPosition.x - position.x) * sensitivity;
```
finetjul, line 78: Shouldn't you also delete previousPosition ? 

Did you try with `model.usePointerLock==false` ?

Sources/Rendering/Core/RenderWindowInteractor/index.js
finetjul: or maybe have `movement` work all the time even when not in pointersLock.
If so, please add documentation

Sources/Interaction/Manipulators/MouseCameraTrackballFirstPersonManipulator/index.js
Jo-Byr: Nothing happends with `model.usePointerLock == false`

Sources/Interaction/Style/InteractorStyleManipulator/index.js
finetjul: Yes, I think that `currentManipulator.onButtonUp()` should be called in `onButtonUp` even in pointerLock mode.

Sources/Interaction/Style/InteractorStyleManipulator/index.js
Jo-Byr: Should it be added to RenderWindowInteractor Constants file or have its own ?

finetjul: Yes, you can use that file if it already exists.

Sources/Rendering/Core/RenderWindowInteractor/index.js
Jo-Byr: What do you mean "have `movement` work all the time" ? It's always packed in the event callData, it's just that the InteractorStyleManipulator has further parsing.

finetjul: Do you mean that "movement" is always in "callData", even when NOT in pointerLock mode ?
Does it always contain the difference between the current mouse cursor position and the position when last clicked ?

Jo-Byr: In `publicAPI.handleMouseMove` of RenderWindowInteractor, the callData has a `movement` property filled by :
image

The movement X and Y values come from the event, we simply pack them together under the `movement` name.

Jo-Byr: I put the `??` purely by security, I did not encounter a case where it was.

finetjul: I suggest to move "movementX/Y" inside the "position" dictionary (in addition to "x", "y" and "z").

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -184,6 +184,8 @@ function vtkRenderWindowInteractor(publicAPI, model) {
       x: scaleX * (source.clientX - bounds.left),
       y: scaleY * (bounds.height - source.clientY + bounds.top),
       z: 0,
+      movementX: source.movementX,
+      movementY: source.movementY,
```
finetjul, line 188: maybe there is a need for "scaling" here as well ? (please try zooming (CTRL++) your chrome page)

Jo-Byr: If you multiply by scale here, the movement speed is different from the current behavior when zoomed in.

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -421,12 +422,24 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
     ) {
       model.currentManipulator.onButtonUp(model._interactor);
       model.currentManipulator.endInteraction();
-      model.currentManipulator = null;
-      model._interactor.cancelAnimation(publicAPI.onButtonDown);
+      if (
+        !(
+          button === MouseButton.LeftButton &&
+          model._interactor.isPointerLocked()
```
finetjul, line 428: It could be any Manipulator that triggered the pointerLocked mode. Let's not assume it is always on a LeftButton.

Maybe `endInteraction()` could return a boolean that tells if the manipulator accepts to be `ended`.

Sources/Rendering/Core/RenderWindowInteractor/index.js
Jo-Byr: @finetjul to be more precise, currently, zooming makes panning slower without affecting the visual, so it could be argued that scaling should be applied, but it will change the existing behavior

finetjul: it's fine to change the existing behavior if it is a "better" behavior

Sources/Interaction/Style/InteractorStyleManipulator/index.js
finetjul: Not sure to understand why you need to store a `canBeEnded` variable.

I let you propose your change and I'll review your code.

Sources/Interaction/Manipulators/MouseCameraTrackballFirstPersonManipulator/index.js
```diff
@@ -15,73 +15,31 @@ function vtkMouseCameraTrackballFirstPersonManipulator(publicAPI, model) {
...
   };
 
   //--------------------------------------------------------------------------
 
   publicAPI.onButtonDown = (interactor, renderer, position) => {
-    internal.previousPosition = position;
-
+    Object.assign(internal, { interactor, renderer });
     if (model.usePointerLock && !interactor.isPointerLocked()) {
       Object.assign(internal, { interactor, renderer });
```
finetjul, line 25: to remove ?

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -420,13 +421,19 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
       model.currentManipulator.getButton() === button
     ) {
       model.currentManipulator.onButtonUp(model._interactor);
-      model.currentManipulator.endInteraction();
-      model.currentManipulator = null;
-      model._interactor.cancelAnimation(publicAPI.onButtonDown);
+      const canEndManipulator = model.currentManipulator.endInteraction();
+      if (canEndManipulator) {
```
finetjul, line 425: let's face it, developers will forget to return something in "endInteraction()". 
Maybe `undefined` (when nothing is returned) should be considered as "ok to be ended".

Sources/Interaction/Manipulators/MouseCameraTrackballFirstPersonManipulator/index.js
```diff
@@ -130,6 +88,9 @@ function vtkMouseCameraTrackballFirstPersonManipulator(publicAPI, model) {
     interactor.requestAnimation(animationRequester);
     animationSub = interactor.onAnimation(() => performStep());
   };
+
+  //----------------------------------------------------------------------------
+  publicAPI.endInteraction = () => !internal.interactor.isPointerLocked();
```
finetjul, line 93: Please add a comment to explain what's going on "In pointer lock mode, manipulator should stay active even after a mouse button release".

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -420,13 +421,19 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
       model.currentManipulator.getButton() === button
     ) {
       model.currentManipulator.onButtonUp(model._interactor);
-      model.currentManipulator.endInteraction();
-      model.currentManipulator = null;
-      model._interactor.cancelAnimation(publicAPI.onButtonDown);
+      const canEndManipulator = model.currentManipulator.endInteraction();
```
finetjul, line 424: that must be documented somewhere the signification of the `endInteraction()` return value.

finetjul (commented):
Thinking more about it:

"endInteraction" is not bound to be called only on mouse button up.
Maybe the "canEndManipulator" could be guessed from onButtonUp instead of endInteraction
Maybe the check on internal.interactor.isPointerLocked() could be applied by default by the interactor style: if we are in a pointer lock mode, then releasing the button does not turn off the manipulator...
WDYT ?

Sources/Interaction/Manipulators/MouseCameraTrackballFirstPersonManipulator/index.js
Jo-Byr: It is necessary to have access to the interactor in `endInteraction`. If we end changing the point where this is done, it may have to be deleted.

finetjul: But you already added that line just above.

Jo-Byr:
> Thinking more about it:
> 
> "endInteraction" is not bound to be called only on mouse button up. Maybe the "canEndManipulator" could be guessed from onButtonUp instead of endInteraction Maybe the check on internal.interactor.isPointerLocked() could be applied by default by the interactor style: if we are in a pointer lock mode, then releasing the button does not turn off the manipulator... WDYT ?

So we would check if button === Left && isPointerLocked and if it's the case we don't end the manipulator ?

I suppose it could work, if there is no instance where we get into pointerLock from another way than LeftMouseButton and have different left mouse button interaction.

finetjul (approved):
LGTM

can you squash your commits in a unique commit ?

PR #3248, opened 2025-04-29
Fix default exports

GalMunGral:
### Context

Trying to port the examples to TypeScript produces the following errors:
Screenshot 2025-04-29 at 2 50 15 PM

It turns out that default exports are misused in many places in the `.d.ts` files:
1. Since the original `.js` files don’t have types, default exports should always be values. But in some cases, an `interface` declaration is incorrectly assigned as the default export, rather than a value that conforms to that interface. This leads to the error shown in the screenshot.
2. Most files correctly default-export a value and separately export the interface type as a named export. However, other `.d.ts` files mistakenly use the default import syntax to import interfaces — which resolves to the exported const values instead of the type definitions.

### Results

Screenshot 2025-04-29 at 3 20 38 PM
The type errors related to default exports are resolved.

### Changes

- Changed import/export statements in in multiple `.d.ts` files.
- Added a `typecheck` npm script and a `tsconfig.json` to ensure the `.d.ts` files are consistent with each other.
- [x] Documentation and TypeScript definitions were updated to match those changes (obviously)

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

No implementation has changed; this should only affect TypeScript-aware IDEs and build tools.

### DIFF
--- a/Utilities/build/macro-shim.d.ts
+++ b/Utilities/build/macro-shim.d.ts
@@ -1,4 +1,4 @@
-import Macro from '../../Sources/macros';
+import Macro from './macros';
 
-export * from '../../Sources/macros';
+export * from './macros';
 export default Macro;
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,6 @@
 {
   "include": ["Sources/**/*.d.ts", "Utilities/**/*.d.ts"],
+  "exclude": ["Utilities/build/macro-shim.d.ts"],
   "compilerOptions": {
     "noEmit": true
   }


daker:
i don't know what's your use case here but they are imported like this:

```javascript
import DataAccessHelper from '@kitware/vtk.js/IO/Core/DataAccessHelper';
// import '@kitware/vtk.js/IO/Core/DataAccessHelper/LiteHttpDataAccessHelper'; // Just need HTTP
import '@kitware/vtk.js/IO/Core/DataAccessHelper/HttpDataAccessHelper'; // HTTP + zip
// import '@kitware/vtk.js/IO/Core/DataAccessHelper/HtmlDataAccessHelper'; // html + base64 + zip
// import '@kitware/vtk.js/IO/Core/DataAccessHelper/JSZipDataAccessHelper'; // zip
```

floryst (commented):
Thanks for adding a typecheck command!

Instead of exporting a declared `const _default` and changing imports to use curly braces, you can declare a value with the same name as the interface (see the following example with HtmlDataAccessHelper), and typescript will infer the type or value:

```
// HtmlDataAccessHelper/index.d.ts
declare const HtmlDataAccessHelper: HtmlDataAccessHelper
export default HtmlDataAccessHelper

// other.d.ts
import HtmlDataAccessHelper from '...'
// HtmlDataAccessHelper can be used as a type or a value

Utilities/build/macro-shim.d.ts
floryst, line 1: This should be left alone, since it gets copied into the dist directory as-is.

floryst:
This is good to go once `macro-shim.d.ts` errors are addressed!

floryst (approved):

PR #3250, opened 2025-05-02
fix(mapper.js): fix color texture coordinate mapping when using logarithmic scale

Antti-Palola:
Scalars array was already mapped with log scaling but texture coordinates were incorrectly mapped using the linear range and value. Log color mapping was working if scalar interpolation was off when scalars were not directly mapped to textures.

Also added some missing types to ColorTransferFunction.

### Results

In the following screenshots the first rectangle has point values of 1 everywhere, second has 100 and the last 1000.

| Linear | Log before | Log after |
| -- | -- | --|
|  Screenshot 2025-05-02 at 11 02 06 |  Screenshot 2025-05-02 at 11 07 54 |  Screenshot 2025-05-02 at 11 02 24 |

### Changes

### PR and Code Checklist

### Testing

- [X] Tested environment:
  - **vtk.js**: master (33.0.1)
  - **OS**: MacOS 15.4.
  - **Browser**: Latest Chrome, Firefox and Safari

The scalar bar actor example doesn't really work at all with interpolated values (because of the geometry?) but the rim of the cone can be seen barely to see the change.

| Before | After | 
| -- | -- |
| Screenshot 2025-05-02 at 12 01 07 | Screenshot 2025-05-02 at 12 00 38 |

### DIFF
--- a/Sources/Rendering/Core/ColorTransferFunction/index.d.ts
+++ b/Sources/Rendering/Core/ColorTransferFunction/index.d.ts
@@ -312,6 +312,30 @@ export interface vtkColorTransferFunction extends vtkScalarsToColors {
    * @param colorMap
    */
   applyColorMap(colorMap: any): void;
+
+  /**
+   * Get the color space of the color transfer function.
+   * @returns {ColorSpace}
+   */
+  getColorSpace(): ColorSpace;
+
+  /**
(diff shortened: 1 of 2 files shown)


Antti-Palola:
@floryst Do you have time to check this out?

floryst (approved):

PR #3266, opened 2025-06-11
feat(Texture): add support for ImageBitmap texture

daker:
### Context

feat(Texture): add support for ImageBitmap texture
perf(WebGPU): improve texture upload using ImageBitmap
refactor(WebGPU): simplify vtkWebGPUCellArrayMapper

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/Texture/index.d.ts
+++ b/Sources/Rendering/Core/Texture/index.d.ts
@@ -15,7 +15,6 @@ export interface ITextureInitialValues {
 }
 
 export interface vtkTexture extends vtkAlgorithm {
-
   /**
    * Returns the canvas used by the texture.
    */
@@ -39,12 +38,12 @@ export interface vtkTexture extends vtkAlgorithm {
   /**
    * Returns the image used by the texture.
    */
-  getImage(): HTMLImageElement;
+  getImage(): Nullable<HTMLImageElement>;
 
   /**
    * Returns an ImageBitmap object.
    */
-  getImageBitmap(): ImageBitmap;
+  getImageBitmap(): Nullable<ImageBitmap>;
 
   /**
    * Returns true if the image is loaded.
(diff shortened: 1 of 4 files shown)


daker:
@sankhesh FYI

sankhesh:
Hey @daker. Just to clarify, would an alternative approach be to read the image bitmap as `vtkImageData` and use that as the texture input? Or am I overlooking something here?

daker:
Here we are going to pass the ImageBitmap(created using createImageBitmap)/HTMLCanvasElement directly to GPU using queue.copyExternalImageToTexture, no need to create a GPUBuffer then use cmdEnc.copyBufferToTexture to upload that buffer.

here is an example of the usage:

```javascript
export async function loadImage(image) {
    const blob = new Blob([image.bufferView.data], { type: image.mimeType });
    const bitmap = await createImageBitmap(blob, {
      colorSpaceConversion: 'none',
    });
    return bitmap;
}

export function createVTKTextureFromGLTFTexture(image) {
  const texture = vtkTexture.newInstance();
  texture.setImageBitmap(image);
  return texture;
}

const image = await loadImage(tex.source);
const diffuseTex = createVTKTextureFromGLTFTexture(
   image
 );
property.setDiffuseTexture(diffuseTex);
```

sankhesh:
Makes sense. Skipping the creation of GPUBuffer and then allocating the buffer would incur more overhead than `GPUQueue::copyExternalImageToTexture`. 

Mind adding a test (and an example)?

daker:
I already have a working example for GLTFImporter, i am just breaking the changes in multiple PRs to simplify the review, and for the test i don't see any tests foundations for WebGPU in the project.

sankhesh:
To ensure that a test just runs on the webgpu backend, see https://github.com/Kitware/vtk-js/blob/071ee0a1e37c6fed34b4b653f4f3d17f92a804ca/Sources/Testing/testIf.js#L8-L11

Though, I think you mean that the CI doesn't test webgpu right now which is fair.

sankhesh (approved):

PR #3272, opened 2025-06-30
feat(InteractorStyleManipulator): update center of rotation on pan

PaulHax:
This enhances pan then rotate interaction with InteractorStyleManipulator.
After panning, move the center of rotation the same vector the camera position moved.

Before, if the user panned the camera, zoomed in, then rotated, they would rotate around a point that was not in the center of the camera.  This made it hard to zoom in on a region of interest and rotate around it.

### Context

vtkInteractorStyleManipulator is used widely, especially in [VolView](https://github.com/Kitware/VolView.git) and in trame-vtk via  [vue-vtk-js](https://github.com/Kitware/vue-vtk-js.git). 

I'm optimistic this change will transparently upgrade their default camera controls for the better.

### Results

To get a feel for the rotation before and after run:
```
npm run example -- InteractorStyleManipulator
```
Try to pan and zoom in on the point of the cone, then rotate around the cone's point.

Every app seems to want their own quirky camera control setup, but this is a transparent upgrade for most?

### Changes

No doc changes methinks?

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Interaction/Style/InteractorStyleManipulator/index.js
+++ b/Sources/Interaction/Style/InteractorStyleManipulator/index.js
@@ -159,7 +159,7 @@ function transformVectorByTransformation(
   afterMatrix,
   vector
 ) {
-  const { matrixA, matrixB, identityMatrix, newCenter } = tempObjects;
+  const { matrixA, matrixB, newCenter } = tempObjects;
 
   // The view matrix from vtk.js is row-major, but gl-matrix expects column-major.
   // We need to transpose them before use.
@@ -171,12 +171,8 @@ function transformVectorByTransformation(
   // Compute delta transformation matrix
   mat4.multiply(matrixA, matrixB, matrixA);
 
-  // Apply transformation if matrix changed
-  if (!mat4.equals(matrixA, identityMatrix)) {
-    vec3.transformMat4(newCenter, vector, matrixA);
-    return newCenter;
-  }
-  return vector;
+  vec3.transformMat4(newCenter, vector, matrixA);
+  return newCenter;
 }
 
 /**
@@ -233,7 +229,6 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
   const tempComputeObjects = {
     matrixA: mat4.create(),
     matrixB: mat4.create(),
-    identityMatrix: mat4.identity(mat4.create()),
     newCenter: vec3.create(),
   };
 
--- a/Sources/Interaction/Style/InteractorStyleManipulator/index.js
+++ b/Sources/Interaction/Style/InteractorStyleManipulator/index.js
@@ -1,6 +1,7 @@
 import macro from 'vtk.js/Sources/macros';
 import { MouseButton } from 'vtk.js/Sources/Rendering/Core/RenderWindowInteractor/Constants';
 import vtkInteractorStyle from 'vtk.js/Sources/Rendering/Core/InteractorStyle';
+import { mat4, vec3, vec4 } from 'gl-matrix';
 
 const { vtkDebugMacro } = macro;
 const { States } = vtkInteractorStyle;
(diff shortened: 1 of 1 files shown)


Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -504,11 +555,26 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
...
       );
+
+      const newCenter = computeNewCenterOfRotation(
+        beforeM,
+        renderer,
+        model.centerOfRotation,
+        tempComputeObjects
+      );
+
+      if (!vec3.equals(model.centerOfRotation, newCenter)) {
```
finetjul: why bother doing a comparison check ? it will be done by the setter automatically, no ?

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -134,6 +135,47 @@ function dollyByFactor(interactor, renderer, factor) {
...
+function getCameraMatrix(renderer, tempMatrix) {
+  const cam = renderer.getActiveCamera();
+  if (cam) {
+    mat4.copy(tempMatrix, cam.getViewMatrix());
+    return tempMatrix;
+  }
+  return null;
+}
+
+function computeNewCenterOfRotation(
```
finetjul, line 189: please document the behavior of this function

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -686,6 +755,16 @@ function vtkInteractorStyleManipulator(publicAPI, model) {
       }
```
finetjul, line 777: use `renderer` instead of `model.getRenderer(callData)`

Sources/Interaction/Style/InteractorStyleManipulator/index.js
```diff
@@ -134,6 +135,47 @@ function dollyByFactor(interactor, renderer, factor) {
...
+  const cam = renderer.getActiveCamera();
+  if (cam) {
+    mat4.copy(tempMatrix, cam.getViewMatrix());
+    return tempMatrix;
+  }
+  return null;
+}
+
+function computeNewCenterOfRotation(
+  beforeM,
```
finetjul: the name of the variable is not self-explanatory

Sources/Interaction/Style/InteractorStyleManipulator/index.js
PaulHax: Yes, setArray macro should take care of it.  fixed.

Sources/Interaction/Style/InteractorStyleManipulator/index.js
PaulHax: added this
/**
 * Computes the new center of rotation based on camera movement.
 * When the camera moves (pan), the center of rotation should move
 * by the same transformation to maintain consistent rotation behavior.
 *
 * @param {Object} tempObjects - Temporary matrices/vectors for computation
 * @param {Object} renderer - VTK renderer
 * @param {mat4} beforeCameraMatrix - Camera view matrix before movement
 * @param {Array} oldCenterOfRotation - Previous center of rotation [x, y, z]
 * @returns {Array} New center of rotation [x, y, z]
 */

Sources/Interaction/Style/InteractorStyleManipulator/index.js
PaulHax: yes.

Sources/Interaction/Style/InteractorStyleManipulator/index.js
PaulHax: changed to beforeCameraMatrix.

Thanks for the quick review

PR #3290, opened 2025-07-21
fix(VolumeMapper): Fixed accessing property with no valid input

yarous224:
### Context

In the calculation of the number of lights, there is no check to determine whether `model.currentValidInputs` is empty. However, later on, `firstVolumeProperty` is used for judgment.
### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


finetjul (approved):
LGTM

PR #3297, opened 2025-07-25
fix(Texture): Prevent GL_INVALID_OPERATION when updating texture data

PaulHax:
### TODO
- [ ] I think we need account for datatype changes, among other tings =/

## Prevent GL_INVALID_OPERATION when updating texture data in WebGL2

I would think this warning has been disccused before but could not find an issue regarding this...

### Context

WebGL2 textures created with `texStorage2D` are immutable, causing browser console warnings:

```
GL_INVALID_OPERATION: glTexStorage2D: Texture is immutable.
```

This occurs when `texture.modified()` triggers recreation of existing textures, such as when updating video textures or other dynamic content. The warning appears because vtk.js attempts to call `texStorage2D` again on already-immutable textures.

Console log showing the error pattern:

- Initial texture creation: `texStorage2D` + `texSubImage2D` (works)
- Texture update: `texture.modified()` → attempts `texStorage2D` again (fails)

Want this to update a texture with a video stream.

### Results

**Before:** Browser console shows GL_INVALID_OPERATION warnings on every texture update

**After:** Clean texture updates without any WebGL errors

The texture data still updates correctly in both cases, but the warnings are eliminated.

### Changes

- **Added:** `update2DFromRaw()` method that uses `texSubImage2D` for data-only updates
- **Modified:** `render()` method to detect when only texture data changes (same dimensions/components)

API additions:

- `publicAPI.update2DFromRaw({ data, dataType, flip })` - Updates existing texture data without recreation

### PR and Code Checklist

- [x] semantic-release commit messages

### Testing

- [x] Tested environment:
  - **vtk.js**: latest master
  - **OS**: Linux
  - **Browser**: Chrome WebGL2

**Test coverage:**

- Unit test verifies no GL_INVALID_OPERATION errors during texture updates

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -143,7 +143,10 @@ function vtkOpenGLTexture(publicAPI, model) {
       const input = model.renderable.getInputData(0);
       if (input && input.getPointData().getScalars()) {
         const ext = input.getExtent();
+        const newWidth = ext[1] - ext[0] + 1;
+        const newHeight = ext[3] - ext[2] + 1;
         const inScalars = input.getPointData().getScalars();
+        const newComponents = inScalars.getNumberOfComponents();
 
         // do we have a cube map? Six inputs
         const data = [];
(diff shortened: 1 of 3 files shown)


floryst:
@sankhesh just an FYI (low priority)

Sources/Rendering/OpenGL/Texture/index.js
```diff
@@ -1091,6 +1119,52 @@ function vtkOpenGLTexture(publicAPI, model) {
...
+      model.target,
+      0, // level
+      0, // xoffset
+      0, // yoffset
+      model.width,
+      model.height,
+      model.format,
+      model.openGLDataType,
+      scaledData[0]
+    );
```
PaulHax, line 1154: Will this work if the openGLDataType changes?

PR #3305, opened 2025-08-06
fix(ESM): add type module to package.json

### LINKED ISSUE #3080: [Bug] ESM doesn't work without "type": "module" in node
### Bug description

Running vtk.js on the backend doesn't work without modifying the package.json.

(Maybe this is a feature request?)

### Steps to reproduce

Try installing @kitware/vtk.js on node when using ESM. The package is incorrectly loaded, and it will run properly when "type": "module" is added.

### Detailed Behavior

_No response_

### Expected Behavior

It works OOTB in node. Is there any other issue that would be caused by adding "type": "module" to the package.json?

### Environment

- vtk.js version: master
- node.js 18+ (but probably before also).

daker:
fixes #3080

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -251,6 +251,7 @@ export default {
           transform(content) {
             const pkg = JSON.parse(content);
             pkg.name = '@kitware/vtk.js';
+            pkg.type = 'module';
             pkg.main = './index.js';
             pkg.module = './index.js';
             pkg.types = './index.d.ts';


daker:
@floryst FYI

floryst (approved):
Good catch!

PR #3308, opened 2025-08-11
feat(PlatonicSolidSource): add vtkPlatonicSolidSource

daker:
### Results

| Example 1    | Example 2|
| -------- | ------- |
| image  | image    |

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


Sources/Filters/Sources/PlatonicSolidSource/index.d.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ *
+ */
+export interface IPlatonicSolidSourceInitialValues {
+  solidType?: SolidType;
+  outputPointsPrecision?: DesiredOutputPrecision;
+}
+
+type vtkPlatonicSolidSourceBase = vtkObject &
```
finetjul, line 14: why do you have to do something custom like that ? can't you simply derive from vtkObject and vtkAlgorithm ?

Sources/Filters/Sources/PlatonicSolidSource/index.js
```diff
@@ -0,0 +1,197 @@
...
+    numPoints: 8,
+    cellSize: 4,
+    numCells: 6,
+    scale: 1.0 / Math.sqrt(3.0),
+  },
+
+  octahedron: {
+    points: [
+      -1.0, -1.0, 0.0, 1.0, -1.0, 0.0, 1.0, 1.0, 0.0, -1.0, 1.0, 0.0, 0.0, 0.0,
+      -1.4142135623731, 0.0, 0.0, 1.4142135623731,
```
finetjul: use sqrt(2) instead of 1.4142135623731

Sources/Filters/Sources/PlatonicSolidSource/index.js
```diff
@@ -0,0 +1,197 @@
...
+  };
+}
+
+// ----------------------------------------------------------------------------
+// Object factory
+// ----------------------------------------------------------------------------
+
+const DEFAULT_VALUES = {
+  solidType: SolidType.VTK_SOLID_TETRAHEDRON,
+  outputPointsPrecision: DesiredOutputPrecision.DEFAULT,
```
finetjul, line 248: you might want to add a global "scale" parameter

Sources/Filters/Sources/PlatonicSolidSource/index.js
```diff
@@ -0,0 +1,197 @@
...
+    } else if (model.outputPointsPrecision === DesiredOutputPrecision.DOUBLE) {
+      pointType = VtkDataTypes.DOUBLE;
+    }
+
+    const points = vtkPoints.newInstance({
+      dataType: pointType,
+      numberOfPoints: solidData.numPoints,
+    });
+
+    const colors = vtkDataArray.newInstance({
```
finetjul: this color array is useless.

When that code was written originally, there was no way to easily get the IDs of each cell. This a changed since then (e.g. `vtkIdFilter`)

Sources/Filters/Sources/PlatonicSolidSource/index.d.ts
daker: We’ll need to omit some methods because of `macro.algo(publicAPI, model, **0**, 1)`. That said, speaking more generally, I’m not satisfied with this approach. I’ll need to think about a more generic Type based solution in a separate PR.

finetjul: ok to deal with that in a separate PR

Sources/Filters/Sources/PlatonicSolidSource/index.js
daker: You mean replace the scale per solid with a global one ?

finetjul: the end user might want to have a cube of size 10, how to easily do that ?

finetjul (approved):
LGTM

PR #3310, opened 2025-08-12
Revert center of rotation change after pan

### LINKED ISSUE #3309: [Breaking change] Center of rotation changes after pan (v34.7.0)
### Bug description

Hello,

After upgrading from v34.6.0 to v34.7.0, my application’s behavior changed due to an intentional modification introduced in this commit:
[5523179c85d2576520ffeb1756bc3b30f6cf5fd6](https://github.com/Kitware/vtk-js/commit/5523179c85d2576520ffeb1756bc3b30f6cf5fd6)

> feat(InteractorStyleManipulator): update center of rotation on pan
> This enhances pan-then-rotate interaction.
> After panning, move the center of rotation by the same vector that the camera position moved.
> Before, if the user panned the camera, zoomed in, then rotated, they would rotate around a point
> that was not in the center of the camera.

**My use case** :

- I display a group of actors.
- The interactor’s center of rotation is set to their center of mass (since vtkCenterOfMass does not exist in vtk.js, I calculate the center of mass manually).
- I want the center of rotation to remain fixed at this center of mass even after panning.

**Problem** :

With the new behavior in v34.7.0, the center of rotation moves after a pan, which breaks this use case.

**Questions / Suggestions** :

- Is there a way to opt-out of this new center-of-rotation update on pan?
- If not, could an option be added to preserve the old behavior?
- Since this change alters the default interaction model, it might be worth marking it as a breaking change in the changelog so users are aware.

Thanks in advance for your help !

### Steps to reproduce

This is an example project https://codesandbox.io/p/devbox/hello-world-cone-forked-grwhjg with my behaviour
For simplification i use the center of the cone instead of the center of mass but the idea is the same

### Detailed Behavior

On start, the cone rotates around its center
After a pan, it rotates around a point on the center of the viewer

### Expected Behavior

On start, the cone rotates around its center
After a pan, it rotates around its center

### Environment

- vtk.js version: 34.7.0
- Browsers: firefox
- OS: Mac os

PaulHax:
Rolls back InteractorStyleManipulator center of rotation behaior to not update on pan.

Adds `vtkMouseCameraTrackballPanManipulatorAutoCenter` for the auto update on center of rotation for mouse only input.

## Test with 
`npm run example InteractorStyleManipulator`

Compare to 
`npm run example InteractorStyleManipulatorRotateCameraCenter`

closes #3309

### DIFF
--- a/Sources/Interaction/Manipulators/MouseCameraTrackballPanManipulatorAutoCenter/index.js
+++ b/Sources/Interaction/Manipulators/MouseCameraTrackballPanManipulatorAutoCenter/index.js
@@ -1,0 +1,144 @@
+import macro from 'vtk.js/Sources/macros';
+import vtkMouseCameraTrackballPanManipulator from 'vtk.js/Sources/Interaction/Manipulators/MouseCameraTrackballPanManipulator';
+import { mat4, vec3 } from 'gl-matrix';
+
+// ----------------------------------------------------------------------------
+// Helper functions for center of rotation adjustment
+// ----------------------------------------------------------------------------
+
+/**
+ * Transforms a vector by the transformation delta between two matrices.
+ *
+ * @param {Object} tempObjects - Temporary matrices/vectors for computation
+ * @param {mat4} beforeMatrix - Matrix before transformation
+ * @param {mat4} afterMatrix - Matrix after transformation
+ * @param {Array} vector - Vector to transform [x, y, z]
+ * @returns {Array} Transformed vector [x, y, z]
+ */
+function transformVectorByTransformation(
+  tempObjects,
+  beforeMatrix,
+  afterMatrix,
+  vector
+) {
+  const { matrixA, matrixB, newCenter } = tempObjects;
+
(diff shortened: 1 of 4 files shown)


Sources/Interaction/Style/InteractorStyleManipulatorRotateWorld/example/index.js
```diff
@@ -0,0 +1,192 @@
...
+// Zoom with middle button
+const middleZoomManipulator =
+  vtkMouseCameraTrackballZoomManipulator.newInstance();
+middleZoomManipulator.setButton(2); // Middle button
+interactorStyle.addMouseManipulator(middleZoomManipulator);
+
+// Also allow pan with shift + left button
+const shiftPanManipulator = vtkMouseCameraTrackballPanManipulator.newInstance();
+shiftPanManipulator.setButton(1);
+shiftPanManipulator.setShift(true);
```
jourdain: This should have an option to enable or disable center of rotation modification. The default should remain the way it was before.

finetjul (approved):
LGTM

PR #3312, opened 2025-08-12
fix(VectorText): fix perLetterFaceColors

### LINKED ISSUE #3311: [Bug] vtkVectorText color not properly set by perLetterFaceColors function
### Bug description

When providing a `perLetterFaceColors` function to `vtkVectorText`, the letter colors are not being set as expected. From my reading of the docstring, and the example [here](https://kitware.github.io/vtk-js/examples/VectorText.html), `perLetterFaceColors` should return an RGB value for each letter, but only the R component has any effect and it seems to do some internal colormapping with the value of the R channel. 

### Steps to reproduce

Here's a sandbox: https://codesandbox.io/p/sandbox/vtkjs-vector-text-zsyq4p 

The sandbox code includes a number of functions you can switch between. For example, with the following:

```javascript
function fixedColor(letterIndex) {
  const clr = [1.0, 0, 0];
  console.log("fixedColor", letterIndex, clr);
  return clr;
}
```

I'd expect all colors to be red, but they are blue:

Image

Or, using the following, I'd expect expect colors to vary from black to red:

```javascript
function varyROnly(letterIndex) {
  const frac = letterIndex / 5;
  const clr = [frac, 0, 0];
  console.log("varyROnly", letterIndex, clr);
  return clr;
}
```

but the colors seem to be mapping to a colormap (a rainbow map of some kind?):

Image

finally, varying only the G and B channels have no effect. e.g.,:

```javascript
function varyGBOnly(letterIndex) {
  const frac = letterIndex / 5;
  const clr = [0, frac, frac];
  console.log("varyGBOnly", letterIndex, clr);
  return clr;
}
```
results in only red letters (the start of whatever colormap is being used?).

Image

### Detailed Behavior

_No response_

### Expected Behavior

I'd expect the RGB value returned by the perLetterFaceColors function to be used to directly set the color. e.g., the docstring for `setPerLetterFaceColors` reads:

```
(method) vtkVectorText.setPerLetterFaceColors(fn: (letterIndex: number) => RGBColor): boolean
Sets the per-letter face color function.
@param fn — Function mapping letter index to [r,g,b] color.
```

If the observed behavior is the desired behavior, it'd be helpful to know how to set the color properly (adjusting colormaps? or?). 

### Environment

- vtk.js version: 34.7.0

daker:
fixes #3311

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/VectorText/Utils.js
+++ b/Sources/Rendering/Core/VectorText/Utils.js
@@ -459,7 +459,8 @@ function addTriangle(
     uvArray.push(uv[0], uv[1]);
   });
   if (colorArray && color) {
-    for (let i = 0; i < 3; ++i) colorArray.push(color[0], color[1], color[2]);
+    for (let i = 0; i < 3; ++i)
+      colorArray.push(color[0] * 255, color[1] * 255, color[2] * 255);
   }
 }
 
@@ -512,7 +513,8 @@ function addQuad(
   uvArray.push(uvs[3][0], uvs[3][1]);
 
   if (colorArray && color) {
-    for (let i = 0; i < 6; ++i) colorArray.push(color[0], color[1], color[2]);
+    for (let i = 0; i < 6; ++i)
+      colorArray.push(color[0] * 255, color[1] * 255, color[2] * 255);
   }
 }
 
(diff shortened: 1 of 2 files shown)


sankhesh (approved):
LGTM

PR #3313, opened 2025-08-14
WIP: drop parallel-webpack in favor of using webpack directly

floryst:
Goal: speed up examples building

### DIFF
--- a/Documentation/config.js
+++ b/Documentation/config.js
@@ -1,5 +1,3 @@
-const path = require('path');
-
 /* eslint-disable no-template-curly-in-string */
 /* eslint-disable no-useless-escape */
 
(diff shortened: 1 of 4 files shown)


daker:
it's related to this commit: https://github.com/kitware/vtk-js/commit/3953b552a2f3d13dc6c1f68ca18fd6648d3018f4#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R125-R128

sankhesh:
@daker, do you mean that this fixes the examples build with lil-gui?

sankhesh:
@jourdain @finetjul Do you see any issues dropping parallel-webpack?

I'm guessing this is just for examples.

jourdain:
Ideally we would move to vite. But the parallel-webpack was to prevent timeout on ci. If that is not an issue, we can proceed.

daker:
@finetjul i will try

PR #3319, opened 2025-08-21
3303 reuse previous output

finetjul:
### Context
Make vtkPolyData and requestData() behavior consistent across VTK.js and VTK C++.

### Results
Reuse previous output data whenever possible.

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 11
  - **Browser**: Chrome

### DIFF
--- a/Sources/Common/DataModel/ImageData/index.js
+++ b/Sources/Common/DataModel/ImageData/index.js
@@ -393,6 +393,18 @@ function vtkImageData(publicAPI, model) {
       .getScalars()
       .getComponent(offsetIndex, comp);
   };
+
+  const superInitialize = publicAPI.initialize;
+  publicAPI.initialize = () => {
+    publicAPI.set({
+      direction: mat3.identity(model.direction),
+      spacing: [1.0, 1.0, 1.0],
+      origin: [0.0, 0.0, 0.0],
+      extent: [0, -1, 0, -1, 0, -1],
+      dataDescription: StructuredType.EMPTY,
+    });
+    return superInitialize();
+  };
 }
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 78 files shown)


jourdain:
There is one issue I see with that. Since you are reusing your previous output as-is, you might have some piece of mesh that are left from previous execution (cells, arrays). You may need to call a `.clear()` or `.reset()`.

If you do it right you could implement it as

```js
const output = outData[0]?.reset() || vtkPolyData.newInstance();
```

jourdain:
Do we have such method on the dataset? Are you going to do the change?

Sources/Common/DataModel/ImageData/index.js
```diff
@@ -393,6 +393,17 @@ function vtkImageData(publicAPI, model) {
       .getScalars()
       .getComponent(offsetIndex, comp);
   };
+
+  const superInitialize = publicAPI.initialize;
+  publicAPI.initialize = () => {
+    publicAPI.setDimensions(0, 0, 0);
```
jourdain: can we simply set the extent like the method is doing instead of creating this indirection.

jourdain (approved):
LGTM pending CI

Sources/Common/DataModel/ImageData/index.js
finetjul: @jourdain do you suggest the following ?

```
publicAPI.set({
  extent: [0, -1, 0, -1, 0, -1],
  dataDescription: StructuredType.EMPTY,
  ...
  });
```

jourdain: yes

floryst (approved):

Sources/Common/DataModel/ImageData/index.js
finetjul: Fair enough.
I find it dangerous if tomorrow `setDimensions()` has additional side-effects...

jourdain: feel free to leave it as-is. I thought there were no reason to just set the object to its initial state instead of calling a function.

PR #3321, opened 2025-08-29
refactor(bounds): fix getBounds to return a copy of model.bounds

finetjul:
A getter should return a copy and not a reference. getBoundsByReference can return the actual model.bounds array.

BREAKING CHANGE: Mappers should overwrite computeBounds instead of getBounds

### DIFF
--- a/Sources/Rendering/Core/ImageResliceMapper/index.js
+++ b/Sources/Rendering/Core/ImageResliceMapper/index.js
@@ -16,22 +16,25 @@ function vtkImageResliceMapper(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkImageResliceMapper');
 
-  publicAPI.getBounds = () => {
-    let bds = [...vtkBoundingBox.INIT_BOUNDS];
+  publicAPI.computeBounds = () => {
     const image = publicAPI.getInputData();
     if (publicAPI.getSlicePolyData()) {
-      bds = publicAPI.getSlicePolyData().getBounds();
+      vtkBoundingBox.setBounds(
+        model.bounds,
+        publicAPI.getSlicePolyData().getBounds()
+      );
     } else if (image) {
-      bds = image.getBounds();
+      vtkBoundingBox.setBounds(model.bounds, image.getBounds());
       if (publicAPI.getSlicePlane()) {
         vtkBoundingBox.cutWithPlane(
-          bds,
+          model.bounds,
           publicAPI.getSlicePlane().getOrigin(),
           publicAPI.getSlicePlane().getNormal()
         );
       }
+    } else {
+      vtkBoundingBox.reset(model.bounds);
     }
-    return bds;
   };
 }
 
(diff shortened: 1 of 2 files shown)


Sources/Common/DataModel/BoundingBox/index.js
```diff
@@ -49,14 +49,23 @@ export function reset(bounds) {
...
-export function addPoint(bounds, x, y, z) {
+export function addPoint(bounds, xOrPoint, y, z) {
   const [xMin, xMax, yMin, yMax, zMin, zMax] = bounds;
-  bounds[0] = xMin < x ? xMin : x;
-  bounds[1] = xMax > x ? xMax : x;
-  bounds[2] = yMin < y ? yMin : y;
-  bounds[3] = yMax > y ? yMax : y;
-  bounds[4] = zMin < z ? zMin : z;
-  bounds[5] = zMax > z ? zMax : z;
+  if (arguments.length === 4) {
```
PaulHax: There be dragons going down the xOrPoint overload path.  If you are commited consiter a `typeof xOrPoint === 'number'` to avoid bug with `addPoint(bounds, point, undefined, undefined)`

Sources/Rendering/Core/Prop3D/index.js
```diff
@@ -289,12 +275,12 @@ function vtkProp3D(publicAPI, model) {
 // ----------------------------------------------------------------------------
 
 const DEFAULT_VALUES = {
+  // bounds: null,
```
PaulHax: I see bounds is lazy init, but maybe uncoment this so the model is consistant before `computeBounds` called and not ever `undefined`

finetjul: If `model.bounds` is `undefined` then `model.bounds == null` (in `componteBounds()` will return true. 
Did I misunderstood your comment ?

Sources/Common/DataModel/BoundingBox/index.js
finetjul: (I forgot why but) I wanted to keep the same API than VTK C++...
Thanks for the tip for the typeof.

Sources/Rendering/Core/Prop3D/index.js
PaulHax: Yes, i see the `== null` guard will work.  Just nitpicking on consistency grounds.  No bigge.

Every other class that has bounds (e.g. AbstractMapper3D) initializes it via macro.getArray(publicAPI, model, ['bounds'], 6) which expects the field to already exist on the model. If model.bounds is undefined, then getBounds() (from macro.getArray) returns undefined before computeBounds is ever called, whereas other mappers would return an uninitialized bounds array.

`// bounds: null` is ambiguous.  Reads like dead code in isolation though i do see thats kind of a convention in vtk.js

finetjul: I see your point. I just enforced initialization.

PR #3328, opened 2025-09-16
Fix typing for `vtkViewport.getViewport()`

ghetolay:
### Context
Fixed typings for `vtkViewport.getViewport()` from `vtkViewport` to `[number, number, number, number]`.

### DIFF
--- a/Sources/Rendering/Core/Viewport/index.d.ts
+++ b/Sources/Rendering/Core/Viewport/index.d.ts
@@ -63,14 +63,19 @@ export interface vtkViewport extends vtkObject {
   getSize(): Size;
 
   /**
-   *
+   * Viewport for the Viewport to draw in the rendering window.
+   * Coordinates are expressed as (xmin,ymin,xmax,ymax), where each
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/Viewport/index.d.ts
```diff
@@ -65,7 +65,7 @@ export interface vtkViewport extends vtkObject {
   /**
    *
    */
-  getViewport(): vtkViewport;
+  getViewport(): [number, number, number, number];
```
finetjul, line 73: For consistency, you must do the same with `getViewportByReference`.

Can you take the opportunity to improve the documentation:

```
/**
 * Viewport for the Viewport to draw in the rendering window.
 * Coordinates are expressed as (xmin,ymin,xmax,ymax), where each
 * coordinate is 0 <= coordinate <= 1.0.
 * @returns {Number[]}
 * @see getViewportByReference()
 * @see setViewport()
 */
 ```

ghetolay: sure

PR #3342, opened 2025-10-15
Fix transform controls widget

Jo-Byr:
### Context
TransformControlsWidget had a problem in TRANSLATE and SCALE mode where we would lose interaction after the first mouse mouve.

### Results
The problem described above was fixed in the first commit of this PR.

### Changes
Added missing return
Added a select in the example to facilitate user interaction

### PR and Code Checklist

### DIFF
--- a/Sources/Widgets/Widgets3D/TransformControlsWidget/example/controller.html
+++ b/Sources/Widgets/Widgets3D/TransformControlsWidget/example/controller.html
@@ -1,0 +1,11 @@
+<table>
+  <tr>
+    <td>
+      <select class="mode" style="width: 100%">
+        <option data-mode="translate" value="0" selected>Translation</option>
+        <option data-mode="rotate" value="1">Rotation</option>
+        <option data-mode="scale" value="2">Scale</option>
+      </select>
+    </td>
+  </tr>
+</table>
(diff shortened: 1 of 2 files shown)


Sources/Widgets/Widgets3D/TransformControlsWidget/behavior.js
```diff
@@ -169,6 +169,8 @@ export default function widgetBehavior(publicAPI, model) {
       translation[axisIndex] = translateState.startPos + delta;
       model.widgetState.getTransform().setTranslation(translation);
     }
+
+    return macro.EVENT_ABORT;
```
finetjul, line 173: should this return event_abort if `!worldCoords.length` ?

Sources/Widgets/Widgets3D/TransformControlsWidget/behavior.js
```diff
@@ -189,6 +191,8 @@ export default function widgetBehavior(publicAPI, model) {
       scales[axisIndex] = scale;
       model.widgetState.getTransform().setScale(scales);
     }
+
+    return macro.EVENT_ABORT;
```
finetjul, line 195: ditto

Issue #3330, opened 2025-09-18
[Bug] ResliceCursorWidgetExample is obscured by axis colors


martinerler:
### Bug description

The `ResliceCursorWidget `is obscured by its axis colors. 

### Steps to reproduce

- View the example ResliceCursorWidget at the [website ](https://kitware.github.io/vtk-js/examples/ResliceCursorWidget.html)
- Or start the example ResliceCursorWidget locally at Version v34.11.2

### Detailed Behavior

The image content is not visible, unless you change the axis rotation in one of the three views.
Last Known Working Version is 34.8.0. 
I noticed that the problem came up between v34.8.0 and v34.8.1.

https://github.com/Kitware/vtk-js/compare/v34.8.0...v34.8.1

### Expected Behavior

The image content should be visible and not covered or distracted by other colors. 

### Environment

- vtk.js version: 34.8.1
- Browsers:
- OS:

Last Known Working Version: 34.8.0

daker:
maybe it does impact the PolyDataNormals examples also ? https://kitware.github.io/vtk-js/examples/PolyDataNormals/index.html

martinerler:
> maybe it does impact the PolyDataNormals examples also ? https://kitware.github.io/vtk-js/examples/PolyDataNormals/index.html

I'm not familiar with the PolyDataNormals example in detail, but it does not seem to be obscured by other colors.

martinerler:
Thank you for analysing the problem and assigning the issue :)

martinerler:
I found a way to deside, if the shift and scale in Glyph3DMapper should be used. This heuristic checks if it's a widget and omit shift and scale for that. 
   
` const numPoints = pts.getNumberOfPoints();`
` const isLikelyWidget = numPoints < 100; // Most widget reps have very few points (lines, spheres, etc.)`
` const shouldApplyShiftScale = useShiftAndScale && !isLikelyWidget;`

Is that kind of heuristic is good?
Or should we fix the order WHEN shift + scale is applied, see Glyph3DMapper line 759:

`// apply shift + scale to primitives AFTER vtkOpenGLPolyDataMapper.buildBufferObjects`

martinerler:
PR #3345 is still pending. Is there anything blocking its resolution? I am not shure if i can help here.

Thanks,
Martin

PR #3345, opened 2025-10-31
Add control for coordinate precision shifting and scaling and fix widgets

sankhesh:
### Context

See #3330 

### Results

Screenshot 2025-10-30 235713

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/OpenGL/Glyph3DMapper/index.js
+++ b/Sources/Rendering/OpenGL/Glyph3DMapper/index.js
@@ -698,7 +698,7 @@ function vtkOpenGLGlyph3DMapper(publicAPI, model) {
     let coordShift = 0.0;
     let coordScale = 1.0;
 
-    if (model.coordShiftAndScaleEnabled && pts) {
+    if (model.renderable.getCoordShiftScaleEnabled() && pts) {
       ({ useShiftAndScale, coordShift, coordScale } =
         computeCoordShiftAndScale(pts));
     }
--- a/Sources/Rendering/OpenGL/PolyDataMapper/index.js
+++ b/Sources/Rendering/OpenGL/PolyDataMapper/index.js
@@ -2022,7 +2022,7 @@ export function extend(publicAPI, model, initialValues = {}) {
   }
 
   // Build VTK API
-  macro.setGet(publicAPI, model, ['context', 'coordShiftScaleEnabled']);
+  macro.setGet(publicAPI, model, ['context']);
 
   model.VBOBuildTime = {};
   macro.obj(model.VBOBuildTime, { mtime: 0 });


Sources/Rendering/OpenGL/PolyDataMapper/index.js
```diff
@@ -2018,7 +2022,7 @@ export function extend(publicAPI, model, initialValues = {}) {
   }
 
   // Build VTK API
-  macro.setGet(publicAPI, model, ['context']);
+  macro.setGet(publicAPI, model, ['context', 'coordShiftScaleEnabled']);
```
finetjul: model.coordShiftScaleEnabled is not initialized on purpose ? 
Therefore false by default ?
Why would coordShiftScaleEnabled have a different value between the API specific mapper (vtkOpenGLPolyDataMapper) and the original mapper (vtkPolyDataMapper)

finetjul: This should be added in the documentation.

Sources/Rendering/Core/Mapper/index.d.ts
```diff
@@ -467,6 +468,15 @@ export interface vtkMapper
   setInterpolateScalarsBeforeMapping(
     interpolateScalarsBeforeMapping: boolean
   ): boolean;
+
+  /**
+   * Control whether the mapper should automatically shift and scale
+   * the coordinates to improve precision. This is enabled by default.
```
finetjul, line 474: please add comment to explain when/why you would want to turn it off.

Sources/Rendering/OpenGL/Glyph3DMapper/index.js
```diff
@@ -694,8 +694,14 @@ function vtkOpenGLGlyph3DMapper(publicAPI, model) {
     const garray = model.renderable.getMatrixArray();
 
     const pts = model.renderable.getInputData(0).getPoints();
-    const { useShiftAndScale, coordShift, coordScale } =
-      computeCoordShiftAndScale(pts);
+    let useShiftAndScale = false;
+    let coordShift = 0.0;
+    let coordScale = 1.0;
+
+    if (model.coordShiftAndScaleEnabled && pts) {
```
finetjul: why don't you check the value of the original mapper (vtkPolyDataMapper.coordShiftAndScaleEnabled), but just the value of the API specific value (vtkOpenGLPolyDataMapper.coordShiftAndScaleEnabled)

Sources/Rendering/OpenGL/Glyph3DMapper/index.js
```diff
@@ -694,8 +694,14 @@ function vtkOpenGLGlyph3DMapper(publicAPI, model) {
...
+      ({ useShiftAndScale, coordShift, coordScale } =
+        computeCoordShiftAndScale(pts));
+    }
```
finetjul, line 704: instead of having a flag, could you detect when shift/scale have "bad" values ?

Sources/Rendering/OpenGL/Glyph3DMapper/index.js
sankhesh: I see what you mean. Good catch 👍

Sources/Rendering/OpenGL/Glyph3DMapper/index.js
finetjul: maybe a metric could be computed with regard to the current camera settings. 
An error in the "display world" could be considered to too large and coord shift/scale be therefore discarded ?

PR #3347, opened 2025-11-03
fix(datasetattributes): Fix active attribute assignment

sankhesh:
This change fixes an issue where setting different attributes after initializing the dataset causes the active attributes to be retained from previous dataset state.

### Context

When dynamically changing TubeFilter properties at runtime, tubes with colormaps lose their colors and render as solid. 

#### Steps to Reproduce

1. Open the official TubeFilter example: https://kitware.github.io/vtk-js/examples/TubeFilter.html
2. Observe the initial tubes with proper colormap (red and yellow tubes) ✅
3. Modify any tube property using the controls (Number of Sides, Radius, etc.) ❌
4. **Bug**: Tubes lose their colors and become solid white/gray

image

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Common/DataModel/DataSetAttributes/index.js
+++ b/Sources/Common/DataModel/DataSetAttributes/index.js
@@ -61,17 +61,9 @@ function vtkDataSetAttributes(publicAPI, model) {
       );
       return -1;
     }
-    let currentAttribute = model[`active${attType}`];
-    if (currentAttribute >= 0 && currentAttribute < model.arrays.length) {
-      if (model.arrays[currentAttribute] === arr) {
-        return currentAttribute;
-      }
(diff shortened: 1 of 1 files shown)


finetjul (approved):
LGTM.
Did you check that your dataset does not bigger and bigger because new arrays get added without being removed when properties are changed ?

sankhesh:
> LGTM. Did you check that your dataset does not get bigger and bigger because new arrays get added without being removed when properties are changed ?

Yes, tested by checking number of point attribute arrays after changing multiple properties on the TubeFilter example. The number of arrays remained the same.

Issue #3349, opened 2025-11-03
scalarBarActor setAxisTextStyle does not work for fontSize


aerogt3:
### Bug description

Using 34.14.0, the command below works great for fontColor and fontFamily, but will not respect fontSize, either as an integer or a string:

`vtkPlaneScalarBarActor.setAxisTextStyle({'fontColor':'black', 'fontSize':8, 'fontFamily':'arial'})`

Feature request: it would also be nice to have the setOrientationToHorizontal as is available in vtk. The auto layout does a nice job of placing things but often puts vertical when horizontal would look better.

### Steps to reproduce

https://codesandbox.io/p/devbox/hello-world-cone-forked-9qdhsm

### Expected Behavior

When changing fontSize, fonts should respond

### Environment

- vtk.js version: 34.14.0
- Browsers: firefox 144.0.2 
- OS: ubuntu 2204

PR #3350, opened 2025-11-04
scalarBar orientation (horizontal/vertical) can be set by user

aerogt3:
### Context
Explain why this change is needed. Please include relevant links supporting this change, such as:
- adds feature suggested/discussed in #3349
- VTK C++ has scalarBar.setOrientationToHorizontal and setOrientationToVertical, so simply extending that functionality to vtk.js

### Results
Colormap orientation can now be overridden by the user

### Changes
scalarBar.get/set for the following added:

1. setOrientation
2. setOrientationToHorizontal
3. setOrientationToVertical

### Testing
Force a scalar bar horizontal with either:

1. scalarBarActor.setOrientation('horizontal')
2. scalarBarActor.setOrientationToHorizontal()

### DIFF
--- a/Sources/Rendering/Core/ScalarBarActor/Constants.js
+++ b/Sources/Rendering/Core/ScalarBarActor/Constants.js
@@ -1,6 +1,6 @@
 export const Orientation = {
-  VERTICAL: 'vertical',
   HORIZONTAL: 'horizontal',
+  VERTICAL: 'vertical',
   AUTO: 'auto',
 };
 
--- a/Sources/Rendering/Core/ScalarBarActor/index.d.ts
+++ b/Sources/Rendering/Core/ScalarBarActor/index.d.ts
@@ -22,7 +22,7 @@ export interface IStyle {
   fontSize?: string;
 }
 
-export type Orientation = 'horizontal' | 'vertical' | null;
+export type Orientation = 'horizontal' | 'vertical' | 'auto' | null;
 
 /**
  *


Sources/Rendering/Core/ScalarBarActor/Constants.js
```diff
@@ -0,0 +1,8 @@
+export const Orientation = {
+  VERTICAL: 'vertical',
+  HORIZONTAL: 'horizontal',
```
finetjul, line 2: you could add "AUTO"

Sources/Rendering/Core/ScalarBarActor/index.d.ts
```diff
@@ -222,6 +224,27 @@ export interface vtkScalarBarActor extends vtkActor {
...
+  /**
+   * Forces the scalar bar to use vertical orientation regardless of aspect ratio
+   */
+  setOrientationToVertical(): boolean;
+
+  /**
+   * Set the orientation of the scalar bar
+   * @param orientation 'horizontal' to force horizontal, 'vertical' to force vertical, or null/undefined for auto
+   */
+  setOrientation(orientation: 'horizontal' | 'vertical' | null): boolean;
```
finetjul: you need to declare a type `Orientation` and use it instead of manually writing them here.

Sources/Rendering/Core/ScalarBarActor/Constants.js
aerogt3: Doing so [would not change the behavior at all](https://github.com/aerogt3/vtk-js/blob/scalarBar_Orientation/Sources/Rendering/Core/ScalarBarActor/index.js#L89-L98)? 

The user could enter vtkPlaneScalarBarActor.setOrientation('banana') and it would revert to auto.

finetjul: You are correct but adding it as an option would make the "auto" behavior "explicit".
It's currently not obvious that the auto option is available by looking at the constant options.

aerogt3: It doesn't need to be obvious by looking at Constants because anyone who looks at Constants will also look at index.js, or can understand that the functionality is apparent in index.js.

IMO nitpicking non-functional changes discourages people from contributing.

Sources/Rendering/Core/ScalarBarActor/index.d.ts
```diff
@@ -22,10 +22,13 @@ export interface IStyle {
   fontSize?: string;
 }
 
+export type Orientation = 'horizontal' | 'vertical' | null;
```
finetjul: Please add 'auto' here as well.

aerogt3: When I first created the issue I was told "feel free to create a PR." Now for this comment, I reply with the same.

I'll create a working .js that I can use on my own, I appreciate the help reaching that point.

finetjul: I'm sorry you take it that way. 

That was the last change I was asking for. Please note that I didn't ask you to create a unit test nor update an example.

If you want to maintain a fork of VTK.js it is indeed possible. It may end up being more effort on your end though.

aerogt3: I thought removing raw strings was the last comment, but then got 3 more about the same thing (even though raw strings 'middle' and 'bottom' exist 10+ times in the same file).

None of this matters to [users like this](https://discourse.vtk.org/t/how-to-change-the-position-of-scalar-bar-on-vtk-js/13082), and a dev doesn't need AUTO to understand.

The options are 
1.) leave the problem as-is
2.) good fix
3.) perfect fix

I've provided option 2, [also for the fontSize bug](https://github.com/Kitware/vtk-js/compare/master...aerogt3:vtk-js:scalarBarFontSizeIssue), and if a maintainer wants to make either of them option 3, great.

I won't do future PRs, just patching an option 2 locally is faster.

`Don't let the perfect be the enemy of the good`

finetjul: > I won't do future PRs, just patching an option 2 locally is faster.

That's fine to me. I hope someone from the VTK.js community will provide you support and guidance when you will face future issues.

PR #3352, opened 2025-11-07
Scalar bar font size issue

finetjul:
Support font sizing.

### Context
Specifying a font size onto the scalar bar actor is ignored.

### Results
A provided font size is now supported.

### Changes

### PR and Code Checklist

### Testing
- [ ] Tested environment:
  - **vtk.js**: master
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/ScalarBarActor/index.js
+++ b/Sources/Rendering/Core/ScalarBarActor/index.js
@@ -280,7 +280,7 @@ function vtkScalarBarActorHelper(publicAPI, model) {
     const newTmAtlas = new Map();
     let maxWidth = 0;
     let totalHeight = 1; // start one pixel in so we have a border
-    applyTextStyle(model.tmContext, model.axisTextStyle, 14);
+    applyTextStyle(model.tmContext, model.axisTextStyle, 18);
     let metrics = model.tmContext.measureText(model.renderable.getAxisLabel());
     let entry = {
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Core/ScalarBarActor/index.js
```diff
@@ -275,7 +280,7 @@ function vtkScalarBarActorHelper(publicAPI, model) {
     const newTmAtlas = new Map();
     let maxWidth = 0;
     let totalHeight = 1; // start one pixel in so we have a border
-    applyTextStyle(model.tmContext, model.axisTextStyle);
+    applyTextStyle(model.tmContext, model.axisTextStyle, 14);
```
finetjul: Please keep the previous default value (18) except if there is a strong reason (and if so, it could be modified in a different commit)

Sources/Rendering/Core/ScalarBarActor/index.js
```diff
@@ -292,7 +297,7 @@ function vtkScalarBarActorHelper(publicAPI, model) {
     // and the ticks, NaN Below and Above
     results.tickWidth = 0;
     results.tickHeight = 0;
-    applyTextStyle(model.tmContext, model.tickTextStyle);
+    applyTextStyle(model.tmContext, model.tickTextStyle, 8);
```
finetjul: why not 14 ? It was the default value.

PR #3353, opened 2025-11-07
fix(OpenGLTexture): fix regression causing RGB DICOM not to display

### LINKED ISSUE #3344: [Bug] Not able to render RGB DICOMs after 34.0.0
### Bug description

I'm not able to render ImageMapper with RGB DICOM after 34.0.0. It renders like this for RGB US for instance

Image

### Steps to reproduce

i have created a sample example here for using vtk.js with drag and drop dicom

https://github.com/sedghi/vtk-js/tree/fix/reproduce

- clone
- npm install 
- npm run example LocalDICOMViewer 
- Drag and Drop from the following [dropbox](https://www.dropbox.com/scl/fo/10vi94iqhiq0t8hy5awxz/ABbuu32Q2ALsC2XNWOJdw3E?rlkey=52j2gy9hysaod5cho79yreppa&dl=0) - ct loads fine, Ultrasound (RGB does not)

### Detailed Behavior

It appears that RGB independent components does not render correctly

### Expected Behavior

Renders like this 

Image

### Environment

- vtk.js version: 34.x
- Browsers: Chrome
- OS: macOS

daker:
### Context

A regression was introduced during the transition to named parameters, where numComps was passed instead of numberOfComponents set to 1 by default. This caused RGB DICOM images to display incorrectly, as only one component was used instead of the intended three.

Ref: https://github.com/kitware/vtk-js/commit/7fbbaa83b697f7fb7b09fe03a05fbca397271496

Fixes #3344

### Results

Check the example in the issue

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/OpenGL/Texture/index.js
+++ b/Sources/Rendering/OpenGL/Texture/index.js
@@ -1625,7 +1625,7 @@ function vtkOpenGLTexture(publicAPI, model) {
       width,
       height,
       dataArray: vtkDataArray.newInstance({
-        numComps,
+        numberOfComponents: numComps,
         dataType,
         values: data,
         ranges,
@@ -1897,7 +1897,7 @@ function vtkOpenGLTexture(publicAPI, model) {
       height,
       depth,
       dataArray: vtkDataArray.newInstance({
-        numComps,
+        numberOfComponents: numComps,
         dataType,
         values: data,
         ranges,


daker:
@finetjul @floryst @sedghi

sedghi:
is that it? haha good catch

Issue #3355, opened 2025-11-12
Direction missing for Index to World conversion in Image marching squares


Joeycho:
### Problem description

https://github.com/Kitware/vtk-js/blob/cb70d12d9426ce592790c0a540c9c11dde5a975e/Sources/Filters/General/ImageMarchingSquares/index.js#L160

https://github.com/Kitware/vtk-js/blob/cb70d12d9426ce592790c0a540c9c11dde5a975e/Sources/Filters/General/ImageMarchingSquares/index.js#L178-L180

Currently it finds the points in ijk coordinates (Index), and save points in world coordinate. However, direction info is missing in this calculation. -> It can be problematic when direction is not standard (a bit tilted..)

----------

### Suggestion
Either pass direction to here, and calculate completely, or return the value in ijk coordinates, and expect people to convert the coordinates from indexToWorld from the viewport.

Based on the suggestion from the community, I will be happy to create a PR to contribute

PR #3362, opened 2025-11-17
fix(ImageMarchingSquares): use indexToWorld function, rather than separate calculation

Joeycho:
Current:
direction missing.
https://github.com/Kitware/vtk-js/issues/3355

PR:
Input image has indexToWorld function. Reuse it.

### DIFF
--- a/Sources/Filters/General/ImageMarchingSquares/index.js
+++ b/Sources/Filters/General/ImageMarchingSquares/index.js
@@ -178,8 +178,8 @@ function vtkImageMarchingSquares(publicAPI, model) {
         }
         if (pId === undefined) {
           const t =
-          (cVal - pixelScalars[edgeVerts[0]]) /
-          (pixelScalars[edgeVerts[1]] - pixelScalars[edgeVerts[0]]);
+            (cVal - pixelScalars[edgeVerts[0]]) /
+            (pixelScalars[edgeVerts[1]] - pixelScalars[edgeVerts[0]]);
           const x0 = pixelPts.slice(edgeVerts[0] * 3, (edgeVerts[0] + 1) * 3);
           const x1 = pixelPts.slice(edgeVerts[1] * 3, (edgeVerts[1] + 1) * 3);
           xyz[0] = x0[0] + t * (x1[0] - x0[0]);


Sources/Filters/General/ImageMarchingSquares/index.js
```diff
@@ -170,14 +170,7 @@ function vtkImageMarchingSquares(publicAPI, model) {
...
         if (pId === undefined) {
-          const t =
-            (cVal - pixelScalars[edgeVerts[0]]) /
-            (pixelScalars[edgeVerts[1]] - pixelScalars[edgeVerts[0]]);
-          const x0 = pixelPts.slice(edgeVerts[0] * 2, (edgeVerts[0] + 1) * 2);
-          const x1 = pixelPts.slice(edgeVerts[1] * 2, (edgeVerts[1] + 1) * 2);
-          xyz[kernelX] = x0[0] + t * (x1[0] - x0[0]);
-          xyz[kernelY] = x0[1] + t * (x1[1] - x0[1]);
-          xyz[model.slicingMode] = z;
+          const xyz = indexToWorld(ijk);
```
finetjul: I do not think it is a drop-in remplacement. 
Previous code was not trying to find the location of ijk in the world coordinate space, but more something the location of cVal on the edgeVerts edge.
Maybe it is getPixelPoints that could use indexToWorld

finetjul (commented):
Please reformat your commit message to match semantic release.

Sources/Filters/General/ImageMarchingSquares/index.js
Joeycho: I double checked, and you are right. @finetjul 

The only update.. happens at

https://github.com/Kitware/vtk-js/blob/cb70d12d9426ce592790c0a540c9c11dde5a975e/Sources/Filters/General/ImageMarchingSquares/index.js#L99-L100

https://github.com/Kitware/vtk-js/blob/cb70d12d9426ce592790c0a540c9c11dde5a975e/Sources/Filters/General/ImageMarchingSquares/index.js#L160

I just updated accordingly.

Sources/Filters/General/ImageMarchingSquares/index.js
```diff
@@ -171,13 +168,13 @@ function vtkImageMarchingSquares(publicAPI, model) {
...
-            (cVal - pixelScalars[edgeVerts[0]]) /
-            (pixelScalars[edgeVerts[1]] - pixelScalars[edgeVerts[0]]);
+          (cVal - pixelScalars[edgeVerts[0]]) /
+          (pixelScalars[edgeVerts[1]] - pixelScalars[edgeVerts[0]]);
           const x0 = pixelPts.slice(edgeVerts[0] * 2, (edgeVerts[0] + 1) * 2);
           const x1 = pixelPts.slice(edgeVerts[1] * 2, (edgeVerts[1] + 1) * 2);
           xyz[kernelX] = x0[0] + t * (x1[0] - x0[0]);
           xyz[kernelY] = x0[1] + t * (x1[1] - x0[1]);
-          xyz[model.slicingMode] = z;
+          xyz[model.slicingMode] = z
```
finetjul: why remove `;` ?

I'm surprised your code is working.

Sources/Filters/General/ImageMarchingSquares/index.js
```diff
@@ -85,60 +85,57 @@ function vtkImageMarchingSquares(publicAPI, model) {
...
-    pixelPts[4] = pixelPts[0]; // 2
-    pixelPts[5] = pixelPts[1] + spacing[kernelY];
+    pixelPts[4] = pixelPts[0]; // 2, i
+    pixelPts[5] = pixelPts[1] + spacing[kernelY]; // j+1
 
-    pixelPts[6] = pixelPts[2]; // 3
-    pixelPts[7] = pixelPts[5];
+    pixelPts[6] = pixelPts[2]; // 3, i+1
+    pixelPts[7] = pixelPts[5]; // j+1
   };
```
finetjul, line 120: I think it is still missing one information: the orientation.

If your image is not axis aligned (has a small tilt), it should increment (i.e. add spacing) not in an axis aligned way.

You can either add a "rotated spacing". Or do first the increment in the ijk world (i.e. +1), and then use indexToWorld to get the world coordinate. In that case, you do not need `spacing` anymore.

Joeycho: Hmm... and then, I think we cannot calculate Z in world coordinates separately (currently, it's like that).

Because, (i, j,k), (i+1, j, k) (i, j+1, k) (i+1, j+1, k) they are on the same k, but not the same z after indexToWorld conversion (a small tilt)

I updated PR accordingly.

Sources/Filters/General/ImageMarchingSquares/index.js
```diff
@@ -86,57 +86,71 @@ function vtkImageMarchingSquares(publicAPI, model) {
...
-    pixelPts[4] = pixelPts[0]; // 2
-    pixelPts[5] = pixelPts[1] + spacing[kernelY];
-
-    pixelPts[6] = pixelPts[2]; // 3
-    pixelPts[7] = pixelPts[5];
+  publicAPI.getPixelPoints = (ijk, kernelX, kernelY, indexToWorld) => {
+    // pixelPts = a flattened world coordinates array of [ (i,j,k), (i+1,j,k), (i,j+1,k), (i+1,j+1,k)]
+
+    // 0: i, j, k
+    let worldPos = indexToWorld(ijk);
```
finetjul: please use the indexToWorld aout parameter for better performance (to avoid instantiating a new array each time)

Sources/Filters/General/ImageMarchingSquares/index.js
```diff
@@ -86,57 +86,71 @@ function vtkImageMarchingSquares(publicAPI, model) {
...
+    pixelPts[0] = worldPos[0];
+    pixelPts[1] = worldPos[1];
+    pixelPts[2] = worldPos[2];
+
+    // 1: i+1, j, k
+    const ijk1 = [...ijk];
```
finetjul: maybe you could do something like the following to avoid instantiating many arrays:
```
  const neighborIJK = [...ijk];
  indexToWorld()
  neighborIJK[kernelX] += 1;
  indexToWorld()
  neighborIJK[kernelY] += 1;
  indexToWorld()
  neighborIJK[kernelX] -= 1;
  indexToWorld()
```

PR #3363, opened 2025-11-17
fix(links): clear Cells and Links arrays when polydata is cleared

### LINKED ISSUE #3361: [Bug] ContourLoopExtraction doesn't work anymore
### Bug description

Open https://kitware.github.io/vtk-js/examples/ContourLoopExtraction/index.html
Use the example

```javascript
Uncaught TypeError: Cannot read properties of null (reading 'forEach')
    at e.buildLinks (ContourLoopExtraction.js:1:175374)
    at e.buildLinks (ContourLoopExtraction.js:1:204265)
    at DT (ContourLoopExtraction.js:1:712029)
```

### Steps to reproduce

Check description

### Detailed Behavior

Check description

### Expected Behavior

Example should work as expected

### Environment

- vtk.js version:
- Browsers:
- OS:

finetjul:
BuildLinks enforces building cells only if they do not exist. When clearing the polydata, cells and links must be cleared accordingly. Same when verts, lines, polys and strips arrays are changed

fix #3361

### Context
Filters reuse their previous output data, but they now clear them.
A crash was happening when trying to access a cell that does not exist in model.cells.
This is because the model.cells array is outdated.

### Results
Enforce the cells array to be recomputed after clearing a polydata.
The ContourLoopExtraction example now works again.

### Changes

### PR and Code Checklist

### Testing
- [x] This change adds or fixes an example
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows 11
  - **Browser**: Chrome

### DIFF
--- a/Sources/Common/DataModel/PolyData/index.js
+++ b/Sources/Common/DataModel/PolyData/index.js
@@ -39,6 +39,11 @@ function vtkPolyData(publicAPI, model) {
       .replace(/\s+/g, '');
   }
 
+  function clearCells() {
+    model.cells = undefined;
+    model.links = undefined;
+  }
+
   // build empty cell arrays and set methods
   POLYDATA_FIELDS.forEach((type) => {
     publicAPI[`getNumberOf${camelize(type)}`] = () =>
@@ -48,6 +53,7 @@ function vtkPolyData(publicAPI, model) {
     } else {
       model[type] = vtk(model[type]);
     }
+    model[`_on${camelize(type)}Changed`] = clearCells;
   });
 
   publicAPI.getNumberOfCells = () =>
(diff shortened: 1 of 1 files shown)


Sources/Common/DataModel/PolyData/index.js
```diff
@@ -180,7 +187,7 @@ function vtkPolyData(publicAPI, model) {
    * topologically complex queries.
    */
   publicAPI.buildLinks = (initialSize = 0) => {
-    if (model.cells === undefined) {
+    if (model.cells == null) {
```
daker, line 190: I did test the code with ContourLoopExtraction example it only works with `model.cells === undefined` and not `model.cells == null`

daker: nvm i think it was a refresh issue on my side, it works

daker (approved):
LGTM

PR #3365, opened 2025-11-18
feat(ColorMaps): Add grayscale sigmoid colormap based on DICOM standard

jadh4v:
# Context

Add a non-linear sigmoid grayscale colormap as an alternative to linear grayscale. This mapping is described in the DICOM standard in section C.11.2.1.3.1 SIGMOID Function and equation C.11-1. We use a normalized window_width of [-1, 1] and window_center = 0. The colormap is generated using a 16-point sampling of this function.

See: https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_c.11.2.html

~~## There are two ways to access the sigmoid function:~~
~~1. Developer can use the newly added sigmoid presets that already are generated on the sigmoid curve, or
2. Use the `ScalarMappingMode` property on `vtkColorTransferFunction` to adapt any color map to sigmoid mapping. the `vtkColorTransferFunction` provides two `ScalarMappingModes`. `LINEAR` continues to be the default value and behavior, and `SIGMOID` allows to switch the scalar to color mapping to the sigmoid function. Additionally, you can control the growth rate factor using `setSigmoidGrowthRate()` which is usually a negative number and equal to `-4` according to the DICOM standard.~~

# Results

A comparison of linear vs sigmoid grayscale using the ImageViewer example.
```
npm run example ImageViewer
```

| linear | sigmoid |
|--|--|
| Screenshot 2025-11-18 at 6 49 43 PM | Screenshot 2025-11-18 at 6 50 00 PM |

The sampled 16-point sigmoid function plotted through matplotlib:
dicom-grayscale-sigmoid-plot

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


Sources/Rendering/Core/ColorTransferFunction/ColorMaps.json
```diff
@@ -529,6 +529,78 @@
...
+   {
+      "ColorSpace" : "RGB",
+      "Name" : "Grayscale-DICOM-sigmoid",
+      "Notes": "This is a 16 point sample of a sigmoid grayscale colormap described in the DICOM standard. See equation C.11-1 in section C.11.2",
+      "NanColor" : [ 1, 0, 0 ],
+      "RGBPoints" : [
+         -1,
+         0.01798620996209156,
+         0.01798620996209156,
+         0.01798620996209156,
```
finetjul, line 543: why not 0 ?

Sources/Rendering/Core/ColorTransferFunction/ColorMaps.json
```diff
@@ -529,6 +529,78 @@
...
+         0.9494698377645865,
+         0.9494698377645865,
+         0.8666666666666667,
+         0.9697243080130495,
+         0.9697243080130495,
+         0.9697243080130495,
+         1,
+         0.9820137900379085,
+         0.9820137900379085,
+         0.9820137900379085
```
finetjul, line 603: why not 1 ?

Sources/Rendering/Core/ColorTransferFunction/ColorMaps.json
```diff
@@ -529,6 +529,78 @@
       "NanColor" : [ 1, 0, 0 ],
       "RGBPoints" : [ 0, 0, 0, 0, 1, 1, 1, 1 ]
    },
+   {
+      "ColorSpace" : "RGB",
```
finetjul, line 533: why not rely on midpoint+sharpness ?
https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Core/ColorTransferFunction/index.js#L721-L725

finetjul: Maybe add support for a "Nodes" key/value in addition to "RGBPoints".
See applyColorMap in ColorTransferFunction/index.js

Sources/Rendering/Core/ColorTransferFunction/ColorMaps.json
jadh4v: Good point. I am using the equation specified by the DICOM standard:
`1 / (1 + e^(4x))`
The multiplying factor of 4 could be increased to make the ends closer to 0 and 1 but they are asymptotic, so they only converge to those values. I think the purpose of the sigmoid is to increase the gradient at the center of the scalar range (window). And it is probably ok to sacrifice some color range at the ends.

jadh4v: I think we could rescale the function to map from 0 to 1 so that it uses the full available grayscale range, but I am not certain if that is ok to do for the DICOM standard. I think we could wait to fix this once we get more feedback on this feature from developers.
Or, I have already added `LINEAR` and `SIGMOID` modes to `vtkColorTransferFunction`, I can add another `RESCALED_SIGMOID`.

finetjul: I guess you could also declare `belowRangeColor` to pure black and `aboveRangeColor` to pure white as an alternative.

Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -493,6 +493,8 @@ function vtkColorTransferFunction(publicAPI, model) {
       ];
     }
 
+    const useSigmoid =
+      publicAPI.getScalarMappingMode() == ScalarMappingMode.SIGMOID;
```
finetjul: please consider the linting errors

Sources/Rendering/Core/ColorTransferFunction/ColorMaps.json
jadh4v: As we discussed in the meeting, using midpoint and sharpness would only be an approximation of the sigmoid, which is not desirable. Even though I haven't actually calculated how far off that curve would be from the sigmoid curve. I did add support for reading `aboveRangeColor` and `belowRangeColor` for mapping the ends to full range.

Sources/Rendering/Core/ColorTransferFunction/ColorMaps.json
jadh4v: I have added support for reading the `belowRangeColor` and `aboveRangeColor` from the preset json.

jadh4v:
@finetjul 
I have updated this PR to drop the TransferFunction related changes. Sigmoid is now only available through the presets. Added equivalent `inverted-grayscale` and `inverted-grayscale-DICOM-sigmoid` presets. I also rebased the branch on the latest main. Should be ready to merge when ci pipeline passes.

finetjul (approved):
LGTM

PR #3366, opened 2025-11-21
3351 fix many actors performance hit

finetjul:
### Context
Rendering is slow when a scene has many actors. This is slow since addMissingNodes() enforceOrder has been added.

### Results
Performance should be back to before what they were before enforceOrder was added.

### Changes
Add caching for OpenGLCamera node and do not enforce data objs to be at the beginning of the children list.

### PR and Code Checklist

### Testing
- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


daker:
@finetjul Does this improve the interaction with the scene when we have many actors ? ex in https://kitware.github.io/vtk-js/examples/IFCImporter/index.html?mergeGeometries=false

PR #3368, opened 2025-11-24
fix(IFCImporter): optimize IFC parsing

daker:
### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/IO/Geometry/IFCImporter/index.js
+++ b/Sources/IO/Geometry/IFCImporter/index.js
@@ -45,14 +45,14 @@ function vtkIFCImporter(publicAPI, model) {
     const pointValues = new Float32Array(vertices.length / 2);
     const normalsArray = new Float32Array(vertices.length / 2);
 
-    for (let i = 0; i < vertices.length; i += 6) {
-      pointValues[i / 2] = vertices[i];
-      pointValues[i / 2 + 1] = vertices[i + 1];
-      pointValues[i / 2 + 2] = vertices[i + 2];
+    for (let i = 0, p = 0; i < vertices.length; p += 3) {
+      pointValues[p] = vertices[i++];
+      pointValues[p + 1] = vertices[i++];
+      pointValues[p + 2] = vertices[i++];
 
-      normalsArray[i / 2] = vertices[i + 3];
-      normalsArray[i / 2 + 1] = vertices[i + 4];
-      normalsArray[i / 2 + 2] = vertices[i + 5];
+      normalsArray[p] = vertices[i++];
+      normalsArray[p + 1] = vertices[i++];
+      normalsArray[p + 2] = vertices[i++];
     }
 
     const nCells = indices.length;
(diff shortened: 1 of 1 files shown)


Sources/IO/Geometry/IFCImporter/index.js
```diff
@@ -99,40 +99,44 @@ function vtkIFCImporter(publicAPI, model) {
...
+        pointValues[p + 2] = vertices[i + 2];
+
+        normalsArray[p] = vertices[i + 3];
+        normalsArray[p + 1] = vertices[i + 4];
+        normalsArray[p + 2] = vertices[i + 5];
+
+        colorArray[p] = color.x;
+        colorArray[p + 1] = color.y;
+        colorArray[p + 2] = color.z;
       }
```
finetjul, line 130: to make it faster, you could set `vertices` to `pointValues` and at the end, do a unique `transformMatrix.apply(pointValues)`

daker: can you check if it's ok ?

Sources/IO/Geometry/IFCImporter/index.js
```diff
@@ -45,14 +45,14 @@ function vtkIFCImporter(publicAPI, model) {
...
-      normalsArray[i / 2 + 1] = vertices[i + 4];
-      normalsArray[i / 2 + 2] = vertices[i + 5];
+    for (let i = 0, p = 0; i < vertices.length; i += 6, p += 3) {
+      pointValues[p] = vertices[i];
+      pointValues[p + 1] = vertices[i + 1];
+      pointValues[p + 2] = vertices[i + 2];
+
+      normalsArray[p] = vertices[i + 3];
+      normalsArray[p + 1] = vertices[i + 4];
+      normalsArray[p + 2] = vertices[i + 5];
```
finetjul: you could do `vertices[i++]` that way you don't have to do `i += 6`

Sources/IO/Geometry/IFCImporter/index.js
finetjul: I think it's fine. But I hope you tried it with a real dataset.

Sources/IO/Geometry/IFCImporter/index.js
```diff
@@ -45,14 +45,16 @@ function vtkIFCImporter(publicAPI, model) {
...
+    for (let i = 0, p = 0; i < vertices.length;) {
+      pointValues[p] = vertices[i++];
+      pointValues[p + 1] = vertices[i++];
+      pointValues[p + 2] = vertices[i++];
+
+      normalsArray[p] = vertices[i++];
+      normalsArray[p + 1] = vertices[i++];
+      normalsArray[p + 2] = vertices[i++];
+
+      p += 3;
```
finetjul: you could have kept it in the for loop

daker: sure

PR #3369, opened 2025-11-24
fix(FullScreenRenderWindow): add warning that this class is for examples only

### LINKED ISSUE #3229: [Bug]
### Bug description

Am I doing this correctly? After clicking the button, it triggers this function to change the size of the mounted DOM element and re-render the model. However, the model seems to remain in its original position, as I can still pick up the actor by clicking the previous location.
` changeBodyScene() {
      let scene = document.getElementsByClassName("scene");
      console.log("BodyScene", scene);
      let canvas = scene[0].getElementsByTagName("canvas")[0];
      if (this.isHeng) {
        scene[0].style.width = "calc(50% - 50px)";
        scene[0].style.height = "calc(100vh)";
        scene[0].style.marginLeft = "calc(50% + 50px)";
        const sceneWidth = scene[0].offsetWidth;
        const sceneHeight = scene[0].offsetHeight;
        canvas.width = sceneWidth;
        canvas.height = sceneHeight;
        this.screenRender.resize();
        this.renderer.resetCamera();
        this.renderWindow.render();
      } else {
        scene[0].style.height = "calc((100vh + 300px) / 2 - 80px - 150px)";
        scene[0].style.marginTop = "calc((100vh - 300px) / 2 + 75px) ";
      }
    },`

### Steps to reproduce

Image

### Detailed Behavior

I can still pick up the actor by clicking the previous location.

### Expected Behavior

Correct approach to re-rendering the model

### Environment

- vtk.js version:32.6.0
- Browsers:chrome
- OS:windows10

daker:
### Context

fixes #3229

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Misc/FullScreenRenderWindow/index.js
+++ b/Sources/Rendering/Misc/FullScreenRenderWindow/index.js
@@ -13,6 +13,8 @@ import 'vtk.js/Sources/Rendering/Core/Actor';
 import 'vtk.js/Sources/Rendering/Core/Mapper';
 import 'vtk.js/Sources/Rendering/Misc/RenderingAPIs';
 
+const { vtkWarningMacro } = macro;
+
 // Process arguments from URL
 const userParams = vtkURLExtract.extractURLParameters();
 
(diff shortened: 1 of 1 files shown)


Sources/Rendering/Misc/FullScreenRenderWindow/index.js
```diff
@@ -221,6 +223,10 @@ export function extend(publicAPI, model, initialValues = {}) {
     'controlContainer',
   ]);
 
+  vtkWarningMacro(
```
finetjul, line 226: We can't really prevent (i.e. bother with a warning message) anyone from using `FullScreenRenderWindow`.

We could think of conditionally testing if hostname is not "localhost" to decide to print a message. 

An alternative is to simply document the class that resizing is not possible.

Issue #1285, opened 2019-12-11
Renderer's `resetCamera()` is biased to height when calculating zoom


mix3d:
I've noticed that there is a render bias for fitting the bounds in the render viewport when using `renderer.resetCamera()`

Example:
Wide viewport scales to fit:
![image](https://user-images.githubusercontent.com/1474137/70650909-2d657d80-1c0d-11ea-9e08-00d81e6df2b6.png)
VS
Narrow viewport crops significantly
![image](https://user-images.githubusercontent.com/1474137/70650975-4837f200-1c0d-11ea-9613-adc88d61755c.png)

mix3d:
Any tips for getting wide volumes in tall viewports to render well?

finetjul:
`Renderer.resetCamera()` does not take into account the viewport size (nor ratio).

Instead it fits the bounding sphere of the scene into a square frustum viewport. 
It seems to work fine because it the vertical axis is always correctly fitted (it could be improved by not using a bounding sphere but the actual bounds).

A solution is to create your own resetCamera function that takes into account the viewer size ratio.

There is useful documentation in code:
https://github.com/Kitware/vtk-js/blob/0a4f91d0adbedb31d21338a8737b9ec2f28a561d/Sources/Rendering/Core/Renderer/index.js#L411-L421

mix3d:
If the bounding sphere is supposed to fit in the view frustum, wouldn't you expect the sphere to always fit, regardless of the viewport frustum ratio? The point of the sphere is to account for the potential oblong-ness or arbitrary orientation of the volume itself, right? So the Sphere is the correct way to go, because as long as the sphere fits, the bounds will be included.

If it's only checking the vertical axis, and that's the one that seems to work all the time, I think a quick check on the frustum to determine if it should use the vertical or horizontal axis would be an easy fix here.

finetjul:
C++ implementation takes into account aspect ratio:
https://gitlab.kitware.com/vtk/vtk/-/blob/master/Rendering/Core/vtkRenderer.cxx#L1138-1157

Interestingly, it used to be possible (or bugged) to choose wether sphere is fit horizontally or vertically:
https://github.com/Kitware/vtk-js/commit/b9956c73cd51947edcd36234c03818802209df9a

Best is to ask original author: @martinken WDYT ?

floryst:
We can add an option to `resetCamera()` to fit vertically or horizontally.

mix3d:
[...]

PR #3371, opened 2025-11-30
feat: add resetCameraScreenSpace method with offsetRatio input, so actors fully fit the screen

aerogt3:
add resetCameraScreenSpace method with offsetRatio input, essentially a resetCamera call that fills the entire screen space (where 90% offsetRatio means visible actors will fill 90% of the screen space)

#1285

### Context
#1285 

### Results
See image here:
https://github.com/Kitware/vtk-js/issues/1285#issuecomment-3510396233

### Changes
Duplicated the resetCamera method, with logic to ensure visible actors fill a percentage of the screen (default: 90%)

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - master
  - ubuntu 2240
  - firefox 144.0.2 & chrome 142.0

### DIFF
--- a/Sources/Rendering/Core/Renderer/index.js
+++ b/Sources/Rendering/Core/Renderer/index.js
@@ -464,6 +464,149 @@ function vtkRenderer(publicAPI, model) {
     return true;
   };
 
+  // Port of VTK C++ vtkRenderer::ZoomToBoxUsingViewAngle.
+  publicAPI.zoomToBoxUsingViewAngle = (box, ratioOrOffsetRatio = 1.0) => {
+    let view = null;
+    if (model._renderWindow && model._renderWindow.getViews) {
+      const views = model._renderWindow.getViews();
+      if (views.length > 0) {
+        view = views[0];
+      }
+    }
+
+    if (!view || !view.getViewportSize) {
+      return;
+    }
+
+    const size = view.getViewportSize(publicAPI);
+    if (!size || size[0] <= 0 || size[1] <= 0) {
+      return;
+    }
+
+    const zf1 = size[0] / box.width;
+    const zf2 = size[1] / box.height;
+    const zoomFactor = Math.min(zf1, zf2);
+    publicAPI.getActiveCamera().zoom(zoomFactor * ratioOrOffsetRatio);
+  };
+
+  // Port of VTK C++ vtkRenderer::ResetCameraScreenSpace.
+  // Uses a screen-space bounding box to zoom closer to the data.
+  publicAPI.resetCameraScreenSpace = (bounds = null, offsetRatio = 0.9) => {
+    let effectiveBounds = bounds;
+    let effectiveOffsetRatio = offsetRatio;
+    if (typeof bounds === 'number') {
+      effectiveOffsetRatio = bounds;
+      effectiveBounds = null;
+    }
+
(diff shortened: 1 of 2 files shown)


Sources/Rendering/Core/Renderer/index.js
```diff
@@ -464,6 +464,149 @@ function vtkRenderer(publicAPI, model) {
...
+      -center[2]
+    );
+
+    // Here to let parallel/distributed compositing intercept
+    // and do the right thing.
+    publicAPI.invokeEvent(RESET_CAMERA_EVENT);
+
+    return true;
+  };
+
```
finetjul, line 634: why not recoding the [VTK C++ implementation](https://gitlab.kitware.com/vtk/vtk/-/blob/master/Rendering/Core/vtkRenderer.cxx?ref_type=heads#L1306) ? 

This ensures you have the same behavior between VTK C++ and VTK.js. Which is important.

aerogt3: 1.) don’t have C++ knowledge
2.) never used VTK C++ (just vtk.js & paraview) Can’t find out how to use/compare resetCameraScreenSpace anywhere in paraview. Either way, #1 is the end of the road.

I think this contribution is better than having nothing, either with the name cameraResetScreenSpace, or maybe cameraZoomExtents. If someone with C++ knowledge can improve it later on (or take this solution to C++ if it’s better) that would be great.

aerogt3: I think this can be approved now?

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -464,144 +464,108 @@ function vtkRenderer(publicAPI, model) {
...
+          );
+          const dx = nd[0] * size[0];
+          const dy = nd[1] * size[1];
+          xmin = Math.min(dx, xmin);
+          xmax = Math.max(dx, xmax);
+          ymin = Math.min(dy, ymin);
+          ymax = Math.max(dy, ymax);
+        }
+      }
     }
```
finetjul, line 624: You can use `vtkBoundingBox.transformBounds(boundsToUse, publicAPI.getActiveCamera().getModelTransformMatrix, boundsToUse)`

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -464,144 +464,108 @@ function vtkRenderer(publicAPI, model) {
...
+    ymax += Math.max(yDiff, 0);
 
-    // update reasonable world to physical values
-    model.activeCamera.setPhysicalScale(radius);
-    model.activeCamera.setPhysicalTranslation(
-      -center[0],
-      -center[1],
-      -center[2]
-    );
+    // ZoomToBoxUsingViewAngle
```
finetjul: to move in its own method

finetjul (changes requested):
Thanks ! That's much easier to maintain.

Sources/Rendering/Core/Renderer/index.js
aerogt3: done I think

aerogt3: The latest changes (including the order of the functions) should better match VTK C++

PR #3384, opened 2026-01-19
feat(actor.objectName): adding objectName as an actor property

GeorgeFuller:
### Context
There is currently no easy way to identify an actor when it is picked out of context of it's pipeline.

### Results
GetObjectName and SetObjectName can now be called on actors.

### Changes
Added objectName to the default values of actor along with a macro for get and set. This provides a feature that the c++ vtk library already has and matches its naming convention.

### Testing
Tested within the context it was required, the CellPicker example, the example now console logs the name applied to the cone source when it is clicked.

### DIFF
--- a/Sources/Rendering/Core/Actor/index.js
+++ b/Sources/Rendering/Core/Actor/index.js
@@ -101,7 +101,6 @@ const DEFAULT_VALUES = {
 
   forceOpaque: false,
   forceTranslucent: false,
-  objectName: '',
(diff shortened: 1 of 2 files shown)


jourdain:
Let's close it for now. If people want that API on vtkObject, we will add it in the macro. But otherwise, vtk.js flexibility let you add anything that don't conflict with the object, so many users should be good.

PR #3389, opened 2026-01-27
docs(Examples): drop parallel-webpack in favor of using webpack directly

daker:
### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


finetjul:
Will it time out ? How about your vitepress PR ? could it be used instead ?

PR #3400, opened 2026-02-04
fix(SplineWidget): fix bug where SplineWidget handles cannot be dragged after creation

TreatTrick:
### Context

1. Execute command: Run `npm run example SplineWidget`.
2. Open the example in your browser, click "Place Widget" in the top-right corner, and add several points. Press Enter to confirm and create the SplineWidget.
3. Hover the mouse over any handle (control point) of the SplineWidget and attempt to drag it using the left mouse button.

### Results

The handles are unresponsive; the control points cannot be dragged or moved.

### Changes

Root Cause Analysis: The issue occurs because the handleMouseMove method in SplineWidget does not return macro.EVENT_ABORT when isHandleMoving is true. Even though the widget is currently in a dragging state, the mouse event continues to propagate to other subscribers. Consequently, the WidgetManager captures the event and triggers the w.deactivateAllHandles() method, which deactivates the currently selected handle and interrupts the drag operation.

Proposed Fix: To resolve this, we must return macro.EVENT_ABORT within the handleMouseMove logic whenever isHandleMoving is true. This ensures that the event is consumed by the active widget and prevented from propagating to the WidgetManager or other listeners.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  latest master
  - **OS**:  Windows 11
  - **Browser**:  Chrome: 144.0.7559.112

### DIFF
--- a/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/SplineWidget/behavior.js
@@ -346,6 +346,7 @@ export default function widgetBehavior(publicAPI, model) {
       if (model.freeHand && model.activeState === model.moveHandle) {
         addPoint();
       }
+      return macro.EVENT_ABORT;
     }
 
     return model.hasFocus ? macro.EVENT_ABORT : macro.VOID;


finetjul (approved):
You wouldn't need it if the widget had the focus while dragging the control point. However it seems that other widgets don't do that.
Let's merge your change but let's keep in mind that widget should probably grab focus in the future.

PR #3420, opened 2026-02-20
Fix/chorded button event ordering

sudomakeinstall:
### Context

In native VTK, mouse button events are registered independently of each other; i.e., if you press the left button, then press the right, then release the left, then release the right, you would get:

- `LeftButtonPress`
- `RightButtonPress`
- `LeftButtonRelease`
- `RightButtonRelease`

However, currently in `vtk.js`, if you went through the same sequence, you would only register the events which occur when no other button is pressed:

- `LeftButtonPress`
- `RightButtonRelease`

This is undesirable because:

- It makes it difficult to capture multi-button events (such as right-left click together).
- It may leave the interactor in an unexpected mode (for example, in the above sequence, it would be left in the "Left Down" mode since the `LeftButtonRelease` was never registered). Although this PR provides an example which shows the detected mouse events, note that you can currently demonstrate this behavior in any example with LeftDown, RightDown, LeftUp, RightUp, which will leave you with no buttons depressed but the object still rotating as you move the mouse.
- It is a change from the native VTK behavior.

The underlying reason for this seems to be the web specification for [Chorded Button Interactions](https://www.w3.org/TR/pointerevents/#chorded-button-interactions):

> Some pointer devices, such as mouse or pen, support multiple buttons. In the [[UIEVENTS](https://www.w3.org/TR/pointerevents/#bib-uievents)] Mouse Event model, each button press produces a mousedown and mouseup event. To better abstract this hardware difference and simplify cross-device input authoring, Pointer Events do not fire overlapping [pointerdown](https://www.w3.org/TR/pointerevents/#dfn-pointerdown) and [pointerup](https://www.w3.org/TR/pointerevents/#dfn-pointerup) events for chorded button presses (depressing an additional button while another button on the pointer device is already depressed).
> 
> Instead, chorded button presses can be detected by inspecting changes to the button and buttons properties. The button and buttons properties are inherited from the [MouseEvent](https://www.w3.org/TR/pointerevents/#dom-mouseevent) interface, but with a change in semantics and values, as outlined in the following sections.

I believe that this is the underlying cause of an issue I previously opened in trame (here: https://github.com/Kitware/trame/issues/803) before I determined the source.

### Results

This PR adds new logic to `RenderWindowInteractor` which inspects `event.buttons` to detect and correctly fire chorded events. As far as I can tell, the behavior now matches native VTK.

### Changes

- New example: `Examples/Rendering/MouseEvents/index.js`
- New test: `Sources/Rendering/Core/RenderWindowInteractor/test/testChordedButtons.js`
- Modified: `Sources/Rendering/Core/RenderWindowInteractor/index.js`

- [ ] Documentation and TypeScript definitions were *not* updated, since in my mind this fix brings behavior in line with expectation, though happy to add additional information to the docs if requested.

### PR and Code Checklist

### Testing

- [X] This change adds a unit test
- [X] Tested environment:
  - **vtk.js version 34.16.3 (latest master)**
  - **OSx version 15.7.3**
  - **FireFox version 147.0.4 **

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -444,9 +444,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
         case 'mouse':
         default: {
           // buttons bitmask: 1=left, 4=middle, 2=right
-          // eslint-disable-next-line no-bitwise
           const buttonBitMap = [1, 4, 2];
-          // eslint-disable-next-line no-bitwise
           const buttonBit = buttonBitMap[event.button] ?? 0;
           const callData = {
             ...getModifierKeysFor(event),


Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -417,9 +425,36 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+            const isRight = (currentButtons & 2) !== 0; // eslint-disable-line no-bitwise
+            const isMiddle = (currentButtons & 4) !== 0; // eslint-disable-line no-bitwise
+            // Only fire for chorded buttons; the primary button (event.button)
+            // is handled by handleMouseUp below.
+            if (!isLeft && wasLeft && event.button !== 0)
+              publicAPI.leftButtonReleaseEvent(callData);
+            if (!isMiddle && wasMiddle && event.button !== 1)
+              publicAPI.middleButtonReleaseEvent(callData);
+            if (!isRight && wasRight && event.button !== 2)
+              publicAPI.rightButtonReleaseEvent(callData);
```
finetjul: what if a button was pressed in the meantime ?

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -417,9 +425,36 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+              deviceType: getDeviceTypeFor(event),
+            };
+            const wasLeft = (previousMouseButtons & 1) !== 0; // eslint-disable-line no-bitwise
+            const wasRight = (previousMouseButtons & 2) !== 0; // eslint-disable-line no-bitwise
+            const wasMiddle = (previousMouseButtons & 4) !== 0; // eslint-disable-line no-bitwise
+            const isLeft = (currentButtons & 1) !== 0; // eslint-disable-line no-bitwise
+            const isRight = (currentButtons & 2) !== 0; // eslint-disable-line no-bitwise
+            const isMiddle = (currentButtons & 4) !== 0; // eslint-disable-line no-bitwise
+            // Only fire for chorded buttons; the primary button (event.button)
+            // is handled by handleMouseUp below.
```
finetjul: any reason to keep handleMouseUp call ?

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -417,9 +442,29 @@ function vtkRenderWindowInteractor(publicAPI, model) {
           publicAPI.handleTouchEnd(event);
           break;
         case 'mouse':
-        default:
+        default: {
+          // buttons bitmask: 1=left, 4=middle, 2=right
+          // eslint-disable-next-line no-bitwise
```
finetjul: there is no bitwise in the next line

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -417,9 +442,29 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+          const buttonBitMap = [1, 4, 2];
+          // eslint-disable-next-line no-bitwise
```
finetjul: there is no bitwise in the next line

Examples/Rendering/MouseEvents/index.js
```diff
@@ -0,0 +1,210 @@
+import '@kitware/vtk.js/favicon';
```
finetjul, line 1: Please add a screenshot for the example gallery. (see https://github.com/Kitware/vtk-js/blob/master/CONTRIBUTING.md#updating-documentation)

PR #3432, opened 2026-02-26
feat(STLReader): use locator-based vertex merging

### LINKED ISSUE #3052: STLReader and ORJReader parsing obj file with incorrect number of points[Bug]
### Bug description

ORJREADER parses more points in the obj file than in the obj file, as well as in the c++parsed and meshlab software displayed；
The stl file also has a problem of a large number of duplicate points

### Steps to reproduce

[demo](https://stackblitz.com/edit/vitejs-vite-asp8lb?file=src%2FOBJReader.vue,src%2FSTLReader.vue&terminal=dev)

### Detailed Behavior

_No response_

### Expected Behavior

1.NumberOfPoints is the actual number of points, without a large number of duplicate points；
2. Expect a clean method to clear duplicate points

### Environment

- vtk.js version: 30.2.0
- Browsers: chrome
- OS:

daker:
### Context

fixes #3052

### Changes

Replace STLReader duplicate-vertex removal logic with locator-based merging.
Add unittests.
Improve the example.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


Sources/IO/Geometry/STLReader/index.js
```diff
@@ -406,6 +452,7 @@ function vtkSTLReader(publicAPI, model) {
 const DEFAULT_VALUES = {
   // baseURL: null,
   // dataAccessHelper: null,
+  locator: null,
```
finetjul: no need to declare null objects

Sources/IO/Geometry/STLReader/index.js
```diff
@@ -406,7 +452,9 @@ function vtkSTLReader(publicAPI, model) {
 const DEFAULT_VALUES = {
   // baseURL: null,
   // dataAccessHelper: null,
+  // locator: null,
   // url: null,
+  merging: false,
```
finetjul: You claimed in documentation that default is "true"? (true in VTK C++ by default)

daker: Do we want to enable it by default like in CPP?

finetjul: I think so, it's best to have the same behavior.

PR #3444, opened 2026-03-24
feat(ClipPolyData): add vtkClipPolyData

daker:
### Context

image

### Changes

- [x] new vtkClipPolyData filter
- [x] triangle and triangle-strip clipping support
- [x] new vtkIncrementalPointLocator

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


Sources/Common/DataModel/TriangleStrip/index.js
```diff
@@ -419,7 +419,54 @@ function vtkTriangleStrip(publicAPI, model) {
...
+      let id2 = i + 1;
+      let id3 = i + 2;
+
+      if (i % 2) {
+        id1 = i + 2;
+        id2 = i + 1;
+        id3 = i;
+      }
+
+      trianglePoints.setPoint(0, ...model.points.getPoint(id1));
```
finetjul: For performance reason, it's faster to provide an already allocated array:
```trianglePoints.setPoint(0, ...model.points.getPoint(id1, tmp));```

File: Sources/Filters/Core/index.js [hunk 2]
```diff
@@ -1,9 +1,11 @@
+import vtkClipPolyData from './ClipPolyData';
```
finetjul: please fix lexicographical order

Sources/Filters/Core/ClipPolyData/index.js
```diff
@@ -0,0 +1,794 @@
...
+   * Evaluate the clip implicit function once per input point.
+   *
+   * @param {vtkPoints} points
+   * @param {*} clipFunction
+   * @returns {Float32Array}
+   */
+  function evaluateImplicitFunction(points, clipFunction) {
+    const numPts = points.getNumberOfPoints();
+
+    const scalars = new Float32Array(numPts);
```
finetjul, line 57: shouldn't it depend on `model.outputPointsPrecision` ?

Sources/Filters/Core/ClipPolyData/index.js
```diff
@@ -0,0 +1,794 @@
...
+    locator.initPointInsertion(newPoints, input.getBounds());
+
+    output.getPointData().initialize();
+    output.getCellData().initialize();
+    output.getFieldData().passData(input.getFieldData());
+    clippedOutput.getPointData().initialize();
+    clippedOutput.getCellData().initialize();
+    clippedOutput.getFieldData().passData(input.getFieldData());
+
+    const triangleScalarsValues = new Float32Array(3);
```
finetjul, line 111: shouldn't it depend on the precision type ?

Sources/Filters/Core/ClipPolyData/index.js
```diff
@@ -0,0 +1,794 @@
...
+  if (!cellArray || cellArray.getNumberOfCells() === 0) {
+    return;
+  }
+
+  const values = cellArray.getData();
+  let offset = 0;
+  let cellId = 0;
+  while (offset < values.length) {
+    const npts = values[offset++];
+    const pointIds = values.subarray(offset, offset + npts);
```
finetjul: subarray calls are not very fast, you could simply pass values, offset and npts to the callback

Sources/Filters/Core/ClipPolyData/index.js
```diff
@@ -0,0 +1,794 @@
...
+   */
+  function createClipState(input, output, clippedOutput) {
+    const newPoints = vtkPoints.newInstance({
+      dataType:
+        model.outputPointsPrecision === DesiredOutputPrecision.SINGLE
+          ? VtkDataTypes.FLOAT
+          : VtkDataTypes.DOUBLE,
+    });
+
+    const locator = model.locator || vtkMergePoints.newInstance();
```
finetjul: shouldn't it call createDefaultLocator() instead ? (createDefaultLocator() would returned model.locator

Sources/Filters/Core/ClipPolyData/index.js
daker: it should not depend on model.outputPointsPrecision, because outputPointsPrecision controls output point-coordinate storage, not clip scalar temporaries. This is also consistent with the C++ reference using vtkFloatArray for temporary clip scalars in https://github.com/Kitware/VTK/blob/master/Filters/Core/vtkClipPolyData.cxx#L139

Sources/Filters/Core/ClipPolyData/index.js
daker: This is consistent with the C++ reference using vtkFloatArray for temporary clip scalars in https://github.com/Kitware/VTK/blob/master/Filters/Core/vtkClipPolyData.cxx#L139

Sources/Filters/Core/ClipPolyData/index.js
finetjul: fair enough :-)

PR #3447, opened 2026-03-24
fix(TextActor): add support for property change

### LINKED ISSUE #3445: [Bug] Can't change vtkTextActor color
### Bug description

Can't seem to change the color of vtkTextActor.

Doing the following does not change the text color, it's always black (full codesandbox linked below):

```js
const actor = vtkTextActor.newInstance();
actor.getProperty().setFontColor(255,255,255); // doctsring says 0-255 for RGB values
```

also tried `actor.getProperty().setColor()`, same behavior. Is there another way I should be setting the text color?

### Steps to reproduce

here's a modification of the textActor example with RGB inputs:

https://codesandbox.io/p/devbox/vtkjs-vector-text-forked-3dkx24?workspaceId=ws_7cQK1iKXMw8NPVoWmC7hHm 

### Detailed Behavior

_No response_

### Expected Behavior

Should be able to set the color of a `vtkTextActor`

### Environment

- vtk.js version: 35.4.0
- Browsers: firefox
- OS: macos

### LINKED ISSUE #3484: [Bug] vtkTextActor ignores newline characters (\n)
### Bug description

Using vtk.js v35.9.0, newline characters passed to `vtkTextActor.setInput()` are not rendered as line breaks. The `\n` character appears to be replaced by a space.

### Steps to reproduce

```
import vtkTextActor from '@kitware/vtk.js/Rendering/Core/TextActor';

...
const actor = vtkTextActor.newInstance();
....
actor.setDisplayPosition(..., ...)
actor.setInput("Line1\nLine2");
```

### Detailed Behavior

The text is rendered on a single line:

`Line1 Line2`

This suggests that the newline character is ignored or converted into a space.

### Expected Behavior

The text should be rendered on two separate lines:

```
Line1
Line2
```

### Environment

- vtk.js version: v35.9.0
- Browsers: Chrome
- OS: Mac

daker:
### Context

fixes : #3445
fixes : #3484

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/TextActor/index.js
+++ b/Sources/Rendering/Core/TextActor/index.js
@@ -14,6 +14,8 @@ function vtkTextActor(publicAPI, model) {
   // Set our className
   model.classHierarchy.push('vtkTextActor');
 
+  const superClass = { ...publicAPI };
+
   publicAPI.makeProperty = vtkTextProperty.newInstance;
 
   const texture = vtkTexture.newInstance({
@@ -38,7 +40,7 @@ function vtkTextActor(publicAPI, model) {
     const backgroundColor = publicAPI.getProperty().getBackgroundColor();
 
     const dpr = Math.max(window.devicePixelRatio || 1, 1);
-    const ctx = canvas.getContext('2d');
+    const ctx = canvas.getContext('2d', { willReadFrequently: true });
 
     // Set the text properties to measure
     const textSize = fontSizeScale(resolution) * dpr;
@@ -110,21 +112,39 @@ function vtkTextActor(publicAPI, model) {
     return ImageHelper.canvasToImageData(canvas);
   }
 
+  function updateTexture() {
+    const image = createImageData(model.input);
+    texture.setInputData(image, 0);
+    model.textureBuildTime.modified();
+  }
+
+  function updateTextureIfNeeded() {
+    if (model.input === undefined) {
+      return;
+    }
+
+    if (
+      model.textureBuildTime.getMTime() < model.mtime ||
+      model.textureBuildTime.getMTime() < publicAPI.getProperty().getMTime()
+    ) {
+      updateTexture();
+    }
+  }
+
   mapper.setInputConnection(plane.getOutputPort());
 
   publicAPI.setMapper(mapper);
   publicAPI.addTexture(texture);
 
-  model._onInputChanged = (_publicAPI, _model, value) => {
-    const image = createImageData(value);
-    texture.setInputData(image, 0);
+  publicAPI.getTextures = () => {
+    updateTextureIfNeeded();
+    return superClass.getTextures();
   };
 }
 
 // Default property values
 const DEFAULT_VALUES = {
-  mapper: null,
-  property: null,
+  textureBuildTime: null,
 };
 
 export function extend(publicAPI, model, initialValues = {}) {
@@ -135,6 +155,8 @@ export function extend(publicAPI, model, initialValues = {}) {
 
   // Build VTK API
   macro.setGet(publicAPI, model, ['input']);
+  model.textureBuildTime = {};
+  macro.obj(model.textureBuildTime, { mtime: 0 });
 
   // Object methods
   vtkTextActor(publicAPI, model);
--- a/Sources/Rendering/Core/TextActor/index.js
+++ b/Sources/Rendering/Core/TextActor/index.js
@@ -28,6 +28,19 @@ function vtkTextActor(publicAPI, model) {
     yResolution: 1,
   });
 
+  /**
+   * Normalize escaped and platform specific line endings, then split into lines.
+   */
+  function splitLines(text) {
+    return text
+      .replace(/\\r\\n/g, '\n')
+      .replace(/\\n/g, '\n')
+      .replace(/\\r/g, '\n')
+      .replace(/\r\n/g, '\n')
+      .replace(/\r/g, '\n')
+      .split('\n');
+  }
+
   function createImageData(text) {
     const fontSizeScale = publicAPI.getProperty().getFontSizeScale();
     const fontStyle = publicAPI.getProperty().getFontStyle();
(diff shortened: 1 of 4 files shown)


Sources/Rendering/Core/TextActor/index.js
```diff
@@ -110,21 +114,66 @@ function vtkTextActor(publicAPI, model) {
...
+
+  model._onInputChanged = () => {
+    updateTexture();
   };
 }
 
 // Default property values
 const DEFAULT_VALUES = {
   mapper: null,
   property: null,
```
finetjul: mapper and property should not be re-declared here because they are already declared in vtkActor2D

Sources/Rendering/Core/TextActor/index.js
```diff
@@ -110,21 +114,66 @@ function vtkTextActor(publicAPI, model) {
...
   publicAPI.addTexture(texture);
 
-  model._onInputChanged = (_publicAPI, _model, value) => {
-    const image = createImageData(value);
-    texture.setInputData(image, 0);
+  publicAPI.getProperty = () => {
+    const property = superGetProperty();
+    if (model.property !== property || !model.propertySubscription) {
+      bindProperty(property);
+    }
```
finetjul: vtkActor2D.getProperty silently setting `model.property` is not clean.
Can you please fix vtkActor2D.getProperty instead ?

Sources/Rendering/Core/TextActor/index.js
```diff
@@ -110,21 +114,66 @@ function vtkTextActor(publicAPI, model) {
...
+  }
+
+  function bindProperty(property) {
+    if (model.propertySubscription) {
+      model.propertySubscription.unsubscribe();
+      model.propertySubscription = null;
+    }
+
+    if (property) {
+      model.propertySubscription = property.onModified(() => {
```
finetjul: vtk.js philosophy is not so much to "update" itself as a reaction to changes. Instead, it should wait to be rendered to update itself.

@sankhesh what event can trigger a texture refresh ?

Sources/Rendering/Core/TextActor/index.js
```diff
@@ -110,21 +114,66 @@ function vtkTextActor(publicAPI, model) {
...
+
+  publicAPI.delete = () => {
+    if (model.propertySubscription) {
+      model.propertySubscription.unsubscribe();
+      model.propertySubscription = null;
+    }
+    superDelete();
+  };
+
+  model._onInputChanged = () => {
```
jourdain: model._onInputChanged = updateTexture;

Sources/Rendering/Core/TextActor/index.js
```diff
@@ -110,21 +114,66 @@ function vtkTextActor(publicAPI, model) {
...
+    updateTexture();
   };
 }
 
 // Default property values
 const DEFAULT_VALUES = {
   mapper: null,
   property: null,
+  propertySubscription: null,
```
jourdain: should that be private?

jourdain: Just a `let propertySubscription` in the closure scope.

Sources/Rendering/Core/TextActor/index.js
```diff
@@ -110,22 +114,53 @@ function vtkTextActor(publicAPI, model) {
...
+
+  function bindProperty(property) {
+    if (propertySubscription) {
+      propertySubscription.unsubscribe();
+      propertySubscription = null;
+    }
+
+    if (property) {
+      propertySubscription = property.onModified(() => {
+        if (model.input !== undefined) {
```
finetjul: if input was valid and becomes invalid, it should probably clear the texture hence it must be updated.

finetjul: @sankhesh

sankhesh: Yes, if the input becomes invalid, we should release the texture.

finetjul: Sorry, I tagged you on the wrong comment, I wanted your input on this instead: https://github.com/Kitware/vtk-js/pull/3447#discussion_r2988901612

Sources/Rendering/Core/TextActor/index.js
sankhesh: You’re right that, in general, vtk prefers to update its internal structures lazily rather than reacting immediately to every change. The usual pattern is that objects refresh their internal representations when they are asked to render or otherwise execute, not when a property is modified.

In this case, the idea would be that when vtkTextActor::Render is called, it should check whether the text property’s MTime has changed since the last render. If it has, the actor can then rebuild or refresh its internal texture before rendering. That way we stay consistent with the vtk philosophy: properties can change freely, and the actual texture update only happens at render time when it’s really needed.

daker: i made a fix, let's hope i get it right this time

Sources/Rendering/Core/TextActor/index.js
```diff
@@ -110,21 +112,55 @@ function vtkTextActor(publicAPI, model) {
...
+    }
+    return changed;
+  };
+
+  publicAPI.getTextures = () => {
+    updateTextureIfNeeded();
+    return superClass.getTextures();
+  };
+
+  publicAPI.delete = () => superClass.delete();
```
finetjul: useless

finetjul (commented):
`model.textureUpdateTime` seems duplicated from `publicAPI.getMTime()`.
Please note that superclass.setXXX() will automatically update the mTime.

I let @sankhesh decide if the update should be done by  `getTextures()` or when the actor is being rendered.

Sources/Rendering/Core/TextActor/index.js
```diff
@@ -26,6 +28,13 @@ function vtkTextActor(publicAPI, model) {
     yResolution: 1,
   });
 
+  function normalizeLineBreaks(text) {
+    return text
+      .replace(/\\r\\n/g, '\n')
+      .replace(/\\n/g, '\n')
```
finetjul, line 37: why replacing \n with \n ?

Sources/Rendering/Core/TextActor/index.js
```diff
@@ -26,6 +28,13 @@ function vtkTextActor(publicAPI, model) {
     yResolution: 1,
   });
 
+  function normalizeLineBreaks(text) {
```
finetjul: please document what does it mean to normalize line breaks. 

(btw, I would have expect all the line breaks to be normalized to "\n" and not "\n" or "\r")

Sources/Rendering/Core/TextActor/index.js
daker: \n will be escpad in javascript and you will get "double\n", ex : i type in the text input not via the setInput "Hello World\nfrom vtk.js" you will get "Hello World double\n from vtk.js"

Sources/Rendering/Core/TextActor/index.js
daker: i have added a small explanation

Sources/Rendering/Core/TextActor/index.js
```diff
@@ -38,30 +52,48 @@ function vtkTextActor(publicAPI, model) {
...
 
     const dpr = Math.max(window.devicePixelRatio || 1, 1);
-    const ctx = canvas.getContext('2d');
+    const ctx = canvas.getContext('2d', { willReadFrequently: true });
 
     // Set the text properties to measure
-    const textSize = fontSizeScale(resolution) * dpr;
+    const normalizedText = normalizeLineBreaks(text);
+    const textSize = fontSizeScale(resolution);
+    const lines = normalizedText.split(/\r\n|\r|\n/);
```
finetjul: now that it is normalized, can't you simply split on \n only ?

daker: yes

finetjul: Maybe that mistake (wrong expected inputs) is a clue that `normalizeLineBreak` should also do the split of the lines (`normalizeLineBreak ` => `splitLines`)

finetjul (approved):
LGTM

daker:
@sankhesh what do you think ?

PR #3453, opened 2026-03-26
Shared WebGL context render window

PaulHax:
Are folks intersted in putting this in the vtk.js repo?  Its an uncommon use case.  We could get away with simply doing something to keep `vtkOpenGLRenderWindow` from resizing the canvas and put SharedRenderWindow in another repo/package.  

### Context
This PR adds an OpenGL shared-context rendering path for vtk.js so it can render into a `WebGL2RenderingContext` owned by another library, instead of always creating and managing its own canvas/context.

The immediate use case is interleaving vtk.js content with host renderers such as MapLibre custom layers, including depth-tested rendering into the host scene. vtk.js also needs an explicit way to opt out of canvas DOM management when the canvas is externally owned.

### Results
Before this change:
- `vtkOpenGLRenderWindow` assumed ownership of the canvas and could resize/restyle it.

After this change:
- vtk.js can render into a host-owned `WebGL2` context through `vtkSharedRenderWindow.createFromContext(canvas, gl)`.

### Changes

- Added `manageCanvas` to `vtkOpenGLRenderWindow` (default `true`) so externally owned canvases can opt out of vtk.js canvas sizing/styling.
- Added `vtkSharedRenderWindow` for rendering into an externally provided `WebGL2RenderingContext`.
- Added `vtkSharedRenderer` for the shared-context rendering path.
- Scoped the vtkSharedRenderer override to the shared render-window instance rather than changing the global OpenGL view-node factory. This avoids leaking shared-context behavior into unrelated normal OpenGL render windows.
- Added a MapLibre example showing shared-context rendering on terrain.

### PR and Code Checklist

### Testing
`npm run example -- SharedContext`

- [x] Tested environment:
  - **OS**: Linux
  - **Browser**: Chrome 

image


Sources/Rendering/OpenGL/ViewNodeFactory/index.js
```diff
@@ -27,12 +27,17 @@ const DEFAULT_VALUES = {};
...
+  // instance-local overrides such as vtkSharedRenderer.
+  model.overrides = Object.create(CLASS_MAPPING);
 
   // Inheritance
   vtkViewNodeFactory.extend(publicAPI, model, initialValues);
 
+  publicAPI.registerOverride = (className, fn) => {
+    model.overrides[className] = fn;
+  };
+
```
finetjul: method definition should be moved in `vtkOpenGLViewNodeFactory()`

finetjul (commented):
@sankhesh I do not have any opinion on the approach, I leave it to you.

Sources/Rendering/OpenGL/ViewNodeFactory/index.js
PaulHax: Ah yes, moved.

Sources/Rendering/OpenGL/SharedRenderWindow/index.js
```diff
@@ -0,0 +1,268 @@
...
+
+    const interactor = getInteractor();
+    if (renderCallback && interactor?.onRenderEvent) {
+      // Render requests flow through the interactor RenderEvent; redirect those
+      // to the host render loop while keeping draw calls inside renderShared().
+      if (savedEnableRender === null && interactor.getEnableRender) {
+        savedEnableRender = interactor.getEnableRender();
+      }
+      interactor?.setEnableRender?.(false);
+      bindRenderEvent(interactor);
```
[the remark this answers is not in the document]
PaulHax: Added a delete override to fix this.

Sources/Rendering/OpenGL/SharedRenderWindow/test/testSharedRenderWindowHostSurvival.js
```diff
@@ -0,0 +1,145 @@
...
+  gl.compileShader(vs);
+
+  const fs = gl.createShader(gl.FRAGMENT_SHADER);
+  gl.shaderSource(fs, FRAG_SRC);
+  gl.compileShader(fs);
+
+  const prog = gl.createProgram();
+  gl.attachShader(prog, vs);
+  gl.attachShader(prog, fs);
+  gl.linkProgram(prog);
```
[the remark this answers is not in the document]
PaulHax: fixed. Detach and delete both shaders after linkProgram.

Sources/Rendering/OpenGL/SharedRenderWindow/test/testSharedRenderWindowHostSurvival.js
```diff
@@ -0,0 +1,145 @@
...
+  return prog;
+}
+
+function createHostVAO(gl, prog) {
+  const verts = new Float32Array([-1, -1, 1, -1, -1, 1, -1, 1, 1, -1, 1, 1]);
+  const vao = gl.createVertexArray();
+  gl.bindVertexArray(vao);
+  const buf = gl.createBuffer();
+  gl.bindBuffer(gl.ARRAY_BUFFER, buf);
+  gl.bufferData(gl.ARRAY_BUFFER, verts, gl.STATIC_DRAW);
```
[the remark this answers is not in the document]
PaulHax: createHostVAO now returns { vao, buf }, and the test calls gl.deleteBuffer(hostBuf) alongside deleteVertexArray at teardown

Sources/Rendering/OpenGL/RenderWindow/index.js
sankhesh: Shouldn't this be done only iff `manageCanvas` is true?

Sources/Rendering/OpenGL/SharedRenderWindow/index.js
```diff
@@ -0,0 +1,300 @@
...
+function isWebGL2Context(gl) {
+  return (
+    typeof WebGL2RenderingContext !== 'undefined' &&
+    gl instanceof WebGL2RenderingContext
+  );
+}
+
+function getDefaultDrawBuffers(gl) {
+  const framebuffer = gl.getParameter(gl.FRAMEBUFFER_BINDING);
+  return framebuffer ? [gl.COLOR_ATTACHMENT0] : [gl.BACK];
```
sankhesh: This is an assumption here. If the external application has bound multiple FBO attachments, assuming attachment 0 might be wrong.

Sources/Rendering/OpenGL/SharedRenderer/index.js
```diff
@@ -0,0 +1,62 @@
...
+    const gl = model.context;
+    const openGLRenderWindow = model._openGLRenderWindow;
+
+    const autoClear = openGLRenderWindow?.getAutoClear?.() ?? true;
+    if (autoClear === false) {
+      const ts = publicAPI.getTiledSizeAndOrigin();
+      gl.enable(gl.SCISSOR_TEST);
+      gl.scissor(ts.lowerLeftU, ts.lowerLeftV, ts.usize, ts.vsize);
+      gl.viewport(ts.lowerLeftU, ts.lowerLeftV, ts.usize, ts.vsize);
+      gl.enable(gl.DEPTH_TEST);
```
sankhesh, line 17: Shouldn't the state be saved  and reset after VTK is done rendering?

Sources/Rendering/OpenGL/SharedRenderer/index.js
```diff
@@ -0,0 +1,62 @@
+import macro from 'vtk.js/Sources/macros';
+import { extend as extendOpenGLRenderer } from 'vtk.js/Sources/Rendering/OpenGL/Renderer';
+
+function vtkSharedRenderer(publicAPI, model) {
+  model.classHierarchy.push('vtkSharedRenderer');
+
+  publicAPI.clear = () => {
+    const gl = model.context;
```
sankhesh, line 8: Is it guaranteed that model.context will be valid always when we get here?

sankhesh (commented):
LGTM overall.

Sources/Rendering/OpenGL/RenderWindow/index.js
PaulHax: Good catch. Added the manageCanvas guard to match the one at line 164.

Sources/Rendering/OpenGL/SharedRenderWindow/index.js
PaulHax: Updated to query DRAW_BUFFER0 to gl.MAX_DRAW_BUFFERS and respect the host's actual draw-buffer state, with [COLOR_ATTACHMENT0] as a fallback when every slot is NONE.

Sources/Rendering/OpenGL/SharedRenderer/index.js
PaulHax: SharedRenderer is assuming a contract where host is responsible for rebinding its GL state after renderShared(). This is the same contract used by three.js (renderer.resetState()), Babylon (engine.wipeCaches()), and deck.gl/luma.gl.   A full state save/restore would add a lot of getParameter calls per render and create a heavier contract.

Sources/Rendering/OpenGL/SharedRenderer/index.js
PaulHax: Its guaranteed. Same assuption as parent vtkOpenGLRenderer.clear()

Sources/Rendering/OpenGL/SharedRenderer/index.js
sankhesh: Not really. You only need to query and cache the state parameters that are being force-changed by vtk. This is what we do, for example in [`vtkExternalOpenGLRenderWindow`](http://vtk.org/doc/nightly/html/classvtkExternalOpenGLRenderWindow.html) that follows the same idea.

PaulHax: What state should we cache (and restore?) I don’t understand the goal. I looked at vtkExternalOpenGLRenderWindow. It pushes framebuffer bindings and queries GL_VIEWPORT + GL_DRAW_BUFFER.  It doesn't snapshot general GL state. 

I see that C++ vtkOpenGLState does the comprehensive Reset + Push + Pop. This PR with vtkRestoringSharedRenderWindow (https://github.com/PaulHax/vtk-js/pull/14) snapshots and restores ~70 host params around renderShared().  Adds ~1.8 ms per render on my machine.  Happy to fold it into this PR, trim the set, or expand it if you'd like a different scope.

PaulHax:
closing in favor of https://github.com/Kitware/vtk-js/pull/3555

PR #3465, opened 2026-03-31
fix(imageResliceMapper): adjust texel size calculation for anisotropi…

sedghi:
### Context

  The label-outline thickness rendered by `vtkImageResliceMapper` appears inconsistent (stretched or compressed) when the input image has anisotropic spacing. This
  happens because the texel-size uniform used for neighbor sampling in the outline shader is computed from voxel dimensions alone, ignoring the physical spacing of
  each axis.

  ### Results

  Before: outline thickness varies per axis on anisotropic data — thicker along the axis with smaller spacing, thinner along the axis with larger spacing.

  After: outline thickness is visually uniform across all axes regardless of spacing differences.

CleanShot 2026-03-31 at 16 55 19@2x

  ### Changes

  - **`Sources/Rendering/OpenGL/ImageResliceMapper/index.js`**: The `texelSize` uniform computation now divides by `inputDims[i] * inputSpacing[i]` instead of just
  `inputDims[i]`, normalizing the texture-coordinate step by physical spacing so the outline kernel samples at consistent physical distances in all directions.
  - No API changes.
 
### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**: 

CC @PaulHax


sedghi:
~not sure if test failing in action my fault or not, since npm test runs fine locally~

Update: updated the screenshot for test

PaulHax:
Good fix.  

This changes outline thickness from voxel space to physical space.  The implication: small spacing produces thicker outlines, large spacing produces thinner ones for the same "outline thickness" paramter.  

Current change is already better methinks, but i'm also consitering:

Normalize by the finest axis (minSpacing / (dims * spacing)) — keeps the anisotropy fix but thickness stays ~1 voxel on the finest axis.

or

Compute texelSize from the view-projection matrix so outlines are a fixed pixel width regardless of spacing or zoom

sedghi:
@PaulHax please check it out

PaulHax:
Looks good, outline thickness operating in "voxel" space makes sense to me.

sedghi:
any chance we can merge this? @sankhesh

finetjul (approved):
LGTM

PR #3466, opened 2026-04-01
feat(CutterMapper): add vtkCutterMapper

### LINKED ISSUE #1967: Geometry cutter on the GPU
### Need

Display the contour of a 3d model when intersected with a plane, to display on top of an image data slice for example.

### Feature

- **suggestion**: create a mapper that can take a `vtkPolyData` and a plane equation (or arbitrary equation) to compute its intersection on the GPU
- **alternative**: can currently do it on the CPU with `vtkCutter` for each slice, but that's slow

daker:
### Context

fixes #1967

### Results

https://github.com/user-attachments/assets/6a7f92bd-6be2-49b9-ad2a-b2e1a5d36d92

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


daker:
@sedghi @finetjul @sankhesh ready for review

finetjul (commented):
Looks cool.

I'm being curious here: could it be made in a way that it can also cut the volume rendering ? Could there be shared parts to accept/reject fragments ?

@sankhesh I let you code review more closely.

Sources/Rendering/Core/CutterMapper/index.js
```diff
@@ -0,0 +1,76 @@
...
+  cutWidth: 1.5,
+};
+
+// ----------------------------------------------------------------------------
+
+export function extend(publicAPI, model, initialValues = {}) {
+  Object.assign(model, DEFAULT_VALUES, initialValues);
+
+  vtkMapper.extend(publicAPI, model, initialValues);
+  publicAPI.setScalarVisibility(false);
```
sankhesh: This is not explicitly required, is it? i.e. the mapper can still support scalar coloring. If not, it should be documented somewhere.

Sources/Rendering/OpenGL/CutterMapper/index.js
```diff
@@ -0,0 +1,358 @@
...
+  const superClass = { ...publicAPI };
+
+  publicAPI.replaceShaderClip = (shaders, ren, actor) => {
+    let VSSource = shaders.Vertex;
+    let FSSource = shaders.Fragment;
+    const functionName = model.renderable.getSupportedImplicitFunctionName();
+
+    if (!functionName) {
+      vtkErrorMacro(
+        `Unsupported implicit function ${functionName} for cutting.`
```
sankhesh:
```suggestion
        `No cut function set.`
```

Sources/Rendering/OpenGL/CutterMapper/index.js
```diff
@@ -0,0 +1,358 @@
...
+          '    if (cutBoxDist > 0.0) {',
+          '      cutBoxInside = false;',
+          '    }',
+          '  }',
+          '  if (cutBoxDist > 0.0) {',
+          '    cutBoxDistance += cutBoxDist * cutBoxDist;',
+          '  }',
+          '}',
+          'cutDistanceVSOutput = cutBoxInside ? cutBoxMinDistance : sqrt(cutBoxDistance);',
+        ];
```
sankhesh, line 150: cutBoxDist is set to a negative value when the point is inside the box (distance to a face is negative from the interior). cutBoxMinDistance accumulates the maximum of these negatives, giving the closest face distance from inside — this is fine.

However: when cutBoxLength == 0.0 for an axis (degenerate flat box), neither the <= 0.5 nor the > 0.5 branch is taken, so cutBoxMinDistance is never updated for that axis. If all three axes are degenerate, cutBoxMinDistance remains -1.0e20 and cutBoxInside may still be true, resulting in cutDistanceVSOutput = -1.0e20. This would discard the entire surface rather than displaying a cut. Guard against a degenerate box or initialize cutBoxMinDistance per-axis rather than globally.

Sources/Rendering/OpenGL/CutterMapper/index.js
```diff
@@ -0,0 +1,358 @@
...
+      case 'vtkCylinder':
+        cutDistanceDec = [
+          'uniform vec3 cutCylinderCenter;',
+          'uniform vec3 cutCylinderAxis;',
+          'uniform float cutCylinderRadius;',
+        ];
+        cutDistanceImpl = [
+          'vec3 cutCylinderDelta = cutFunctionPoint - cutCylinderCenter;',
+          'float cutCylinderProjection = dot(cutCylinderAxis, cutCylinderDelta);',
+          'cutDistanceVSOutput = dot(cutCylinderDelta, cutCylinderDelta) - cutCylinderProjection * cutCylinderProjection - cutCylinderRadius * cutCylinderRadius;',
```
sankhesh: The formula |delta|² - (delta·axis)² - r² is only geometrically correct when axis is a **unit vector**. Normalize `cutCylinderAxis`

Sources/Rendering/WebGPU/CutterMapper/index.js
```diff
@@ -0,0 +1,375 @@
...
+
+    const renderable = model.renderable;
+    const cutFunction = renderable.getCutFunction();
+    const functionName = renderable.getSupportedImplicitFunctionName();
+    if (!functionName || !cutFunction) {
+      return;
+    }
+
+    const bufferShift = model.WebGPUActor.getBufferShift(model.WebGPURenderer);
+    model.UBO.setValue('CutWidth', vtkMath.max(renderable.getCutWidth(), 0.0));
```
sankhesh:
```suggestion
    model.UBO.setValue('CutWidth', Math.max(renderable.getCutWidth(), 0.0));
```

Sources/Rendering/WebGPU/CutterMapper/index.js
```diff
@@ -0,0 +1,375 @@
...
+        apex[1] = 0.0;
+        apex[2] = 0.0;
+        axis[0] = 1.0;
+        axis[1] = 0.0;
+        axis[2] = 0.0;
+        axisPoint[0] = 1.0;
+        axisPoint[1] = 0.0;
+        axisPoint[2] = 0.0;
+
+        if (inverseTransform?.transformPoint) {
```
sankhesh, line 296: Should there be a warning macro here for when the inverse transform is null?

Sources/Rendering/Core/CutterMapper/index.js
daker: it's not a requirement, i did enable it with explanation.

sankhesh (approved):
LGTM

PR #3472, opened 2026-04-05
feat(DataArray): add preserveTypedArrays option to getState()

PaulHax:
### Context

`vtkDataArray.getState()` calls `Array.from()` on typed array values, which creates a plain JS Array with Numbers. For large arrays this causes an out-of-memory crash (RangeError: Invalid array length).

Real-world case: [Kitware/VolView#852](https://github.com/Kitware/VolView/issues/852) — saving a session with a 1024×1024×256 labelmap (268M Uint8 elements) crashes because `Array.from()` tries to allocate ~1.6GB of heap for the boxed Number array.

The `Array.from()` conversion in the base `getState()` was added in [#2927](https://github.com/Kitware/vtk-js/pull/2927) to fix JSON serialization of `Float64Array` properties like `direction` ([Kitware/glance#480](https://github.com/Kitware/glance/issues/480)).

Related: see [this Discourse thread on web workers](https://discourse.vtk.org/t/vtk-and-web-workers/5536) and [this thread on REST API serialization](https://discourse.vtk.org/t/vtk-js-how-to-serialize-and-deserialize-vtkdataset-for-rest-apis/6634). The `preserveTypedArrays` option formalizes that pattern in the existing API.

If this gets merged, should open another breaking change PR to remove `getStateArrayMapFunc` from `macros.js`.

### Results

**Before:** No built-in way to serialize DataArrays without copying TypedArrays to plain Arrays.
**After:** `getState({ preserveTypedArrays: true })` preserves TypedArray values without converting and copying. Default `getState()` behavior is unchanged.

```js
// Default — JSON-safe, same as before
const jsonState = dataset.getState();
JSON.stringify(jsonState); // works

// New option — efficient for structured clone / postMessage
const state = dataset.getState({ preserveTypedArrays: true });
worker.postMessage(state);
```

### Changes

- **`macros.js`**: Added `preserveTypedArrays` option to the base `getState()`. When true, skips `Array.from()` on TypedArrays and propagates the option to nested objects.
- **`DataArray/index.js`**: Updated `getState()` to conditionally skip `Array.from()` on values when `preserveTypedArrays` is true.
- **`FieldData.js`**: Updated `getState()` to forward options to nested DataArrays.
- **`interfaces.d.ts`**: Added `GetStateOptions` type and updated `getState()` signature.
- **`DataArray/index.d.ts`**: Updated `getState()` signature.
- **`testDataArray.js`**: Added tests for `preserveTypedArrays` option.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **OS**: Linux
  - **Browser**: Chrome Headless 138.0.0.0

### DIFF
--- a/Sources/macros.js
+++ b/Sources/macros.js
@@ -368,10 +368,11 @@ export function obj(publicAPI = {}, model = {}) {
   };
 
   // Add serialization support
-  publicAPI.getState = () => {
+  publicAPI.getState = ({ preserveTypedArrays = false } = {}) => {
     if (model.deleted) {
       return null;
     }
+    const options = { preserveTypedArrays };
     const jsonArchive = { ...model, vtkClass: publicAPI.getClassName() };
 
     // Convert every vtkObject to its serializable form
@@ -383,11 +384,16 @@ export function obj(publicAPI = {}, model = {}) {
       ) {
         delete jsonArchive[keyName];
       } else if (jsonArchive[keyName].isA) {
-        jsonArchive[keyName] = jsonArchive[keyName].getState();
+        jsonArchive[keyName] = jsonArchive[keyName].getState(options);
       } else if (Array.isArray(jsonArchive[keyName])) {
-        jsonArchive[keyName] = jsonArchive[keyName].map(getStateArrayMapFunc);
+        jsonArchive[keyName] = jsonArchive[keyName].map((item) =>
+          item && item.isA ? item.getState(options) : item
+        );
       } else if (isTypedArray(jsonArchive[keyName])) {
-        jsonArchive[keyName] = Array.from(jsonArchive[keyName]);
+        if (!preserveTypedArrays) {
+          jsonArchive[keyName] = Array.from(jsonArchive[keyName]);
+        }
+        // else: keep TypedArray as-is for structured clone / postMessage
       }
     });
 
--- a/Sources/Common/Core/DataArray/index.d.ts
+++ b/Sources/Common/Core/DataArray/index.d.ts
@@ -1,4 +1,4 @@
-import { vtkObject, vtkRange } from '../../../interfaces';
+import { vtkObject, vtkRange, GetStateOptions } from '../../../interfaces';
 import { float, int, Nullable, Range, TypedArray } from '../../../types';
 
 /**
@@ -274,9 +274,12 @@ export interface vtkDataArray extends vtkObject {
 
   /**
    * Get the state of this array.
+   *
+   * Pass `{ preserveTypedArrays: true }` to keep TypedArray values
+   * without converting and copying to a plain Array.
    * @returns {object}
    */
-  getState(): object;
+  getState(options?: GetStateOptions): object;
 
   /**
    * Deep copy of another vtkDataArray into this one.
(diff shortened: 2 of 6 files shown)


Sources/macros.js
```diff
@@ -383,10 +384,12 @@ export function obj(publicAPI = {}, model = {}) {
...
       } else if (jsonArchive[keyName].isA) {
-        jsonArchive[keyName] = jsonArchive[keyName].getState();
+        jsonArchive[keyName] = jsonArchive[keyName].getState(options);
       } else if (Array.isArray(jsonArchive[keyName])) {
-        jsonArchive[keyName] = jsonArchive[keyName].map(getStateArrayMapFunc);
-      } else if (isTypedArray(jsonArchive[keyName])) {
+        jsonArchive[keyName] = jsonArchive[keyName].map((item) =>
+          item && item.isA ? item.getState(options) : item
+        );
+      } else if (!preserveTypedArrays && isTypedArray(jsonArchive[keyName])) {
```
finetjul: what if isTypedArray and preserveTypedArrays are both true ? maybe add a comment as "else" to emphasize to leave it as-is

Sources/macros.js
```diff
@@ -383,10 +384,12 @@ export function obj(publicAPI = {}, model = {}) {
       ) {
         delete jsonArchive[keyName];
       } else if (jsonArchive[keyName].isA) {
-        jsonArchive[keyName] = jsonArchive[keyName].getState();
+        jsonArchive[keyName] = jsonArchive[keyName].getState(options);
       } else if (Array.isArray(jsonArchive[keyName])) {
-        jsonArchive[keyName] = jsonArchive[keyName].map(getStateArrayMapFunc);
-      } else if (isTypedArray(jsonArchive[keyName])) {
+        jsonArchive[keyName] = jsonArchive[keyName].map((item) =>
```
finetjul, line 389: should `getStateArrayMapFunc` be deleted ?

finetjul: alternatively, use `bind()` ?

Sources/macros.js
PaulHax: Changed to 
```
      } else if (isTypedArray(jsonArchive[keyName])) {
        if (!preserveTypedArrays) {
          jsonArchive[keyName] = Array.from(jsonArchive[keyName]);
        }
        // else: keep TypedArray as-is for structured clone / postMessage
      }
```

Sources/macros.js
PaulHax: > should getStateArrayMapFunc be deleted ?

Yes, suggest in follow up for major version release as `getStateArrayMapFunc` is exported

> alternatively, use bind() ?

Currying, I like your style.  `bind` is a bit akward here as this call would be something like  `getStateArrayMapFunc.bind(null, options)`

If you think the indirection is worth it, perhaps better to simply define a local fuction?
```
  publicAPI.getState = ({ preserveTypedArrays = false } = {}) => {
    const mapStateItem = (item) => (item && item.isA ? item.getState(options) : item);
    ...
       jsonArchive[keyName].map(mapStateItem )
```

finetjul (approved):
LGTM

PaulHax:
@finetjul  Good to merge?

PR #3481, opened 2026-04-12
fix(ImageMapper): check cardinal neighbors in label outline border detection

### LINKED ISSUE #3480: [Bug] WebGL ImageMapper label outline skips cardinal direction neighbor checks
### Bug description

While working on the WebGPU ImageMapper label outline, I came across something that I believe may be a bug.

In WebGL ImageMapper the label outline border detection loop uses `if (i == 0 || j == 0) { continue; }`

This skips all pixels on the horizontal and vertical axes through the center (the entire cross/plus pattern). Only diagonal neighbors are checked. The correct condition should be `if (i == 0 && j == 0) { continue; }` like in ImageResliceMapper

## Correct
https://github.com/Kitware/vtk-js/blob/920f0ce148ebd61c9680f140ca966c2634127fb4/Sources/Rendering/OpenGL/ImageResliceMapper/index.js#L1177

## Incorrect
https://github.com/Kitware/vtk-js/blob/920f0ce148ebd61c9680f140ca966c2634127fb4/Sources/Rendering/OpenGL/ImageMapper/index.js#L392

https://github.com/Kitware/vtk-js/blob/920f0ce148ebd61c9680f140ca966c2634127fb4/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl#L949

### Steps to reproduce

TBD

### Detailed Behavior

TBD

### Expected Behavior

TBD

### Environment

- vtk.js version:
- Browsers:
- OS:

PaulHax:
The label outline border detection loop used `if (i == 0 || j == 0)` which skips all cardinal-direction neighbors (up/down/left/right), only checking diagonals. Changed to `if (i == 0 && j == 0)` to skip only the center pixel.

Fixes Kitware/vtk-js#3480

### DIFF
--- a/Sources/Rendering/Core/ImageMapper/test/testLabelOutlineCardinal.js
+++ b/Sources/Rendering/Core/ImageMapper/test/testLabelOutlineCardinal.js
@@ -1,0 +1,173 @@
+import test from 'tape';
+import 'vtk.js/Sources/Rendering/Misc/RenderingAPIs';
+
+import vtkImageMapper from 'vtk.js/Sources/Rendering/Core/ImageMapper';
+import vtkImageSlice from 'vtk.js/Sources/Rendering/Core/ImageSlice';
+import vtkRenderer from 'vtk.js/Sources/Rendering/Core/Renderer';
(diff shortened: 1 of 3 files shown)


daker (approved):
LGTM

PR #3490, opened 2026-04-28
Guard volume mapper when shader program is unavailable

nutzlastfan:
## Summary
- make OpenGL/VolumeMapper.updateShaders return false when shader program creation or binding returns no program
- skip mapper shader parameter binding when no program is available
- stop renderPieceDraw before drawing and deactivate textures when shader setup fails

## Context
ShaderCache.readyShaderProgramArray(...) and readyShaderProgram(...) can return null when shader compilation or binding fails.

VolumeMapper currently continues into setMapperShaderParameters, where program.isAttributeUsed(...) can throw because cellBO.getProgram() is null.

This keeps the original shader compile failure as the root error while preventing the follow-up null dereference and avoiding a draw with an invalid shader program.

## Test
- Syntax checked Sources/Rendering/OpenGL/VolumeMapper/index.js as an ES module with Node 18:
  node --input-type=module --check

I did not run the full vtk-js test suite locally.

### DIFF
--- a/Sources/Rendering/OpenGL/VolumeMapper/index.js
+++ b/Sources/Rendering/OpenGL/VolumeMapper/index.js
@@ -507,6 +507,12 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
           shaders.Geometry
         );
 
+      if (!newShader) {
+        vtkErrorMacro('Error compiling volume mapper shader program.');
+        cellBO.setProgram(null);
+        return false;
+      }
+
       // if the shader changed reinitialize the VAO
       if (newShader !== cellBO.getProgram()) {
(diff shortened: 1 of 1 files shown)


finetjul (changes requested):
LGTM but please reword your commit title to follow semantic-release pattern. you can use Commitizen to create your commit. See CONTRIBUTING.md

nutzlastfan:
Thanks, updated the commit title to follow the semantic-release pattern:

fix(VolumeMapper): guard unavailable shader program

Force-pushed the amended single commit branch at f598d30e.

finetjul (approved):
LGTM

PR #3493, opened 2026-04-29
fix(glyph): clear stale source shift scale

PaulHax:
Bug fix.  Clear stale coordinate shift/scale state on glyph source VBOs when `vtkGlyph3DMapper` no longer needs shift/scale normalization for glyph centers.

### Details

`vtkGlyph3DMapper` applies shift/scale normalization to glyph center matrices for large coordinate ranges. It also applies the same transform to glyph source primitives after the base `PolyDataMapper` builds source VBOs.

Previously, when a later frame no longer required center shift/scale, the mapper stopped applying the transform to center matrices but did not clear the previous transform from the source primitive CABOs. That left glyph centers and glyph source geometry using different coordinate transforms.

This updates `Glyph3DMapper` to always set the source primitive shift/scale state: pass the computed shift/scale when active, otherwise pass `null, null` to clear stale state.

### DIFF
--- a/Sources/Rendering/Core/Glyph3DMapper/test/testGlyph3DMapperShiftScale.js
+++ b/Sources/Rendering/Core/Glyph3DMapper/test/testGlyph3DMapperShiftScale.js
@@ -1,0 +1,98 @@
+import test from 'tape';
+import testUtils from 'vtk.js/Sources/Testing/testUtils';
+
+import 'vtk.js/Sources/Rendering/Misc/RenderingAPIs';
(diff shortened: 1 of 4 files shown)


finetjul (approved):
LGTM but @jadh4v may know more than me about shift/scale.

finetjul (commented):
Can you also do it in SphereMapper ?

PaulHax:
Sphere mapper updated.

finetjul (approved):
LGTM

PR #3500, opened 2026-05-05
perf(GLTFImporter): improve memory usage

daker:
### Context

Open "ABeautifulGame" then check tab memory usage 
https://kitware.github.io/vtk-js/examples/GLTFImporter/index.html?model=ABeautifulGame&viewAPI=WebGPU

### Results

74% reduction of memory usage and 64% faster in rendering

Top tab: with PR
Bottom tab: main branch

image

### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/IO/Geometry/GLTFImporter/index.js
+++ b/Sources/IO/Geometry/GLTFImporter/index.js
@@ -13,6 +13,7 @@ import {
 import parseGLB from 'vtk.js/Sources/IO/Geometry/GLTFImporter/Decoder';
 import { createAnimationMixer } from 'vtk.js/Sources/IO/Geometry/GLTFImporter/Animations';
 import { BINARY_HEADER_MAGIC } from 'vtk.js/Sources/IO/Geometry/GLTFImporter/Constants';
+import { clearImageCaches } from 'vtk.js/Sources/IO/Geometry/GLTFImporter/Utils';
 
 const { vtkDebugMacro, vtkErrorMacro } = macro;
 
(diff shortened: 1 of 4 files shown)


Sources/IO/Geometry/GLTFImporter/index.js
```diff
@@ -221,6 +222,44 @@ function vtkGLTFImporter(publicAPI, model) {
 
     await Promise.all(promises);
   };
+
+  publicAPI.releaseGraphicsResources = () => {
+    if (model.renderer) {
+      model.actors?.forEach((actor) => {
+        model.renderer.removeActor(actor);
```
finetjul: releasing graphic resources should call releaseGraphicsResources on the actors, not delete the actors.

It seems that what you want to do is to "reset"/"clear" the importer insted.

daker: yes it's more resetting the importer in case you want to reuse it to import another model, should i just do :

```diff
model.actors?.forEach((actor) => {
+        actor.releaseGraphicsResources?.();
         model.renderer.removeActor(actor);
-        actor.delete?.();
       });
       model.lights?.forEach((light) => {
+        light.releaseGraphicsResources?.();
         model.renderer.removeLight(light);
-        light.delete?.();
       });
```

daker: hmm releaseGraphicsResources is not possible on actors, so my question is if we delete the importer should we also delete all the actors that this importer created ?

daker: i did use `clear` and removed the actors/lights deletion

finetjul (approved):
LGTM

Issue #3216, opened 2025-02-12
Failed to set up webGPU backend.


MountainAndMorning:
I try to use webGPU backend:

```js
import '@kitware/vtk.js/favicon';

// Load the rendering pieces we want to use (for both WebGL and WebGPU)
import '@kitware/vtk.js/Rendering/Profiles/Volume';

import Constants from '@kitware/vtk.js/Rendering/Core/ImageMapper/Constants';
import vtkGenericRenderWindow from "@kitware/vtk.js/Rendering/Misc/GenericRenderWindow";
import vtkImageMapper from '@kitware/vtk.js/Rendering/Core/ImageMapper';
import vtkImageSlice from '@kitware/vtk.js/Rendering/Core/ImageSlice';
import vtkInteractorStyleImage from '../utils/vtkInteractorStyleImage';
import vtkImageResliceMapper from '@kitware/vtk.js/Rendering/Core/ImageResliceMapper';
import vtkPlane from '@kitware/vtk.js/Common/DataModel/Plane';
import ITKHelper from '@kitware/vtk.js/Common/DataModel/ITKHelper'
import vtkColorMaps from '@kitware/vtk.js/Rendering/Core/ColorTransferFunction/ColorMaps';
import vtkColorTransferFunction from "@kitware/vtk.js/Rendering/Core/ColorTransferFunction";
import vtkPiecewiseFunction from "@kitware/vtk.js/Common/DataModel/PiecewiseFunction";
const { convertItkToVtkImage } = ITKHelper;
const { SlicingMode } = Constants;

    context.genericRenderWindow = vtkGenericRenderWindow.newInstance({ background: [0,0,0] })
    context.genericRenderWindow.setContainer(_container)
    context.renderWindow = context.genericRenderWindow.getRenderWindow()
    context.renderWindow.setDefaultViewAPI('WebGPU')
    const openglRenderWindow = context.genericRenderWindow.getApiSpecificRenderWindow()
    console.log(openglRenderWindow.getContext(), 'console.log(openglRenderWindow); '); 
    openglRenderWindow.buildPass(true)
```

But the console still show the webgl context:

```console
[...]

Issue #3139, opened 2024-10-07
[Bug] getDefaultViewAPI always return WebGL even if viewAPI params is set to WebGPU


daker:
### Bug description

`getDefaultViewAPI()` returns the wrong viewAPI

### Steps to reproduce

- Open https://kitware.github.io/vtk-js/examples/TubeFilter/index.html?viewAPI=WebGPU (see the WebGPU param)
- Open the console then type in : `globalThis.renderer.getRenderWindow().getDefaultViewAPI()`
- You get WebGL instead of WebGPU

### Detailed Behavior

![image](https://github.com/user-attachments/assets/20f1fa85-c093-4971-ba3c-36abec1619c2)

### Expected Behavior

Calling `globalThis.renderer.getRenderWindow().getDefaultViewAPI()` should return the used viewAPI, in this case WebGPU

### Environment

- vtk.js version: master
- Browsers: latest Chrome
- OS:

finetjul:
Not a bug according to the documentation. 
You might want to call `getApiSpecificRenderWindow()` instead.
 (look at the code of FullScreenRenderWindow how it is used)

daker:
Ok but is there a way to get the viewAPI being used as plain string ? WebGL or WebGPU instead of detecting the class type ? i want to use it for workaround

finetjul:
Not that I know of.

I guess you could add a convenient function in Core/RenderWindow/index.js that converts a class instance into a plain string (by parsing the classname instance ?).

PR #3502, opened 2026-05-07
fix(types): correct defaultViewAPI typing

PaulHax:
`defaultViewAPI` is a string at runtime ('WebGL' / 'WebGPU'), but the types
disagreed:

- `DEFAULT_VIEW_API` was a `const enum` with numeric values, so
  `setDefaultViewAPI(DEFAULT_VIEW_API.WebGPU)` would pass `1` and silently
  fall back to WebGL.
- `IFullScreenRenderWindowInitialValues.defaultViewAPI` was typed `boolean`.

Switch `DEFAULT_VIEW_API` to a string union (same exported name) and reuse it.
No runtime changes. Related: #3216, #3139.

### DIFF
--- a/Sources/Rendering/Core/RenderWindow/index.d.ts
+++ b/Sources/Rendering/Core/RenderWindow/index.d.ts
@@ -29,10 +29,7 @@ interface IStatistics {
   str: string;
 }
 
-export const enum DEFAULT_VIEW_API {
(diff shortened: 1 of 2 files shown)


daker (approved):
LGTM

finetjul (approved):
LGTM

PR #3503, opened 2026-05-07
fix(Property): stop setSpecularPower from overwriting roughness

PaulHax:
Drop the `setSpecularPower` override that also writes `roughness`, leaving
the plain `macro.setGet` setter behind.

I think the override existed to auto-derive a PBR `roughness` for WebGL code
that only set the Phong `specularPower`. But Phong and PBR are parallel
rendering paths, not a migration — so the bridge silently corrupts `roughness`
for any caller that sets both.


finetjul (approved):
LGTM

daker (approved):
LGTM

PR #3511, opened 2026-05-18
fix(vtkWebGPUImageCPRMapper): projection direction should use vtkImag…

TreatTrick:
### Context

`vtkWebGPUImageCPRMapper` already uses `MCTCMatrix` to transform CPR sample positions from model coordinates to texture coordinates. That matrix is built from `image.getWorldToIndex()`, so it includes the `vtkImageData` direction matrix.

This is the same class of issue as the one addressed for `vtkOpenGLImageCPRMapper` in #3508. The WebGPU implementation had the same size-only projection direction conversion path.

However, the projection slab direction was still converted to texture space by dividing by `VolumeSizeMC`:

```wgsl
let volumeSize = max(mapperUBO.VolumeSizeMC.xyz, vec3<f32>(0.000001));
let scaledDirection = projectionDirection / volumeSize;
```

That only accounts for volume size. It does not account for image direction, so CPR projection sampling can follow the wrong texture-space direction for oriented images, including flipped or rotated `vtkImageData`.

### Results

Before this change, the CPR sample position used `MCTCMatrix`, but the projection direction used a separate size-only conversion. This could make the projection slab sample along the wrong axis or sign when `vtkImageData` direction is not identity.

After this change, both the CPR sample position and the projection slab direction are transformed through `MCTCMatrix`. Positions use `w = 1.0`, while direction vectors use `w = 0.0` so the image orientation and spacing are applied without applying translation.

### Changes

Transform the projection slab direction with `MCTCMatrix` using `w = 0.0` so image orientation is applied to direction vectors without applying translation.

Replace the size-only `projectionDirection / VolumeSizeMC` conversion with:

```wgsl
(mapperUBO.MCTCMatrix * vec4<f32>(projectionDirection, 0.0)).xyz
```

so projection sampling follows the correct direction for oriented `vtkImageData`.

Remove the `VolumeSizeMC` UBO entry and its CPU-side setup because the projection direction conversion now comes directly from `MCTCMatrix`.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  master
  - **OS**:  Window 11 25H2 26200.8457
  - **Browser**:   148.0.3967.70

### DIFF
--- a/Sources/Rendering/WebGPU/ImageCPRMapper/index.js
+++ b/Sources/Rendering/WebGPU/ImageCPRMapper/index.js
@@ -110,8 +110,8 @@ fn sampleOpacityRow(value: f32, row: f32) -> f32
 
 fn getProjectedValue(volumePosTC: vec3<f32>, projectionDirection: vec3<f32>) -> vec4<f32>
 {
-  let volumeSize = max(mapperUBO.VolumeSizeMC.xyz, vec3<f32>(0.000001));
-  let scaledDirection = projectionDirection / volumeSize;
+  let scaledDirection =
+    (mapperUBO.MCTCMatrix * vec4<f32>(projectionDirection, 0.0)).xyz;
   let projectionStep = mapperUBO.ProjectionParams.z * scaledDirection;
   let projectionStart = volumePosTC + mapperUBO.ProjectionParams.y * scaledDirection;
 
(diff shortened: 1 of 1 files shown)


daker:
The WebGPU version is a pure port of the WebGL one, so if the other PR passes it's ok to pass this one

finetjul (approved):
LGTM

PR #3512, opened 2026-05-18
fix(PolyDataMapper2D): restore cull face state

PaulHax:
## Summary

This fixes a WebGL state leak in `vtkOpenGLPolyDataMapper2D`. The mapper enables `CULL_FACE` while rendering 2D polydata, but it did not restore that state afterward. In a parent render window with multiple child render windows sharing one WebGL context, the leaked cull state can carry into the next child renderer and fully cull back-facing image slice quads, leaving the child viewport black on the next render.

The fix restores cull-face state after `renderPieceFinish()` and adds a regression test that renders two child render windows twice, with a `vtkActor2D` in the second child, then verifies the first child slice remains visible and `CULL_FACE` is disabled after rendering.

## Bisect

First bad commit:

`bdc7e0163bb5dc002cee7c446126a7dae2e4d9bd`

`feat(polydatamapper2d): cell scalar based coloring for PolyDataMapper2D`

### DIFF
--- a/Sources/Rendering/OpenGL/Actor/index.js
+++ b/Sources/Rendering/OpenGL/Actor/index.js
@@ -142,10 +142,16 @@ function vtkOpenGLActor(publicAPI, model) {
     if (prepass) {
       model.context.depthMask(true);
       publicAPI.activateTextures();
-    } else if (model.activeTextures) {
-      for (let index = 0; index < model.activeTextures.length; index++) {
-        model.activeTextures[index].deactivate();
+    } else {
+      if (model.activeTextures) {
+        for (let index = 0; index < model.activeTextures.length; index++) {
+          model.activeTextures[index].deactivate();
+        }
       }
+      // Restore baseline cull state so a mapper that enabled culling
+      // cannot leak into later actors or the next renderer's clear().
+      // Mirrors C++ vtkOpenGLProperty::PostRender.
+      model._openGLRenderWindow.disableCullFace();
     }
   };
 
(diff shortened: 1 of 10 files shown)


finetjul (changes requested):
None of the polydata mappers (poly, glyph, surfacelic) disable face culling after rendering.

It seems that it is to "cache" face culling state between mappers.

Maybe what you want to do should be done at the "renderer" or "renderwindow" level.

I note that `gl.cullFace` could be "cached" as well similarly. I did not check how it is done in VTK C++.

PaulHax:
Good notes finetjul. This stayed subtle because the 3D mappers paper over it: each rewrites cull state on entry from the actor's properties, so a following actor with the default `disableCullFace` clears the leak before it shows; `PolyDataMapper2D` exposes it by enabling cull unconditionally, and an `ImageMapper` rendered next never touches cull, so nothing resets it.

I checked VTK C++: cull is managed around each actor in `vtkOpenGLProperty::Render` / `PostRender`, not in the mappers or at the renderer level. Updated PR code matches that: `vtkOpenGLActor` and `vtkOpenGLActor2D` now disable cull in their postpass hook (`apply(renderPass, prepass=false)`), after every actor's mapper.

Separately, I cached `gl.cullFace` as you suggested: a `cullFaceMode` field and `setCullFaceMode()` on `vtkOpenGLRenderWindow`.

finetjul (approved):
LGTM

sankhesh (approved):
LGTM

PR #3516, opened 2026-05-21
Restore cull state after 2D mapper render

sankhesh:
### Context

Before this change, the 2D mapper would enable backface culling but never disable it after finishing. That caused other mappers in the same view to have to deal with the culled state and unintentionally create rendering artifacts.

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


jourdain (approved):

sankhesh:
Rebased onto #3512

PR #3517, opened 2026-05-22
test: re-enable four tests skipped during past tooling migrations

### LINKED ISSUE #2102: Average Intensity Projection Volume Rendering test fails

PaulHax:
Four tests were skipped over the years, each during a test-harness or build migration rather than for a genuine feature failure. A skipped test is green in CI, so none produced a signal and all silently rotted. Re-enable them.

testLabelmapOutlineManyRenderer
  Skipped 2022-03-11 in commit 50ace42c8be (PR #2308, "Make test
  harnesses the same for release and PRs"); the commit gave no
  reason. While skipped the code rotted: createLabelPipeline called
  backgroundImageData.getDimensionsByReference(), which is not a
  function on vtkImageData. Use getDimensions(). Passes against the
  existing baseline, so no baseline change.

testEdgeVisibility
  Skipped 2021-11-12 in commit 13b9be1847e (PR #2161, "Update test
  framework"), flagged in a pixelmatch-threshold-0 audit (PR #2098,
  closed unmerged) as a "bad image test" (issue #2104). The 2017
  baseline predates commit a8873e5cd9a (2021-02-19,
  "feat(vtkopenglpolydatamapper): add support for backface
  property"), which changed the edge draw path from lit
  (ambient = getAmbient(), diffuse = getDiffuse()) to flat
  (ambient = 1.0, diffuse = 0.0): edges now render at the full
  configured edge color, the intended behavior. The render has been
  stable since at least 2021-06 (a steady 1.80% from the 2017
  baseline at 2021-06, 2022, and 2024). Refresh the baseline; the
  tolerance is unchanged at 1% (chromium ~0%, firefox passes).

testAverageIntensityProjection
  Skipped 2021-11-12 in commit 13b9be1847e (PR #2161), tracked by
  issue #2102. It did not drift or regress. Before the 2021
  test-framework work the test was effectively a no-op: the old
  compareImages passed its threshold argument (here 1.5) straight to
  pixelmatch, whose threshold range is 0 to 1, so pixelmatch reported
  zero differing pixels for any image and the test passed vacuously.
  The pixelmatch-threshold-0 audit (PR #2098, closed unmerged)
  exposed the real comparison, and the test then "failed" because its
  baseline had never actually been checked.

  The committed baseline is the original 2019 image from aae2f554559;
  commit 8a50448cbfb only renamed it OpenGL/ to Core/ (100% rename).
  It is a near-blank volume the real render never matched. Rendering
  against old Sources shows a stable, correct head volume back to
  2021-06 (~1.5% from the corrected baseline, ~0% on current master).

  Replace the baseline with the actual render (chromium ~0%, firefox
  0.01%, within the 1% tolerance) and re-enable the test. Also
  release resources via .finally; gc.releaseResources had been passed
  as an unused 5th argument to compareImages.

testSetTable
  Skipped 2026-05-21 by the Vite/Vitest migration (c0d988751e6),
  which rewrote test.onlyIfWebGL(...) as it.skip(...) instead of the
  it.skipIf(__VTK_TEST_NO_WEBGL__)(...) form used for every other
  migrated WebGL test, silently disabling a passing test. Restore the
  intended WebGL-conditional skip.

Closes #2102


sankhesh (approved):
LGTM

PR #3518, opened 2026-05-24
docs(IDE): enable IntelliSense

daker:
### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,0 +1,15 @@
+{
+  "compilerOptions": {
+    "ignoreDeprecations": "6.0",
+    "checkJs": false,
+    "module": "preserve",
+    "moduleResolution": "bundler",
+    "resolveJsonModule": true,
+    "baseUrl": ".",
+    "paths": {
+      "vtk.js/Sources/*": ["Sources/*/index.js", "Sources/*.js", "Sources/*"],
+      "vtk.js/*": ["./*"]
+    }
+  },
+  "include": ["Sources/**/*.js", "Examples/**/*.js", "Utilities/**/*.js"]
+}


.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
+  "recommendations": [
```
sankhesh: These recommendations seems user-preference. Should they be committed?

.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
...
+      "EditorConfig.EditorConfig",
+      "yoavbls.pretty-ts-errors"
```
sankhesh: Same idea - IMHO, this is very much user-preference.

.vscode/settings.json
```diff
@@ -0,0 +1,16 @@
+{
```
sankhesh: Same question here. Do we need vscode preferences at the project level?

jsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "ignoreDeprecations": "6.0",
+    "checkJs": false,
+    "baseUrl": ".",
+    "moduleResolution": "bundler",
```
sankhesh, line 6: Could you please share why "bundler" was chosen? I can see why the path mappings help editor navigation, but I'd like to understand this resolution.

package.json
```diff
@@ -108,7 +108,6 @@
     "build:esm": "npm run build:pre && cross-env BUILD_TARGET=esm vite build",
     "build:umd": "npm run build:pre && cross-env BUILD_TARGET=umd vite build",
     "build:release": "npm run lint && npm run build",
-    "release:create-packages": "node ./Utilities/ci/build-npm-package.js",
```
sankhesh, line 113: This looks like an unrelated change. Was it meant to be removed?

daker: Yes, `build-npm-package.js` doesn't exist anymore.

.vscode/extensions.json
daker: This is only to inform users about the recommended VS Code extensions. Without `EditorConfig.EditorConfig`, the `.editorconfig` file in the project will not be applied, which can lead to PR issues such as incorrect EOLs, indentation, and other formatting inconsistencies.

.vscode/settings.json
daker: These are fallback defaults for users who do not have the `EditorConfig` extension installed.
Additionally, with the `jsconfig.json` config, VS Code defaults to opening definitions from the `.d.ts` file instead of the actual `.js` source file, which is why `"js/ts.preferGoToSourceDefinition": true` is required.

jsconfig.json
daker: Like TS, we have few options for moduleResolution: "Node", "Classic", "Node16", "NodeNext" and "Bundler"
"bundler" makes the IDE resolution behave like bundlers and avoids Node/ESM specific false errors on extensionless imports.

https://code.visualstudio.com/docs/languages/jsconfig#_what-is-jsconfigjson
https://www.typescriptlang.org/tsconfig/#moduleResolution:~:text=but%20unlike%20the%20Node.js%20resolution%20modes%2C%20bundler%20never%20requires%20file%20extensions%20on%20relative%20paths%20in%20imports

.vscode/extensions.json
daker: They are not installed by default. When developers open VS Code, they will receive a notification prompting them to install the recommended extensions.
https://code.visualstudio.com/docs/configure/extensions/extension-marketplace#_workspace-recommended-extensions

finetjul (changes requested):
Discussed with @jourdain @PaulHax :
 - please move the package.json change into a different "commit".
 - do not commit .vscode files in the repository, but you can add those in the github page documentation.
 - ok to keep jcsonfig.json
 - the commit message is not a "feat" but a "docs" or "chore (to not trigger a new release).

Thanks !

finetjul (approved):
LGTM

PR #3523, opened 2026-06-01
refactor(WebGL): remove support for WebGL1

daker:
### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Common/Core/CellArray/index.js
+++ b/Sources/Common/Core/CellArray/index.js
@@ -106,7 +106,16 @@ function vtkCellArray(publicAPI, model) {
       cellPointIds = cell;
     }
     const cellId = publicAPI.getNumberOfCells();
-    publicAPI.insertNextTuples([cellPointIds.length, ...cellPointIds]);
+    // Build the [count, ...pointIds] record without the spread operator, which
+    // would walk the iterator protocol and allocate an extra intermediate for
+    // every inserted cell.
+    const numberOfPoints = cellPointIds.length;
+    const cellRecord = new Array(numberOfPoints + 1);
+    cellRecord[0] = numberOfPoints;
+    for (let i = 0; i < numberOfPoints; ++i) {
+      cellRecord[i + 1] = cellPointIds[i];
+    }
+    publicAPI.insertNextTuples(cellRecord);
     // By computing the number of cells earlier, we made sure that numberOfCells is defined
     ++model.numberOfCells;
     if (model.cellSizes != null) {
--- a/Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
+++ b/Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
@@ -26,6 +26,156 @@ function shouldApplyCoordShiftAndScale(coordShift, coordScale) {
   );
 }
 
+function canUseIndexedVBO(options) {
+  // Indexed (shared point) VBOs avoid expanding every cell vertex. They only
+  // work for point based attributes. Per cell attributes are baked per vertex
+  // in the flattened path, and cell accurate hardware selection needs the
+  // flattened layout too (WebGL2 has no gl_PrimitiveID), so callers ask for it
+  // explicitly via options.forceFlatten.
+  return (
+    !options.forceFlatten &&
+    !options.haveCellScalars &&
+    !options.haveCellNormals &&
+    !options.useTCoordsPerCell
+  );
+}
+
+// Walk one cell, emitting each output vertex via emit(pointId, cellId). The
+// indexed and flattened VBO paths share these so the primitive assembly logic
+// lives in exactly one place; they differ only in what emit() does (push an
+// index vs. expand a full vertex).
+const CELL_BUILDERS = {
+  // easy, every input point becomes an output point
+  anythingToPoints(numPoints, cellPts, offset, cellId, emit) {
+    for (let i = 0; i < numPoints; ++i) {
+      emit(cellPts[offset + i], cellId);
+    }
+  },
+  linesToWireframe(numPoints, cellPts, offset, cellId, emit) {
+    // for lines we add a bunch of segments
+    for (let i = 0; i < numPoints - 1; ++i) {
+      emit(cellPts[offset + i], cellId);
+      emit(cellPts[offset + i + 1], cellId);
+    }
+  },
+  polysToWireframe(numPoints, cellPts, offset, cellId, emit) {
+    // for polys we add a bunch of segments and close it
+    if (numPoints > 2) {
+      for (let i = 0; i < numPoints; ++i) {
+        emit(cellPts[offset + i], cellId);
+        emit(cellPts[offset + ((i + 1) % numPoints)], cellId);
+      }
+    }
+  },
+  stripsToWireframe(numPoints, cellPts, offset, cellId, emit) {
+    if (numPoints > 2) {
+      // for strips we add a bunch of segments and close it
+      for (let i = 0; i < numPoints - 1; ++i) {
+        emit(cellPts[offset + i], cellId);
+        emit(cellPts[offset + i + 1], cellId);
+      }
+      for (let i = 0; i < numPoints - 2; i++) {
+        emit(cellPts[offset + i], cellId);
+        emit(cellPts[offset + i + 2], cellId);
+      }
+    }
+  },
+  polysToSurface(numPoints, cellPts, offset, cellId, emit) {
+    for (let i = 0; i < numPoints - 2; i++) {
+      emit(cellPts[offset], cellId);
+      emit(cellPts[offset + i + 1], cellId);
+      emit(cellPts[offset + i + 2], cellId);
+    }
+  },
+  stripsToSurface(numPoints, cellPts, offset, cellId, emit) {
+    for (let i = 0; i < numPoints - 2; i++) {
+      emit(cellPts[offset + i], cellId);
+      emit(cellPts[offset + i + 1 + (i % 2)], cellId);
+      emit(cellPts[offset + i + 1 + ((i + 1) % 2)], cellId);
+    }
+  },
+};
+
+// Number of output vertices a single cell produces, matching CELL_BUILDERS.
+const CELL_COUNTERS = {
+  anythingToPoints(numPoints) {
+    return numPoints;
+  },
+  linesToWireframe(numPoints) {
+    return numPoints > 1 ? (numPoints - 1) * 2 : 0;
+  },
+  polysToWireframe(numPoints) {
+    return numPoints > 2 ? numPoints * 2 : 0;
+  },
+  stripsToWireframe(numPoints) {
+    return numPoints > 2 ? numPoints * 4 - 6 : 0;
+  },
+  polysToSurface(numPoints) {
+    return numPoints > 2 ? (numPoints - 2) * 3 : 0;
+  },
+  stripsToSurface(numPoints) {
+    return numPoints > 2 ? (numPoints - 2) * 3 : 0;
+  },
+};
+
+// Pick the build/count pair for an input cell type + output representation.
+function getCellHandlers(inRep, outRep) {
+  if (outRep === Representation.POINTS || inRep === 'verts') {
+    return {
+      build: CELL_BUILDERS.anythingToPoints,
+      count: CELL_COUNTERS.anythingToPoints,
+    };
+  }
+  if (outRep === Representation.WIREFRAME || inRep === 'lines') {
+    return {
+      build: CELL_BUILDERS[`${inRep}ToWireframe`],
+      count: CELL_COUNTERS[`${inRep}ToWireframe`],
+    };
+  }
+  return {
+    build: CELL_BUILDERS[`${inRep}ToSurface`],
+    count: CELL_COUNTERS[`${inRep}ToSurface`],
+  };
+}
+
+// Total number of output vertices across every cell in the array.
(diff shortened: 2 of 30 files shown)


finetjul (commented):
How does it compare with #2212 ?

I find it a bit misleading that claiming it remove WebGL1 support but it still render as in "WebGL1" would do, e.g. it duplicates points for each cells (see vtkCellArrayBufferObject). This is big bottleneck when it comes to rendering initialization and memory usage.

It can be done in a second pass though.

daker:
@finetjul it's the same as #2212 but rebased. i get your point i will try to see if i can do something

Sources/Common/Core/CellArray/index.js
```diff
@@ -106,7 +106,16 @@ function vtkCellArray(publicAPI, model) {
...
+    // Build the [count, ...pointIds] record without the spread operator, which
+    // would walk the iterator protocol and allocate an extra intermediate for
+    // every inserted cell.
+    const numberOfPoints = cellPointIds.length;
+    const cellRecord = new Array(numberOfPoints + 1);
+    cellRecord[0] = numberOfPoints;
+    for (let i = 0; i < numberOfPoints; ++i) {
+      cellRecord[i + 1] = cellPointIds[i];
+    }
+    publicAPI.insertNextTuples(cellRecord);
```
finetjul, line 118: The following is faster and shorter:
```
const cellTuple = new Uint32Array(numberOfPoints + 1);
cellTuple[0] = numberOfPoints;
cellTuple.set(cellPointIds, 1); // bulk copy, faster than a loop
```

finetjul (commented):
@Jo-Byr can you have a look at the WebGL2 implementation ?

@daker do you see an improvement of performance (TTFF)?

Sources/Common/Core/CellArray/index.js
daker: new Uint32Array(numberOfPoints +1) zero fills the buffer first, then .set() overwrites it, i benchmarked this Array + forloop is faster

Chrome:
image

Firefox
image

finetjul: Interesting.
Did you try on Chrome or Firefox?
How about:
`const cellTuple = Array.from(cellPointIds).unshift(numberOfPoints)` ?

daker: [benchmark.html](https://github.com/user-attachments/files/29330992/benchmark.html)

Chrome:
image

FF:
image

Edge:
image

finetjul: Sounds good, thanks for the comparisons...

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -26,6 +26,165 @@ function shouldApplyCoordShiftAndScale(coordShift, coordScale) {
...
+  for (let index = 0; index < cellData.length; index += cellData[index] + 1) {
+    total += count(cellData[index]);
+  }
+  return total;
+}
+
+// Number of cells encoded in a raw VTK.js cellarray (connectivity layout). This
+// works for any input exposing getData() (vtkCellArray or vtkDataArray), so
+// callers that pass a plain vtkDataArray of connectivity keep working.
+function countCells(cellData) {
```
finetjul: why not use vtkCellArray.getNumberOfCells() ?
What does it mean to count the number of cells of a vtkDataArray? If the array contains cells, then it must be a vtkCellArray, no ?

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -26,6 +26,165 @@ function shouldApplyCoordShiftAndScale(coordShift, coordScale) {
...
+  let total = 0;
+  for (let index = 0; index < cellData.length; index += cellData[index] + 1) {
+    total += 1;
+  }
+  return total;
+}
+
+// Walk every cell in the array, calling build() with a running cell id.
+function forEachCell(cellData, cellOffset, build, emit) {
```
finetjul, line 152: maybe it could return the new cellId

finetjul (commented):
Should we get better performance even when rendering with a cell data array ?

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
daker: @finetjul you mean something like vtkCellArray.getNumberOfCells(cellArray.getData()); ?

finetjul: `cellArray.getNumberOfCells()` where `cellArray` is a `vtkCellArray`

daker: Yes, but this breaks the code because there are several places (around 10) in the codebase where cells are built as plain `vtkDataArray` objects and passed directly to `createVBO`. With this change, those code paths crash, causing the code and tests to fail.

https://github.com/Kitware/vtk-js/blob/07c3e9a5eae23188bcceed6ae11417981493d928/Sources/Rendering/OpenGL/Convolution2DPass/index.js#L248

https://github.com/Kitware/vtk-js/blob/07c3e9a5eae23188bcceed6ae11417981493d928/Sources/Rendering/OpenGL/ImageMapper/index.js#L1404

finetjul: Let's change those places to `vtkCellArray.newInstance` instead of `vtkDataArray.newInstance`. 
`createVBO` is costly and we must do all we can to improve speed.

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -26,6 +27,156 @@ function shouldApplyCoordShiftAndScale(coordShift, coordScale) {
   );
 }
 
+function canUseIndexedVBO(options) {
+  // Indexed (shared point) VBOs avoid expanding every cell vertex. They only
+  // work for point based attributes. Per cell attributes are baked per vertex
+  // in the flattened path, and cell accurate hardware selection needs the
```
finetjul, line 32: It seems that we have a path where cell colors are inside a texture.

https://github.com/Kitware/vtk-js/commit/a72b8efa2c7a9b2ce97b92c1c0dc3b0a6cacc8c0

Can't you use an indexedVBO in such case ?

daker: I don't think so, with cell data the texture only moves where the colors are stored, the lookup key (the tcoord into that texture) is still per cell, so a point shared between two cells would need two different tcoords in the VBO. Any per-cell attribute in the vertex stream forces the flattened layout, which is why canUseIndexedVBO rejects useTCoordsPerCell as well as haveCellScalars. VTK.js avoids this by recovering the cell id in the fragment shader via gl_PrimitiveID, but WebGL2 doesn't expose it. Indexed VBOs are still used when coloring by point scalars, including through the color texture, since those tcoords are per point.

finetjul: sounds good. Thanks for the explanation.

Sources/Rendering/OpenGL/ImageMapper/index.js
```diff
@@ -1401,7 +1402,7 @@ function vtkOpenGLImageMapper(publicAPI, model) {
       cellArray[5] = 0;
       cellArray[6] = 3;
       cellArray[7] = 2;
-      const cells = vtkDataArray.newInstance({
+      const cells = vtkCellArray.newInstance({
         numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/ImageCPRMapper/index.js
```diff
@@ -501,7 +502,7 @@ function vtkOpenGLImageCPRMapper(publicAPI, model) {
         offset += 5;
         ptIdx += 4;
       }
-      const cells = vtkDataArray.newInstance({
+      const cells = vtkCellArray.newInstance({
         numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/ImageResliceMapper/index.js
```diff
@@ -628,14 +629,17 @@ function vtkOpenGLImageResliceMapper(publicAPI, model) {
         values: model.resliceGeom.getPoints().getData(),
       });
       points.setName('points');
-      const cells = vtkDataArray.newInstance({
+      const cells = vtkCellArray.newInstance({
         numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/OrderIndependentTranslucentPass/index.js
```diff
@@ -75,14 +76,17 @@ function vtkOpenGLOrderIndependentTranslucentPass(publicAPI, model) {
       values: tcoordArray,
     });
     tcoords.setName('tcoords');
-    const cells = vtkDataArray.newInstance({
+    const cells = vtkCellArray.newInstance({
       numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/RadialDistortionPass/index.js
```diff
@@ -212,14 +213,17 @@ function vtkRadialDistortionPass(publicAPI, model) {
       values: tcoordArray,
     });
     tcoords.setName('tcoords');
-    const cells = vtkDataArray.newInstance({
+    const cells = vtkCellArray.newInstance({
       numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/Skybox/index.js
```diff
@@ -113,14 +114,17 @@ function vtkOpenGLSkybox(publicAPI, model) {
       cellArray[5] = 0;
       cellArray[6] = 3;
       cellArray[7] = 2;
-      const cells = vtkDataArray.newInstance({
+      const cells = vtkCellArray.newInstance({
         numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/SurfaceLIC/LineIntegralConvolution2D/pingpong.js
```diff
@@ -34,7 +35,7 @@ function getQuadPoly(openGLRenderWindow) {
     values: ptsArray,
   });
   points.setName('points');
-  const cells = vtkDataArray.newInstance({
+  const cells = vtkCellArray.newInstance({
     numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/SurfaceLIC/SurfaceLICInterface/index.js
```diff
@@ -53,7 +54,7 @@ function getQuadPoly(openGLRenderWindow) {
     values: ptsArray,
   });
   points.setName('points');
-  const cells = vtkDataArray.newInstance({
+  const cells = vtkCellArray.newInstance({
     numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -1864,13 +1863,16 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
         values: ptsArray,
       });
       points.setName('points');
-      const cells = vtkDataArray.newInstance({
+      const cells = vtkCellArray.newInstance({
         numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/Convolution2DPass/index.js
```diff
@@ -245,14 +246,17 @@ function vtkConvolution2DPass(publicAPI, model) {
       values: tcoordArray,
     });
     tcoords.setName('tcoords');
-    const cells = vtkDataArray.newInstance({
+    const cells = vtkCellArray.newInstance({
       numberOfComponents: 1,
```
finetjul: `numberOfComponents` is not needed for a `vtkCellArray`

Sources/Rendering/OpenGL/CellArrayBufferObject/index.js
```diff
@@ -2,6 +2,7 @@ import { vec3 } from 'gl-matrix';
 
 import macro from 'vtk.js/Sources/macros';
 import vtkBufferObject from 'vtk.js/Sources/Rendering/OpenGL/BufferObject';
+import vtkCellArray from 'vtk.js/Sources/Common/Core/CellArray';
```
finetjul: There is an unused warning here

finetjul (approved):
LGTM

PR #3526, opened 2026-06-03
feat(WebGPU): add WebGPU support to VolumePassFSQ

daker:
### Changes

### PR and Code Checklist

### Testing

- [ ] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Examples/Rendering/LabelmapEdgeProjection/index.js
+++ b/Examples/Rendering/LabelmapEdgeProjection/index.js
@@ -37,8 +37,12 @@ const { BlendMode } = Constants;
  * New Blend Mode: [LABELMAP_EDGE_PROJECTION_BLEND]
  *
  */
+const viewAPI =
+  new URLSearchParams(window.location.search).get('viewAPI') || 'WebGL';
+
 const fullScreenRenderer = vtkFullScreenRenderWindow.newInstance({
   background: [0, 0, 0],
+  defaultViewAPI: viewAPI,
 });
 const renderer = fullScreenRenderer.getRenderer();
 const renderWindow = fullScreenRenderer.getRenderWindow();
@@ -53,10 +57,20 @@ actor.setMapper(mapper);
 // ----------------------------------------------------------------------------
 const gui = new GUI();
 const guiParams = {
+  viewAPI,
   blendMode: 'regular',
   thickness1: 3,
   thickness2: 3,
 };
+
+gui
+  .add(guiParams, 'viewAPI', ['WebGL', 'WebGPU'])
+  .name('Renderer')
+  .onChange((api) => {
+    const query = new URLSearchParams(window.location.search);
+    query.set('viewAPI', api);
+    window.location.search = query.toString();
+  });
 
 let imageData;
 let dims;
@@ -65,6 +79,9 @@ let radius1;
 let radius2;
 let edgeLabelmapActor = null;
 let regularLabelmapActor = null;
+// Pristine single component scalars of the base image; both blend modes
+// rebuild their scalar array from this deep copy so switching stays repeatable.
+let baseScalarData = null;
 
 // ----------------------------------------------------------------------------
 // Common functions
@@ -164,8 +181,7 @@ function createRegularLabelmap(imgData, d, c, r1, r2) {
 }
 
 function createAdvancedMIPLabelmap(imgData, d, c, r1, r2) {
-  const array = imgData.getPointData().getArray(0);
-  const baseData = array.getData();
+  const baseData = baseScalarData.getData();
 
   const numberOfComponents = 2;
   const cubeData = new Float32Array(numberOfComponents * baseData.length);
@@ -216,6 +232,13 @@ function createAdvancedMIPLabelmap(imgData, d, c, r1, r2) {
 function createBasePipeline() {
   renderer.removeVolume(actor);
   mapper.setInputData(imageData);
+
+  // restore the pristine single component scalars and the default color mix
+  // (edge mode replaces them with a two component image + labels array and
+  // the additive preset)
+  const array = imageData.getPointData().getArray(0);
+  array.deepCopy(baseScalarData);
+  actor.getProperty().setColorMixPreset(ColorMixPreset.DEFAULT);
 
   renderer.addVolume(actor);
   setupTransferFunctions();
@@ -296,6 +319,9 @@ const reader = vtkHttpDataSetReader.newInstance({ fetchGzip: true });
 reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
   reader.loadData().then(() => {
     imageData = reader.getOutputData();
+    const scalars = imageData.getPointData().getArray(0);
+    baseScalarData = scalars.newClone();
+    baseScalarData.deepCopy(scalars);
 
     dims = imageData.getDimensions();
     center = dims.map((d) => Math.floor(d / 2));
--- a/Sources/Rendering/WebGPU/VolumePassFSQ/index.js
+++ b/Sources/Rendering/WebGPU/VolumePassFSQ/index.js
@@ -164,6 +164,117 @@ fn getDependentGradient(vTex: texture_3d<f32>, tpos: vec4<f32>, vNum: i32, numCo
   return result;
 }
 
+fn getViewVector(posVC: vec3<f32>) -> vec3<f32>
+{
+  if (rendererUBO.cameraParallel != 0u)
+  {
+    return vec3<f32>(0.0, 0.0, 1.0);
+  }
+  return normalize(-posVC);
+}
+
+fn applyLighting(tColor: vec3<f32>, posVC: vec3<f32>, normalVC: vec3<f32>, vNum: i32) -> vec3<f32>
+{
+  if (rendererUBO.LightCount <= 0)
+  {
+    return tColor;
+  }
+
+  let lighting = volumeSSBO.values[vNum].lighting;
+  let ambient = lighting.x;
+  let diffuseCoeff = lighting.y;
+  let specularCoeff = lighting.z;
+  let specularPower = lighting.w;
+  let viewDir = getViewVector(posVC);
+
+  var diffuse = vec3<f32>(0.0);
+  var specular = vec3<f32>(0.0);
+
+  var i: i32 = 0;
+  loop
+  {
+    if (i >= rendererUBO.LightCount) { break; }
+
+    let lightColor =
+      rendererLightSSBO.values[i].LightColor.rgb *
+      (rendererLightSSBO.values[i].LightColor.w * 0.2);
+    let lightType = i32(rendererLightSSBO.values[i].LightData.x);
+
+    var lightDirection = vec3<f32>(0.0);
+    var attenuation = 1.0;
+
+    if (lightType == 0)
+    {
+      let lightPosVC = rendererLightSSBO.values[i].LightPos.xyz;
+      let lightOffset = posVC - lightPosVC;
+      let lightDistance = length(lightOffset);
+      if (lightDistance <= 0.0)
+      {
+        i++;
+        continue;
+      }
+      lightDirection = lightOffset / lightDistance;
+    }
+    else if (lightType == 1)
+    {
+      lightDirection = -normalize(
+        (rendererUBO.WCVCNormals *
+          vec4<f32>(normalize(rendererLightSSBO.values[i].LightDir.xyz), 0.0)).xyz
+      );
+    }
+    else if (lightType == 2)
+    {
+      let lightPosVC = rendererLightSSBO.values[i].LightPos.xyz;
+      let lightOffset = posVC - lightPosVC;
+      let lightDistance = length(lightOffset);
+      if (lightDistance <= 0.0)
+      {
+        i++;
+        continue;
+      }
+      lightDirection = lightOffset / lightDistance;
+
+      let spotDirVC = -normalize(
+        (rendererUBO.WCVCNormals *
+          vec4<f32>(normalize(rendererLightSSBO.values[i].LightDir.xyz), 0.0)).xyz
+      );
+      let theta = dot(spotDirVC, lightDirection);
+      let innerCone = rendererLightSSBO.values[i].LightData.y;
+      let outerCone = rendererLightSSBO.values[i].LightData.z;
+      let coneRange = max(0.000001, innerCone - outerCone);
+      attenuation = clamp((theta - outerCone) / coneRange, 0.0, 1.0);
+      if (attenuation <= 0.0)
+      {
+        i++;
+        continue;
+      }
+    }
+    else
+    {
+      i++;
+      continue;
+    }
+
+    let ndotL = max(dot(normalVC, lightDirection), 0.0);
+    if (ndotL > 0.0)
+    {
+      diffuse += ndotL * attenuation * lightColor;
+
+      let reflectDir = normalize(lightDirection - 2.0 * ndotL * normalVC);
+      let vdotR = max(dot(viewDir, reflectDir), 0.0);
+      if (vdotR > 0.0)
+      {
+        specular += pow(vdotR, specularPower) * attenuation * lightColor;
+      }
+    }
+
+    i++;
+  }
+
+  return tColor * (ambient + diffuseCoeff * diffuse) +
+    specularCoeff * specular;
+}
+
 fn processVolume(vTex: texture_3d<f32>, vNum: i32, rowStart: i32, posSC: vec4<f32>, tfunRows: f32) -> vec4<f32>
 {
   var outColor: vec4<f32> = vec4<f32>(0.0, 0.0, 0.0, 0.0);
@@ -176,6 +287,7 @@ fn processVolume(vTex: texture_3d<f32>, vNum: i32, rowStart: i32, posSC: vec4<f3
   let numComp: u32 = u32(volumeSSBO.values[vNum].componentInfo.x);
   let independent = volumeSSBO.values[vNum].componentInfo.y > 0.5;
   let sample: vec4<f32> = getTextureValue(vTex, tpos);
+  let posVC = (rendererUBO.SCVCMatrix * posSC).xyz;
 
   if (independent)
   {
@@ -208,7 +320,7 @@ fn processVolume(vTex: texture_3d<f32>, vNum: i32, rowStart: i32, posSC: vec4<f3
 
       if (volumeSSBO.values[vNum].shade[0] > 0.0)
       {
-        color = color*abs(normal.z);
+        color = applyLighting(color, posVC, normal.xyz, vNum);
       }
 
       let mixWeight = componentSSBO.values[rowIdx].mixWeight;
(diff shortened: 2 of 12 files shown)


daker:
Draft to get some initial feedback

daker:
@sankhesh mind doing an initial review ?

daker:
@sankhesh any chance to get an initial review ?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -27,10 +27,68 @@ const volFragTemplate = `
...
+    maxCoord
+  );
+  let nearestValue = textureLoad(vTex, nearestCoord, 0);
+
+  if ((forceNearestMask & 1) != 0) { value.x = nearestValue.x; }
+  if ((forceNearestMask & 2) != 0) { value.y = nearestValue.y; }
+  if ((forceNearestMask & 4) != 0) { value.z = nearestValue.z; }
+  if ((forceNearestMask & 8) != 0) { value.w = nearestValue.w; }
+
+  return value;
```
sankhesh: Why does getTextureValue need to perform filtering as well now?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -27,10 +27,68 @@ const volFragTemplate = `
...
+  {
+    return value;
+  }
+
+  let dims = textureDimensions(vTex, 0);
+  let maxCoord = vec3<i32>(dims) - vec3<i32>(1);
+  let nearestCoord = clamp(
+    vec3<i32>(floor(tpos.xyz * vec3<f32>(dims))),
+    vec3<i32>(0),
+    maxCoord
```
sankhesh: This does not match the exact texel-center convention used by filter sampling, especially near boundaries.

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -27,10 +27,68 @@ const volFragTemplate = `
...
+fn getComponentValue(vTex: texture_3d<f32>, tpos: vec4<f32>, vNum: i32, component: u32) -> f32
+{
+  return getComponent(getTextureValue(vTex, tpos, vNum), component);
+}
+
+fn getDependentOpacityValue(sample: vec4<f32>, numComp: u32) -> f32
+{
+  if (numComp == 1u) { return sample.x; }
+  if (numComp == 2u) { return sample.y; }
+  if (numComp == 3u) { return length(sample.xyz); }
```
sankhesh, line 84: Why is opacity the `length(sample.xyz)` here? Was this intentional?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -73,14 +131,14 @@ fn intersectRayBoundsWithClipPlanes(vNum: i32, minPosSC: vec4<f32>, rayStepSC: v
...
 {
   var result: vec4<f32>;
 
   var tstep: vec4<f32> = volumeSSBO.values[vNum].tstep;
-  result.x = getTextureValue(vTex, tpos + vec4<f32>(tstep.x, 0.0, 0.0, 1.0)) - scalar;
-  result.y = getTextureValue(vTex, tpos + vec4<f32>(0.0, tstep.y, 0.0, 1.0)) - scalar;
-  result.z = getTextureValue(vTex, tpos + vec4<f32>(0.0, 0.0, tstep.z, 1.0)) - scalar;
+  result.x = getComponentValue(vTex, tpos + vec4<f32>(tstep.x, 0.0, 0.0, 1.0), vNum, component) - scalar;
+  result.y = getComponentValue(vTex, tpos + vec4<f32>(0.0, tstep.y, 0.0, 1.0), vNum, component) - scalar;
+  result.z = getComponentValue(vTex, tpos + vec4<f32>(0.0, 0.0, tstep.z, 1.0), vNum, component) - scalar;
```
sankhesh: I know the webgl mapper uses forward differences only. Did you try central differences here? I think that would produce better results.

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -102,7 +160,476 @@ fn getGradient(vTex: texture_3d<f32>, tpos: vec4<f32>, vNum: i32, scalar: f32) -
...
+      normalDotRay = -normalDotRay;
+    }
+
+    var currTC = (volumeSSBO.values[vNum].VCTCMatrix * vec4<f32>(posVC, 1.0)).xyz;
+    let rayStepTC = (volumeSSBO.values[vNum].VCTCMatrix *
+      vec4<f32>(rayDirectionVC * mapperUBO.SampleDistance, 0.0)).xyz;
+    var visibility = 1.0;
+
+    var j: i32 = 0;
+    loop
```
sankhesh, line 514: This is a nested sampling loop that is known to have quite an overhead. Have you profiled this?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -102,7 +160,476 @@ fn getGradient(vTex: texture_3d<f32>, tpos: vec4<f32>, vNum: i32, scalar: f32) -
...
+    weightSum = weightSum + rayWeight;
+    i++;
+  }
+
+  if (weightSum == 0.0)
+  {
+    return 1.0;
+  }
+
+  return clamp(visibilitySum / weightSum, 0.3, 1.0);
```
sankhesh, line 547: Why the 0.3 min?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -102,7 +160,476 @@ fn getGradient(vTex: texture_3d<f32>, tpos: vec4<f32>, vNum: i32, scalar: f32) -
...
+  let startDistance = max(intersectionDistances.x, 0.0);
+  let endDistance = min(intersectionDistances.y, startDistance + lightReach);
+  if (endDistance <= startDistance)
+  {
+    return 1.0;
+  }
+
+  var currentDistance = startDistance;
+  var shadow = 1.0;
+  loop
```
sankhesh, line 801: This combined with the outer loops is quite a nesting. If you see decent performance, I'd leave it alone. But if you start seeing performance issues, this is where to start refactoring.

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -102,7 +160,476 @@ fn getGradient(vTex: texture_3d<f32>, tpos: vec4<f32>, vNum: i32, scalar: f32) -
...
+  let dMinPerAxis = min(distancesTo0, distancesTo1);
+  let dMaxPerAxis = max(distancesTo0, distancesTo1);
+  let distanceMin = max(dMinPerAxis.x, max(dMinPerAxis.y, dMinPerAxis.z));
+  let distanceMax = min(dMaxPerAxis.x, min(dMaxPerAxis.y, dMaxPerAxis.z));
+  return vec2<f32>(distanceMin, distanceMax);
+}
+
+fn computeVolumeShadow(vTex: texture_3d<f32>, posVC: vec3<f32>, lightDirVC: vec3<f32>, vNum: i32, rowStart: i32) -> f32
+{
+  let shadowStepLength = volumeSSBO.values[vNum].shadow.y * 2.25;
```
sankhesh: Where does 2.25 come from?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -102,7 +160,476 @@ fn getGradient(vTex: texture_3d<f32>, tpos: vec4<f32>, vNum: i32, scalar: f32) -
...
+fn applyAllLighting(vTex: texture_3d<f32>, fragPos: vec4<f32>, tColor: vec3<f32>, alpha: f32, posVC: vec3<f32>, normalVC: vec4<f32>, vNum: i32, rowStart: i32) -> vec3<f32>
+{
+  if (rendererUBO.LightCount <= 0)
+  {
+    return tColor;
+  }
+
+  let volCoeff = volumeSSBO.values[vNum].scattering.x *
+    (1.0 - alpha * 0.5) *
+    (1.0 - atan(normalVC.w) * 0.0795774715);
```
sankhesh, line 842: I am not sure I follow this. Could you please explain or add documentation?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -182,21 +930,37 @@ fn adjustBounds(tpos: vec4<f32>, tstep: vec4<f32>, numSteps: f32) -> vec2<f32>
...
+}
+
+fn getRadonColor(scalar: f32, rowIdx: i32, tfunRows: f32) -> vec4<f32>
+{
+  let coord = vec2<f32>(
+    scalar * componentSSBO.values[rowIdx].cScale + componentSSBO.values[rowIdx].cShift,
+    getTFunRowCoord(rowIdx, tfunRows)
+  );
+  let color = textureSampleLevel(tfunTexture, clampSampler, coord, 0.0).rgb;
+  return vec4<f32>(color, 1.0);
```
sankhesh, line 1223: Shouldn't the alpha be calculated as well?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -380,10 +1148,59 @@ fn traverseAdditive(vTex: texture_3d<f32>, vNum: i32, cNum: i32, rayLengthSC: f3
   }
 
   // process to get the color and opacity
-  traverseVals[vNum] = getSimpleColor(sumVal, vNum, cNum);
+  traverseVals[vNum] = getSimpleColor(sumVal, rowIdx, tfunRows);
+}
+
+fn traverseRadon(vTex: texture_3d<f32>, vNum: i32, rowIdx: i32, rayLengthSC: f32, minPosSC: vec4<f32>, rayStepSC: vec4<f32>)
```
sankhesh: This seems different than the opengl mapper. Could you please explain/document this function?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -452,6 +1269,19 @@ fn main(
 
 const tmpMat4 = new Float64Array(16);
 const tmp2Mat4 = new Float64Array(16);
+const tmp3Mat4 = new Float64Array(16);
+const tmp4Mat4 = new Float64Array(16);
+
+function transformPoint(mat, x, y, z) {
```
sankhesh: IMHO, there are math functions already that do this.

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -815,6 +1706,42 @@ function vtkWebGPUVolumePassFSQ(publicAPI, model) {
...
+        ? 1.0
+        : 0.0;
+      componentInfoArray[vidx * 4 + 2] = vprop.getColorMixPreset();
+      let forceNearestMask = 0;
+      for (let component = 0; component < numComp; component++) {
+        if (vprop.getForceNearestInterpolation(component)) {
+          forceNearestMask |= 1 << component;
+        }
+      }
+      componentInfoArray[vidx * 4 + 3] = forceNearestMask;
```
sankhesh, line 2651: I don't follow this. Could you please explain?

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -815,6 +1706,42 @@ function vtkWebGPUVolumePassFSQ(publicAPI, model) {
...
+
+        const ofun = vprop.getScalarOpacity(
+          vprop.getIndependentComponents() ? component : 0
+        );
+        const oRange = ofun.getRange();
+        opacityScaleArray[vidx * 4 + component] =
+          sscale / (oRange[1] - oRange[0]);
+        opacityShiftArray[vidx * 4 + component] =
+          -oRange[0] / (oRange[1] - oRange[0]);
+      }
```
sankhesh, line 2685: Maybe add a check to ensure the ranges are valid before dividing by them

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
```diff
@@ -949,12 +1906,44 @@ function vtkWebGPUVolumePassFSQ(publicAPI, model) {
...
   publicAPI.updateBuffers = () => {
     superClassUpdateBuffers();
+    if (!model.jitterSSBOSent) {
+      model.jitterSSBO.clearData();
+      model.jitterSSBO.setNumberOfInstances(32 * 32);
+      model.jitterSSBO.addEntry('value', 'f32');
+      const jitterArray = new Float32Array(32 * 32);
+      for (let i = 0; i < jitterArray.length; i++) {
+        jitterArray[i] = Math.random();
+      }
```
sankhesh: I thought we used a perlin noise texture in the opengl mapper. This works too, but this buffer will be computed for each volume in the window.

sankhesh (commented):
This is a big change. Took me a while to go through it. I've left some questions but overall, needs more documentation and testing.

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
daker: mirrors the OpenGL mapper https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl#L289-L291

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
daker: 2.25 is the midpoint of the OpenGL mapper's random mix(1.5, 3.0, seed) step factor

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
daker: Radon output is an opaque intensity image: the accumulated attenuation is already encoded in the lookup coordinate, so alpha is 1.0 (matches the OpenGL mapper, which also hardcodes alpha here).

- vtkVolumeFS.glsl:331-334 -> getRadonColor(float) returns vec3 only.
- vtkVolumeFS.glsl:1592 -> thin volume branch: gl_FragData[0] = vec4(getRadonColor(normalizedRayIntensity), 1.0);
- vtkVolumeFS.glsl:1631 -> main radon loop: same vec4(getRadonColor(...), 1.0).

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
daker: Matchs OpenGL mapper
https://github.com/Kitware/vtk-js/blob/07c3e9a5eae23188bcceed6ae11417981493d928/Sources/Rendering/OpenGL/glsl/vtkVolumeFS.glsl#L687

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
daker: Each volume being rendered has an SSBO (shader storage buffer) that carries per volume parameters up to the GPU. componentInfoArray is a `vec4<f32>` per volume - 4 float slots:

- [0] = number of components
- [1] = independent components flag (1.0 / 0.0)
- [2] = color mix preset
- [3] = the force-nearest bitmask <- line 1771

What lines 1765–1771 do

```js
let forceNearestMask = 0;
for (let component = 0; component < numComp; component++) {
  if (vprop.getForceNearestInterpolation(component)) {
    forceNearestMask |= 1 << component;
  }
}
componentInfoArray[vidx * 4 + 3] = forceNearestMask;
```
A volume can have up to 4 scalar components (e.g. R, G, B, A). Each component individually can be flagged to use nearest neighbor texture sampling instead of the default linear interpolation (ForceNearestInterpolation).

Rather than send 4 separate booleans, the loop packs them into a single integer bitmask, one bit per component:

- bit 0 set -> component 0 forces nearest
- bit 1 set -> component 1 forces nearest
- etc.

1 << component produces the bit for that component (1, 2, 4, 8), and |= ORs it into the accumulator. So if components 0 and 2 want nearest, the mask is 0b0101 = 5.

That single number is stored in the 4th slot of componentInfo. On the GPU, the shader (in getTextureValue) unpacks it - checking whether bit N is set - to decide per component whether to do a nearest fetch and override the interpolated value.

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
daker: `getTextureValue` filters in the shader because VTK.js allows forcing individual components to nearest interpolation using vtkVolumeProperty.setForceNearestInterpolation(index, value), and a GPU sampler can only apply one filter mode to all four components at once, so the shader samples linearly, re-samples at the texel center, and mixes the two according to the bitmask. It does not duplicate filtering done elsewhere; volumes that never set the force-nearest mask still get all their filtering from the fixed-function sampler alone.

Sources/Rendering/WebGPU/VolumePassFSQ/index.js
daker: | Config | WebGPU (this PR) | WebGL |
|---|---|---|
| LAO off | 18.7 ms/frame | 16.7 ms/frame (vsync capped) |
| LAO on, kernel 15x7 (defaults) | 37.7 ms | 31.3 ms |
| LAO on, kernel 32x16 (max) | 129.8 ms | 111.5 ms |

Yes, with AI help. The loop is only reached when the user enables `localAmbientOcclusion` on the volume property; with it off the function early-returns before the loop and the frame time equals the pre-LAO baseline (18.7 ms). With LAO at the default kernel (15 rays x 7 steps) the frame goes from 18.7 ms to 37.7 ms, and at the maximum kernel (32x16) to 129.8 ms. That is a 2.0x / 6.9x cost, which matches the existing OpenGL implementation of the same nested loop on the same scene (16.7 -> 31.3 ms and 111.5 ms, i.e. 1.9x / 6.7x before accounting for the vsync-capped baseline). So the overhead is real but opt-in, scales with the user-chosen kernel size, and is at parity with the WebGL backend rather than a WebGPU-specific regression. The inner loop also exits early on volume-bounds and on visibility saturation, so the worst case kernelSize x kernelRadius fetch count is rarely reached.

[Benchmark Script](https://github.com/user-attachments/files/29804681/index.js)

sankhesh (approved):
LGTM

Examples/Rendering/LabelmapEdgeProjection/index.js
```diff
@@ -296,6 +320,7 @@ const reader = vtkHttpDataSetReader.newInstance({ fetchGzip: true });
 reader.setUrl(`${__BASE_PATH__}/data/volume/LIDC2.vti`).then(() => {
   reader.loadData().then(() => {
     imageData = reader.getOutputData();
+    baseScalarData = imageData.getPointData().getArray(0).getData().slice();
```
finetjul: if needed, vtkDataArray has a `deepCopy()` feature...

finetjul (approved):
LGTM

PR #3531, opened 2026-06-10
Compose a series of monotonic piecewise functions and a color transfer function into a single color transfer function

jadh4v:
### Context

Imaging applications may require multiple scalar transforms on pixel data before it is rendered. These transforms can be done directly on the image pixel array before passing the image to the GPU. However, the transform can necessitate a data type conversion to higher number of bytes to maintain precision. This can consume a lot of memory and make pixel arrays very large if the application is handling many such images simultaneously.

We implement a feature to compose multiple scalar transforms (piecewise functions) and one color transfer function into a single color transfer function that represents the entire transform from original input scalars to output display colors. This allows the transforms to be applied on the GPU via the composed transfer function's lookup at render time, rather than storing the transformed pixel arrays inflated on the CPU side.

### Results

We have added an example that shows how three different scalar transforms (e.g. modality transforms, values-of-interest, color-window adjustments, etc) can be composed along with a color function.

image

### Changes

- **`vtkPiecewiseFunction.findX(y)`** (new method): inverse of `getValue()`. Locates the first segment whose endpoint values bracket `y` and bisects it through forward evaluation, so midpoint and sharpness are honored. Returns `null` when no `x` maps to `y` (e.g. out-of-range values with clamping off).
- **`compose(fnList, colorFn, outputFn, errorThreshold?, maxSubdivisionDepth?)`** (new helper in `Common/DataModel/PiecewiseFunction/helpers`): composes an ordered chain of piecewise value-transform functions with a final-stage color transfer function into a single output color transfer function. Breakpoints from every stage are pulled back through the upstream inverses (`findX`) into the source domain, then forward-propagated to sample the composed color at each one.
  - Transform functions must be monotonic (`getType() !== 'Varied'`) so breakpoints can be pulled back unambiguously. Without an `errorThreshold` they must also be piecewise linear (midpoint 0.5, sharpness 0 on every segment), since the output interpolates linearly between the collected breakpoints; with an `errorThreshold`, shaped (non-linear) segments in any stage are captured instead by recursively subdividing each output segment at its midpoint until linear interpolation matches the true composed color to within the threshold on every color channel. The subdivision is depth-first, so the threshold alone cannot bound the node count: `maxSubdivisionDepth` (default 24) caps the worst case at 2^depth − 1 extra nodes per segment and guarantees termination at discontinuities. The color transfer function itself is never restricted.
  - Returns the maximum measured midpoint deviation between the composed output and the true composed chain (0 means exact; a value above `errorThreshold` means the depth cap stopped refinement somewhere), or `null` when an input is rejected (with a warning, leaving `outputFn` untouched). Note that 0 is falsy: test the result against `null` for success, never its truthiness.
  - Usage: `compose([modalityFn, voiFn, userFn], colorFn, resultFn, 0.01)` — see the new example.
- **`vtkColorTransferFunction.getDataPointer()`** (new method): returns the function's nodes, for parity with the VTK/C++ API.
- **`vtkPiecewiseFunction` typings fixed**: `getNodeValue`/`setNodeValue` were declared to return `void` but return `-1` (index out of range) or `1` (success); `val` narrowed from `any[]` to `number[]`.
- **New example** `Rendering/ComposePiecewiseFunctions`: DICOM slice viewer chaining modality, VOI, and interactive window/level transforms with a color map preset into a single transfer function.

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**:  latest master
  - **OS**:  macOS 26.5.1 (25F80) Tahoe
  - **Browser**:  Chrome Version 149.0.7827.55 (Official Build) (arm64)


Examples/Rendering/ComposePiecewiseFunctions/index.js
```diff
@@ -0,0 +1,603 @@
...
+  camera.setPosition(cx - normal[0], cy - normal[1], cz - normal[2]);
+  camera.setViewUp(0, -1, 0);
+  renderer.resetCamera();
+}
+
+// ----------------------------------------------------------------------------
+// UI helpers
+// ----------------------------------------------------------------------------
+
+const body = document.querySelector('body');
```
daker, line 162: We do use lil-gui now, check other examples

Sources/Common/DataModel/PiecewiseFunction/index.js
```diff
@@ -338,6 +338,41 @@ function vtkPiecewiseFunction(publicAPI, model) {
...
+      if (y >= minY && y <= maxY) {
+        return x0 + ((y - y0) / (y1 - y0)) * (x1 - x0);
+      }
+    }
+    /* eslint-enable no-continue */
+
+    if (model.clamping && nodes.length > 0) {
+      const first = nodes[0];
+      const last = nodes[nodes.length - 1];
+      if (y <= first.y) {
```
finetjul: If your piecewise function is a "decreasing function", you won't have the expected result here. Maybe you could: 
 - check what y is higher between the first and the last point
 - or better: check the tangent at the first/last points.

(please create a unit test to enforce this scenario)

Sources/Common/DataModel/PiecewiseFunction/index.js
```diff
@@ -338,6 +338,41 @@ function vtkPiecewiseFunction(publicAPI, model) {
...
+  // getValue(x) === y, walking each segment and linearly interpolating
+  // within the first matching one. Returns null if y is outside the
+  // function's output range and clamping is off.
+  publicAPI.findX = (y) => {
+    const { nodes } = model;
+    /* eslint-disable no-continue */
+    for (let i = 0; i < nodes.length - 1; i++) {
+      const { x: x0, y: y0 } = nodes[i];
+      const { x: x1, y: y1 } = nodes[i + 1];
+      if (y0 === y1) {
```
finetjul: you claim that you will return the "first" maching one, but if y is == y0, then you won't return the first matching x.
(Please create a unit test to check this scenario.)

Sources/Common/DataModel/PiecewiseFunction/helpers.js
```diff
@@ -0,0 +1,67 @@
+function getColorFunctionXValues(cfun) {
+  const ret = [];
+  const v = [0, 0, 0, 0, 0, 0]; // [x, r, g, b, midpoint, sharpness]
+  for (let i = 0; i < cfun.getSize(); i++) {
+    if (cfun.getNodeValue(i, v) === 1) {
```
finetjul: can you please take the opportunity to fix index.d.ts to document the return value type of `getNodeValue`

Sources/Common/DataModel/PiecewiseFunction/helpers.js
```diff
@@ -0,0 +1,67 @@
+function getColorFunctionXValues(cfun) {
+  const ret = [];
```
finetjul: for performance reason, you could allocate it to the size of `cfun.getSize()`

Sources/Common/DataModel/PiecewiseFunction/helpers.js
finetjul: for performance reason, I wouldn't bother check the return value...

Sources/Common/DataModel/PiecewiseFunction/index.js
```diff
@@ -338,6 +338,41 @@ function vtkPiecewiseFunction(publicAPI, model) {
...
+        continue;
+      }
+      const minY = Math.min(y0, y1);
+      const maxY = Math.max(y0, y1);
+      if (y >= minY && y <= maxY) {
+        return x0 + ((y - y0) / (y1 - y0)) * (x1 - x0);
```
finetjul: here you ignore midpoint and sharpness and assume it is linear but it may not

Sources/Common/DataModel/PiecewiseFunction/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+        if (x == null) break;
+      }
+      if (x != null) xSet.add(x);
+    }
+  });
+
+  // Also reverse compute from x-values of the final-stage color transfer function,
+  // and add those to our xSet so that we don't miss any break points defined
+  // within the color transfer function.
+  const colorXs = getColorFunctionXValues(colorFn);
```
finetjul, line 92: why not create a `getDataPointer` convenient method for a color function ? that would bring some consistency...

Sources/Common/DataModel/PiecewiseFunction/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Compose a chain of piecewise (value) transform functions and a color
+ * transfer function into a single output color transfer function.
+ *
+ * Collects all x-positions across all transform functions and chains their
+ * outputs through to the color function, producing equivalent break points
+ * in the composed result. h(g(f(x))): g's x-values live in f's output domain
+ * and must be pulled back through f-inverse; h's x-values need g-inverse then
+ * f-inverse, and so on.
```
finetjul: you assume piecewise functions to be linear between nodes, but they may not.

finetjul (changes requested):

finetjul:
Did you check ColorTransferFunction/CssFilters.js ? Maybe it already does what you need...

Sources/Common/DataModel/PiecewiseFunction/index.js
jadh4v: Great catch. I completely missed that.

Examples/Rendering/ComposePiecewiseFunctions/index.js
```diff
@@ -0,0 +1,603 @@
...
+
+/**
+ *
+ * @param {vtkPiecewiseFunction} fn
+ * @returns Output range of the given function as a tuple, using its maximum input range.
+ */
+function getOutputRange(fn) {
+  const inputRange = fn.getRange();
+  return [fn.getValue(inputRange[0]), fn.getValue(inputRange[1])];
+}
```
sankhesh, line 78: This assumes the min and max output values always occur at the end points of the input range. Does vtkPiecewiseFunction guarantee monotonic behavior over [minX, maxX]? If not, getOutputRange() can return an incorrect [minY, maxY] for non-monotonic functions or functions with interior extrema. It might be safer to derive the output range from the function nodes (or otherwise document the monotonicity requirement for callers).

Sources/Common/DataModel/PiecewiseFunction/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+    const data = fn.getDataPointer();
+    if (!data) return;
+    for (let i = 0; i < data.length; i += 2) {
+      let x = data[i];
+      for (let j = idx - 1; j >= 0; j--) {
+        x = fnList[j].findX(x);
+        if (x == null) break;
+      }
+      if (x != null) xSet.add(x);
+    }
```
sankhesh, line 86: Is this behavior guaranteed to be correct? In particular:

What happens when a function is non‑linear between nodes or uses midpoint/sharpness so the interpolation isn’t purely linear?
What happens when the same output value appears on multiple segments (non‑monotonic shapes or plateaus), so there are multiple valid values for a given y?

If it was intentional to always take the first segment that matches  I think we should be document it clearly here and in the findX() docs, since compose() is depending on that choice.

Sources/Common/DataModel/PiecewiseFunction/test/testHelpers.js
```diff
@@ -0,0 +1,126 @@
+import { it, expect } from 'vitest';
```
sankhesh, line 1: I think we are missing a test case here where a non-linear function (or a function with repeated output values on different segments) is added.

Sources/Common/DataModel/PiecewiseFunction/index.js
jadh4v: Fixed by checking minY and maxY with both first and last nodes. Added two tests that use decreasing monotonic functions.

Sources/Common/DataModel/PiecewiseFunction/index.js
jadh4v: Removed this condition so that we always pass the first matching node. This condition was a relic from trying to avoid a divide-by-zero, but now we are doing iterative forward evaluations to invert the function.

Sources/Common/DataModel/PiecewiseFunction/helpers.js
jadh4v: done

Sources/Common/DataModel/PiecewiseFunction/index.js
jadh4v: Fixed by using an iterative forward evaluation approach which doesn't make any assumption about the function except: the function is monotonic between two consecutive nodes.

Sources/Rendering/Core/ColorTransferFunction/index.js
```diff
@@ -1273,6 +1273,11 @@ function vtkColorTransferFunction(publicAPI, model) {
 
     return modifiedInvoked || callModified;
   };
+
+  //----------------------------------------------------------------------------
+  publicAPI.getDataPointer = () => {
+    return model.nodes;
+  };
```
jadh4v, line 1280: @finetjul  Is it ok to return a direct reference to `model.nodes` or should we make a copy of the array?

Examples/Rendering/ComposePiecewiseFunctions/index.js
jadh4v: Fixed to explicitly find minY and maxY from the internal nodes.

Sources/Common/DataModel/PiecewiseFunction/helpers.js
jadh4v: Removed the assumption that pieces are linear (we use iterative evaluation of f(x) to find y). This also automatically handles the effect of midpoint and sharpness since we treat f(x) as a monotonic black box between consecutive nodes.

We also now restrict the functionality to monotonic piecewise functions. The compose() function will reject the inputs if they are not monotonic.

While I am confident that the `findX()` now actually works for even non-monotonic piecewise functions, I am not sure if the composition i.e. the compose() function output is actually correct or desirable in that case. Restricting the input functions to monotonic is probably good for the currently desired feature.

Sources/Common/DataModel/PiecewiseFunction/test/testHelpers.js
jadh4v: Added a test case here: https://github.com/Kitware/vtk-js/pull/3531/changes#diff-47a378a2daaafab91d1c7fddc65c27eeb33d07bcc1ab361127c9c148a87f29fcR59

jadh4v:
> Did you check ColorTransferFunction/CssFilters.js ? Maybe it already does what you need...

Interesting. These filters look like direct color transforms (i.e. transforming 5d color vectors [R, G,B, A, 1] whereas we are working mostly with scalar value functions/transforms that are not restricted to linear matrix transforms. If I understand correctly the CssFilters transforms are entirely represented using 5D matrices or a composition of them.

finetjul:
> > Did you check ColorTransferFunction/CssFilters.js ? Maybe it already does what you need...
> 
> Interesting. These filters look like direct color transforms (i.e. transforming 5d color vectors [R, G,B, A, 1] whereas we are working mostly with scalar value functions/transforms that are not restricted to linear matrix transforms. If I understand correctly the CssFilters transforms are entirely represented using 5D matrices or a composition of them.

Indeed, what you want to do is non-linear. Those 5D matrices won't work for your use case. Nevermind

Sources/Rendering/Core/ColorTransferFunction/index.js
finetjul: In VTK.JS world, it can be ok (just for performance reasons), but it is done with 2 separate methods: the default one returns a copy and another method with the same name but suffixed with "ByReference" returns a direct reference. (maybe add a warning in the documentation to not modify the array). You can look at the getArray function in macros.js

finetjul:
@jadh4v please let us know when it is a good time for us to do a code review.

jadh4v:
> @jadh4v please let us know when it is a good time for us to do a code review.

@finetjul @sankhesh The PR is ready for re-review. I have made some significant changes to my previous approach. Primarily taking an iterative numerical approach for computing inverse as well as additional bisection of composed function.

Sources/Common/DataModel/PiecewiseFunction/index.js
```diff
@@ -338,6 +339,75 @@ function vtkPiecewiseFunction(publicAPI, model) {
...
+      const maxY = Math.max(y0, y1);
+      if (y > minY && y < maxY) {
+        // Within a segment, forward evaluation is monotonic from y0 to
+        // y1 regardless of midpoint and sharpness, so bisect until
+        // floating-point convergence.
+        const increasing = y1 > y0;
+        let lo = x0;
+        let hi = x1;
+        let mid = 0.5 * (lo + hi);
+        while (lo < mid && mid < hi) {
```
sankhesh, line 372: Should there be another stop condition (like an epsilon) for *near-flat* slopes?

Examples/Rendering/ComposePiecewiseFunctions/index.js
```diff
@@ -0,0 +1,430 @@
+import '@kitware/vtk.js/favicon';
```
finetjul, line 1: if you add an example, please add a screenshot for the example gallery

Sources/Common/DataModel/PiecewiseFunction/index.js
```diff
@@ -338,6 +339,75 @@ function vtkPiecewiseFunction(publicAPI, model) {
...
+  publicAPI.findX = (y) => {
+    const { nodes } = model;
+    for (let i = 0; i < nodes.length - 1; i++) {
+      const { x: x0, y: y0 } = nodes[i];
+      const { x: x1, y: y1 } = nodes[i + 1];
+      if (y === y0) {
+        return x0;
+      }
+      if (y === y1) {
+        return x1;
```
finetjul, line 356: if the segment is very sharp (i.e. 1), the first x might not be x1, but midpoint.

Sources/Common/DataModel/PiecewiseFunction/helpers.js
```diff
@@ -0,0 +1,158 @@
...
+  // Also reverse compute from x-values of the final-stage color transfer function,
+  // and add those to our xSet so that we don't miss any break points defined
+  // within the color transfer function.
+  const colorXs = getColorFunctionXValues(colorFn);
+  colorXs.forEach((x) => {
+    const t = pullBack(fnList.length - 1, x);
+    if (t != null) xSet.add(t);
+  });
+
+  // Now use the gathered x values to propogate through the entire set of
```
finetjul: propogate =>  propagate

jadh4v:
https://github.com/Kitware/vtk-js/blob/master/Documentation/.vitepress/sidebar.ts is an empty file but is still tracked. It gets updated if you build documentation, and can throw off devs into thinking that it needs to be committed. Should we remove it from being tracked?

Examples/Rendering/ComposePiecewiseFunctions/index.js
jadh4v: added a snapshot

Sources/Common/DataModel/PiecewiseFunction/index.js
jadh4v: Fixed by changing the early termination towards the smooth case, and letting the bisection loop handle the step case.

Sources/Common/DataModel/PiecewiseFunction/helpers.js
jadh4v: fixed

Sources/Common/DataModel/PiecewiseFunction/index.js
jadh4v: The loop termination relies on the x-axis search window convergence (lo, mid, hi) rather than testing for the y-value directly since an exact evaluation to target_y may not actually happen.

finetjul (approved):
LGTM

PR #3535, opened 2026-06-18
docs(BREAKING_CHANGES): note standalone bundles now load as ES modules

PaulHax:
The Vite migration switched standalone apps (e.g. OfflineLocalView, GeometryViewer) from an IIFE classic <script> to a deferred ESM <script type="module">. These apps expose their API as a global assigned by their own source (e.g. `window.OfflineLocalView = { load }`). Because the module is deferred, a consumer that synchronously calls that global at parse time (as trame-vtk did) runs before the module executes and throws. Mitigation: await the module or wait for DOMContentLoaded. Links Kitware/trame-vtk#120.


daker (approved):
LGTM

finetjul (approved):
LGTM

PR #3537, opened 2026-06-18
fix(RenderWindowInteractor): add option to disable preventDefault on wheel

jourdain:
Even if we don't plan to consume the wheel event, we consume it and prevent the user to be able to scroll through the page.


jourdain:
I found a workaround `v_on_wheel="window.scrollBy(0, $event.deltaY)",`

PR #3538, opened 2026-06-25
Fix modeltransform

sankhesh:
### Context

When applying the model transform matrix, it was observed that the matrix was being applied in the view space leading to unintended orientation specific artifacts. This change fixes the order of matrix multiplication for the transform and adds testing to ensure that the transform is applied camera-orientation-agnostically in the world space. 

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/Camera/index.js
+++ b/Sources/Rendering/Core/Camera/index.js
@@ -516,7 +516,7 @@ function vtkCamera(publicAPI, model) {
   publicAPI.getViewMatrix = () => {
     if (model.viewMatrix) {
       if (model.modelTransformMatrix) {
-        mat4.multiply(tmpMatrix, model.viewMatrix, model.modelTransformMatrix);
+        mat4.multiply(tmpMatrix, model.modelTransformMatrix, model.viewMatrix);
         return tmpMatrix;
       }
       return model.viewMatrix;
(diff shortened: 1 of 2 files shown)


Sources/Rendering/Core/Camera/index.js
```diff
@@ -533,7 +533,7 @@ function vtkCamera(publicAPI, model) {
 
     const result = new Float64Array(16);
     if (model.modelTransformMatrix) {
-      mat4.multiply(result, tmpMatrix, model.modelTransformMatrix);
+      mat4.multiply(result, model.modelTransformMatrix, tmpMatrix);
```
finetjul: instead of creating another temporary variable (`result`), can't you reuse `tmpMatrix`?

finetjul (commented):
Why setModelTransformMatrix and getModelTransformMatrix are not generated with `macro.setGet` ?

finetjul (approved):
LGTM

PaulHax:
An good fix here.  Recomend not doing the perf optimziaiotn returning static array and changing tests: https://github.com/PaulHax/vtk-js/pull/30

To do avoid array copy, let getters accept `out=inArray` arg.

sankhesh:
Closing in favor of #3544

PR #3544, opened 2026-07-08
Model transform view matrix fix

PaulHax:
When applying the model transform matrix, it was observed that the matrix was being applied in the view space leading to unintended orientation specific artifacts. This change fixes the order of matrix multiplication for the transform and adds testing to ensure that the transform is applied camera-orientation-agnostically in the world space.

Supersedes https://github.com/Kitware/vtk-js/pull/3538

Review comments here https://github.com/PaulHax/vtk-js/pull/30


finetjul (commented):
LGTM.

You can check vtkCamera.getViewMatrix() usage, there are a few places where you can already pass an argument to avoid new variable instantation.

sankhesh (approved):
LGTM

sankhesh:
@finetjul @PaulHax Can we merge this please?

PR #3546, opened 2026-07-09
fix(interaction): catch pointerlock failures

sankhesh:
The catch handler silently handles pointer lock failures allowing interaction to proceed without causing application level disruption

### Context

When pointer lock fails on an unsupported browser, the application would crash with an error. This change allows the application to proceed by silently catching the failure.

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:


finetjul (approved):
LGTM

PR #3547, opened 2026-07-10
fix(WebGPU): apply projection once in ImageMapper and refresh imageData textures

sedghi:
### Context
The WebGPU `ImageMapper` vertex shader applies `rendererUBO.SCPCMatrix` twice (regression from 7fbc87a98). With close-in cameras — e.g. a parallel camera one unit in front of the slice, as medical viewers set up — the doubly-projected quad falls outside the clip volume and the slice silently renders nothing. The same commit also stores a clip-space position in `vertexSC`, which the clip-plane code expects in stabilized coordinates.

Separately, `TextureManager.getTextureForImageData` keys its texture cache on the scalar array mtime alone, so writing scalars in place and signaling with `imageData.modified()` (the invalidation the OpenGL backend keys texture rebuilds on) keeps returning the stale texture.

Both were found while rendering image stacks on the WebGPU view API from Cornerstone3D.

### Results
- Image slices render again with offset parallel cameras (before: nothing drawn; reproducible with the `ImageMapper` example and `?viewAPI=WebGPU`).
- Clip planes on `ImageMapper` are evaluated in stabilized coordinates again.
- Textures refresh when scalars are updated in place and only `imageData.modified()` is called.

### Changes
- `WebGPU/ImageMapper`: capture `vertexSC` before projecting, apply `SCPCMatrix` once (matches `ImageResliceMapper`/`ImageCPRMapper`).
- `WebGPU/TextureManager`: key the `getTextureForImageData` cache on the newest of the imageData and scalar array mtimes.
- `vitest.config.js`: pass `--enable-unsafe-webgpu` to headless Chromium for `WEBGPU=1` runs (plus the SwiftShader WebGPU adapter on CI) so the WebGPU tests can acquire an adapter.
- No public API changes.
- [x] Documentation and TypeScript definitions were updated to match those changes (none needed — no API changes)

### PR and Code Checklist

### Testing
Two regression tests, one per fix; both fail without the fixes:

```
WEBGPU=1 NO_WEBGL=1 npx vitest run \
  Sources/Rendering/WebGPU/ImageMapper/test/testImageMapperCameraProjection.js \
  Sources/Rendering/WebGPU/TextureManager/test/testImageDataTextureCacheInvalidation.js
```

- [x] Tested environment:
  - **vtk.js**: master (07c3e9a5e)
  - **OS**: macOS 15
  - **Browser**: Chromium (Playwright, headless) and Chrome, Apple Metal adapter

### DIFF
--- a/Sources/Rendering/WebGPU/ImageMapper/index.js
+++ b/Sources/Rendering/WebGPU/ImageMapper/index.js
@@ -665,12 +665,14 @@ function vtkWebGPUImageMapper(publicAPI, model) {
       lines.push('var tcoord : vec3<f32> = (mapperUBO.SCTCMatrix * pos).xyz;');
     }
     lines.push(
+      // Clip planes are evaluated in stabilized coordinates, so capture the
+      // position before it is projected.
+      'output.vertexSC = pos;',
       'pos = rendererUBO.SCPCMatrix * pos;',
       // Match the OpenGL coincident topology constant offset scale (~1 / 2^16 = 0.000016)
       'pos.z = clamp(pos.z - 0.000016 * mapperUBO.CoincidentOffset * pos.w, 0.0, pos.w);',
       'output.tcoordVS = tcoord;',
-      'output.vertexSC = pos;',
-      'output.Position = rendererUBO.SCPCMatrix * pos;'
+      'output.Position = pos;'
     );
     code = vtkWebGPUShaderCache.substitute(
       code,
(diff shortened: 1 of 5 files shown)


daker (approved):
+1 it will fix MultiSliceImageMapper example

sedghi:
if you run the example in main it will be for webgpu view

CleanShot 2026-07-10 at 09 17 48@2x

but in this branch will become

CleanShot 2026-07-10 at 09 18 33@2x

PaulHax:
Nice fix. +2

No test yet for the vertexSC clip-plane space restoration (no clip planes in the scene) but something for later.

PR #3548, opened 2026-07-10
fix(WebGPU): pass depth-equal fragments in the translucent pass

sedghi:
### Context
The order-independent translucent pass depth-tests with `depthCompare: 'greater'` (reversed z) against the opaque pass depth buffer, so translucent fragments that are exactly coplanar with already-rendered opaque geometry are rejected and silently disappear. The common case is a translucent overlay on an image slice — e.g. a segmentation labelmap or fusion layer — which renders on the OpenGL backend (equal depths pass in its translucent pass) but not on WebGPU.

Found while running Cornerstone3D segmentation overlays on the WebGPU view API. Related: #3547 (independent fixes from the same effort).

### Results
- Coplanar translucent geometry (labelmap/fusion overlays on slices) renders on WebGPU, matching the OpenGL backend.
- Regular translucent-behind-opaque occlusion is unchanged (`greater-equal` still rejects strictly farther fragments).

### Changes
- `WebGPU/OrderIndependentTranslucentPass`: `depthCompare: 'greater'` → `'greater-equal'`.
- New regression test: opaque slice + exactly coplanar 50%-opacity overlay; asserts the overlay blends over the slice (renders black without the fix).
- No API changes.
- [x] Documentation and TypeScript definitions were updated to match those changes (none needed — no API changes)

### PR and Code Checklist

### Testing
```
WEBGPU=1 NO_WEBGL=1 npx vitest run \
  Sources/Rendering/WebGPU/OrderIndependentTranslucentPass/test/testCoplanarTranslucentDepth.js
```
Fails without the fix (`expected 0 to be greater than 60`), passes with it. Requires the WebGPU adapter flags for headless Chromium added in #3547's `vitest.config.js` change when running before that lands.

- [x] Tested environment:
  - **vtk.js**: master (07c3e9a5e)
  - **OS**: macOS 15
  - **Browser**: Chromium (Playwright, headless) and Chrome, Apple Metal adapter

### DIFF
--- a/Sources/Rendering/WebGPU/OrderIndependentTranslucentPass/index.js
+++ b/Sources/Rendering/WebGPU/OrderIndependentTranslucentPass/index.js
@@ -178,7 +178,11 @@ function vtkWebGPUOrderIndependentTranslucentPass(publicAPI, model) {
       primitive: { cullMode: 'none' },
       depthStencil: {
         depthWriteEnabled: false,
-        depthCompare: 'greater',
+        // greater-equal (not greater): translucent geometry that is exactly
+        // coplanar with already-rendered opaque geometry (e.g. a labelmap
+        // overlay on an image slice) must not be depth-rejected. Matches the
+        // OpenGL backend, where equal depths pass in the translucent pass.
+        depthCompare: 'greater-equal',
         format: 'depth32float',
       },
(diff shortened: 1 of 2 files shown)


daker (approved):
LGTM

PR #3552, opened 2026-07-16
Fix view proxy render window name

Jo-Byr:
### Context
ProxyManager example did not work.

### Results
The example is fixed and ViewProxy uses apiSpecificRenderWindow instead of _openGLRenderWindow

### Changes
Replace model._openGLRenderWindow by model.apiSpecificRenderWindow in ViewProxy
Fixed argument of render windows' setSize in ProxyManager example

### PR and Code Checklist

### DIFF
--- a/Sources/Proxy/Core/ViewProxy/index.js
+++ b/Sources/Proxy/Core/ViewProxy/index.js
@@ -37,11 +37,11 @@ function vtkViewProxy(publicAPI, model) {
   model.renderer = vtkRenderer.newInstance({ background: [0, 0, 0] });
   model.renderWindow.addRenderer(model.renderer);
 
-  model._openGLRenderWindow = model.renderWindow.newAPISpecificView();
-  model.renderWindow.addView(model._openGLRenderWindow);
+  model._apiSpecificRenderWindow = model.renderWindow.newAPISpecificView();
+  model.renderWindow.addView(model._apiSpecificRenderWindow);
 
   model.interactor = vtkRenderWindowInteractor.newInstance();
-  model.interactor.setView(model._openGLRenderWindow);
+  model.interactor.setView(model._apiSpecificRenderWindow);
 
   model.interactorStyle3D = vtkInteractorStyleManipulator.newInstance();
   model.interactorStyle2D = vtkInteractorStyleManipulator.newInstance();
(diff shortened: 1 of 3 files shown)


finetjul (commented):
Can you please explain why the name of the variable is problem ?

Why not keeping it protected? (with the _ prefix in the name)

Jo-Byr:
> Can you please explain why the name of the variable is problem ?
> 
> Why not keeping it protected? (with the _ prefix in the name)

It's not the name directly but the setMacro, there were error thrown due to getApiSpecificRenderWindow not being found.

I removed the '_' to uniformize with other variables but if there is a reason to leave it, I'll put it back.

Sources/Proxy/Core/ViewProxy/index.js
finetjul: Replace `openGLRenderWindow` with `apiSpecificRenderWindow`

Sources/Proxy/Core/ProxyManager/example/index.js
```diff
@@ -70,7 +70,7 @@ const view3DProxy = proxyManager.createProxy('Views', 'View3D');
 view3DProxy.setContainer(view3DContainer);
 view3DProxy
   .getApiSpecificRenderWindow()
-  .setSize([view3DContainer.clientWidth, view3DContainer.clientHeight]);
+  .setSize(view3DContainer.clientWidth, view3DContainer.clientHeight);
```
finetjul: "array" setters should support both signatures (passing an array or passing all the arguments). Please fix the problem at its root.

finetjul (commented):
Yes, please keep the variable "protected" (to prevent infinite loop).

finetjul:
> It's not the name directly but the setMacro, there were error thrown due to getApiSpecificRenderWindow not being found.

Your commit message should not be "replace openGLRenderWindow by apiSpecificRenderWindow" because 
 1) it does not bring additional information than reading the diff 
 2) it does not carry the information about what your commit brings to the "user" (i.e. why should I want that commit)

but instead "fix(ViewProxy): add missing function getApiSpecificRenderWindow()" (or something similar).

Sources/Proxy/Core/ProxyManager/example/index.js
Jo-Byr: Are you sure ? All definitions of setSize I see take 2 arguments (VolumeController, PiecewiseGaussianFilter, ImageStream, OpenGL and WebGPU RenderWindow) except SUrfaceLICInterface.

finetjul: > Are you sure ?

Yes:
  https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/RenderWindow/index.js#L1396
  https://github.com/Kitware/vtk-js/blob/master/Sources/macros.js#L654

Jo-Byr: So I should add

```
if (Array.isArray(width)) {
  [width, height] = width;
}
```

To all setSize definitions (all 5 listed previously) ?

finetjul:
```
  const superSetSize = publicAPI.setSize;
  publicAPI.setSize = (...args) => {
    const modified = superSetSize(...args);
    if (modified) {
      publicAPI.invokeWindowResizeEvent({ width: model.size[0], height: model.size[1] });
    }
    return modified;
  };
```

or even better:
```
  model._onSizeChanged = () => {
    publicAPI.invokeWindowResizeEvent({ width: model.size[0], height: model.size[1] });
  };
```

finetjul (approved):
LGTM

PR #3555, opened 2026-07-16
feat(RenderWindow): add manageCanvas for externally owned canvases

PaulHax:
When the canvas belongs to another library (MapLibre, three.js, a plain app canvas), vtk.js must not resize or restyle it or take over webglcontextlost/webglcontextrestored handling. manageCanvas (default true, no behavior change) opts out of canvas DOM management.

For example, a vtk.js custom layer sharing MapLibre's canvas syncs its size from the drawing buffer inside MapLibre's frame, so a managed canvas would get its width/height attributes rewritten mid-frame, resetting the drawing buffer and wiping the just-drawn basemap on every resize.

Screenshot captures with an explicit size or scale go through the two-pass resize path, so with manageCanvas=false the requested size is ignored with a warning and the capture happens at the current size instead of corrupting the host canvas.

Supersedes https://github.com/Kitware/vtk-js/pull/3453


finetjul (approved):
LGTM

PR #3557, opened 2026-07-16
Fix webgpu vtktexture cache invalidation

PaulHax:
### Context

PR #3547 (e072af2fe1) fixed the stale-texture cache for `getTextureForImageData`: the shared `_fillRequest` helper derives the cache hash from the scalar array mtime alone, so writing scalars in place and signaling with `imageData.modified()` kept returning the stale GPU texture.

The sibling entry point `getTextureForVTKTexture` has the identical clobber: it seeds the hash with the `vtkTexture` mtime, but for imageData-backed textures `_fillRequest` overwrites it. Streaming new pixels into a texture draped over geometry (`actor.addTexture(...)`), a background, or a skybox silently keeps rendering the old image unless `scalars.modified()` is called directly — the OpenGL backend refreshes in all of these cases.

### Results

- imageData-backed `vtkTexture`s refresh when scalars are updated in place and the change is signaled through `imageData.modified()` or `texture.modified()` alone (before: the WebGPU view kept the stale texture until `scalars.modified()`; WebGL updated).
- Non-imageData texture sources (image, jsImageData, imageBitmap, canvas) are unaffected — their hash already used the texture mtime.
- Failing texture-cache assertions now report their message instead of a stack overflow: `expect(...).toBe(vtkTexture)` made the reporter serialize the circular device–texture graph through `toJSON()`/`getState()` until `RangeError: Maximum call stack size exceeded` masked the actual failure.
- Device-level WebGPU tests no longer need `try`/`finally` cleanup: `createWebGPUTestDevice` destroys its device when the test finishes, pass or fail.

### Changes

- `WebGPU/TextureManager`: `getTextureForVTKTexture` hashes cache entries on the newest of the texture, imageData, and scalar array mtimes, mirroring e072af2fe1.
- New regression test `TextureManager/test/testVTKTextureCacheInvalidation.js`.
- `TextureManager/test/testImageDataTextureCacheInvalidation.js`: compare cached texture identities as booleans so assertion failures stay readable.
- `Testing/testUtils.js`: `createWebGPUTestDevice` registers `onTestFinished(() => handle.destroy())`; removed per-test device cleanup from the device-level tests (destroying a device destroys every resource created from it).
- No public API changes.

### PR and Code Checklist

### Testing

```
WEBGPU=1 npx vitest run \
  Sources/Rendering/WebGPU/TextureManager \
  Sources/Rendering/WebGPU/Texture/test/testWriteSubImageData.js
```

The new regression test fails without the fix (`imageData.modified() alone must invalidate the cached texture: expected true to be false`). Note the WebGPU tests are skipped unless `WEBGPU=1` is set, so CI does not exercise them yet.

- [x] Tested environment:
  - **vtk.js**: master (ca498ee75)
  - **OS**: Linux (WSL2 Ubuntu 24.04)
  - **Browser**: Chromium (Playwright, headless)

### DIFF
--- a/Sources/Rendering/WebGPU/Texture/index.js
+++ b/Sources/Rendering/WebGPU/Texture/index.js
@@ -155,7 +155,7 @@ function vtkWebGPUTexture(publicAPI, model) {
   };
 
   publicAPI.create = (device, options) => {
-    model.device = device;
+    model._device = device;
     model.width = options.width;
     model.height = options.height;
     model.depth = options.depth ? options.depth : 1;
@@ -173,7 +173,7 @@ function vtkWebGPUTexture(publicAPI, model) {
       : GPUTextureUsage.TEXTURE_BINDING | GPUTextureUsage.COPY_DST;
     /* eslint-enable no-undef */
     /* eslint-enable no-bitwise */
-    model.handle = model.device.getHandle().createTexture({
+    model.handle = model._device.getHandle().createTexture({
       size: [model.width, model.height, model.depth],
       format: model.format, // 'rgba8unorm',
       usage: model.usage,
@@ -184,7 +184,7 @@ function vtkWebGPUTexture(publicAPI, model) {
   };
 
   publicAPI.assignFromHandle = (device, handle, options) => {
-    model.device = device;
+    model._device = device;
     model.handle = handle;
     model.width = options.width;
     model.height = options.height;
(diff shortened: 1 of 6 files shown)


finetjul (approved):
LGTM

FYI You can avoid toJSON()/getState() circular device-texture graph by using "protected" `model` member variables. To do so, you need to prefix their names with '_' or use `macros.makeProtected()`.

finetjul (approved):
LGTM

PaulHax:
> FYI You can avoid toJSON()/getState() circular device-texture graph by using "protected" `model` member variables.

Thanks, vtkWebGPUTexture holds _device now, so getState() skips it and the cycle never forms; getDevice()/setDevice() are unchanged. Rewrote the branch to drop my boolean-comparison workaround, so the tests are back to plain toBe().

PR #3558, opened 2026-07-16
Avoid GL readbacks for perf 

PaulHax:
Two hot-path `gl.getParameter` readbacks of dynamic GL state removed.

### Context

Reading dynamic state back is a synchronization point with the GPU process. On an uncontended context browsers answer from CPU-side state and the query is nearly free; standalone vtk.js scenes show no measurable change from this PR. But when vtk.js shares a context with a busy host (e.g. rendering inside a MapLibre custom layer), Chromium answers each query with a GPU-process round-trip (~0.2 ms), and the first query of the frame additionally blocks until the host's queued GPU work drains (20-45 ms measured mid-frame over a terrain-heavy map, spikes past 100 ms). A per-draw query keeps re-paying smaller stalls as each draw refills the queue.

Demo, no vtk.js involved ([`gl-readback-stall-demo` branch](https://github.com/PaulHax/vtk-js/blob/gl-readback-stall-demo/getparameter-stall-demo.html), one self-contained HTML file; serve it and open): a MapLibre terrain page whose custom layer just issues N × `getParameter(DEPTH_WRITEMASK)` per frame, the pattern vtk.js's draw path had. On Chromium (ANGLE D3D12): N=0 → 30 fps; N=200, i.e. a 200-actor scene before this PR → 12 fps, with the first query of each frame costing 43 ms p50 / 137 ms max.

### Changes

- **Helper:** `drawArrays` read `DEPTH_WRITEMASK` unconditionally on every draw call, but only the point-picking pass uses the value. The readback (and restore) is now scoped to `pointPicking`; ordinary rendering issues no query.
- **Framebuffer:** `saveCurrentBindings` queried `FRAMEBUFFER_BINDING` on every save (e.g. each ForwardPass zbuffer capture). `vtkOpenGLRenderWindow` now carries a JS-side mirror (`set`/`getFramebufferBinding`, gated by an explicit `getFramebufferBindingKnown()` flag). The mirror is inert until first seeded, so stock behavior is byte-identical. Once a caller seeds it (e.g. an embedding host that knows what it bound), saves use the mirror and `vtkFramebuffer` bind/restore keep it in sync. A unit test covers both modes, including nested save/bind/restore through the mirror.

### Results

A/B benchmark (200-actor scene in a MapLibre terrain custom layer, continuous camera motion, Chromium/ANGLE D3D12; harness pairs this branch with the external-context render window follow-up): mean vtk render time **79 ms → 14 ms** per frame, whole-frame **91 ms → 35 ms** (11 → 29 fps). The measured win is dominated by the per-draw Helper readback; the Framebuffer mirror removes the queries that remain once an embedding host declares its framebuffer (any single remaining query inherits the whole stall), so the steady-state render path can issue zero readbacks.

### DIFF
--- a/Sources/Rendering/OpenGL/Framebuffer/index.js
+++ b/Sources/Rendering/OpenGL/Framebuffer/index.js
@@ -14,6 +14,16 @@ function vtkFramebuffer(publicAPI, model) {
   // publicAPI.getDrawMode = () => model.context.DRAW_FRAMEBUFFER;
   // publicAPI.getReadMode = () => model.context.READ_FRAMEBUFFER;
 
+  // Keep the render window's JS-side binding mirror in step with binds made
+  // through this class. Inert until someone seeds the tracker (see
+  // OpenGLRenderWindow.setFramebufferBinding).
+  function syncTrackedBinding(binding) {
+    const renderWindow = model._openGLRenderWindow;
+    if (renderWindow && renderWindow.getFramebufferBinding() !== undefined) {
+      renderWindow.setFramebufferBinding(binding);
+    }
+  }
+
   publicAPI.saveCurrentBindingsAndBuffers = (modeIn) => {
     const mode =
       typeof modeIn !== 'undefined' ? modeIn : publicAPI.getBothMode();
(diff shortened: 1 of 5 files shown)


PaulHax:
superseded by https://github.com/Kitware/vtk-js/pull/3594

I am closing this PR because it combines the measured GL Helper change with a larger public framebuffer-state API.

PR #3562, opened 2026-07-21
chore(RemoteView): add dependencies to example readme

Jo-Byr:
### Context
RemoteView's example fails due to a wslink version mismatch between js and Python. I added a specification of the necessary version in the README of the example.

### Results
The example works locally. But it still does not on the example page since there's no remote view running. We should probably remove it from the page.

### Changes
Added dependencies list in the example's README file

### PR and Code Checklist

### DIFF
--- a/Sources/Rendering/Misc/RemoteView/example/index.md
+++ b/Sources/Rendering/Misc/RemoteView/example/index.md
@@ -3,6 +3,8 @@
 The RemoteView can be used with either the VTK-Web or ParaView-Web servers.
 The current example will not work until you start one of the servers below on
 your local machine.
+Both vtk and pv servers require wslink==1.12.4 installed in your python environment.
+For the vtk server, you need to install vtk[web] as well.
 
 ## Using VTK as server
 


jourdain:
Maybe, we should just bump wslink dependency in vtk.js so that the latest version of wslink would work everywhere?

https://www.npmjs.com/package/@kitware/wslink => `2.5.7`

jourdain:
The issue here is that the example get built with `wslink<2` and therefore force the server to have a `wslink<2`. But if the example was built with `wslink>=2`, then the python will just work because the latest wslink will be picked up.

Jo-Byr:
> we should just bump wslink dependency 

Is updating the version enough or are there changes to do in WSLinkClient ? wslink does not seem to used anywhere else than this class and RemoteView, and the latter works fine with a simple version bump

jourdain:
updating the version should be enough. The "breaking change" is internal and just impose the matching (1 vs 2) version on the server.

Jo-Byr:
I created a [PR](https://github.com/Kitware/vtk-js/pull/3577) to bump wslink, I'll close this PR.

PR #3573, opened 2026-07-24
fix(PaintWidget): fix widget deactivation when not painting

Jo-Byr:
### Context
It was impossible to use paint mode in PaintWidget example.

### Results
The problem is fixed.

### Changes
Replaced macro.VOID return by macro.EVENT_ABORT return in handleEvent to avoid call to deactivateAllWidgets in macros invoke

### PR and Code Checklist

### DIFF
--- a/Sources/Widgets/Widgets3D/PaintWidget/behavior.js
+++ b/Sources/Widgets/Widgets3D/PaintWidget/behavior.js
@@ -76,7 +76,7 @@ export default function widgetBehavior(publicAPI, model) {
         model.activeState.get('origin', 'up', 'right', 'direction', 'scale1')
       );
     } else {
-      return macro.VOID;
+      return macro.EVENT_ABORT;
     }
 
     publicAPI.invokeInteractionEvent();


daker:
This was changed in https://github.com/Kitware/vtk-js/pull/3000 maybe you need to check the usecase

Sources/Widgets/Widgets3D/PaintWidget/behavior.js
```diff
@@ -76,7 +76,7 @@ export default function widgetBehavior(publicAPI, model) {
         model.activeState.get('origin', 'up', 'right', 'direction', 'scale1')
       );
     } else {
-      return macro.VOID;
+      return macro.EVENT_ABORT;
```
finetjul, line 79: if you are not "painting", what are you doing ? Why don't you want other widgets to "use" the event ?

Jo-Byr: We're updating the cursor circle. But `return model.hasFocus ? macro.EVENT_ABORT : macro.VOID;` may be more appropriate.

Jo-Byr:
> This was changed in #3000 maybe you need to check the usecase

The problem was actually introduced in 1d56bbb02c947925c18fd74c0fbe632cc8ebcd80.

With it, if the widget returns VOID, it's deactivated.

finetjul:
> The problem was actually introduced in [1d56bbb](https://github.com/Kitware/vtk-js/commit/1d56bbb02c947925c18fd74c0fbe632cc8ebcd80).

That commit handles the case when you have multiple renderers. Is it your case ? (i.e. do you have an orientation marker widget ?)

That commit might be too strict. Maybe it should `deactivateAllWidgets` ONLY if:
 - there are more than one renderer (quick fix)
 - another renderer widget "eats" the event (better fix)

If I understand well:
 - move event can happen when we are actually painting (model.painting = true) or simply hovering. When hovering, we need to update the stroke circle representation.
 - the mouse cursor position can be invalid (e.g. outside the image to paint, but inside the renderer)
 - the problem of returning `macro.EVENT_ABORT` is that it may prevent other widgets from "using" the event. Say you have a line widget that you would want to "highlight" when moving hover it while having the paint widget active.
 - the problem of deactivating the widget is that maybe we temporarily moved outside a legit place for painting, but we could need, in the next event, to keep painting/updating the cursor. Therefore the widget should not be deactivated when temporarily moving in a non-supported-for-painting position (i.e. world coords is null).
 -  if the paint widget does not have focus, it makes sense to return VOID. But if it has focus (but not painting), it would also make sense to return VOID.

Jo-Byr:
No there's only one renderer.

Replacing the simple `else` by `else if (model._renderWindow.getRenderers()?.length > 1)` does fix the problem when returning VOID, but I don't see how we could check if another renderer "eats" the event.

PR #3581, opened 2026-07-27
Fix interactor style trackball camera mouse wheel event

Jo-Byr:
### Context

InteractorStyleTrackballCamera's mouseWheel to dolly factor formula has 2 issues:
1) It's not symetrical (f(x) != 1/f(-x)) which makes a chain of zoom + dezoom different from a no-op.
2) It prevents dezooming at zoomFactor === 1 (the current formula is 1 - spin / zoomFactor, which returns 0 0 for zoomFactor = 1 and x = 1, which is the value of a mouseWheel dezoom event by default)

### Results
The new formula is symetrical and never reaches 0 to always allow zooming and dezooming.

### Changes
Replaced the formula `dollyFactor = 1 - spin / zoomFactor` by `dollyFactor = exp(- spin / zoomFactor)`

### PR and Code Checklist

### DIFF
--- a/Sources/Interaction/Style/InteractorStyleTrackballCamera/index.js
+++ b/Sources/Interaction/Style/InteractorStyleTrackballCamera/index.js
@@ -421,7 +421,7 @@ function vtkInteractorStyleTrackballCamera(publicAPI, model) {
 
   //----------------------------------------------------------------------------
   publicAPI.handleMouseWheel = (callData) => {
-    const dyf = 1 - callData.spinY / model.zoomFactor;
+    const dyf = Math.exp(-callData.spinY / model.zoomFactor);
     publicAPI.dollyByFactor(model.getRenderer(callData), dyf);
   };
 
--- a/Sources/Interaction/Style/InteractorStyleTrackballCamera/test/testMouseWheel.js
+++ b/Sources/Interaction/Style/InteractorStyleTrackballCamera/test/testMouseWheel.js
@@ -1,0 +1,53 @@
+import { it, expect } from 'vitest';
+import vtkActor from 'vtk.js/Sources/Rendering/Core/Actor';
+import vtkConeSource from 'vtk.js/Sources/Filters/Sources/ConeSource';
+import vtkMapper from 'vtk.js/Sources/Rendering/Core/Mapper';
+import vtkFullScreenRenderWindow from 'vtk.js/Sources/Rendering/Misc/FullScreenRenderWindow';
+import vtkInteractorStyleTrackballCamera from 'vtk.js/Sources/Interaction/Style/InteractorStyleTrackballCamera';
+
+function setup() {
+  const fullScreenRenderer = vtkFullScreenRenderWindow.newInstance({
+    background: [0.2, 0.3, 0.4],
+  });
+  const renderWindow = fullScreenRenderer.getRenderWindow();
+  const renderer = fullScreenRenderer.getRenderer();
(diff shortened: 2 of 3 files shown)


daker (approved):
LGTM, i let @finetjul do a final check

Sources/Interaction/Style/InteractorStyleTrackballCamera/index.js
```diff
@@ -421,7 +421,7 @@ function vtkInteractorStyleTrackballCamera(publicAPI, model) {
 
   //----------------------------------------------------------------------------
   publicAPI.handleMouseWheel = (callData) => {
-    const dyf = 1 - callData.spinY / model.zoomFactor;
+    const dyf = Math.exp(-callData.spinY / model.zoomFactor);
```
finetjul, line 424: is this the same as in VTK C++ ?

Jo-Byr: No. In C++ this is used

```
this->Dolly(pow(1.1, +/- this->MotionFactor * 0.2 * this->MouseWheelMotionFactor));
```

+/- depending on mousewheel foward or backward. It does not use the event's spin value.

finetjul:
A unit test to enforce that mouse wheel events are symmetrical would be welcomed (in case someone decide to change slightly your formula)

Jo-Byr:
> A unit test to enforce that mouse wheel events are symmetrical would be welcomed (in case someone decide to change slightly your formula)

Does that mean you approve the change ?

finetjul:
> Does that mean you approve the change ?

Yes with a unit test. (I don't know where that "exp(...)" formula is coming from. It seems logical and reasonable but at the same time I wouldn't be surprised if the formula can be improved/changed. Therefore it should be "enforced" that a formula change does not break the rule of zoom "symmetry"

Sources/Interaction/Style/InteractorStyleTrackballCamera/test/testMouseWheel.js
```diff
@@ -0,0 +1,64 @@
...
+
+  interactorStyle.handleMouseWheel({
+    spinY: -1,
+    pokedRenderer: renderer,
+  });
+  interactorStyle.handleMouseWheel({
+    spinY: 1,
+    pokedRenderer: renderer,
+  });
+  console.log(camera.getPosition());
```
finetjul: can be removed.

Sources/Interaction/Style/InteractorStyleTrackballCamera/test/testMouseWheel.js
```diff
@@ -0,0 +1,64 @@
...
+  interactorStyle.handleMouseWheel({
+    spinY: -0.5,
+    pokedRenderer: renderer,
+  });
+
+  expect(
+    camera.getPosition(),
+    'Make sure chaining 2 mouse wheel events of opposite values is equivalent to a no-op.'
+  ).toEqual(position);
+});
```
finetjul, line 53: what about:
  spinY: 0.5 -> spinY: 0.5 ->  spinY: -0.5 -> spinY: -0.5 ? 
  spinY: 0.5 -> spinY: 0.7 ->  spinY: -0.7 -> spinY: -0.5 ? 
  spinY: 0.5 -> spinY: 0.7 ->  spinY: -0.5 -> spinY: -0.7 ?

finetjul (approved):
LGTM

PR #3583, opened 2026-07-27
chore(DepthTest): fix example initial render

Jo-Byr:
### Context
The initial render of the DepthTest example is wrong, the first interaction with the camera makes the plane appear

Initial render

image

After first interaction

image

### Results

The first render is correct and moving the camera modifies it in a coherent way.

image

### Changes
Added resetCamera and render calls.

### PR and Code Checklist


daker (approved):
LGTM

PR #3584, opened 2026-07-27
fix(FPSMonitor): fix FPSMonitor for float numbers

Jo-Byr:
### Context
Replace custom display logic by Number.toLocaleString

### Results
Simplify code.

Fix digit display.

### PR and Code Checklist

### DIFF
--- a/Sources/Interaction/UI/FPSMonitor/index.js
+++ b/Sources/Interaction/UI/FPSMonitor/index.js
@@ -3,18 +3,8 @@ import style from 'vtk.js/Sources/Interaction/UI/FPSMonitor/FPSMonitor.module.cs
 
 const noOp = Function.prototype;
 
-function formatNumbers(n) {
-  const sections = [];
-  let size = n;
-  while (size > 1000) {
-    sections.push(`000${size % 1000}`.slice(-3));
-    size = Math.floor(size / 1000);
-  }
-  if (size > 0) {
-    sections.push(size);
-  }
-  sections.reverse();
-  return sections.join("'");
+function formatNumber(n) {
+  return n.toLocaleString();
 }
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


sankhesh (approved):
LGTM

Sources/Interaction/UI/FPSMonitor/index.js
```diff
@@ -7,9 +7,13 @@ function formatNumbers(n) {
   const sections = [];
   let size = n;
   while (size > 1000) {
-    sections.push(`000${size % 1000}`.slice(-3));
+    sections.push(`000${Math.floor(size) % 1000}`.slice(-3));
     size = Math.floor(size / 1000);
   }
+  if (size % 1 !== 0) {
+    // float
```
finetjul: "float" is not very accurate. "decimal" is a better name.

Sources/Interaction/UI/FPSMonitor/index.js
```diff
@@ -7,9 +7,13 @@ function formatNumbers(n) {
   const sections = [];
   let size = n;
   while (size > 1000) {
```
finetjul: Shouldn't it be `while (size >= 1000) {`

Sources/Interaction/UI/FPSMonitor/index.js
finetjul: To rename into formatNumber ?

finetjul: It's not clear to me why not using `Number.toLocaleString(n)` instead.
`formatNumbers` is slightly faster than `Number.toLocaleString` (or `Intl.NumberFormat().format()`) but `Number.toLocaleString()` makes a better formating and I doubt that performance is very important here.

@jourdain

finetjul: it does not seem to handle negative numbers well

jourdain: Could be fine, I don't remember why we made our own.

finetjul (approved):
LGTM

Issue #2574, opened 2022-09-14
Does vtkClipClosedSurface support lookuptable applyColorMap?


supriome:
I need to clip a Polydata then display it with different ColorMap. But when i use vtkClipClosedSurface, i can only set three types of pure color for clipColor/baseColor/activePlaneColor.

But in the example [GeometryViewer](https://kitware.github.io/vtk-js/examples/GeometryViewer/GeometryViewer.html?fileURL=https://data.kitware.com/api/v1/item/59de9de58d777f31ac641dc5/download), I can set different color to display a vtp modal.

I want to custom my clipped modal's baseColor just like below.

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/19944032/190053648-ea1ae2e2-bccd-4646-b9d8-ef8f0cdc62c0.png">

And show the clipped section as a solid color closed face.

<img width="1750" alt="image" src="https://user-images.githubusercontent.com/19944032/190054052-00a4b51c-6e2f-41b8-812f-2306ff548a45.png">

Can I do this?

supriome:
@DavidBerger98

finetjul:
Hi @supriome , 
with GeometryViewer, you use a point data array for coloring.
with vtkClipClosedSurface, the clipped color are on a cell data array. 
You can't mix both in vtkClipClosedSurface.

One easy thing would be for vtkClipClosedSurface to support 1-component color for basecolor, clipcolor and activeplancecolor. That way, it wouldn't enforce the input polydata to have RGB scalar cell data array, but scalar cell data array of any tuple size.
PR are welcomed :-)

indominus-hub:
I am also trying to find a solution for this...Would it work for a slice instead of a clip?

PR #3585, opened 2026-07-27
chore(ClipClosedSurface): fix surface colors not showing

Jo-Byr:
### Context
Surface colors are incorrect in ClipClosedSurface example

image

### Results
The sphere and plane colors are now correct.

image

### Changes
Added `mapper.setColorModeToDirectScalars();` in the example to use the scalars set by the filter.

### DIFF
--- a/Sources/Filters/General/ClipClosedSurface/example/index.js
+++ b/Sources/Filters/General/ClipClosedSurface/example/index.js
@@ -35,6 +35,7 @@ const actor = vtkActor.newInstance();
 renderer.addActor(actor);
 
 const mapper = vtkMapper.newInstance({ interpolateScalarBeforeMapping: true });
+mapper.setColorModeToDirectScalars();
 actor.setMapper(mapper);
 
 const cam = vtkCamera.newInstance();


daker (approved):
LGTM

finetjul:
Can you please explain why changing to direct scalar mode fixes the problem? Could it be the active attribute field that is not correctly set?

Jo-Byr:
> Can you please explain why changing to direct scalar mode fixes the problem? Could it be the active attribute field that is not correctly set?

From my understanding, it's necessary because we set the colors as the scalars directly, rather than use a lookuptable.

daker:
@Jo-Byr There is a bug related to this i think #2574 ?

Jo-Byr:
To be more complete: 

There's 2 coloring modes: colors and labels

If we are in color mode, we can have only 3 colors: base, clip and active. The example didn't use the colors correctly because we didn't tell the mapper to use the scalars as colors directly, so it tried to use the colors as scalars, reading their first component as a float scalar resulting in the binary blue/red we saw due to saturation of the default lut that maps from 0 to 1 (colors are in uint8 so we could only get both ends).

In label mode, we have the same issue if we don't change the LUT because the filter uses labels 0, 1 and 2 arbitrarily without allowing to change these values:

```
if (model.scalarMode === ScalarMode.COLORS) {
  numberOfScalarComponents = 3;
  createColorValues(
    model.baseColor,
    model.clipColor,
    model.activePlaneColor,
    colors
  );
} else if (model.scalarMode === ScalarMode.LABELS) {
  colors[0][0] = 0;
  colors[1][0] = 1;
  colors[2][0] = 2;
}
```

So if we want to be able to still show scalars, we would need to be able to use default scalars instead of this overwrite for "base" right ?

finetjul (approved):
Understood! In the `ScalarMode.COLORS` mode, the clip filter produces an output "Colors" cell data array with an RGB UInt8 color for each face. Hence the "direct mode" for rendering.

PR #3586, opened 2026-07-27
Fix camera model transform bounds

sankhesh:
### Context

#3538 fixed the composition order so the transform is applied in world space. But the camera pose is *consumed after* the transform: `getViewMatrix()` builds `lookAt(position, focalPoint, viewUp) · M`, so `position`/`focalPoint`/`clippingRange` are interpreted in transformed space, while prop bounds are in world space. Nothing bridged the two.

This is exactly how VTK C++ behaves — `vtkCamera.cxx` composes `ModelView = ViewTransform · ModelTransformMatrix`, and `vtkCamera.h` documents that "Clipping distance is measured in world coordinate unless a scale factor exists in camera's ModelTransformMatrix." The difference is that C++ has `vtkRenderer::ExpandBounds` (`vtkRenderer.cxx`), which pushes bounds through the transform before deriving the camera pose from them, called from `ResetCamera`, `ResetCameraClippingRange` and `ResetCameraScreenSpace`. vtk-js had no equivalent.

While adding it, I found a second, unrelated bug: the model transform was being applied transposed.

### Results

**`resetCamera` / `resetCameraClippingRange`.** Before, both derived the camera pose and clipping range from untransformed bounds. With a 10x Z exaggeration on bounds `[-1000, 1000, -1000, 1000, 90, 110]`, `resetCamera` aimed the camera at world Z 100 while the rendered scene center was at 1000 — the scene was mis-framed and clipped. The error is proportional to `(factor − 1)` and vanishes at 1x, which is why it went unnoticed. After, the camera is aimed at the transformed center and the clipping range covers the transformed scene.

**Transposed model transform.** `getViewMatrix` composed `view · Mᵀ` instead of `view · M`. Probing a camera at `(0,0,10)` looking at the origin, with `M = translate(1, 0, 0)` from `vtkTransform`:

```
eye of world origin, actual : [0, 0, -10]
                    view · M   : [1, 0, -10]
                    view · Mᵀ  : [0, 0, -10]   <- what we were producing
```

A pure scale is symmetric, so `M = Mᵀ` and vertical exaggeration worked either way — but rotations and translations were applied transposed.

### Changes

**Behavior change:** `setModelTransformMatrix` now interprets its argument as gl-matrix column-major, consistent with `vtkTransform`, `vtkMatrixBuilder` and `userMatrix` on `vtkProp3D` — and with what `index.d.ts` already declared (`mat4`). This is a silent change for anyone who passed a rotation or translation and pre-transposed to compensate; pure scales are unaffected.

### PR and Code Checklist

- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Arch Linux
  - **Browser**: Chromium

### DIFF
--- a/Sources/Rendering/Core/Camera/test/testModelTransformMatrix.js
+++ b/Sources/Rendering/Core/Camera/test/testModelTransformMatrix.js
@@ -1,10 +1,25 @@
 import { describe, it, expect, beforeEach } from 'vitest';
-import { mat4 } from 'gl-matrix';
+import { mat4, vec3 } from 'gl-matrix';
 import { areEquals } from 'vtk.js/Sources/Common/Core/Math';
 import vtkCamera from 'vtk.js/Sources/Rendering/Core/Camera';
 import vtkTransform from 'vtk.js/Sources/Common/Transform/Transform';
 
 let camera;
+
+// getViewMatrix returns vtk's row-major order; transpose to get a matrix that
+// can be applied to points with gl-matrix.
+function asPointTransform(rowMajor) {
+  return mat4.transpose(mat4.create(), rowMajor);
+}
+
+// Where does world point p land in eye coordinates?
+function toEye(cam, p) {
+  return vec3.transformMat4(
+    vec3.create(),
+    p,
+    asPointTransform(cam.getViewMatrix())
+  );
+}
 
 describe('Camera Model Transform Matrix', () => {
   beforeEach(() => {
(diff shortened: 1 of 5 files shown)


sankhesh:
@daker @finetjul Could you please review?

daker (approved):
LGTM, shouldn't this be a BREAKING CHANGE ?

daker:
@sankhesh One thing for a follow up, not for this PR. After this change the camera pose (`position`, `focalPoint`, `clippingRange`) is in transformed space, while props and `renderer.worldToView` are in world space. Consumers that read the pose as a world point are wrong when a model transform is set:

- https://github.com/Kitware/vtk-js/tree/master/Sources/Rendering/Core/Picker/index.js#L251 (`publicAPI.pick`): `cameraFP` goes into `renderer.worldToNormalizedDisplay`, and the ray is `pickPosition (world) - cameraPos (transformed)`. Picking selects the wrong cell. The clipping range used for the ray endpoints is also a distance in transformed space.
- https://github.com/Kitware/vtk-js/tree/master/Sources/Interaction/Manipulators/MouseCameraTrackballPanManipulator/index.js#L64: `computeDisplayToWorld` results are added to `setPosition` / `setFocalPoint`, so panning drifts. Same shape in the unicam and dolly to position manipulators. Rotate, roll and zoom stay inside the pose and are fine.
- https://github.com/Kitware/vtk-js/tree/master/Sources/Rendering/Core/Renderer/index.js#L52 (`updateLightsGeometryToFollowCamera`): a head light copies the camera pose, then https://github.com/Kitware/vtk-js/tree/master/Sources/Rendering/OpenGL/PolyDataMapper/index.js:1370 rotates light directions by the view matrix, which now carries `M`, so the transform is applied twice.
- https://github.com/Kitware/vtk-js/tree/master/Sources/Rendering/Core/Follower/index.js#L51, plus `Widgets/Manipulators/AbstractManipulator` and `Widgets/Core/WidgetManager`, which intersect and scale in world space from the pose.

Sources/Rendering/Core/Camera/test/testModelTransformMatrix.js
```diff
@@ -1,53 +1,80 @@
...
       const transform = vtkTransform.newInstance();
       transform.translate(1, 0, 0);
       const modelTransform = transform.getMatrix();
 
+      camera.setModelTransformMatrix(modelTransform);
+
+      // The world origin is moved to (1, 0, 0) by the model transform, so it
+      // must land one unit off the view axis rather than on it.
+      const eye = toEye(camera, [0, 0, 0]);
+      expect(eye[0]).toBeCloseTo(1, 10);
```
finetjul, line 47: use vtkMath.areEquals ?

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -392,13 +405,22 @@ function vtkRenderer(publicAPI, model) {
...
+      boundsToUse,
+      model.activeCamera.getModelTransformMatrix()
+    );
 
-    let w1 = boundsToUse[1] - boundsToUse[0];
-    let w2 = boundsToUse[3] - boundsToUse[2];
-    let w3 = boundsToUse[5] - boundsToUse[4];
+    center[0] = (expandedBounds[0] + expandedBounds[1]) / 2.0;
+    center[1] = (expandedBounds[2] + expandedBounds[3]) / 2.0;
+    center[2] = (expandedBounds[4] + expandedBounds[5]) / 2.0;
```
finetjul, line 419: vtkBoundingBox.getCenter()

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -392,13 +405,22 @@ function vtkRenderer(publicAPI, model) {
...
+
+    let w1 = expandedBounds[1] - expandedBounds[0];
```
finetjul, line 421: vtkBoundingBox.getLength(expandedBounds, 0)

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -35,6 +35,19 @@ function vtkRenderer(publicAPI, model) {
     renderer: publicAPI,
   };
 
+  // Counterpart of vtkRenderer::ExpandBounds: transform the 8 corners of bounds
+  // by matrix and take the axis-aligned bounds of the result.
+  // matrix is in gl-matrix column-major order, matching the camera's
+  // modelTransformMatrix and what vtkBoundingBox.transformBounds expects.
+  // Unlike C++, where ModelTransformMatrix is always allocated, vtk-js leaves it
+  // null by default, so a null matrix is a no-op rather than an error.
```
finetjul, line 43: That seems like a too verbose comment... Maybe it could go into the PR but not really in the code...

daker: maybe also add it to the commit description

PR #3595, opened 2026-08-09
fix(RenderWindowInteractor): preserve first mouse move

PaulHax:
`handleMouseMove` groups pointer moves into bursts with a 200 ms timer. Since 0564e3d2 introduced the burst markers, the first move of a burst fired only `startMouseMoveEvent`, replacing the `mouseMoveEvent` that fired for every move before that change. `MouseMove` subscribers never see the first move after 200 ms of pointer rest:

- Camera styles diff against their stored previous position, so a drag picks the motion up one event late.
- Per-event consumers, such as `WidgetManager` hover picking, miss single moves entirely: one small pointer nudge after rest produces no `MouseMove` at all.

This change fires `mouseMoveEvent` for every move. `startMouseMoveEvent` still marks the burst boundary and the end timer is unchanged. Subscribers now receive both `StartMouseMove` and `MouseMove` for the first event of a burst.

Existing consumers that use `StartMouseMove`/`EndMouseMove` as animation-lifecycle markers keep working unchanged. Consumers of `MouseMove` (Camera styles, `WidgetManager` hover picking) now also see the first move.

A new test in `testChordedButtons.js` asserts the first move of a burst produces `StartMouseMove` then `MouseMove`, and a following move produces only `MouseMove`.

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -756,9 +756,11 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     if (model.moveTimeoutID === 0) {
       publicAPI.startMouseMoveEvent(callData);
     } else {
-      publicAPI.mouseMoveEvent(callData);
       clearTimeout(model.moveTimeoutID);
     }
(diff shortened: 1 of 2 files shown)


finetjul (approved):
Nice catch !
LGTM

PR #3596, opened 2026-08-10
feat(RenderWindowInteractor): add Tap and LongTap gestures

UlysseDurand:
### Context

There is a need for mobile web applications made with vtk-js to have more control interactions, especially because right click isn't available. The LongTap event can replace the right click, Tap comes along the way.

### Results

- Touch/pen pointers now emit dedicated `TapEvent` and `LongTap` events through the standard `handledEvents` mechanism, alongside the existing press/move/release events.
- New configurable options allow tuning gesture detection per application.

### Changes

- `Sources/Rendering/Core/RenderWindowInteractor/index.js`
  - Added `Tap`, and `LongTap` to `handledEvents`.
  - Added gesture tracking (pointer ID, position, timers) for single touch/pen pointers, canceling gestures on multitouch or movement beyond the long-press distance.
- `Sources/Rendering/Core/RenderWindowInteractor/index.d.ts`
  - Added new events to `handledEvents` enum, `invokeTap`, `invokeLongTap` and the corresponding `on*`/`*Event` types, plus the new initial values/getters.
- `Examples/Applications/TapGestureDemo/index.js`
  - New demo application to try the three gestures.
- New configurable values (defaults):
  - `recognizeTapGestures` (true)
  - `longTapDuration` (500 ms)
  - `longTapDistance` (30 px)

### PR and Code Checklist

### Testing

- Demo: run the `TapGestureDemo` example and tap and long-tap with a touch device or pen.
- No unit tests were added: existing Pan/Pinch/Rotate gestures are not unit-tested either.
- [x] Tested example with:
  - **vtk.js**: master
  - **OS**: Linux
  - **Browser**: Chrome on an Android tablet


Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
```diff
@@ -41,6 +41,8 @@ declare enum handledEvents {
   'Interaction',
   'EndInteraction',
   'AnimationFrameRateUpdate',
+  'Tap',
+  'LongTap',
```
finetjul: please add them after "Pan" events

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -68,6 +68,8 @@ const handledEvents = [
   'Interaction',
   'EndInteraction',
   'AnimationFrameRateUpdate',
+  'Tap',
+  'LongTap',
```
finetjul: to move after "Pan" events

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -84,6 +86,10 @@ function pointerCacheToPositions(cache) {
   return positions;
 }
 
+function distanceBetweenPositions(a, b) {
+  return Math.sqrt((a.x - b.x) * (a.x - b.x) + (a.y - b.y) * (a.y - b.y));
```
finetjul: don't compute sqrt, it is expensive: distanceBetweenPositions --> distance2BetweenPositions.

It can be a separate commit (because you would need to consider the pinchDistance, rotateDistance and panDistance as well)

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,67 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
 
+  // Tap / double-tap / long-press gesture tracking for touch and pen pointers.
+  let tapPointerId = null;
+  let tapStartPosition = null;
+  let longTapTimer = null;
+  let longTapFired = false;
+  let tapGestureActive = false;
+
+  function cancelTapGesture() {
+    if (longTapTimer !== null) {
```
finetjul: `longTapTimer != null`  to support `longTapTimer=undefined`

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,67 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+    }
+    tapPointerId = null;
+    tapStartPosition = null;
+    longTapFired = false;
+    tapGestureActive = false;
+  }
+
+  function startLongTapTimer() {
+    if (longTapTimer !== null) {
+      clearTimeout(longTapTimer);
```
finetjul: longTapTimer = null

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,67 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+      longTapTimer = null;
+    }
+    tapPointerId = null;
+    tapStartPosition = null;
+    longTapFired = false;
+    tapGestureActive = false;
+  }
+
+  function startLongTapTimer() {
+    if (longTapTimer !== null) {
```
finetjul: longTapTimer != null

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,67 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+      clearTimeout(longTapTimer);
+    }
+    longTapTimer = setTimeout(() => {
+      longTapTimer = null;
+      if (tapGestureActive && tapPointerId !== null) {
```
finetjul: why trigger the timeout in the first place if tapGesture is not active ?

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,67 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+          position: tapStartPosition,
+          deviceType: 'touch',
+        };
+        publicAPI.longTapEvent(callData);
+      }
+    }, model.longTapDuration);
+  }
+
+  function finalizeTapGesture(event) {
+    if (!tapGestureActive) {
```
finetjul: don't call `finalizeTapGesture()` if `tapGestureActive` is `false`

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -893,6 +962,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     const pointers = [...pointerCache.values()];
     // If multitouch
     if (model.recognizeGestures && pointers.length > 1) {
+      cancelTapGesture();
```
finetjul: `cancelTapGesture()` is called in the `if` and in the `else`, maybe it can be factorized then

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,67 @@ function vtkRenderWindowInteractor(publicAPI, model) {
   // another button is held are signaled via pointermove with updated `buttons`.
   let previousMouseButtons = 0;
 
+  // Tap / double-tap / long-press gesture tracking for touch and pen pointers.
+  let tapPointerId = null;
+  let tapStartPosition = null;
+  let longTapTimer = null;
+  let longTapFired = false;
+  let tapGestureActive = false;
```
finetjul: Move those variables in a unique JS Object, e.g. `let tapObject = {pointId, startPosition, longTimer, longTapFired} | undefined | null;`

if `tapObject != null`, it means it is "active", you do not need a variable for that

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -1331,6 +1420,9 @@ const DEFAULT_VALUES = {
   preventDefaultOnPointerDown: false,
   preventDefaultOnPointerUp: false,
   mouseScrollDebounceByPass: false,
+  recognizeTapGestures: true,
+  longTapDuration: 500,
+  longTapDistance: 30,
```
finetjul: longTapDistance -> `longTapMaximumDistance`, `longTapDistanceThreshold` `longTapMoveThreshold`

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,67 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+    if (!tapGestureActive) {
+      return;
+    }
+    if (longTapTimer !== null) {
+      clearTimeout(longTapTimer);
+      longTapTimer = null;
+    }
+    tapGestureActive = false;
+    tapPointerId = null;
+    if (longTapFired) {
```
finetjul: if you delete `tapObject` when you fire a long tap, you wouldn't have to keep the information of "fired" or not.

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
```diff
@@ -1025,6 +1083,25 @@ export interface vtkRenderWindowInteractor extends vtkObject {
    */
   setRecognizeGestures(recognizeGestures: boolean): boolean;
 
+  /**
+   * Enable/Disable recognition of tap and long-press gestures.
+   * @param recognizeTapGestures
+   */
+  setRecognizeTapGestures(recognizeTapGestures: boolean): boolean;
```
finetjul: why do you want to add Tap gesture recognition granularity ? 
Or said differently, by recognizing tap gestures, do you change the behavior of regular gestures ? do you decrease performance ?

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,67 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+  let longTapFired = false;
+  let tapGestureActive = false;
+
+  function cancelTapGesture() {
+    if (longTapTimer !== null) {
+      clearTimeout(longTapTimer);
+      longTapTimer = null;
+    }
+    tapPointerId = null;
+    tapStartPosition = null;
```
finetjul: can't you reuse `startingEventPositions` ?

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -84,6 +86,10 @@ function pointerCacheToPositions(cache) {
   return positions;
 }
 
+function distanceBetweenPositions(a, b) {
```
finetjul: you can take the opportunity in the file to call this function where the math is done manually (e.g. pinchDistance, rotateDistance and panDistance) . It can be a separate commit.

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -911,16 +981,34 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+        startLongTapTimer();
+      }
       publicAPI.leftButtonPressEvent(callData);
     }
   };
 
   publicAPI.handleTouchMove = (event) => {
     const pointers = [...pointerCache.values()];
     if (model.recognizeGestures && pointers.length > 1) {
+      cancelTapGesture();
```
finetjul: should probably have been already cancelled when a second pointer was made down.

finetjul:
Can you consider writing a test ?

e.g. https://github.com/Kitware/vtk-js/pull/3595/changes#diff-ce1dbea0013f09679bf8d00604fc66d9da245c796ed66f0440467f9c123d2435R147

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
```diff
@@ -186,6 +190,21 @@ export interface vtkRenderWindowInteractor extends vtkObject {
    */
   getMouseScrollDebounceByPass(): boolean;
 
+  /**
+   * @default true
+   */
+  getRecognizeTapGestures(): boolean;
```
finetjul: Does not seem needed anymore

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,40 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+  function cancelTapGesture() {
+    if (tapObj.timer != null) clearTimeout(tapObj.timer);
+    tapObj.timer = null;
+  }
+
+  function startTapTimer(pointerId) {
+    cancelTapGesture();
+    tapObj.pointerId = pointerId;
+    tapObj.timer = setTimeout(() => {
+      cancelTapGesture();
```
finetjul: I would call it after longTapEvent

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -891,21 +933,36 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+            deviceType: getDeviceTypeFor(event),
+          };
+          publicAPI.leftButtonReleaseEvent(callData);
+        }
+        // handle the gesture
+        publicAPI.recognizeGesture('TouchStart', positions);
+      } else if (pointers.length === 1) {
+        preventDefault(event);
+        cancelTapGesture();
+        startTapTimer(event.pointerId);
```
finetjul, line 961: startTapTimer will cancelTapGesture already, you don't have to call it here also

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,40 @@ function vtkRenderWindowInteractor(publicAPI, model) {
   // another button is held are signaled via pointermove with updated `buttons`.
   let previousMouseButtons = 0;
 
+  // Tap / LongTap gesture tracking for touch and pen pointers.
+  let tapObj = { timer: null, pointerId: null };
```
finetjul: tapObj -> tapInformation

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+afterEach(() => {
+  vi.useRealTimers();
+});
+
+// Tests ------------------------------------------------------------------
+
+function tapTest(container, interactor, time, distance, expected) {
+  const finalX = 100 + distance / Math.sqrt(2);
```
finetjul: move 100 into its own variable:
const origX = 100;
const finalX = origX + distance / Math.sqrt(2);

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
```diff
@@ -0,0 +1,115 @@
...
+  subs.forEach((s) => s.unsubscribe());
+}
+
+it('Test RenderWindowInteractor handles taps well', () => {
+  const env = setupInteractor();
+  const { container, interactor } = env;
+
+  tapTest(container, interactor, 490, 28, ['Tap']);
+  tapTest(container, interactor, 490, 32, []);
+  tapTest(container, interactor, 510, 28, ['LongTap']);
```
finetjul: test when adding an additional touch
test when restarting a tap (and make sure a longtap still works)

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -1132,12 +1191,10 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     // of movement it is and then deal with it.
     // calculate the distances
     const originalDistance = Math.sqrt(
```
finetjul, line 1193: I think you can work with `originalDistance2`, you don't need the sqrt

UlysseDurand: I think I need `originalDistance` (calculated with the sqrt),

Let $d_1$ be `originalDistance`, $d_2$ be `newDistance`, $T$ be `thresh`
(so $d_1^2$ is `originalDistance2`, $d_2^2$ is `newDistance2` and $T^2$ is `thresh2`)

The test to see if the gesture is a pinch is $|d_2-d_1| < T$ ($\iff |d2-d1|^2 < T^2$ as in the code)
I can't figure out a way of having an equivalent test using $d_1^2$ and $d_2^2$ only, $d_1$ and $d_2$ are necessary.

Plus, if we only had $d_1^2$ and $d_2^2$, we would still need a sqrt to calculate the pinch `scale` = $\sqrt{\frac{d_2^2}{d_1^2}}$

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
UlysseDurand: Done in https://github.com/Kitware/vtk-js/pull/3596/changes/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: This is done in https://github.com/Kitware/vtk-js/pull/3596/changes/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: this test is removed in https://github.com/Kitware/vtk-js/pull/3596/changes/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: this test is removed in https://github.com/Kitware/vtk-js/pull/3596/changes/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: this information of whether the timer is still running is moved to `tapInformation.timer` being null or not in https://github.com/Kitware/vtk-js/pull/3596/changes/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620.

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
UlysseDurand: I included Tap and LongTap as gestures (detected only if recognizeGestures is true) in https://github.com/Kitware/vtk-js/pull/3596/changes/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: I can, this is done in https://github.com/Kitware/vtk-js/pull/3596/changes/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.js
finetjul: Fair enough.
(I did not check the math, but I believe you.)

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: Sure, this is done in https://github.com/Kitware/vtk-js/pull/3596/changes/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

UlysseDurand:
> Can you consider writing a test ?
> 
> e.g. https://github.com/Kitware/vtk-js/pull/3595/changes#diff-ce1dbea0013f09679bf8d00604fc66d9da245c796ed66f0440467f9c123d2435R147

A test is added in https://github.com/Kitware/vtk-js/commit/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.d.ts
UlysseDurand: Sure, this is removed in https://github.com/Kitware/vtk-js/commit/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: This is changed in https://github.com/Kitware/vtk-js/commit/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: I changed the name in https://github.com/Kitware/vtk-js/commit/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
UlysseDurand: This is done in https://github.com/Kitware/vtk-js/commit/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
UlysseDurand: This is done in this commit https://github.com/Kitware/vtk-js/commit/8eae7c17c1df5c6cd4eb0a88ac6e67902667b620

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: I used this function as often as possible in https://github.com/Kitware/vtk-js/pull/3596/changes/4f9e9a428b3db20de567a0d4bce5fea2534dbccc

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: This was discussed in https://github.com/Kitware/vtk-js/pull/3596#discussion_r3811374633

Examples/Applications/TapGestureDemo/index.js
```diff
@@ -0,0 +1,64 @@
+import '@kitware/vtk.js/favicon';
```
finetjul, line 1: if you add an example, you also need to add a screenshot in the gallery

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
```diff
@@ -0,0 +1,164 @@
...
+  container.dispatchEvent(
+    makePointerEvent('pointerup', { button: 1, buttons: 2, pointerId: 2 })
+  );
+  container.dispatchEvent(
+    makePointerEvent('pointerup', { button: 2, buttons: 0 })
+  );
+  expect(events).toEqual([]);
+  events.splice(0);
+
+  // Test LongTap still works after a failed tap
```
finetjul: you already tested it line 123

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
```diff
@@ -0,0 +1,164 @@
...
+  tapTest(container, 510, 32);
+  expect(events).toEqual([]);
+  events.splice(0);
+
+  // Test Tap fail because an additional pointer is added
+  container.dispatchEvent(
+    makePointerEvent('pointerdown', { button: 1, buttons: 1 })
+  );
+  vi.advanceTimersByTime(480);
+  container.dispatchEvent(
```
finetjul, line 136: you could have extended the tapText signature to support extra events:
`tapTest(container, 600, 0, extraEvents={480: makePointerEvent(...), 580: [ makePointerEvent()});`

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
```diff
@@ -0,0 +1,164 @@
...
+  const events = [];
+  const subs = [
+    interactor.onTap(() => events.push('Tap')),
+    interactor.onLongTap(() => events.push('LongTap')),
+  ];
+
+  vi.useFakeTimers();
+
+  // Test Tap
+  tapTest(container, 490, 28);
```
finetjul, line 112: This test will be long. You might want to lower the RWI tapDuration to keep the test as short as possible.

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -110,6 +116,40 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+  // Tap / LongTap gesture tracking for touch and pen pointers.
+  let tapInformation = { timer: null, pointerId: null };
+
+  function cancelTapGesture() {
+    if (tapInformation.timer != null) clearTimeout(tapInformation.timer);
+    tapInformation.timer = null;
+  }
+
+  function startTapTimer(pointerId) {
+    cancelTapGesture();
```
finetjul, line 128: shouldn't it cancel tap gesture for a given pointerId ?
what if we startTapTimer on 2 different pointerIds ?

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -1183,40 +1240,41 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
-            model.container.clientWidth * model.container.clientWidth +
-              model.container.clientHeight * model.container.clientHeight
-          );
-        if (thresh < 15.0) {
-          thresh = 15.0;
+        let thresh2 =
+          0.0001 *
+          (model.container.clientWidth * model.container.clientWidth +
+            model.container.clientHeight * model.container.clientHeight);
+        if (thresh2 < 225.0) {
```
finetjul: write 15*15 instead because 225 is not obvious

Examples/Applications/TapGestureDemo/index.js
UlysseDurand: A screenshot is added in https://github.com/Kitware/vtk-js/pull/3596/commits/d38d2ce8a623ff35c8cc1a28d3a12cfa6e843fd3

Documentation/public/gallery/TapGestureDemo.png
finetjul, line 1: Please use the same image ratio than the other screenshots

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
UlysseDurand: I checked that the vi.advanceTimersByTime artificially advance timers and don't affect the test time, I tried with 5000ms for a long tap and the test still runs in ~2s

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
UlysseDurand: I don't understand why it could be useful, where could we use extra events ?

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: in the current state the second startTapTimer on a different pointerId will erase the first timer and replace it, isn't it the intended behaviour ? Or do you mean we should be able to make multiple taps in the same time with different pointers ?


As of now
Scenario 1 where pointer 1 press, then pointer 2 press: multitouch detected -> cancelTapGesture
Scenario 2 where pointer 1 press and release, then pointer 2 press and release: tap event triggered -> cancelTapGesture, pointer 2 press makes a new timer and when released it erases it and triggers a Tap event.

Sources/Rendering/Core/RenderWindowInteractor/test/testGestures.js
UlysseDurand: Yes you're right, I remove it in https://github.com/Kitware/vtk-js/pull/3596/commits/2eaa55f9b8d28b2eca4acc1c48c8d9a20d8306cf

Sources/Rendering/Core/RenderWindowInteractor/index.js
UlysseDurand: Sure, I write it here https://github.com/Kitware/vtk-js/pull/3596/commits/e844323db4f2ab37585a5fb512e8d305700ff953

Documentation/public/gallery/TapGestureDemo.png
UlysseDurand: This is done in https://github.com/Kitware/vtk-js/pull/3596/commits/9cc4d70a598b14995226c4d4da29686aeb166d38

finetjul (approved):
LGTM

PR #3599, opened 2026-08-17
chore(WidgetRepresentation): fix example representations scale and initial render

### LINKED ISSUE #3582: [Bug] WidgetRepresentation example - glyphs scale with abs(originX)
### Bug description

In WidgetRepresentation's example, modifying OriginX for any of the widget groups scales them proportionnaly to abs(OriginX).

### Steps to reproduce

Launch `npm run example WidgetRepresentation`

Move any of the OriginX sliders.

### Detailed Behavior

I think the bug is due to a combination of factors:

1) The handle representations used (sphere and cube) have `scaleInPixels` at `true` by default, switching it to `false` fixes the issue

2) The scaling happens in GlyphRepresentation's `scale1` function, more precisly in its call to `getPixelWorldHeightAtCoord`. The `displayScaleParams` of `GlyphRepresentation` are never updated and as such `cameraPosition` is always [0.0, 0.0, 0.0] and `cameraDir` is always [1.0, 0.0, 0.0].

### Expected Behavior

I am not certain of the purpose of the `scaleInPixels` behavior, but I think `displayScaleParams` should change with camera modifications.

### Environment

- vtk.js version: 675a10946f26eaeff643f02db5dca5a37775cdbf
- Browsers: Chrome and Firefox
- OS: Windows

Jo-Byr:
### Context
There are 2 issues in the example of WidgetRepresentation:

- The initial lil gui widgets state is not coherent with what is displayed in the scene (unchecking Active at first does nothing)
- The size of the representations is proportionnal to their Origin X

### Results
The scene is now initialized correctly with lil-gui widgets' state and representations' scale is only dictated by the Scale sliders.

### Changes
Calling applyStateField at initialization fixes the 1st point.

Setting scaleInPixels to false fixes the second.

This PR fixes #3582

### DIFF
--- a/Sources/Widgets/Representations/WidgetRepresentation/example/index.js
+++ b/Sources/Widgets/Representations/WidgetRepresentation/example/index.js
@@ -71,11 +71,13 @@ const compositeState = vtkStateBuilder
 const widgetSphereRep = vtkSphereHandleRepresentation.newInstance();
 widgetSphereRep.setInputData(compositeState);
 widgetSphereRep.setLabels(['all']);
+widgetSphereRep.setScaleInPixels(false);
 widgetSphereRep.getActors().forEach(renderer.addActor);
 
 const widgetCubeRep = vtkCubeHandleRepresentation.newInstance();
 widgetCubeRep.setInputData(compositeState);
 widgetCubeRep.setLabels('all');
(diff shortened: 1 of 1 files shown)


finetjul (approved):
LGTM

PR #3601, opened 2026-08-18
fix(camera): fix setPositionAndFocalPoint

Jo-Byr:
### Context
vtkCamera's setPositionAndFocalPoint introduced in #3554 does not work as intended. setFocalPoint is not called if setPosition returns true, due to the if structure.

### Results
Both methods are now always called.

### Changes
Separated the calls from the if.


finetjul (approved):
LGTM

PR #3602, opened 2026-08-18
chore(LineWidget): add missing initial render in example

Jo-Byr:
### Context
LineWidget's example seems initially empty due to a missing render call.

### Results
The cube is now visible without any interaction.

### Changes
Added a .render()


finetjul (approved):
LGTM

PR #3603, opened 2026-08-18
fix(ResliceCursorWidget): add missing render in examples

Jo-Byr:
### Context
The 3D view in ResliceCursorWidget's and ResliceCursorWidgetLabelmap's examples is not rendered properly on first initialization.

### Results
3D views are properly rendered initially.

### Changes
Added a missing render call.


finetjul (approved):
LGTM

PR #3604, opened 2026-08-18
chore(VolumeContour): add missing listen to example GUI

Jo-Byr:
### Context
In VolumeContour's example, the initial iso value is not updated when the data is finished loading.

### Results
ISO value is correctly initialized to 1308.6666666666667.

### Changes
Added a `.listen()` to the GUI element so that updating `params.IsoValue` updates the UI.


daker (approved):
LGTM

finetjul (approved):
LGTM

PR #3606, opened 2026-08-18
chore(PBR): fix initial render in example

Jo-Byr:
### Context
PBR example's initial render is sometimes incomplete (missing textures)

### Results
This PR ensures the first render is correct.

image

image

### Changes
Await textures loading before rendering


finetjul (commented):
LGTM

PR #3607, opened 2026-08-18
fix(VolumeMapper): support non-rigid volume transforms

### LINKED ISSUE #3605: vtkOpenGLVolumeMapper scaling and shear broken for non-rigid `userMatrix`
### Bug description

When applying a non-rigid user matrix to a volume, the voxels are misalligned in the 3D texture.

WebGPU render backend is not affected 
The issue exists since version v33.0.0

### Steps to reproduce

- Create two vtkImageData volumes with differently colored marker cubes and different spacing
- Use setUserMatrix to scale the smaller volume to the size of the larger one

**Result**: Marker points are on different positions
**Expected**: Volumes should match exactly

Live example:
https://codesandbox.io/p/devbox/hello-world-cone-forked-hyk4jv

### Detailed Behavior

Marker cubes are on different positions.

### Expected Behavior

Marker cubes should overlap exactly after scaling

### Environment

- vtk.js version: 35.5.3
- Browsers: chrome
- OS: windows 11

daker:
### Context

Texture positions used the actor normal matrix, which is an inverse transpose. It gives the correct result only for rigid transforms, so an actor matrix with scaling or shear inversely scaled or sheared the texture coordinates. Position vectors now use the linear part of the model to view matrix, and a new normalISToVCMatrix uniform keeps the inverse transpose transform for lighting normals. The sample count and the ray length now come from the actor bounds, so they match the rendered size of a scaled volume.

Fixes #3605

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/OpenGL/VolumeMapper/index.js
+++ b/Sources/Rendering/OpenGL/VolumeMapper/index.js
@@ -44,6 +44,8 @@ const { vtkWarningMacro, vtkErrorMacro } = macro;
 const preAllocatedMatrices = {
   idxToView: mat4.identity(new Float64Array(16)),
   vecISToVCMatrix: mat3.identity(new Float64Array(9)),
+  normalISToVCMatrix: mat3.identity(new Float64Array(9)),
+  linearModelToViewMatrix: mat3.identity(new Float64Array(9)),
   modelToView: mat4.identity(new Float64Array(16)),
   projectionToView: mat4.identity(new Float64Array(16)),
   projectionToWorld: mat4.identity(new Float64Array(16)),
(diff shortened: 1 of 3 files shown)


Sources/Rendering/OpenGL/VolumeMapper/index.js
```diff
@@ -864,7 +889,7 @@ function vtkOpenGLVolumeMapper(publicAPI, model) {
     );
     program.setUniform3fv(`${uniformPrefix}.originVC`, originVC);
 
-    const diagonalLength = vec3.length(sizeVC);
+    const diagonalLength = vtkBoundingBox.getDiagonalLength(actor.getBounds());
```
sankhesh, line 892: `sizeVC` is in view coordinates whereas actor bounds are in world coordinates. Please confirm that this is actually right.

sankhesh (approved):
Looks good overall with one thread.

Sources/Rendering/OpenGL/VolumeMapper/index.js
daker: sizeVC is misnamed, sizeVC is not yet in view coordinates when computed maybe it should be renamed ? SizeMC?

Issue #2536, opened 2022-07-26
CameraLightMatrix does not transform camera light position properly


JiayiXuu:
### High-level description
Camera light is not properly implemented

### Expected behavior

Light should follow camera. If we use a simple sphere with one light src, the area lit up should remain static when we pan/rotate/dolly.  Right now light src ends up in an unexpected location when we interact with the window.

_Left expected, right current behavior:_
![image](https://user-images.githubusercontent.com/43657251/181044034-653b47bd-766f-4c1d-a73d-a67aa8e66b7d.png)

### Steps taken to fix the bug:

Correct `camera.computeLightCameraTransform()`. The result is verified by rendering exactly the same scene in VTK & vtk.js and printing out the matrix value.
```
  publicAPI.computeCameraLightTransform = () => {
    mat4.copy(tmpMatrix, publicAPI.getViewMatrix());
    mat4.invert(tmpMatrix, tmpMatrix);
    
    mat4.scale(tmpMatrix2,tmpMatrix,[
      model.distance,
      model.distance,
      model.distance,
    ]);
    tmpMatrix2[3] = tmpMatrix[3];
    tmpMatrix2[7] = tmpMatrix[7];
    tmpMatrix2[11] = tmpMatrix[11];

    mat4.identity(tmpMatrix);
    tmpMatrix[3] = 0.0;
    tmpMatrix[7] = 0.0;
    tmpMatrix[11] = -1.0;
    mat4.multiply(tmpMatrix, tmpMatrix, tmpMatrix2);

    mat4.copy(model.cameraLightTransform, tmpMatrix)
  };
```

Correct `light.getTransformedPosition()` and `light.getTransformedFocalPoint()`. `vec3.transformMat4` pre-multiply positional vector with matrix rather than post-multiply. Function should be something like the following code snippet. Result is verified by rendering exactly the same scene in VTK & vtk.js and printing out the transformed position value.
```
  publicAPI.getTransformedPosition = () => {
    if (model.transformMatrix) {
      var x = model.position[0],y = model.position[1],z = model.position[2];
      var m = model.transformMatrix;
      tmpVec[0] = m[0] * x + m[1] * y + m[2] * z + m[3];
      tmpVec[1] = m[4] * x + m[5] * y + m[6] * z + m[7];
      tmpVec[2] = m[8] * x + m[9] * y + m[10] * z + m[11];
  
    } else {
      vec3.set(tmpVec, model.position[0], model.position[1], model.position[2]);
    }
[...]

PR #3608, opened 2026-08-18
chore(Convolution2DPass): fix incorrect initial render in example

Jo-Byr:
Add missing renderer.updateLightsGeometryToFollowCamera() to fix incorrect initial render and the first camera interaction changing the lighting of the scene

### Context
In Convolution2DPass example, the ligting of the scene changes after the first camera interaction (rotation or zoom).

image
image

### Results
The lighting is correct at the first render.

### Changes
Added missing `renderer.updateLightsGeometryToFollowCamera();`

### DIFF
--- a/Sources/Rendering/OpenGL/Renderer/index.js
+++ b/Sources/Rendering/OpenGL/Renderer/index.js
@@ -21,6 +21,7 @@ function vtkOpenGLRenderer(publicAPI, model) {
         return;
       }
 
+      model.renderable.updateLightGeometry();
       publicAPI.updateLights();
       publicAPI.prepareNodes();
       publicAPI.addMissingNode(model.renderable.getActiveCamera());
--- a/Sources/Rendering/WebGPU/Renderer/index.js
+++ b/Sources/Rendering/WebGPU/Renderer/index.js
@@ -48,6 +48,7 @@ function vtkWebGPURenderer(publicAPI, model) {
 
       model.camera = model.renderable.getActiveCamera();
 
+      model.renderable.updateLightGeometry();
       publicAPI.updateLights();
       publicAPI.prepareNodes();
       publicAPI.addMissingNode(model.camera);


finetjul (approved):
LGTM

PaulHax:
Drive-by observation: this looks like a general gap rather than an example issue. `renderer.updateLightGeometry()` exists but nothing calls it, so lights only get refreshed from the interaction path. Anything that moves the camera without a mouse ends up stale: view preset buttons, restored camera state, and a fair number of examples in here that call elevation/azimuth after setup.

For reference, VTK C++ calls it from `vtkOpenGLRenderer::DeviceRender()`, right next to `UpdateLights()`:

    this->UpdateCamera();
    this->UpdateLightGeometry();
    this->UpdateLights();
    this->UpdateGeometry();

vtk.js has the equivalent OpenGL/Renderer buildPass calling only `updateLights()`, so the light geometry half never happens.

#2536 may be the same root cause. This one looks related too: https://discourse.vtk.org/t/issues-with-lighting-and-clipping-in-multi-renderer-vtk-js-setup/15495 where the workaround was setFocusedRenderer(), which fits since only the poked renderer gets its lights updated.

Might be fixable in one place, either there or by updating lights when the camera is modified. Happy to open an issue if that is useful.

Jo-Byr:
The issue I have is that in C++, vtkOpenGLRenderer inherits from vtkRenderer, and as such has access to UpdateLightGeometry, but in vtk-js, vtkOpenGLRenderer extends vtkViewNode.

vtkOpenGLRenderer holds vtkRenderer instances indirectly through `model.renderable.getRenderWindow().getRenderers()`.

Should we call `updateLightGeometry` on each of them ?

finetjul:
> vtkOpenGLRenderer holds vtkRenderer instances indirectly through model.renderable.getRenderWindow().getRenderers().

Isn't `model.renderable` what you are looking for (i.e. the vtkRenderer paired with vtkOpenGLRenderer)?

Jo-Byr:
> > vtkOpenGLRenderer holds vtkRenderer instances indirectly through model.renderable.getRenderWindow().getRenderers().
> 
> Isn't `model.renderable` what you are looking for (i.e. the vtkRenderer paired with vtkOpenGLRenderer)?

Absolutely, I fixed it

finetjul (commented):
Is the light geometry updated sufficiently often ?
i.e. is `buildPass()` called for each new camera position ?

Jo-Byr:
> is `buildPass()` called for each new camera position ?

Yes, maybe even too much. It's called every few ms as long as the left click is held down (in Convolution2DPass and SimpleCone examples).

finetjul:
> > is `buildPass()` called for each new camera position ?
> 
> Yes, maybe even too much. It's called every few ms as long as the left click is held down (in Convolution2DPass and SimpleCone examples).

Sounds good then.

finetjul (approved):
LGTM

finetjul:
Is there an example to test if #2536 is fixed ?

finetjul (dismissed):
Apparently some test(s) are failing

PR #3609, opened 2026-08-19
chore(ImageResliceMapper): add missing resetCamera and render

Jo-Byr:
### Context
In ImageResliceMapper's example, interacting with the scene before the data is loaded can prevent a proper display of the data.

Current initial render
image

Current display after data is finished loading if we dezoom before it is finished
image

Fixed initial display
image

### Results
Camera is properly reset and data is loaded correctly.

### Changes
Added initial resetCamera and render calls.


finetjul (approved):
LGTM

PR #3613, opened 2026-08-21
fix(widgets): do not deactivate widgets when none consume events

finetjul:
Fixes the PaintWidget example where a mouse hover anywhere on the window was deactivating the widget. It still seems unclear whether widgets should be deactivated automatically in case of an event over a different renderer, but that behavior was kept.

### Context
See #3573

### Results
Closes #3573

### Changes
Because the "Animating" mode was on, the widgets were deactivated. That need was not expressively stated in https://github.com/Kitware/vtk-js/commit/1d56bbb02c947925c18fd74c0fbe632cc8ebcd80 

### PR and Code Checklist

### Testing
- [x] Tested environment:
  - **vtk.js**: master
  - **OS**: Windows
  - **Browser**: Firefox

### DIFF
--- a/Sources/Widgets/Core/WidgetManager/index.js
+++ b/Sources/Widgets/Core/WidgetManager/index.js
@@ -195,17 +195,16 @@ function vtkWidgetManager(publicAPI, model) {
   };
 
   const handleEvent = async (callData, fromTouchEvent = false) => {
-    if (
-      !model.isAnimating &&
-      model.pickingEnabled &&
-      callData.pokedRenderer === model._renderer
-    ) {
+    if (callData.pokedRenderer !== model._renderer) {
+      deactivateAllWidgets();
+      return macro.VOID;
+    }
+    if (!model.isAnimating && model.pickingEnabled) {
       const callID = Symbol('UpdateSelection');
       model._currentUpdateSelectionCallID = callID;
       await updateSelection(callData, fromTouchEvent, callID);
-    } else {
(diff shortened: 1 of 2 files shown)


Sources/Widgets/Core/WidgetManager/index.js
```diff
@@ -195,17 +195,16 @@ function vtkWidgetManager(publicAPI, model) {
...
+      return macro.VOID;
+    }
+    if (!model.isAnimating && model.pickingEnabled) {
       const callID = Symbol('UpdateSelection');
       model._currentUpdateSelectionCallID = callID;
       await updateSelection(callData, fromTouchEvent, callID);
-    } else {
-      deactivateAllWidgets();
     }
+    return macro.VOID;
```
Jo-Byr, line 207: Why not

```
    if (callData.pokedRenderer !== model._renderer) {
      deactivateAllWidgets();
    }
    else if (
      !model.isAnimating &&
      model.pickingEnabled
    ) {
      const callID = Symbol('UpdateSelection');
      model._currentUpdateSelectionCallID = callID;
      await updateSelection(callData, fromTouchEvent, callID);
    }
    return macro.VOID;
```

?

finetjul: I generally like "early returns" in code. I find it easier to read.

Jo-Byr:
LGTM

PR #3618, opened 2026-08-28
fix(OpenGL): release poly data mapper buffers when the view node is deleted

PaulHax:
### Context

`vtkOpenGLPolyDataMapper`, `vtkOpenGLPolyDataMapper2D` and `vtkOpenGLGlyph3DMapper` have no way to free the VBOs, VAOs and color texture they own on the GL context. Deleting a view node, which happens when an actor is removed or a render window closes, leaves them allocated.

Kitware/vtk-js#2887 gave the image and volume mappers `unregisterGraphicsResourceUser` chained into `delete()`, so their *shared* textures are freed when a view closes. The poly data mappers never got the per-view equivalent. Applications cannot recover the leak themselves: when views share one WebGL context, a child view's `releaseGraphicsResources()` is forwarded to the root view and releases the resources of every view on the context, so an application closing one view would break the views that stay open.

Found while tightening per-view GPU teardown in [Kitware/VolView#932](https://github.com/Kitware/VolView/pull/932), which documents this as a known gap that needs a vtk.js-side fix: its layout swaps add and remove actors on views sharing one WebGL context, so the orphaned buffers accumulate across the session.

### Results

Before: fifty `addActor`/`removeActor` cycles orphan 150 live WebGL objects on a context that stays alive, and `getGraphicsMemoryInfo()` reports none of it (it only walks the shared-resource map). After: deleting the view node returns the live-object count to the empty-scene baseline. Separately, releasing a live mapper's resources and rendering again produces an identical image.

### Changes

- `vtkOpenGLPolyDataMapper`, `vtkOpenGLPolyDataMapper2D` and `vtkOpenGLGlyph3DMapper` gain a public `releaseGraphicsResources()`, chained into `delete()`, so a view node frees its own buffers and nothing else. The release logic is shared between the two poly data mappers, which own the same buffer-holding model shape.
- `Testing/testUtils.js` gains `trackWebGLObjects` (counts live WebGL objects by wrapping create/delete pairs on the rendering context prototypes) and `createRenderContainer`; the tracked-view setup and the same-image-after-release assertion live in a new `Testing/renderTestUtils`.

### PR and Code Checklist


Sources/Rendering/OpenGL/PolyDataMapper/test/testReleaseGraphicsResources.js
```diff
@@ -0,0 +1,131 @@
...
+    actors.forEach((actor) => renderer.addActor(actor));
+    renderer.resetCamera();
+    renderWindow.render();
+    expect(tracker.count()).toBeGreaterThan(emptySceneObjects);
+
+    actors.forEach((actor) => renderer.removeActor(actor));
+    renderWindow.render();
+    expect(tracker.count()).toBe(emptySceneObjects);
+
+    gc.releaseResources();
```
sankhesh, line 76: Since this is inside an async call, this may never get called if the test fails early leaking resources. Recommend enclosing inside an `afterEach` to ensure to ensure cleanup.

Sources/Rendering/OpenGL/PolyDataMapper/test/testReleaseGraphicsResources.js
```diff
@@ -0,0 +1,131 @@
...
+    rootRenderWindow.removeRenderWindow(closing.childRenderWindow);
+    closing.childRenderWindow.delete();
+
+    const survivingAfter = surviving.childView.captureNextImage();
+    rootRenderWindow.render();
+
+    expect(tracker.count()).toBeLessThan(bothViewsObjects);
+    expect(await survivingAfter).toBe(await survivingBefore);
+
+    gc.releaseResources();
```
sankhesh, line 127: Same as before.

Sources/Testing/renderTestUtils.js
```diff
@@ -0,0 +1,50 @@
...
+
+  view.getViewNodeFor(actor.getMapper()).releaseGraphicsResources(view);
+  expect(tracker.count()).toBe(emptySceneObjects);
+
+  const afterRelease = view.captureNextImage();
+  renderWindow.render();
+  expect(tracker.count()).toBeGreaterThan(emptySceneObjects);
+  expect(await afterRelease).toBe(await beforeRelease);
+
+  gc.releaseResources();
```
sankhesh, line 47: Same thing here.

Sources/Rendering/OpenGL/PolyDataMapper/test/testReleaseGraphicsResources.js
```diff
@@ -0,0 +1,131 @@
...
+    const surviving = addChildView();
+    rootRenderWindow.render();
+
+    const bothViewsObjects = tracker.count();
+    const survivingBefore = surviving.childView.captureNextImage();
+    rootRenderWindow.render();
+    expect(tracker.count()).toBe(bothViewsObjects);
+
+    rootRenderWindow.removeRenderWindow(closing.childRenderWindow);
+    closing.childRenderWindow.delete();
```
sankhesh, line 119: Wouldn't this lead to double deletion if the garbage collector already tracks resources?

Sources/Rendering/OpenGL/PolyDataMapper/test/testReleaseGraphicsResources.js
PaulHax: thanks for taking a look!

Agreed, a trailing `gc.releaseResources()` is skipped whenever the test doesn't reach the end.

Put the cleanup in `createGarbageCollector` instead of `afterEach`:

```js
onTestFinished(releaseResources);
```

`expectSameImageAfterRelease` creates its own `gc`, so an `afterEach` in the test file would have nothing to release. Registering it in `createGarbageCollector` covers all three spots and matches `createWebGPUTestDevice` in the same file. Releasing twice is a no-op.

Sources/Rendering/OpenGL/PolyDataMapper/test/testReleaseGraphicsResources.js
PaulHax: No. The child render windows aren't gc-registered, and neither are their view nodes or renderers. The gc holds the root render window, the root view, the containers, and each child's cone source, mapper and actor, so `releaseResources` never reaches `closing.childRenderWindow`.

`removeRenderWindow()` plus the next build pass is what deletes the view node and frees the buffers. The explicit `delete()` is the application-side disposal that goes with closing one of several views sharing a GL context.

finetjul (approved):
LGTM

sankhesh (approved):
LGTM

PR #3620, opened 2026-08-28
fix(Widgets): release owned picking resources on deletion

PaulHax:
### Context

`vtkWidgetManager` owns its per-view widgets and its hardware selector, and `delete()` left both alive. The selector holds a full-window framebuffer, color texture and depth renderbuffer per view; a focused widget keeps an animation request on the interactor that only losing focus cancels, so a deleted manager could pin the requestAnimationFrame loop forever.

### Results

Deleting a widget manager, including mid-capture or while a widget has focus, returns the live WebGL object count to baseline, stops the focused widget's animation request, and no longer lets a stale capture serve selections. `setRenderer` frees the selector built for the previous wiring instead of leaking it.

### Changes

- `vtkWidgetManager.delete()` releases focus, removes the view widgets, and deletes the selector once any in-flight capture settles. Losing focus re-enables picking, so deletion also suppresses the capture that `renderWidgets` would otherwise start on the way out.
- `vtkOpenGLHardwareSelector` chains `releaseGraphicsResources()` into `delete()` so its framebuffer goes with it.
- `setRenderer` drops the captured buffers and in-flight capture belonging to the previous wiring; a capture that settles after the manager was deleted *or re-targeted* no longer publishes stale buffers (previously an in-flight capture for renderer A could serve selections after a switch to renderer B). `setRenderer` with the *same* renderer intentionally remains a full re-wire: it re-reads camera/interactor/view off the renderer and rebuilds the selector.
- `getSelectedDataForXY` handles a capture that produced nothing to select against (selector bailed on a half-torn-down view) instead of throwing from the mousemove chain.

### PR and Code Checklist


finetjul (approved):
LGTM

PR #3622, opened 2026-09-02
Render pass teardown

PaulHax:
### Context

Render passes are not view nodes and are not registered graphics resources, so none of the teardown walkers reached them. Deleting a `vtkOpenGLRenderWindow` left every framebuffer its passes owned on the GPU, and nothing reclaimed them later: `deleteGLContext` only decrements a counter and the context outlives the render window. The same passes were stranded when `setRenderPasses` swapped them out, and a child render window, which has no context of its own, skipped the release entirely. `vtkOpenGLVolumeMapper` had the same gap for the objects it allocates for itself, and `vtkConvolution2DPass` stranded a vertex array on every kernel dimension change.

With one translucent actor, the order independent translucent pass alone leaves a framebuffer, three textures, a vertex array and a vertex buffer on the GPU for the life of the context, per deleted view. The WebGPU render window already frees its passes on delete.

Stacked on #3620

### Results

Deleting a render window, replacing its render passes, deleting a child render window, or deleting the root before its children returns the live WebGL object count to baseline. A view that releases its graphics resources draws the same image and reads depth again on the next render. Removing a volume from a scene frees the mapper's interaction framebuffer, jitter texture and vertex buffers. Changing a convolution kernel dimension no longer leaks a vertex array.

### Changes

- `vtkRenderPass` gains `releaseGraphicsResources(viewNode)`, which reaches the passes it delegates to. `vtkForwardPass`, `vtkOpenGLOrderIndependentTranslucentPass`, `vtkConvolution2DPass` and `vtkRadialDistortionPass` chain their own release onto it. The translucent pass had such a method already, unreachable and throwing when called; it now drops its shader reference instead of releasing a program the shader cache owns, and keeps its helper so the next traverse can rebuild.
- `vtkOpenGLRenderWindow` releases its render passes on `delete()` and from `releaseGraphicsResources()`, through a module local walk because the public method is proxied to the root window. The context-wide release now walks child render windows too. A child releases while its root is still alive, which the view node tree guarantees when the root is deleted first.
- `vtkOpenGLRenderWindow.setRenderPasses()` releases the passes absent from the incoming list. A pass kept at another index keeps what it has. Membership is compared at the top level only.
- `vtkOpenGLVolumeMapper` gains `releaseGraphicsResources()`, chained into `delete()`.
- `vtkConvolution2DPass` resets its vertex array via `shaderProgramChanged()` on a kernel dimension change instead of replacing it.
- Both release walks skip deleted passes, so an application that tears down a pass it owns before the view still gets the view torn down.

### PR and Code Checklist

### Testing

New tests count live WebGL objects by wrapping the create/delete pairs on the rendering context prototypes, across view teardown for the translucent, convolution and radial distortion passes, across `setRenderPasses`, across child window deletion in both orders, and across a volume leaving the scene. 

Image tests check that a released view draws the same image and reads depth again. One test deletes a pass before the view and checks the view still tears down.

- [x] Tested environment:
  - **vtk.js**: master at 52fff31926 plus this stack
  - **OS**: Linux (WSL2)
  - **Browser**: Playwright 1.60.0 headless Chromium (chromium-1234); full suite passes


finetjul (approved):
LGTM

Sources/Rendering/OpenGL/OrderIndependentTranslucentPass/index.js
```diff
@@ -371,16 +371,12 @@ function vtkOpenGLOrderIndependentTranslucentPass(publicAPI, model) {
...
-      model.copyShader = null;
-    }
-    if (model.tris) {
-      model.tris.releaseGraphicsResources(viewNode);
-      model.tris = null;
-    }
+    // The shader cache owns the programs it hands out, so only drop the
+    // reference.
+    model.copyShader = null;
+    model.tris.releaseGraphicsResources(viewNode);
```
sankhesh, line 377: Why remove the null check guards before calling release?

sankhesh (approved):
LGTM except one question

Sources/Rendering/OpenGL/OrderIndependentTranslucentPass/index.js
PaulHax: tris is created in extend() and must remain available after releasing its GPU resources.  The next render call expects  `model.tris` to exist.  The flow:
- Construct pass
- Create the tris helper
- Build its VBO
- Release its VAO and CABO resources
- Keep the tris helper
- Mark the pass modified
- Rebuild the VBO on the next render

(and no need to null check model.copyShader before setting it to null)

PaulHax:
I think this is a good time to merge this stack if you agree @finetjul

PR #3623, opened 2026-09-02
feat(RenderWindowInteractor): add optional mouseWheelSpinYBuffering

yehor-murza-lx:
## Motivation

High-frequency wheel input devices (e.g. trackpads) can emit many small fractional spinY deltas in quick succession. Consumers of tkRenderWindowInteractor that map wheel events directly to camera zoom/dolly can end up with jittery or overly sensitive interactions as a result.

## Change

- Adds an opt-in mouseWheelSpinYBuffering flag (default alse). When enabled, fractional spinY deltas are accumulated per interactor instance and a mouseWheelEvent is only dispatched once the buffered spin reaches a full step (rounded via Math.sign). The buffer resets when the scroll direction reverses and when the wheel-end debounce fires.
- Adds a configurable wheelEndDebounceDelay (default 200, matching the previous hardcoded value) so consumers can tune how long to wait after the last wheel event before the wheel-end event fires.
- Updates the corresponding TypeScript definitions (index.d.ts).

## Backward compatibility

Both new properties default to values that reproduce the current behavior exactly (mouseWheelSpinYBuffering: false disables buffering entirely; wheelEndDebounceDelay: 200 matches the previous hardcoded timeout), so no existing consumer behavior changes unless they opt in.

## Testing

- 
pm run reformat and lint (oxlint) pass with no issues.
- Manually verified against a downstream application that previously carried this behavior as a local patch; setting mouseWheelSpinYBuffering(true) reproduces that smoothing behavior.

### DIFF
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -116,6 +116,8 @@ function vtkRenderWindowInteractor(publicAPI, model) {
   // Accumulates fractional spinY deltas when mouseWheelSpinYBuffering is enabled.
   let scrollBuffer = 0;
 
+  let wheelTimeoutID = 0;
+
   // Track mouse button bitmask for detecting chorded button interactions.
   // Per W3C Pointer Events spec §10, pointerdown/pointerup only fire for the
   // first press / last release. Chorded (additional) button changes while
@@ -326,9 +328,10 @@ function vtkRenderWindowInteractor(publicAPI, model) {
   const _unbindEvents = () => {
     // Clear any previous timeouts and state variables that control mouse / touchpad behavior.
     clearTimeout(model.moveTimeoutID);
-    clearTimeout(model.wheelTimeoutID);
+    clearTimeout(wheelTimeoutID);
     model.moveTimeoutID = 0;
-    model.wheelTimeoutID = 0;
+    wheelTimeoutID = 0;
+    scrollBuffer = 0;
     wheelCoefficient = 1.0;
 
     const { container } = model;
@@ -879,7 +882,7 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     // mouse wheel events have absolute spin values higher than 1.
     // Here the first spin value is "recorded", and used to normalize
     // all the following mouse wheel events.
-    if (model.wheelTimeoutID === 0) {
+    if (wheelTimeoutID === 0) {
       // we attempt to distinguish between trackpads and mice
       // .3 will be larger than the first trackpad event,
       // but small enough to detect some common edge case mice
@@ -893,6 +896,12 @@ function vtkRenderWindowInteractor(publicAPI, model) {
     }
     callData.spinY /= wheelCoefficient;
 
+    if (wheelTimeoutID === 0) {
+      publicAPI.startMouseWheelEvent(callData);
+    } else {
+      clearTimeout(wheelTimeoutID);
+    }
+
     if (model.mouseWheelSpinYBuffering) {
       // Reset the buffer when the scroll direction reverses so a direction
       // change is never delayed by leftover buffer from the previous one.
@@ -903,42 +912,27 @@ function vtkRenderWindowInteractor(publicAPI, model) {
         scrollBuffer = 0;
       }
       scrollBuffer += callData.spinY;
-    }
 
-    if (model.wheelTimeoutID === 0) {
-      publicAPI.startMouseWheelEvent(callData);
-      if (!model.mouseWheelSpinYBuffering) {
+      if (Math.abs(scrollBuffer) >= SCROLL_THRESHOLD) {
+        callData.spinY = Math.trunc(scrollBuffer);
+        scrollBuffer -= callData.spinY;
         publicAPI.mouseWheelEvent(callData);
       }
     } else {
-      if (!model.mouseWheelSpinYBuffering) {
-        publicAPI.mouseWheelEvent(callData);
-      }
-      clearTimeout(model.wheelTimeoutID);
-    }
-
-    // Only fire once the buffered spin reaches a full step, rounding fractional
-    // high-frequency deltas (e.g. trackpads) into a single integer-sized event.
-    if (
-      model.mouseWheelSpinYBuffering &&
-      Math.abs(scrollBuffer) >= SCROLL_THRESHOLD
-    ) {
-      callData.spinY = Math.sign(scrollBuffer);
-      scrollBuffer -= callData.spinY;
       publicAPI.mouseWheelEvent(callData);
     }
 
-    if (model.mouseScrollDebounceByPass) {
+    if (model.mouseScrollDebounceByPass || model.wheelEndDebounceDelay === 0) {
       publicAPI.extendAnimation(600);
       publicAPI.endMouseWheelEvent();
-      model.wheelTimeoutID = 0;
+      wheelTimeoutID = 0;
       scrollBuffer = 0;
     } else {
       // start a timer to keep us animating while we get wheel events
-      model.wheelTimeoutID = setTimeout(() => {
+      wheelTimeoutID = setTimeout(() => {
         publicAPI.extendAnimation(600);
         publicAPI.endMouseWheelEvent();
-        model.wheelTimeoutID = 0;
+        wheelTimeoutID = 0;
         scrollBuffer = 0;
       }, model.wheelEndDebounceDelay);
     }
@@ -1418,7 +1412,6 @@ const DEFAULT_VALUES = {
   animationRequest: null,
   lastFrameTime: 0.1,
   recentAnimationFrameRate: 10.0,
-  wheelTimeoutID: 0,
   moveTimeoutID: 0,
   lastGamepadValues: {},
   preventDefaultOnPointerDown: false,
--- a/Sources/Rendering/Core/RenderWindowInteractor/index.js
+++ b/Sources/Rendering/Core/RenderWindowInteractor/index.js
@@ -11,6 +11,9 @@ const { vtkWarningMacro, vtkErrorMacro, normalizeWheel, vtkOnceErrorMacro } =
 // Global methods
 // ----------------------------------------------------------------------------
 
+// Spin amount (in normalized wheel units) needed to trigger a buffered wheel event.
+const SCROLL_THRESHOLD = 1;
+
 const EMPTY_MOUSE_EVENT = {
   ctrlKey: false,
   altKey: false,
@@ -109,6 +112,9 @@ function vtkRenderWindowInteractor(publicAPI, model) {
 
   // Factor to apply on wheel spin.
   let wheelCoefficient = 1;
+
+  // Accumulates fractional spinY deltas when mouseWheelSpinYBuffering is enabled.
+  let scrollBuffer = 0;
 
   // Track mouse button bitmask for detecting chorded button interactions.
   // Per W3C Pointer Events spec §10, pointerdown/pointerup only fire for the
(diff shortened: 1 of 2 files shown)


Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -887,25 +893,54 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+    // Only fire once the buffered spin reaches a full step, rounding fractional
+    // high-frequency deltas (e.g. trackpads) into a single integer-sized event.
+    if (
+      model.mouseWheelSpinYBuffering &&
+      Math.abs(scrollBuffer) >= SCROLL_THRESHOLD
+    ) {
+      callData.spinY = Math.sign(scrollBuffer);
+      scrollBuffer -= callData.spinY;
+      publicAPI.mouseWheelEvent(callData);
+    }
```
finetjul, line 923: Let's move things around to cleanup the logic:
If (wheeleventid ==) {
  Startmousewheelevent();
} Else {
ClearTimeout();
}

If (not wheelbuferring or ...)
  Mousewheelevent();
}

...

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -887,25 +893,54 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
       clearTimeout(model.wheelTimeoutID);
     }
 
+    // Only fire once the buffered spin reaches a full step, rounding fractional
+    // high-frequency deltas (e.g. trackpads) into a single integer-sized event.
+    if (
+      model.mouseWheelSpinYBuffering &&
+      Math.abs(scrollBuffer) >= SCROLL_THRESHOLD
+    ) {
+      callData.spinY = Math.sign(scrollBuffer);
```
finetjul: I think spinY should be 3 (and not 1) if the scroll buffer is 3.2 (maybe the last event added 2.3 to the buffer)

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -887,25 +893,54 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
       publicAPI.extendAnimation(600);
       publicAPI.endMouseWheelEvent();
       model.wheelTimeoutID = 0;
+      scrollBuffer = 0;
     } else {
       // start a timer to keep us animating while we get wheel events
       model.wheelTimeoutID = setTimeout(() => {
         publicAPI.extendAnimation(600);
         publicAPI.endMouseWheelEvent();
         model.wheelTimeoutID = 0;
```
finetjul: Can you please take the opportunity to move wheeltimeoutid out of model?

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -1431,6 +1468,8 @@ export function extend(publicAPI, model, initialValues = {}) {
     'preventDefaultOnPointerDown',
     'preventDefaultOnPointerUp',
     'mouseScrollDebounceByPass',
+    'wheelEndDebounceDelay',
```
finetjul, line 1464: I like the idea of exposing the delay. But in such a case it duplicates the debouncebypass. Please consolidate both (e.g. a delay of null or 0 is equivalent to a debounce by pass). Let's not break the api for now and simply create a warning that the debounce by pass is now obsolete.

finetjul:
We now have a case when a startmousewheelevent may not be followed by a mouse event. Please check in the code base that it is fine. I think I remember that assumption to be expected

Sources/Rendering/Core/RenderWindowInteractor/index.js
YehorMurza-elucid: Kept `mouseScrollDebounceByPass `for backward compatibility but marked it deprecated in the TypeScript API. `wheelEndDebounceDelay `is now the canonical API. A delay of 0 performs the same immediate wheel-end behavior as the legacy bypass flag.
In [203b687](https://github.com/Kitware/vtk-js/pull/3623/commits/203b6873bdb126504badb5a4c6f54f33e222cec4)

Sources/Rendering/Core/RenderWindowInteractor/index.js
YehorMurza-elucid: Moved `wheelTimeoutID `into the `vtkRenderWindowInteractor `closure so it is private per interactor instance. It is cleared during unbind and is no longer exposed through `DEFAULT_VALUES `or `IRenderWindowInteractorInitialValues`.
In [203b687](https://github.com/Kitware/vtk-js/pull/3623/commits/203b6873bdb126504badb5a4c6f54f33e222cec4)

Sources/Rendering/Core/RenderWindowInteractor/index.js
YehorMurza-elucid: Updated the buffered event to preserve the complete integer portion of the accumulated scroll amount. For example, a buffer of 3.2 now emits spinY: 3 and retains 0.2; negative values are handled symmetrically with Math.trunc.
In [203b687](https://github.com/Kitware/vtk-js/pull/3623/commits/203b6873bdb126504badb5a4c6f54f33e222cec4)

Sources/Rendering/Core/RenderWindowInteractor/index.js
YehorMurza-elucid: Updated the wheel handling flow as suggested. Starting/clearing the wheel timeout is now handled independently from buffering, and mouseWheelEvent is dispatched either immediately when buffering is disabled or after the buffer reaches the threshold when buffering is enabled.
In [203b687](https://github.com/Kitware/vtk-js/pull/3623/commits/203b6873bdb126504badb5a4c6f54f33e222cec4)

YehorMurza-elucid:
> We now have a case when a startmousewheelevent may not be followed by a mouse event. Please check in the code base that it is fine. I think I remember that assumption to be expected

The buffering behavior intentionally allows startMouseWheelEvent to occur before the first mouseWheelEvent when the initial wheel delta is fractional and does not yet reach the threshold. This is required to retain the accumulated input and avoid dispatching a partial event. The buffer is reset on direction changes, wheel-end, and unbind. Existing mouse-move handling already uses the same burst-marker pattern, where a start event can precede later interaction events.

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -887,25 +896,45 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+      ) {
+        scrollBuffer = 0;
+      }
+      scrollBuffer += callData.spinY;
+
+      if (Math.abs(scrollBuffer) >= SCROLL_THRESHOLD) {
+        callData.spinY = Math.trunc(scrollBuffer);
+        scrollBuffer -= callData.spinY;
+        publicAPI.mouseWheelEvent(callData);
+      }
```
finetjul, line 924: Let's call mouseWheelEvent only once:
```
let triggerMouseWheelEvent = true;
if (model.mouseWheelSpinYBuffering) {
  triggerMouseWheelEvent = false;
  ...
  if (Math.abs(scrollBuffer) >= SCROLL_THRESHOLD) {
    ...
    triggerMouseWheelEvent = true;
}
if (triggerMouseWheelEvent) {
  publicAPI.mouseWheelEvent(callData);
}
```

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -887,25 +896,45 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+      wheelTimeoutID = setTimeout(() => {
         publicAPI.extendAnimation(600);
         publicAPI.endMouseWheelEvent();
-        model.wheelTimeoutID = 0;
-      }, 200);
+        wheelTimeoutID = 0;
+        scrollBuffer = 0;
+      }, model.wheelEndDebounceDelay);
     }
   };
```
finetjul, line 946: Let's warn with `mouseScrollDebounceByPass` usage:
publicAPI.mouseScrollDebounceByPass = macro.chain(vtkWarningMacro(
        'mouseScrollDebounceByPass is deprecated. Use wheelEndDebounceDelay instead.'), publicAPI.mouseScrollDebounceByPass);

finetjul (commented):
Thanks looks good.
Minor points though.

Sources/Rendering/Core/RenderWindowInteractor/index.js
YehorMurza-elucid: added in [02dd7f6](https://github.com/Kitware/vtk-js/pull/3623/commits/02dd7f604ceba552126a3d86d6e7b5da7d0f740d)

Sources/Rendering/Core/RenderWindowInteractor/index.js
YehorMurza-elucid: changed in [02dd7f6](https://github.com/Kitware/vtk-js/pull/3623/commits/02dd7f604ceba552126a3d86d6e7b5da7d0f740d)

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -887,25 +896,51 @@ function vtkRenderWindowInteractor(publicAPI, model) {
...
+      }
+    }
+
+    if (triggerMouseWheelEvent) {
       publicAPI.mouseWheelEvent(callData);
-      clearTimeout(model.wheelTimeoutID);
     }
 
-    if (model.mouseScrollDebounceByPass) {
+    if (model.mouseScrollDebounceByPass || model.wheelEndDebounceDelay === 0) {
```
finetjul: `if (model.wheelEndDebounceDelay === 0) {`

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -1383,12 +1418,13 @@ const DEFAULT_VALUES = {
   animationRequest: null,
   lastFrameTime: 0.1,
   recentAnimationFrameRate: 10.0,
-  wheelTimeoutID: 0,
   moveTimeoutID: 0,
   lastGamepadValues: {},
   preventDefaultOnPointerDown: false,
   preventDefaultOnPointerUp: false,
   mouseScrollDebounceByPass: false,
```
finetjul: let's remove `mouseScrollDebounceByPass`

Sources/Rendering/Core/RenderWindowInteractor/index.js
```diff
@@ -1431,11 +1467,23 @@ export function extend(publicAPI, model, initialValues = {}) {
...
   macro.moveToProtected(publicAPI, model, ['view']);
 
+  // mouseScrollDebounceByPass is deprecated; warn while preserving the
+  // original setter's boolean return value (macro.chain would return an array).
+  const setMouseScrollDebounceByPass = publicAPI.setMouseScrollDebounceByPass;
+  publicAPI.setMouseScrollDebounceByPass = (...args) => {
+    vtkWarningMacro(
+      'mouseScrollDebounceByPass is deprecated. Use wheelEndDebounceDelay instead.'
+    );
+    return setMouseScrollDebounceByPass(...args);
```
finetjul: apologies for my erroneous suggestion earlier, I meant:
```
publicAPI.setMouseScrollDebounceByPass = (byPass) => {
    vtkWarningMacro(
      'mouseScrollDebounceByPass is deprecated. Use wheelEndDebounceDelay instead.'
    );
    return publicAPI.setWheelEndDebounceDelay(byPass ? 0 : DEFAULT_VALUES.wheelEndDebounceDelay);
```

YehorMurza-elucid: changed in [35c8c21](https://github.com/Kitware/vtk-js/pull/3623/commits/35c8c211701520dfbf0621bf15d97cb1fed96ac9) and removed the typescript definitions as well

Sources/Rendering/Core/RenderWindowInteractor/index.js
YehorMurza-elucid: removed in [35c8c21](https://github.com/Kitware/vtk-js/pull/3623/commits/35c8c211701520dfbf0621bf15d97cb1fed96ac9)

Sources/Rendering/Core/RenderWindowInteractor/index.js
YehorMurza-elucid: changed in [35c8c21](https://github.com/Kitware/vtk-js/pull/3623/commits/35c8c211701520dfbf0621bf15d97cb1fed96ac9)

finetjul (approved):
LGTM

PR #3624, opened 2026-09-02
fix(WebGPU): apply camera physical scale to custom projection

daker:
### Context

WebGPU ignores both the explicit projection matrix and physicalScale. The matrix remains unchanged for every scale.

### Changes

### PR and Code Checklist

### Testing

- [x] Tested environment:
  - **vtk.js**: 
  - **OS**: 
  - **Browser**:

### DIFF
--- a/Sources/Rendering/Core/Camera/index.d.ts
+++ b/Sources/Rendering/Core/Camera/index.d.ts
@@ -1,6 +1,6 @@
 import { mat4 } from 'gl-matrix';
 import { vtkObject } from '../../../interfaces';
-import { Bounds, Vector3, Range } from '../../../types';
+import { Bounds, Nullable, Vector3, Range } from '../../../types';
 
 /**
  *
@@ -141,6 +141,12 @@ export interface vtkCamera extends vtkObject {
   getDistance(): number;
 
   /**
+   * Get the explicit projection matrix, or null when the camera computes
+   * its projection from its standard parameters.
+   */
+  getExplicitProjectionMatrix(): Nullable<mat4>;
+
+  /**
    * Get the focal of the camera in world coordinates.
    */
   getFocalPoint(): Vector3;
@@ -241,8 +247,15 @@ export interface vtkCamera extends vtkObject {
    * @param {Number} aspect Camera frustum aspect ratio.
    * @param {Number} nearz Camera frustum near plane.
    * @param {Number} farz Camera frustum far plane.
+   * @param {mat4} [out] Optional output matrix to mutate and return, instead
+   * of allocating a new one.
    */
-  getProjectionMatrix(aspect: number, nearz: number, farz: number): mat4;
+  getProjectionMatrix(
+    aspect: number,
+    nearz: number,
+    farz: number,
+    out?: mat4
+  ): mat4;
 
   /**
    * Not implemented yet
@@ -566,10 +579,15 @@ export interface vtkCamera extends vtkObject {
   setPositionAndFocalPoint(position: number[], focalPoint: number[]): boolean;
 
   /**
-   *
-   * @param {mat4} mat
+   * Set the explicit projection matrix, or null to use a projection computed
+   * from the camera's standard parameters.
    */
-  setProjectionMatrix(mat: mat4): boolean;
+  setExplicitProjectionMatrix(mat: Nullable<mat4>): boolean;
+
+  /**
+   * @deprecated Use setExplicitProjectionMatrix instead.
+   */
+  setProjectionMatrix(mat: Nullable<mat4>): boolean;
 
   /**
    * Set the roll angle of the camera about the direction of projection.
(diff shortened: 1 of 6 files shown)


Sources/Rendering/Core/Camera/index.js
```diff
@@ -557,9 +557,16 @@ function vtkCamera(publicAPI, model) {
   };
 
   publicAPI.setProjectionMatrix = (mat) => {
+    if (model.projectionMatrix === mat) {
+      return false;
+    }
     model.projectionMatrix = mat;
+    publicAPI.modified();
+    return true;
   };
```
finetjul, line 571: why not `macros.setArray()`

Sources/Rendering/Core/Camera/index.js
```diff
@@ -557,9 +557,16 @@ function vtkCamera(publicAPI, model) {
...
 
+  publicAPI.getProjectionMatrixByReference = () => model.projectionMatrix;
```
finetjul: I think it's a problem that `getProjectionMatrixByReference()` and `getProjectionMatrix()` do not return the same thing.

It's also a problem that a "get" does not return the same value than the "set".

Therefore, publicAPI.getProjectionMatrix should probably be renamed (or model.projectMatrix must be renamed)

daker: i renamed it to customProjectionMatrix, maybe you have a better name ?

Sources/Rendering/Core/Camera/index.js
```diff
@@ -556,19 +556,18 @@ function vtkCamera(publicAPI, model) {
     return out;
   };
 
-  publicAPI.setProjectionMatrix = (mat) => {
-    model.projectionMatrix = mat;
-  };
+  // Backward compatible alias for the former, ambiguous setter name.
+  publicAPI.setProjectionMatrix = publicAPI.setCustomProjectionMatrix;
```
finetjul: let's print a developer warning in addition to call setCustomProjectionMatrix. 
At some point we will remove that "setProjectionMatrix" method

daker: done

finetjul (approved):
LGTM

@sankhesh @PaulHax  good with you too ?

Sources/Rendering/Core/Camera/index.d.ts
```diff
@@ -566,10 +578,15 @@ export interface vtkCamera extends vtkObject {
   setPositionAndFocalPoint(position: number[], focalPoint: number[]): boolean;
 
   /**
-   *
-   * @param {mat4} mat
+   * Set the custom projection matrix, or null to use a projection computed
+   * from the camera's standard parameters.
+   */
+  setCustomProjectionMatrix(mat: Nullable<mat4>): boolean;
```
PaulHax: Lets consider naming this after C++ VTK's ExplicitProjectionTransformMatrix

PaulHax: Should we follow VTK C++ and call this SetExplicitProjectionTransformMatrix (so the thing is `explicitProjectionMatrix`)?

Sources/Rendering/WebGPU/Camera/index.js
```diff
@@ -15,6 +15,16 @@ function vtkWebGPUCamera(publicAPI, model) {
...
     mat4.identity(outMat);
+
+    const customProjectionMatrix =
+      model.renderable.getCustomProjectionMatrixByReference?.();
+    if (customProjectionMatrix) {
+      const scale = 1 / model.renderable.getPhysicalScale();
+      mat4.copy(outMat, customProjectionMatrix);
+      mat4.scale(outMat, outMat, [scale, scale, scale]);
+      return;
+    }
```
PaulHax: WebGL and WebGPU use opposite depth conventions: WebGL is near→−1/far→+1, WebGPU is reversed-Z (near→1/far→0).

This copies a WebGL-style matrix straight in with no remap. Overlapping objects can draw in the wrong order, and geometry near the camera can get clipped.

Sources/Rendering/WebGPU/Camera/index.js
```diff
@@ -15,6 +15,16 @@ function vtkWebGPUCamera(publicAPI, model) {
 
   publicAPI.getProjectionMatrix = (outMat, aspect, cRange, windowCenter) => {
     mat4.identity(outMat);
```
PaulHax: This identiy call is not doing anyting if we early return in the explict matrix case.

PaulHax (commented):
Dat WebGPU getting better and better 🥰

Do we have any bandwidth for some tests?

This code has some fixing ideas  https://github.com/PaulHax/vtk-js/pull/39

daker:
@PaulHax do you want to submit the PR or want me to cherry pick the commit?

PaulHax:
if you want to copy in whats useful from the other pr, that might be best.

finetjul (approved):

PaulHax (approved):
🚀

PaulHax:
Just a note: user suplied explicit matrix must use Core/OpenGL clip-depth convention (near = −1, far = +1). Supplying an already WebGPU-native 0…1 matrix would be remapped again and break depth ordering.  

I like how this smooths transitions between backends, but users do need to keep that in mind (and its also a departure from C++ VTK which uses the explict matrix verbatim)

PR #3627, opened 2026-09-03
feat(ImageInterpolation): port VTK ImageInterpolator cubic interpolation implementation

Jo-Byr:
### Context
Add missing cubic interpolation implementation in ImageInterpolator, 1:1 from VTK C++ version.


Sources/Imaging/Core/ImageInterpolator/example/index.js
```diff
@@ -0,0 +1,175 @@
+import '@kitware/vtk.js/favicon';
```
finetjul, line 1: please add a galery screenshot (e.g. 500x280px with 50% JPEG compression) when you add an example.

Sources/Imaging/Core/ImageInterpolator/index.js
```diff
@@ -178,13 +178,154 @@ function vtkImageInterpolator(publicAPI, model) {
...
+        // loop over z
+        let ifz = fZ[k];
+        let factz = factZ[k];
+        let j = j1;
+        do {
+          // loop over y
+          let ify = fY[j];
+          let fzy = ifz * ify;
+          let factzy = factz + factY[j];
+          // loop over x is unrolled (significant performance boost)
```
finetjul, line 307: are you sure this statement holds in Javascript ? 
Can you quickly check in benchmark it ?
If it does not hold, then lets simply use regular for loops

Sources/Imaging/Core/ImageInterpolator/index.js
```diff
@@ -178,13 +178,154 @@ function vtkImageInterpolator(publicAPI, model) {
...
+    const minX = inExt[0];
+    const maxX = inExt[1];
+    const minY = inExt[2];
+    const maxY = inExt[3];
+    const minZ = inExt[4];
+    const maxZ = inExt[5];
+
+    let factX = Array(4);
+    let factY = Array(4);
+    let factZ = Array(4);
```
finetjul, line 214: array creation is somehow expensive, move their creations outside the scope of the function.

Jo-Byr: The other implementations use separate variables. Would it be better to use arrays that are created only once, or 4 variables created every time ?

finetjul: I guess it depends how many variables/arrays :-) Something to bench...

Sources/Imaging/Core/ImageInterpolator/index.js
Jo-Byr: It seems to allow a slight performance gain. In the example I added, with the current version, it takes 41.6ms on average to interpolate the image, vs 46ms with a triple loop (10 repetitions on each for the average).

Documentation/examples/gallery.js
```diff
@@ -0,0 +1,1089 @@
+// This file is auto-generated by generate-gallery.mjs
```
finetjul, line 1: you shouldn't need to add this file. screenshots are automatically considered by the existing script.

finetjul (changes requested):

Sources/Imaging/Core/ImageInterpolator/index.js
```diff
@@ -178,13 +178,154 @@ function vtkImageInterpolator(publicAPI, model) {
...
+    const floorZ = vtkInterpolationMathFloor(point[2]);
+
+    let inIdX0 = floorX.floored;
+    let inIdY0 = floorY.floored;
+    let inIdZ0 = floorZ.floored;
+    const fx = floorX.error;
+    const fy = floorY.error;
+    const fz = floorZ.error;
+    let inIdX1 = inIdX0 + (fx !== 0);
+    let inIdY1 = inIdY0 + (fy !== 0);
```
finetjul: apparently there are unused variables, please check github report

finetjul (approved):
LGTM

PR #3634, opened 2026-09-05
fix(DataArray): validate structural mutations and preserve ranges

### LINKED ISSUE #3628: vtkDataArray.getValue reads past array end on non-first-component indices
# `vtkDataArray.getValue` reads past array end on non-first-component indices

## Summary

`vtkDataArray.getValue(valueIdx)` computes its tuple index as
`valueIdx / model.numberOfComponents` without flooring. Any `valueIdx`
that isn't a multiple of `numberOfComponents` reads out of bounds and
returns `undefined` (or the wrong value if adjacent bytes happen to
contain a live entry).

## File / line

`Sources/Common/Core/DataArray/index.js:152-156`

```js
publicAPI.getValue = (valueIdx) => {
  const idx = valueIdx / model.numberOfComponents;   // ← not floored
  const comp = valueIdx % model.numberOfComponents;
  return publicAPI.getComponent(idx, comp);
};
```

Downstream `getComponent(idx, comp)` at
`Common/Core/DataArray.js:145-147` computes
`values[idx * numberOfComponents + comp]`, so a non-integer `idx`
becomes a non-integer array subscript, resolving to `undefined` under
`TypedArray[]` access.

## Minimal reproducer

```html
<!doctype html>
<script type="module">
import '@kitware/vtk.js/Rendering/Profiles/Geometry';
import vtkDataArray from '@kitware/vtk.js/Common/Core/DataArray';

const a = vtkDataArray.newInstance({
  numberOfComponents: 3,
  values: Float32Array.from([
    0, 1, 2,    // tuple 0
    3, 4, 5,    // tuple 1
    6, 7, 8,    // tuple 2
    9, 10, 11,  // tuple 3
  ]),
});

// Expected: 11 (last live element)
// Actual  : undefined (reads values[13], off the end)
console.log('getValue(11) =', a.getValue(11));

// Every valueIdx where valueIdx % 3 !== 0 is broken. For a 3-comp
// array of 4 tuples, valueIdx ∈ {1,2,4,5,7,8,10,11} — that's 8 of
// the 12 slots, i.e. 66% of the flat-view accesses.
for (let i = 0; i < 12; i++) {
  console.log(i, '=>', a.getValue(i));
}
</script>
```

## Fix

```js
const idx = Math.floor(valueIdx / model.numberOfComponents);
```

One-line change, no API change, no behavior change on the correct
`valueIdx % numberOfComponents === 0` path.

## Impact

Every VTK.js consumer using `.getValue(i)` to iterate a flat view of
a multi-component array gets `undefined` for ~`(components - 1) /
components` of accesses. Single-component arrays are unaffected.

## Workaround

Use `getComponent(tupleIdx, compIdx)` — always correct.

## Cross-substrate note

Filed by the vtk-wasm project after our tri-substrate parity harness
surfaced this defect while comparing wasi VTK to VTK.js on a
`vtkFloatArray` (`tests/parity/data-array.parity.mjs`). Native VTK's
equivalent method is `vtkAOSDataArrayTemplate::GetValue(vtkIdType)`
which correctly floors via integer division.

daker:
### Context

Prevent invalid tuple insertions, resize operations, and component counts from corrupting array invariants. Invalidate cached ranges after mutations and preserve independently cloned ranges in newClone() and deepCopy().

Require setData() to receive a typed array.

Fixes #3628

### PR Checklist

- [x] Test coverage added


daker:
@finetjul

Sources/Common/Core/DataArray/test/testDataArray.js
```diff
@@ -581,3 +582,160 @@ it('Test vtkDataArray getState preserveTypedArrays option', () => {
...
+    values: new Float32Array([0, 1, 2, 3, 4, 5, 6, 7, 8]),
+  });
+
+  expect(dataArray.getValue(1)).toBe(1);
+  expect(dataArray.getValue(5)).toBe(5);
+
+  dataArray.setValue(1, 10);
+  expect(dataArray.getData()).toEqual(new Float32Array([0, 10, 2, 3, 4, 5]));
+  expect(dataArray.get()).toMatchObject({
+    values: new Float32Array([0, 10, 2, 3, 4, 5, 6, 7, 8]),
```
finetjul: this is internal details, it should not be tested. Internal implementation may differ.

Other valid options could be:
`[publicAPI.mouseScrollDebounceByPass](values: new Float32Array([0, 10, 2, 3, 4, 5, 0, 0, 0]),)`
`[publicAPI.mouseScrollDebounceByPass](values: new Float32Array([0, 10, 2, 3, 4, 5, 6, 7, 8, 0, 0, 0, 0, 0, 0]),)`

Sources/Common/Core/DataArray/index.d.ts
```diff
@@ -42,19 +42,38 @@ export interface vtkDataArray extends vtkObject {
 
   /**
    * Get the component for a given tupleIdx.
+   *
+   * Unlike `getValue()`, this method addresses the entire allocated backing
+   * array, including capacity beyond `getNumberOfValues()`.
    * @param {Number} tupleIdx
    * @param {Number} [componentIndex] (default: 0)
    */
   getComponent(tupleIdx: number, componentIndex?: number): number;
```
finetjul: Shall we create a new "type" to be an "index" (i.e. positive integer)? 
I understand we can't enforce it (without overhead), but it might give enough clues to the developers.

Sources/Common/Core/DataArray/index.js
```diff
@@ -184,11 +178,16 @@ function vtkDataArray(publicAPI, model) {
...
+    if (!Number.isInteger(extraNumTuples) || extraNumTuples < 0) {
+      return;
+    }
     resize(publicAPI.getNumberOfTuples() + extraNumTuples);
   };
 
   publicAPI.resize = (requestedNumTuples) => {
-    resize(requestedNumTuples);
+    if (!resize(requestedNumTuples)) {
+      return false;
```
finetjul, line 181: why bother doing a check and early return ?

Sources/Common/Core/DataArray/index.js
```diff
@@ -225,16 +224,13 @@ function vtkDataArray(publicAPI, model) {
     }
   };
 
-  publicAPI.getValue = (valueIdx) => {
-    const idx = valueIdx / model.numberOfComponents;
-    const comp = valueIdx % model.numberOfComponents;
-    return publicAPI.getComponent(idx, comp);
-  };
+  publicAPI.getValue = (valueIdx) => model.values[valueIdx];
```
finetjul, line 220: you also need to change StringArray and VariantArray that overwrote only `getComponent()`.

Sources/Common/Core/DataArray/index.js
```diff
@@ -322,6 +318,7 @@ function vtkDataArray(publicAPI, model) {
     for (let i = 0; i < model.numberOfComponents; i++) {
       model.values[offset + i] = tuple[i];
     }
+    publicAPI.dataChange();
```
finetjul: it should not trigger dataChange if nothing changes.

Sources/Common/Core/DataArray/index.js
```diff
@@ -330,22 +327,36 @@ function vtkDataArray(publicAPI, model) {
     for (let j = 0; j < last; ) {
       model.values[i++] = tuples[j++];
     }
+    if (last > 0) {
+      publicAPI.dataChange();
```
finetjul, line 340: it should not trigger dataChange if nothing changes.

Sources/Common/Core/DataArray/index.js
finetjul: can return true if something changes, false otherwise

Sources/Common/Core/DataArray/index.js
```diff
@@ -330,22 +327,36 @@ function vtkDataArray(publicAPI, model) {
...
   publicAPI.insertTuple = (idx, tuple) => {
+    if (!Number.isInteger(idx) || idx < 0) {
+      throw new RangeError('idx must be a nonnegative integer');
+    }
     if (model.size <= idx * model.numberOfComponents) {
-      model.size = (idx + 1) * model.numberOfComponents;
       resize(idx + 1);
+      model.size = (idx + 1) * model.numberOfComponents;
     }
     publicAPI.setTuple(idx, tuple);
```
finetjul: It should trigger `dataChanged()` if `size` was increased and if `setTuple()` does not call `dataChanged`

Sources/Common/Core/DataArray/index.js
```diff
@@ -330,22 +327,36 @@ function vtkDataArray(publicAPI, model) {
...
+        'tuples.length must be a multiple of numberOfComponents'
+      );
+    }
     const end = idx + tuples.length / model.numberOfComponents;
     if (model.size < end * model.numberOfComponents) {
-      model.size = end * model.numberOfComponents;
       resize(end);
+      model.size = end * model.numberOfComponents;
     }
     publicAPI.setTuples(idx, tuples);
```
finetjul: It should trigger `dataChanged()` if `size` was increased and if `setTuples()` does not call `dataChanged`

Sources/Common/Core/DataArray/index.js
```diff
@@ -508,6 +544,8 @@ function vtkDataArray(publicAPI, model) {
     } else {
       publicAPI.setData(other.getData().slice());
     }
+    model.ranges = ranges;
+    model.rangeTuple = [0, 0];
```
finetjul, line 565: you probably want the ranges to already be up-to-date when modified() will be triggerred (in case an observer wants to react on the change)

Sources/Common/Core/DataArray/index.d.ts
daker: i renamed the types and used NonNegativeInteger which more generic and can be used in other places

Sources/Common/Core/DataArray/index.js
daker: i check requestedNumTuples only now

Sources/Common/Core/DataArray/index.d.ts
```diff
@@ -42,18 +48,44 @@ export interface vtkDataArray extends vtkObject {
 
   /**
    * Get the component for a given tupleIdx.
+   *
+   * Unlike `getValue()`, this method addresses the entire allocated backing
+   * array, including capacity beyond `getNumberOfValues()`.
```
finetjul, line 53: I wouldn't advertise it.
This is an implementation detail. 
I can't see a legitimate reason why to access it.

Sources/Common/Core/DataArray/index.d.ts
```diff
@@ -42,18 +48,44 @@ export interface vtkDataArray extends vtkObject {
...
+  getComponent(
+    tupleIdx: NonNegativeInteger,
+    componentIndex?: NonNegativeInteger
+  ): number;
 
   /**
    * Set the component value for a given tupleIdx and componentIndex.
+   *
+   * Unlike `setValue()`, this method addresses the entire allocated backing
+   * array, including capacity beyond `getNumberOfValues()`.
```
finetjul, line 66: same here, let's not advertise it.

Sources/Common/Core/DataArray/index.js
```diff
@@ -319,35 +308,75 @@ function vtkDataArray(publicAPI, model) {
 
   publicAPI.setTuple = (idx, tuple) => {
     const offset = idx * model.numberOfComponents;
+    let changed = false;
     for (let i = 0; i < model.numberOfComponents; i++) {
-      model.values[offset + i] = tuple[i];
+      const valueIdx = offset + i;
+      const previousValue = model.values[valueIdx];
+      model.values[valueIdx] = tuple[i];
+      if (!Object.is(model.values[valueIdx], previousValue)) {
```
finetjul, line 316: why bother using `Object.is` ? Isn't `!=` sufficient ?

daker:
```js
Object.is(NaN, NaN)   // true
Object.is(-0, 0)      // false
```
Loose != would:
```js
NaN != NaN            // true  - falsely reports a change
-0 != 0               // false - misses a change
'1' != 1              // false - performs coercion
null != undefined     // false
```
differs for NaN and signed zero

finetjul: I meant `!==`. How expensive is `Object.is` wrt `!==`? This is on a critical path.
I can live with `-0 === 0`, it's a bit more problematic that `NaN !== NaN` though.

daker: object.is is a bit expensive 

```js
!== 673.70 ms 6.74 ns/op 100000000
Object.is 898.70 ms 8.99 ns/op 100000000
```

Benchmark script:

```js
function bench(label, fn, iterations = 100_000_000) {
  // Warm up JIT
  for (let i = 0; i < 1_000_000; i++) fn(i)
  const start = performance.now()
  let result = 0
  for (let i = 0; i < iterations; i++) {
    result += fn(i)
  }
  const end = performance.now()
  console.log(
    label,
    `${(end - start).toFixed(2)} ms`,
    `${((end - start) * 1e6 / iterations).toFixed(2)} ns/op`,
    result
  )
}

const a = 123
const b = 456

bench("!==", () => a !== b)
bench("Object.is", () => !Object.is(a, b))
```

finetjul: Ok to keep `Object.is()`. There is still a fast path to modify the array through `getData()`

daker: fixed

Sources/Common/Core/DataArray/index.js
```diff
@@ -225,16 +217,13 @@ function vtkDataArray(publicAPI, model) {
...
-    const comp = valueIdx % model.numberOfComponents;
-    return publicAPI.getComponent(idx, comp);
-  };
+  publicAPI.getValue = (valueIdx) => model.values[valueIdx];
 
   publicAPI.setValue = (valueIdx, value) => {
-    const idx = valueIdx / model.numberOfComponents;
-    const comp = valueIdx % model.numberOfComponents;
-    publicAPI.setComponent(idx, comp, value);
+    if (value !== model.values[valueIdx]) {
```
finetjul: Object.is ?

daker: i will change in both setComponent & setValue

finetjul (approved):
LGTM

PR #3635, opened 2026-09-07
fix(renderwindowinteractor): do not trigger mouse press event for the…

finetjul:
… first mouse down button

The RWI may not catch a mouse press event (e.g. because it was caught by a div over the RW), in that case the "chorded event" must not apply and moving the mouse over the RW must not trigger a mouse down event

### Context
If you add a slider over the RW, a click and drag on the slider may trigger some mouse move events on the RW if the mouse moves outside the slider. Currently, due to the chorded event mechanism, the first mouse move event was triggering a mouse press event that was then enabling the interactor style to rotate the camera while the slider was being interacted with.

### Results
The interactor style interaction use case described above no longer happens because the mouse press event is no longer triggered.

### Changes
The chorded event mechanism is not applied for the "first" button(s), only when there is at least one button down.

### PR Checklist
- [ ] GitHub Actions CI passed: [semantic-release](https://github.com/semantic-release/semantic-release) commit messages, lint, and tests
- [x] Test coverage added

### DIFF
--- a/Examples/Rendering/MouseEvents/index.js
+++ b/Examples/Rendering/MouseEvents/index.js
@@ -163,36 +163,42 @@ clearBtn.addEventListener('click', () => {
 // ----------------------------------------------------------------------------
 
 interactor.onLeftButtonPress(() => {
+  const inconsistent = buttonState.left;
   buttonState.left = true;
   updateIndicators();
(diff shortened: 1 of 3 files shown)


Thibault-Pelletier (approved):
LGTM 👍

PR #3637, opened 2026-09-07
Fix vtkCutter and add test for ContourLoopExtraction

Jo-Byr:
### Context
vtkCutter yields invalid results if the plane intersects a point of the cut mesh. This is visible on vtkContourLoopExtraction's example where the loop extraction returns 8 loops (visibly 3: pink, green, pink but each actually has multiple layers under).

Screenshot 2026-09-07 154550

### Results
By properly handling cases where the plane intersects a point, we can fix the issue and have only 2 loops (red and green):

image

### Changes
Properly differentiate the 3 cases: negative sign, null sign (on plane) and positive sign

Dedupe points, result of a same point being counted twice.

image

For the green triangle, intersected in one of its vertices by the red plane, 2 edges will be counted as intersected on the same position.

Added a non-regression test on ContourLoopExtraction

Modified ContourLoopExtraction example's parameters to fit with the model's bounds.

### PR Checklist

- [x] Test coverage added

### DIFF
--- a/Sources/Filters/Core/Cutter/index.js
+++ b/Sources/Filters/Core/Cutter/index.js
@@ -82,6 +82,8 @@ function vtkCutter(publicAPI, model) {
     const newLinesData = [];
     const newPolysData = [];
     const newPointData = {}; // TODO: cell data must also be processed
+    const sign = (x) =>
+      x - model.cutValue === 0 ? 0 : x < model.cutValue ? -1 : 1;
 
     // Initialize arrays
     const numberOfArrays = pointData.getNumberOfArrays();
@@ -121,16 +123,24 @@ function vtkCutter(publicAPI, model) {
       }
 
       // Get associated scalar of points that constitute the current cell
-      for (let i = 0; i < it.cellSize; ) {
-        cellPointsScalars[i] = model.cutScalars[it.cell[i++]];
+      let sideFirstPoint = null;
+      let firstPointIdx = null;
+      for (let i = 0; i < it.cellSize; i++) {
+        cellPointsScalars[i] = model.cutScalars[it.cell[i]];
+        if (sideFirstPoint === null && cellPointsScalars[i] !== 0) {
+          sideFirstPoint = sign(cellPointsScalars[i]);
+          firstPointIdx = i;
+        }
       }
 
       // Check if all cell points are on same side (same side == cell not crossed by cut function)
       // TODO: won't work if one point scalar is = 0 ?
-      const sideFirstPoint = cellPointsScalars[0] > 0;
       let allPointsSameSide = true;
-      for (let i = 1; i < it.cell.length; i++) {
-        const sideCurrentPoint = cellPointsScalars[i] > 0;
+      for (let i = 0; i < it.cell.length; i++) {
+        if (i === firstPointIdx) {
+          continue;
+        }
+        const sideCurrentPoint = sign(cellPointsScalars[i]);
         if (sideCurrentPoint !== sideFirstPoint) {
           allPointsSameSide = false;
           break;
@@ -143,7 +153,7 @@ function vtkCutter(publicAPI, model) {
       }
 
       // Find and compute edges which intersect cells
-      const intersectedEdgesList = [];
+      let intersectedEdgesList = [];
       for (let i = 0; i < it.cellSize; i++) {
         const idNext = i + 1 === it.cellSize ? 0 : i + 1;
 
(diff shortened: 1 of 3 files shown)


Sources/Filters/Core/Cutter/index.js
```diff
@@ -82,6 +82,7 @@ function vtkCutter(publicAPI, model) {
     const newLinesData = [];
     const newPolysData = [];
     const newPointData = {}; // TODO: cell data must also be processed
+    const sign = (x) => (x - model.cutValue === 0 ? 0 : x < 0 ? -1 : 1);
```
finetjul: shouldn't it be `x < model.cutValue` instead of `x < 0`?

finetjul: `x - model.cutValue === 0`  => `x === model.cutValue` ?

Sources/Filters/Core/Cutter/index.js
```diff
@@ -250,6 +265,22 @@ function vtkCutter(publicAPI, model) {
         }
       }
 
+      // Remove duplicate points
```
finetjul: you only remove duplicate cells, not duplicate points, do you ? (when you remove a cell, you are not removing its associated points)

Sources/Filters/Core/Cutter/index.js
```diff
@@ -250,6 +265,22 @@ function vtkCutter(publicAPI, model) {
...
+            intersectedEdgesList[i].newPointID ===
+            intersectedEdgesList[j].newPointID
+          ) {
+            toRemove.push(i);
+          }
+        }
+      }
+      intersectedEdgesList = intersectedEdgesList.filter(
+        (_, idx) => !toRemove.includes(idx)
+      );
```
finetjul: any chance to use a Set instead ? 
An can you populate such Set before instead of creating another loops?

finetjul:
Does it mean the VTK C++ implementation also has the problem ? If so, please report a bug in VTK

Sources/Filters/Core/Cutter/index.js
Jo-Byr: intersectedEdgesList lists intersected edges of the current cell, with associated intersection point. So I'm removing a point of this list so that it's treated as a line and not as a poly with a duplicated point.

Jo-Byr:
> Does it mean the VTK C++ implementation also has the problem ? If so, please report a bug in VTK

I didn't look into the C++ implementation but it does not produce this invalid result, I used a local python vtk as a truth test.

Sources/Filters/General/ContourLoopExtraction/test/testContourLoopExtraction.js
```diff
@@ -0,0 +1,70 @@
...
+    [1.0, 1.0, 1.0],
+  ];
+  const origins = [
+    [0.0, 0.0, 0.0],
+    [-100.0, 0.0, 0.0],
+    [1.2, 0.0, 0.0],
+    [0.0, 0.0, -0.7],
+    [0.0, 5.0, 0.0],
+  ];
+  const expectedNbLoops = [2, 0, 3, 1];
```
sankhesh: Why not test all 5 loops?

Sources/Filters/Core/Cutter/index.js
```diff
@@ -121,16 +123,24 @@ function vtkCutter(publicAPI, model) {
...
       }
 
       // Check if all cell points are on same side (same side == cell not crossed by cut function)
       // TODO: won't work if one point scalar is = 0 ?
-      const sideFirstPoint = cellPointsScalars[0] > 0;
       let allPointsSameSide = true;
-      for (let i = 1; i < it.cell.length; i++) {
-        const sideCurrentPoint = cellPointsScalars[i] > 0;
+      for (let i = 0; i < it.cell.length; i++) {
+        if (i === firstPointIdx) {
```
sankhesh, line 140: This new `sign()-based` crossing detection is definitely better than what we had before but I'm curious what would happen if all points of the cell are exactly on the cut plane? I'd assume `firstPointIdx` remains `null`. It might be a good test case too.

sankhesh (approved):
LGTM with a couple of questions

finetjul:
> > Does it mean the VTK C++ implementation also has the problem ? If so, please report a bug in VTK
> 
> I didn't look into the C++ implementation but it does not produce this invalid result, I used a local python vtk as a truth test.

Isn't the vtkCutter JS implementation a port of the VTK C++ implementation ? Maybe it would be wise to check how it differs...

Jo-Byr:
> Isn't the vtkCutter JS implementation a port of the VTK C++ implementation ? Maybe it would be wise to check how it differs...

From my understanding of the C++ implementation, it has a special implementation for plane cutting with vtkPlaneCutter, itself going into vtkPolyDataPlaneCutter if the inputData is a polydata, and a generic DatasetCutter method for cutting a vtkPolyData by something else than a plane.

This method itself uses vtkContourHelper to get the contour out of the evaluated cutter function on every point of the input data that indirectly falls back to vtkTriangle::Contour that finds intersection points along edges and draws a line between them.

It all relies on the cut scalars indirectly. For plane cutter it uses a triple state too:

```
// Point is either above(=2), below(=1), or on(=0) the plane.
// NOLINTNEXTLINE(readability-avoid-nested-conditional-operator)
*inOutItr = (eval > zero ? 2 : (eval < zero ? 1 : 0));
```

I don't know if it's generic in C++ but it seems a natural approach to me.

Sources/Filters/General/ContourLoopExtraction/test/testContourLoopExtraction.js
Jo-Byr: It's an overlook.

Sources/Filters/Core/Cutter/index.js
Jo-Byr: I tested with a cube and it works fine with a cut plane in the same plane as one of the faces since their points will be discovered when testing the intersection of other cells that share these points.

The only case where I can see it failing is a single cell whose points are not shared with any other cell, that is colinear to the plane. But even if we added a fallback `firstPointIdx = firstPointIdx ?? 0` it would get caught by `allPointsSameSide`.

finetjul (approved):
LGTM

PR #3638, opened 2026-09-08
perf(Renderer): accumulate actors and volumes without concat

PaulHax:
### Context

`getActors()` and `getVolumes()` collect each prop's objects using repeated
`concat` calls. Each concat call copies the growing result array. 
This quadratic work repeats on every `getActors` call, including `getActors()`
during picking without an explicit pick list.

Batching geometry into fewer actors is the vtk.js way, but separate
actors are useful when datasets load and unload independently. This change came
out of loading tiled point clouds, where chunks of points are swapped in and out as the view changes.

### Results

Recorded headless Chromium medians per collection call, spanning actor and
volume results:

| Count | Before | After |
| ---: | ---: | ---: |
| 400 | 0.05–0.06 ms | 0.005–0.010 ms |
| 2,000 | 1.08–1.21 ms | 0.035–0.055 ms |

At 2,000 objects, this saves roughly 1 ms per call.

### Changes

- Use a shared `push` helper for linear collection, preserving array and
  non-array prop answers.
- Preserve order and fresh snapshots; by-reference getters return the latest
  snapshot.

### PR Checklist

- [x] Test coverage added

### DIFF
--- a/Sources/Rendering/Core/Renderer/index.js
+++ b/Sources/Rendering/Core/Renderer/index.js
@@ -17,8 +17,8 @@ function notImplemented(method) {
 // Append array elements individually; append other values as one element.
 function collectFromProps(props, getFromProp) {
   const collected = [];
-  props.forEach((prop) => {
-    const items = getFromProp(prop);
+  for (let i = 0; i < props.length; i++) {
+    const items = getFromProp(props[i]);
     if (Array.isArray(items)) {
       for (let j = 0; j < items.length; j++) {
         collected.push(items[j]);
@@ -26,7 +26,7 @@ function collectFromProps(props, getFromProp) {
     } else {
       collected.push(items);
     }
-  });
+  }
   return collected;
 }
 
--- a/Sources/Rendering/Core/Renderer/index.js
+++ b/Sources/Rendering/Core/Renderer/index.js
@@ -13,6 +13,26 @@ function notImplemented(method) {
   return () => vtkErrorMacro(`vtkRenderer::${method} - NOT IMPLEMENTED`);
 }
 
+// Props may return an array or a single object (e.g. vtkImageSlice).
+// Append array elements individually; append other values as one element.
+function collectFromProps(props, getFromProp) {
+  const collected = [];
+  props.forEach((prop) => {
+    const items = getFromProp(prop);
+    if (Array.isArray(items)) {
+      for (let j = 0; j < items.length; j++) {
+        collected.push(items[j]);
+      }
+    } else {
+      collected.push(items);
+    }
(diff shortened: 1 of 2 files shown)


sankhesh (approved):
LGTM 👍

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -13,6 +13,26 @@ function notImplemented(method) {
   return () => vtkErrorMacro(`vtkRenderer::${method} - NOT IMPLEMENTED`);
 }
 
+// Props may return an array or a single object (e.g. vtkImageSlice).
+// Append array elements individually; append other values as one element.
+function collectFromProps(props, getFromProp) {
+  const collected = [];
+  props.forEach((prop) => {
```
finetjul: Sometimes `.forEach` is slower than a regular `for` loop. It might not be true each time though (Chrome, Firefox)

Sources/Rendering/Core/Renderer/index.js
```diff
@@ -138,10 +158,7 @@ function vtkRenderer(publicAPI, model) {
   };
 
   publicAPI.getActors = () => {
-    model.actors = [];
-    model.props.forEach((prop) => {
-      model.actors = model.actors.concat(prop.getActors());
-    });
+    model.actors = collectFromProps(model.props, getPropActors);
```
finetjul, line 161: wouldn't you get even better performances if you
reuse the array ? : 
```
  collectFromProps = (props, collector, collected) {
  let c = 0;
  ...
  collected[c++] = items; // was .push before
  ...
  collected.length = c;
  return c;
}
...

   collectFromProps(model.props, getPropActors, model.actors);
```

Sources/Rendering/Core/Renderer/index.js
PaulHax: Yes, adopted.  Its significantly faster in FireFox.

Sources/Rendering/Core/Renderer/index.js
PaulHax: This did not show any perf improvment.  https://gist.github.com/PaulHax/48b960e570d834004a2aa763c68bfaff

finetjul (approved):
LGTM

Sources/Rendering/Core/Renderer/index.js
finetjul: 20% faster is an improvement 🫣

PaulHax: that 25% was for the `for` loop change

PR #3639, opened 2026-09-08
perf(OpenGL): draw only the polydata primitives that own elements

PaulHax:
### Context

`vtkOpenGLPolyDataMapper` visits all six primitive helpers on every draw, including helpers with no elements.  Thats slow.

### Results

400 polys-only actors, 300 measured frames per run. 

| Metric | Before | After |
| --- | ---: | ---: |
| Median render-call time | 29.0 ms | 23.8 ms |

### Changes

Record the active helpers in `updateBufferObjects` immediately after
`buildBufferObjects`, including when a subclass overrides `buildBufferObjects`.
Use that list in `renderPieceDraw`, retaining the element-count check for helpers
released between rebuilds.

Note: A subclass that replaces `updateBufferObjects` without calling the base
implementation will not populate the active list.

### PR Checklist

- [x] Test coverage added


sankhesh (approved):
LGTM

PR #3640, opened 2026-09-08
fix(DiskSource): fix Uint8 overflow

Jo-Byr:
### Context
For every circurmferential resolution (CR), there's a radial resolution (RR) past which the connectivity of the disk is wrong

image

(CR=6, RR=42)

image

(CR=4, RR=64)

image

(CR=10, RR=25)

### Results
The connectivity is correct for every combination of CR and RR.

### Changes
Replace Uint8Array by Uint32Array to fix overflow error with disks with more than 50 points

### PR Checklist

- [x] Test coverage added


finetjul (approved):
LGTM

PR #3641, opened 2026-09-08
chore(ImageMarchingCubes): fix example clipping

Jo-Byr:
### Context
The produces sphere in ImageMarchingCubes example can clip with the sampler bounds for a big enough radius or ISO value.

image

### Results
Bounds are updated when changing radius or ISO to prevent clipping.

### Changes
Update sample bounds on radius and iso value updates to avoid clipping of output with bounds

Set initial GUI radius to actual sphere's initial radius

### PR Checklist

- [x] Test coverage added

### DIFF
--- a/Sources/Filters/General/ImageMarchingCubes/example/index.js
+++ b/Sources/Filters/General/ImageMarchingCubes/example/index.js
@@ -45,10 +45,20 @@ mapper.setInputConnection(mCubes.getOutputPort());
 // ----------------------------------------------------------------------------
 // UI control handling
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


finetjul (commented):
I do not think there is a problem in the first place. The clipping makes sense and helps understand what the algorithm really does. 
Showing the "sample" volume (don't do it) could be more helpful than changing the bounds.

Jo-Byr:
> I do not think there is a problem in the first place. The clipping makes sense and helps understand what the algorithm really does.

So should I close this PR ?

finetjul:
Yes, we can close the issue.

PR #3649, opened 2026-09-11
feat(WebGPU): add vtkWebGPUConfiguration

### LINKED ISSUE #3648: [Feature] Support shared GPUDevice instances across WebGPURenderWindows
### Motivation

Applications with multiple VTK.js WebGPU viewports currently create a device per RenderWindow or Render through a shared canvas and read it back per viewport.
The shared canvas approach requires GPU fences and introduces avoidable latency.
WebGPU supports one GPUDevice backing multiple canvas contexts, so each viewport can render directly to its own canvas.

### Detailed Description

Allow vtkWebGPURenderWindow.setDevice(gpuDevice) before initialization.
Windows given the same native GPUDevice should share one internal vtkWebGPUDevice wrapper and its buffer, texture, shader, pipeline, and bindgroup caches.
Each window retains its own canvas context, swap chain, renderer tree, and interaction handling.
Deviceloss notification should be coordinated through the shared wrapper. Applications supply a replacement device for externally owned devices.

```js
const adapter = await navigator.gpu.requestAdapter({ powerPreference: 'high-performance' });
const device = await adapter.requestDevice({ requiredFeatures, requiredLimits });

const windowA = vtkWebGPURenderWindow.newInstance();
windowA.setDevice(device); 
const windowB = vtkWebGPURenderWindow.newInstance();
windowB.setDevice(device);
```

daker:
### Context

Fixes #3648

### PR Checklist

- [x] Test coverage added


daker:
@PaulHax

Sources/Rendering/WebGPU/Configuration/index.d.ts
```diff
@@ -0,0 +1,185 @@
...
+   * @defaultValue 'high-performance'
+   */
+  powerPreference?: PowerPreference;
+
+  /**
+   * Limits that the adapter and the device must supply. When this is null,
+   * initialization asks for the largest buffer sizes that the adapter offers.
+   * @defaultValue null
+   */
+  requiredLimits?: Nullable<RequiredLimits>;
```
PaulHax: This might be a departure from the repo conventions. We don't have to litigate it. So I don't have a strong opinion here, but I would prefer to just have things as undefined rather than undefined or null or a valid value. Already we see that power preference and required limits here is inconsistent. We have the question mark, so it could be undefined for power preference. So why not nullable? And then required limits here also has it could be undefined, and it also could be nullable, also could be null. My loosely held opinion for this repo is that we transition to just valid value or undefined.

Sources/Rendering/WebGPU/Configuration/index.d.ts
```diff
@@ -0,0 +1,185 @@
+import { vtkObject } from '../../../interfaces';
+import { Nullable } from '../../../types';
+
+/**
+ * Power preference sent to the browser when an adapter is requested.
+ *
+ * - `high-performance` selects the discrete GPU on a system that has one.
+ * - `low-power` selects the integrated GPU to save battery.
+ */
+export type PowerPreference = 'low-power' | 'high-performance';
```
PaulHax: I'm sorry, I didn't look into the details, but these seem like these wrappers on WebGPU APIs. Maybe our TypeScript settings could be updated, or maybe we can pull in a WebGPU types dependency. I think we might have done that actually. So instead of defining our own power preference values here, we could just reuse the maintained types API. 
Or are these just follwing the C++ VTK API for good reason, thats OK too.

PaulHax (commented):
Thanks for pushing on this. I don't have a project using WebGPU (at the moment) so I'm not really in a good position to give good feedback on this important foundational work.

Sources/Rendering/WebGPU/Configuration/index.d.ts
daker: i have pulled the webgpu type in devdeps

Sources/Rendering/WebGPU/Configuration/index.d.ts
daker: i have applied that

daker:
@finetjul can i get a review on this one when you have time ?

finetjul:
> @finetjul can i get a review on this one when you have time ?

That's more a topic for @sankhesh

Sources/Rendering/WebGPU/Configuration/index.js
```diff
@@ -0,0 +1,168 @@
...
+  };
+
+  publicAPI.isInitialized = () => model.deviceReady;
+
+  publicAPI.finalize = () => {
+    model.initializationPromise = null;
+    model.device = null;
+    model.adapter = null;
+    model.deviceReady = false;
+  };
```
sankhesh, line 135: Shouldn't finalize call `device.destroy()`? The GPU buffers, textures and pipelines stay alive until the GC collects the device, which can hold a lot of VRAM.

Sources/Rendering/WebGPU/Configuration/index.js
```diff
@@ -0,0 +1,168 @@
...
+      return await model.initializationPromise;
+    } finally {
+      model.initializationPromise = null;
+    }
+  };
+
+  publicAPI.isInitialized = () => model.deviceReady;
+
+  publicAPI.finalize = () => {
+    model.initializationPromise = null;
```
sankhesh, line 131: This wouldn't be enough for an async initialize in-flight.

Sources/Rendering/WebGPU/Configuration/index.js
```diff
@@ -0,0 +1,168 @@
...
+        if (!adapter) {
+          vtkErrorMacro('Failed to acquire a WebGPU adapter.');
+          return false;
+        }
+        model.adapter = adapter;
+
+        // Exact storage of 16 bit integers uses r32float. Request filter
+        // support when the adapter offers it, as vtkWebGPUConfiguration does
+        // for implementation-supported features.
+        const optionalFeatures = ['float32-filterable'];
```
sankhesh, line 85: Maybe worth making this list configurable and exposing `hasFeature` / `getFeature`.

Sources/Rendering/WebGPU/Configuration/index.js
```diff
@@ -0,0 +1,168 @@
...
+          vtkErrorMacro('Failed to acquire a WebGPU device.');
+          return false;
+        }
+        if (!hasRequiredLimits(requiredLimits, device.limits, 'device')) {
+          return false;
+        }
+
+        model.device = device;
+        model.deviceReady = true;
+        watchDeviceLoss(device);
```
sankhesh, line 113: Also add an uncaptured error listener on the device. As the owner of the device, the configuration should route validation errors to vtkErrorMacro.

Sources/types.d.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="@webgpu/types" />
```
sankhesh, line 1: But the package is only a devDependency ("dev": true in the lockfile). Consumers without it get TS2688 or unresolved GPU* types. Make it a dependency or peerDependency, or avoid the global reference. 

`getAapter()` and `getDevice()` are still typed any, and `requiredLimits?: GPUSupportedLimits` should be `Record<string, number>`.

sankhesh (changes requested):

sankhesh:
Several windows now share one vtkWebGPUDevice (pipelines, shader and buffer/texture caches). No cache is released or reference-counted when one window releases. Please confirm that a window's `releaseGraphicsResources()` can't destroy cached GPU objects another window still uses. It would be good to add a two-window test that includes device loss.

daker:
@sankhesh i have addressed your concerns

sankhesh (approved):
LGTM

